PR #227, opened 2014-12-12
Fix CPU_SUSPEND when invoked with affinity level higher than get_max_aff...

soby-mathew:
This patch fixes the assertion failure when CPU_SUSPEND is invoked with
an affinity level higher than supported by the platform by adding suitable
checks for affinity level within `psci_cpu_suspend`. Also added suitable
bound checks within `psci_aff_map_get_idx` to prevent indexing beyond array
limits.

Fixes ARM-software/tf-issues#260

Change-Id: I04b75c49729e6c6d1983add590f60146c8fc3630


danh-arm:
Please use shorter commit message titles in future (<~50 chars). I'll merge this one, though.

PR #236, opened 2015-01-19
Remove coherent memory

soby-mathew:
This patchset incrementally removes trusted firmware's dependency on coherent memory, culminating in a patch that optionally allows removal of the coherent memory section altogether.


danh-arm:
A short description summarising this patchset would be helpful

danh-arm:
It's OK, I've done it myself

PR #240, opened 2015-01-22
Remove coherent memory v2

danh-arm:
This patchset incrementally removes trusted firmware's dependency on coherent memory, culminating in a patch that optionally allows removal of the coherent memory section altogether.


danh-arm:
Supersedes #236

PR #241, opened 2015-01-22
Clean-up PSCI code and introduce early validation of parameters v2

danh-arm:
This patch-set incrementally re-factors the Trusted Firmware's implementation of PSCI specification to validate the PSCI parameters early enough so that error can be returned safely back to the PSCI client. It also optimizes the argument list for platform hooks and introduces early return from CPU_SUSPEND in-case a wake-up interrupt is detected prior to power-down sequences.


danh-arm:
Supersedes #237

PR #242, opened 2015-01-22
Rework PSCI migrate APIs and implement the FEATURES API

danh-arm:
This patch-set reworks the MIGRATE, MIGRATE_INFO_TYPE and MIGRATE_INFO_UP_CPU PSCI APIs. It also implements the FEATURES API which is a mandatory API in PSCI 1.0 specification.


danh-arm:
Supersedes #238

PR #245, opened 2015-01-26
Increment the PSCI VERSION to 1.0 (PR v2)

danh-arm:
This patch:

   * Bumps the PSCI VERSION to 1.0. This means that
     the PSCI_VERSION API will now return the value 0x00010000
     to indicate the version as 1.0. The firmware remains
     compatible with PSCI v0.2 clients.

   * The firmware design guide is updated to document the
     APIs supported by the Trusted Firmware generic code.

   * The FVP Device Tree Sources (dts) and Blobs(dtb) are also
     updated to add "psci-1.0" and "psci-0.2" to the list of
     compatible PSCI versions.

Change-Id: Iafc2f549c92651dcd65d7e24a8aae35790d00f8a


danh-arm:
Supersedes #239

PR #250, opened 2015-01-29
Fix the Cortex-A57 reset handler register usage

soby-mathew:
The CPU specific reset handlers no longer have the freedom
of using any general purpose register because it is being invoked
by the BL3-1 entry point in addition to BL1. The Cortex-A57 CPU
specific reset handler was overwriting x20 register which was being
used by the BL3-1 entry point to save the entry point information.
This patch fixes this bug by reworking the register allocation in the
Cortex-A57 reset handler to avoid using x20. The patch also
explicitly mentions the register clobber list for each of the
callee functions invoked by the reset handler

Change-Id: I28fcff8e742aeed883eaec8f6c4ee2bd3fce30df


danh-arm:
To be clear, this is regression introduced by #247

PR #258, opened 2015-02-19
Juno: clear DMA-330 SMMU security state

rmurphy-arm:
By default the SMMU for the DMA-330 is configured to mark some stream IDs
as always belonging to the Secure world. As a result, if EL1 software turns
the SMMU on, certain Non-Secure accesses get rewritten as Secure, making
them bypass translation and access Secure physical addresses directly.

Since the current Juno board firmware configures the DMA-330 hardware as
Non-Secure, rewrite the SMMU's default SSD table as well to prevent any
unexpected behaviour in EL1.

Change-Id: Iaa81d883eecf28d80eb182b9ce475684bf9c718c

### DIFF
--- a/plat/juno/juno_def.h
+++ b/plat/juno/juno_def.h
@@ -244,4 +244,10 @@
 #define PRIMARY_CPU_SHIFT	8
 #define PRIMARY_CPU_MASK	0xf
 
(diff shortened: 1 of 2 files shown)


danh-arm:

PR #260, opened 2015-02-27
TBB: use SHA256 to generate the certificate signatures

jcastillo-arm:
This patch replaces SHA1 by SHA256 in the 'cert_create' tool, so
certificate signatures are generated according to the NSA Suite B
cryptographic algorithm requirements.

Documentation updated accordingly.

Change-Id: I7be79e6b2b62dac8dc78a4f4f5006e37686bccf6


danh-arm:
retest this please

PR #263, opened 2015-03-11
TBB: remove PolarSSL SHA1 functions from the binary

jcastillo-arm:
Commit ea4ec3aad5e1 ("TBB: use SHA256 to generate the certificate
signatures") updated the cert_create tool to generate the
signatures using SHA256 instead of SHA1. Therefore, SHA1 is no
longer required.

This patch removes the SHA1 option from the PolarSSL configuration
file. The source file sha1.c is no longer needed and has been
excluded from the build. The SHA1 functions are no longer included
in the binary, reducing the memory footprint of BL1 and BL2 by
approximately 6 KB.

Change-Id: I72ea2cff03c0964c3eaadce148ec2ad2c6dde2e3


danh-arm:
retest this please

PR #271, opened 2015-03-19
TLK dispatcher

vwadekar:
This topic branch adds the TLK dispatcher to the mix. This dispatcher helps TLK to run with the Trusted Firmware.

Verification: Code compiles with the master branch.

### DIFF
--- a/acknowledgements.md
+++ b/acknowledgements.md
@@ -4,6 +4,7 @@ Contributor Acknowledgements
 Companies
 ---------
 Linaro Limited
+NVIDIA Corporation
 
 Individuals
 -----------
--- a/docs/change-log.md
+++ b/docs/change-log.md
@@ -105,7 +105,9 @@ New features
     added.  Details of using it with ARM Trusted Firmware can be found in
     [OP-TEE Dispatcher]
 
-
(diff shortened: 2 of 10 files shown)


danh-arm:

vwadekar:
It seems there is a problem with one of the commits in the list. How should
I fix the problem?  Should I just fix the issue and rewrite the topic
branch history? What's the general way of fixing such problems?

danh-arm:
Hi Varun

While your PR is still open and there are no comments on it, you are free to force push to your topic branch (vwadekar:tlkd) and the PR will be automatically updated.

If there are significant comments on the commits in your PR, then it's usually better to raise another PR on another branch and close the first PR. The comment history will then be retained.

Once a PR is merged, the commits (and comments against them) are fixed so it doesn't matter what you do to the topic branch.

Finally, I notice that the target branch of this PR is ARM-software:master (the default). You should make this ARM-software:integration instead. You might need to raise another PR to do this.

Regards

Dan.

vwadekar:
Thanks Dan. Will generate another PR.

vwadekar:
Created another pull request against the integration branch rather than master.

PR #277, opened 2015-03-27
Optimize the bakery lock implementation

soby-mathew:
This patch-set applies several optimizations to bakery lock implementation that uses coherent memory. It also adds assertion checks to prevent recursive acquisition for the implementation that uses normal memory.


danh-arm:
retest this please

PR #282, opened 2015-04-02
Pass arguments/results between EL3/S-EL1 via CPU registers (x0-x7)

vwadekar:
This patch removes the need for a shared buffer between the EL3 and S-EL1
levels. We now use the CPU registers, x0-x7, while passing data between
the two levels. Since TLK is a 32-bit Trusted OS, tlkd has to unpack the
arguments in the x0-x7 registers. TLK in turn gets these values via r0-r7.

### DIFF
--- a/include/bl32/payloads/tlk.h
+++ b/include/bl32/payloads/tlk.h
@@ -52,7 +52,6 @@
 #define TLK_PREEMPTED		(0x32000002 | (1 << 31))
 #define TLK_ENTRY_DONE		(0x32000003 | (1 << 31))
 #define TLK_VA_TRANSLATE	(0x32000004 | (1 << 31))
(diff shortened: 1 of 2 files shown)


danh-arm:

vwadekar:
Hi, any feedback on this patch?

danh-arm:
Hi Varun. We've just had 2 public holidays and a weekend here. We'll look into it soon. Dan.

vwadekar:
Ok. Thanks for the update Dan.

danh-arm:
test this please

danh-arm:
Just minor comments from me, Varun

vwadekar:
Thanks Dan. Will push a new patch tomorrow.

vwadekar:
Pushed v1 to fix Dan's comments.

PR #284, opened 2015-04-08
Add support to indicate size and end of assembly functions

kpet:
In order for the symbol table in the ELF file to contain the size of
functions written in assembly, it is necessary to report it to the
assembler using the .size directive.

To fulfil the above requirements, this patch introduces an 'endfunc'
macro which contains the .endfunc and .size directives. It also adds
a .func directive to the 'func' assembler macro.

The .func/.endfunc have been used so the assembler can fail if
endfunc is omitted.

Fixes ARM-Software/tf-issues#295

Change-Id: If8cb331b03d7f38fe7e3694d4de26f1075b278fc
Signed-off-by: Kévin Petit <kevin.petit@arm.com>

### DIFF
--- a/bl1/aarch64/bl1_entrypoint.S
+++ b/bl1/aarch64/bl1_entrypoint.S
@@ -172,3 +172,4 @@ func bl1_entrypoint
 	bl	bl1_main
 panic:
 	b	panic
+endfunc bl1_entrypoint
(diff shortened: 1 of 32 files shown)


danh-arm:

danh-arm:
test this please

PR #285, opened 2015-04-09
Pass arguments/results between EL3/S-EL1 via CPU registers (x0-x7)

vwadekar:
This patch removes the need for a shared buffer between the EL3 and S-EL1
levels. We now use the CPU registers, x0-x7, while passing data between
the two levels. Since TLK is a 32-bit Trusted OS, tlkd has to unpack the
arguments in the x0-x7 registers. TLK in turn gets these values via r0-r7.


danh-arm:

danh-arm:
Looks good now.

Just one small thing. When you close PRs and raise a new one, can you add a link to the old one for reference (i.e. #281 in this case).

vwadekar:
Yes, will do. I am new to this review process, so please bear with me till
get it right.

PR #287, opened 2015-04-13
Panic if platform specific BL3-0 handling fails v2

danh-arm:
The return value of bl2_plat_handle_bl30() used to be ignored.
This patch modifies the function load_bl30() so that it now
checks this return value and returns it to bl2_main().

This patch also unifies the error handling code across the
load_blx() functions so that they return a status code in all
cases and bl2_main() has the sole responsibility of panicking
if appropriate.

Change-Id: I2b26cdf65afa443b48c7da1fa7da8db956071bfb

### DIFF
--- a/bl2/bl2_main.c
+++ b/bl2/bl2_main.c
@@ -205,23 +205,28 @@ static int load_bl30(void)
 		       &bl30_image_info,
 		       NULL);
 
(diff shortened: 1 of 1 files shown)


danh-arm:
Supersedes #286

PR #291, opened 2015-04-17
mt8173: Add initial implementation of BL31

mtk09422:
Change-Id: Id9f75cbd2f56a5c765ae54592161b6b510d7051e
Signed-off-by: CC Ma <cc.ma@mediatek.com>


danh-arm:

danh-arm:
Commit msg title doesn't convey what the patch is doing. Suggest "Initial platform port for MediaTek mt8173".

PR #292, opened 2015-04-20
Juno: Move to the new ARM SCP Messaging Interfaces

sandrine-bailleux:
The communication protocol used between the AP cores and the SCP
has undergone a number of changes.

The first patch makes the required modifications to the SCP Boot Protocol,
SCPI Protocol and MHU driver code in the Juno platform code so that the
AP cores are still able to communicate with the SCP.

The second patch detects any SCP version incompatibility in case the user
forgot to update the SCP ROM firmware.


danh-arm:
Superseded by #296

PR #293, opened 2015-04-22
Issue296 v8

omnium21:
I've updated the code with the fixes from the review comments in the v7 branch.  And I rebased it to the latest master branch.

I've run it through checkpatch and I only see two warnings that I've ignored:

drivers/arm/sp804/sp804.c:75: WARNING: line over 80 characters
include/drivers/timer.h:36: WARNING: do not add new typedefs


danh-arm:

omnium21:
Sorry Dan, I forgot to move the #defines from platform_def.h to fvp_def.h in this version, so I'll send another pull request for v9 now...

PR #299, opened 2015-05-11
Preserve bl31 boot registers

vwadekar:
In case of platforms using custom BL1/BL2 boot loaders and setting
RESET_TO_BL31, we need to preserve the x0-x3 registers that might
have been used to pass input parameters to the BL31 image.


vwadekar:
Hi Dan et al, I'm not sure if we should be seeing the commits where I merged from the Master branch to get to top-of-tree. Please let me know if I should be doing things in a different way.

danh-arm:

vwadekar:
Dan, thanks for pointing me to the issue #304. Do you think we could merge this change in the interim as you have not added a timeline #304? The change itself is very small and wont interfere with the actual code re-structure that you are planning.

And thanks for helping me out with the merge v rebase problem.

danh-arm:
Hi Varun. No, we're not going to merge what amounts to a hack, especially as I don't think it would work as you intend without further changes. The function will not preserve the x0-x3 registers (when RESET_TO_BL31 is true) with this change.

The phase 1 work in #304 will be implemented as a priority.

vwadekar:
Ok, thanks. Will wait for the implementation of phase 1. Also, it would be good if you can add an ETA for phase 1 in #304.

PR #300, opened 2015-05-13
Support for NVIDIA's Tegra T210 SoCs

vwadekar:
T210 is the latest chip in the Tegra family of SoCs from NVIDIA. It is an
ARM v8 dual-cluster (A57/A53) SoC, with any one of the clusters being active
at a given point in time.

This patch adds support to boot the Trusted Firmware on T210 SoCs. The patch
also adds support to boot secondary CPUs, enter/exit core power states for
all CPUs in the slow/fast clusters. The support to switch between clusters
is still not available in this patch and would be available later.


vwadekar:
Hello Dan, please let me know if there's some more documentation needed.

danh-arm:

vwadekar:
Hello Dan. I have obviously messed up the copyrights, but will wait for
your feedback for the code as well before pushing another patch. Thanks.

vwadekar:
https://github.com/ARM-software/arm-trusted-firmware/pull/302 is the new PR with some minor code comment and copyright fixes.

PR #304, opened 2015-05-19
Tegra SoC support v2

vwadekar:
Support for booting Trusted Firmware with Tegra T210 SoCs

### DIFF
--- a/drivers/ti/uart/16550_console.S
+++ b/drivers/ti/uart/16550_console.S
@@ -1,0 +1,204 @@
+/*
+ * Copyright (c) 2015, ARM Limited and Contributors. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * Neither the name of ARM nor the names of its contributors may be used
+ * to endorse or promote products derived from this software without specific
+ * prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <arch.h>
+#include <asm_macros.S>
+#include <uart_16550.h>
+
+	.globl	console_init
+	.globl	console_putc
+	.globl	console_core_init
+	.globl	console_core_putc
+	.globl	console_getc
+
+	/*
+	 *  The console base is in the data section and not in .bss
+	 *  even though it is zero-init. In particular, this allows
+	 *  the console functions to start using this variable before
+	 *  the runtime memory is initialized for images which do not
+	 *  need to copy the .data section from ROM to RAM.
+	 */
+.section .data.console_base ; .align 3
+	console_base: .quad 0x0
+
+	/* -----------------------------------------------
+	 * int console_init(unsigned long base_addr,
+	 * unsigned int uart_clk, unsigned int baud_rate)
+	 * Function to initialize the console without a
+	 * C Runtime to print debug information. It saves
+	 * the console base to the data section.
+	 * In: x0 - console base address
+	 *     w1 - Uart clock in Hz
+	 *     w2 - Baud rate
+	 * out: return 1 on success.
+	 * Clobber list : x1 - x3
+	 * -----------------------------------------------
+	 */
+func console_init
+	adrp	x3, console_base
+	str	x0, [x3, :lo12:console_base]
+	b	console_core_init
+endfunc console_init
+
+	/* -----------------------------------------------
+	 * int console_core_init(unsigned long base_addr,
+	 * unsigned int uart_clk, unsigned int baud_rate)
+	 * Function to initialize the console without a
+	 * C Runtime to print debug information. This
+	 * function will be accessed by console_init and
+	 * crash reporting.
+	 * In: x0 - console base address
+	 *     w1 - Uart clock in Hz
+	 *     w2 - Baud rate
+	 * Out: return 1 on success
+	 * Clobber list : x1, x2, x3
+	 * -----------------------------------------------
+	 */
+func console_core_init
+	/* Check the input base address */
+	cbz	x0, init_fail
+	/* Check baud rate and uart clock for sanity */
+	cbz	w1, init_fail
+	cbz	w2, init_fail
+
+	/* Program the baudrate */
+	/* Divisor =  Uart clock / (16 * baudrate) */
+	lsl	w2, w2, #4
+	udiv	w2, w1, w2
+	and	w1, w2, #0xff		/* w1 = DLL */
+	lsr	w2, w2, #8
+	and	w2, w2, #0xff		/* w2 = DLLM */
+	ldr	w3, [x0, #UARTLCR]
+	orr	w3, w3, #UARTLCR_DLAB
+	str	w3, [x0, #UARTLCR]	/* enable DLL, DLLM programming */
+	str	w1, [x0, #UARTDLL]	/* program DLL */
+	str	w2, [x0, #UARTDLLM]	/* program DLLM */
+	mov	w2, #~UARTLCR_DLAB
+	and	w3, w3, w2
+	str	w3, [x0, #UARTLCR]	/* disable DLL, DLLM programming */
+
+	/* 8n1 */
+	mov	w3, #3
+	str	w3, [x0, #UARTLCR]
+	/* no interrupt */
+	mov	w3, #0
+	str	w3, [x0, #UARTIER]
+	/* enable fifo, DMA */
+	mov	w3, #(UARTFCR_FIFOEN | UARTFCR_DMAEN)
+	str	w3, [x0, #UARTFCR]
+	/* DTR + RTS */
+	mov	w3, #3
+	str	w3, [x0, #UARTMCR]
+	mov	w0, #1
+init_fail:
+	ret
+endfunc console_core_init
+
+	/* ---------------------------------------------
+	 * int console_putc(int c)
+	 * Function to output a character over the
+	 * console. It returns the character printed on
+	 * success or -1 on error.
+	 * In : x0 - character to be printed
+	 * Out : return -1 on error else return character.
+	 * Clobber list : x1, x2
+	 * ---------------------------------------------
+	 */
+func console_putc
+	adrp	x2, console_base
+	ldr	x1, [x2, :lo12:console_base]
+	b	console_core_putc
+endfunc console_putc
+
+	/* --------------------------------------------------------
+	 * int console_core_putc(int c, unsigned int base_addr)
+	 * Function to output a character over the console. It
+	 * returns the character printed on success or -1 on error.
+	 * In : w0 - character to be printed
+	 *      x1 - console base address
+	 * Out : return -1 on error else return character.
+	 * Clobber list : x2
+	 * --------------------------------------------------------
+	 */
+func console_core_putc
+	/* Check the input parameter */
+	cbz	x1, putc_error
+
+	/* Prepend '\r' to '\n' */
+	cmp	w0, #0xA
+	b.ne	2f
+	/* Check if the transmit FIFO is full */
+1:	ldr	w2, [x1, #UARTLSR]
+	and	w2, w2, #(UARTLSR_TEMT | UARTLSR_THRE)
+	cmp	w2, #(UARTLSR_TEMT | UARTLSR_THRE)
+	b.ne	1b
+	mov	w2, #0xD		/* '\r' */
+	str	w2, [x1, #UARTTX]
+	ldr	w2, [x1, #UARTFCR]
+	orr	w2, w2, #UARTFCR_TXCLR
+	str	w2, [x1, #UARTFCR]
+
+	/* Check if the transmit FIFO is full */
+2:	ldr	w2, [x1, #UARTLSR]
+	and	w2, w2, #(UARTLSR_TEMT | UARTLSR_THRE)
+	cmp	w2, #(UARTLSR_TEMT | UARTLSR_THRE)
+	b.ne	2b
+	str	w0, [x1, #UARTTX]
+	ldr	w2, [x1, #UARTFCR]
+	orr	w2, w2, #UARTFCR_TXCLR
+	str	w2, [x1, #UARTFCR]
+	ret
+putc_error:
+	mov	w0, #-1
+	ret
+endfunc console_core_putc
+
+	/* ---------------------------------------------
+	 * int console_getc(void)
+	 * Function to get a character from the console.
+	 * It returns the character grabbed on success
+	 * or -1 on error.
+	 * Clobber list : x0, x1, x2
+	 * ---------------------------------------------
+	 */
+func console_getc
+	adrp	x0, console_base
+	ldr	x1, [x0, :lo12:console_base]
+	cbz	x1, getc_error
+
+	/* Check if the receive FIFO is empty */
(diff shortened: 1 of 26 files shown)


danh-arm:

vwadekar:
PR to fix comments from https://github.com/ARM-software/arm-trusted-firmware/pull/302

vwadekar:
Dan, could you please take a look?

achingupta:
Hi Varun,

Dan's on holiday this week. I will review your latest pull request and get back latest by early next week. We are also in the middle of implementing Phase 1 of tf-issues#304. We will upstream the Nvidia port once that is done. I hope that sounds reasonable to you?

cheers,
Achin

vwadekar:
Thanks Achin. Do you have an ETA for the phase1 completion? If it's going
to take too much time (>1 week), I would request you to help upstream the
NV port and I will get it compatible with the new code later. There are
some consumers who would benefit from this, hence the push.

achingupta:
Hi Varun,

I expect phase 1 to complete latest by the end of next week. Does that work for you? 

cheers,
Achin

plat/nvidia/tegra/common/tegra_bl31_setup.c
```diff
@@ -0,0 +1,228 @@
...
+			ro_size,
+			MT_MEMORY | MT_RO | MT_SECURE);
+#if USE_COHERENT_MEM
+	mmap_add_region(coh_start, coh_start,
+			coh_size,
+			MT_DEVICE | MT_RW | MT_SECURE);
+#endif
+
+	/* add MMIO space */
+	plat_mmio_map = plat_get_mmio_map();
```
achingupta: There is a case for the `arm_configure_mmu_elX()` function to take the platform mmap as an argument instead of retrieving it directly inside the function. This way you could have simply re-used that function instead of having to split it to accomodate `plat_mmio_map` instead of `plat_arm_mmap`. Maybe in a separate patch.

achingupta:
Hi Varun, Could you please double check that all the headers are included in an alphabetical order e.g. in plat/nvidia/tegra/common/tegra_pm.c. Thanks a lot!

include/lib/aarch64/cortex_a_common.h
```diff
@@ -0,0 +1,46 @@
...
+#define ACTLR_ENB_L2CTLR_ACCESS		(1 << 4)
+#define ACTLR_ENB_CPUECTLR_ACCESS	(1 << 1)
+#define ACTLR_ENB_CPUACTLR_ACCESS	(1 << 0)
+#define ACTLR_ACCESS_CTRL		(ACTLR_ENB_L2ACTLR_ACCESS | \
+					 ACTLR_ENB_L2ECTLR_ACCESS | \
+					 ACTLR_ENB_L2CTLR_ACCESS | \
+					 ACTLR_ENB_CPUECTLR_ACCESS | \
+					 ACTLR_ENB_CPUACTLR_ACCESS)
+
+#endif /* __CORTEX_A_COMMON_H__ */
```
achingupta: I guess you meant to place this file in include/lib/cpus/aarch64. A couple of nits:

1.  The prefix `ACTLR_` seems too generic. ACTLR_EL1 does not match ACTLR_EL3 and ACTLR_EL2. Furthermore, ACTLR_EL2 should not need to be set from EL3. It can be done in EL2. So I suggest we rename the prefix to `ACTLR_EL3_`

2.  The constant names should not include possible operations on the bit i.e. enable as indicated by "ENB". This should be defined by macro which uses these bits. So we should have:

```
#define ACTLR_L2ACTLR_BIT 
.
.
.
#define ACTLR_EL3_ENABLE_ALL_ACCESS    <blah>
```

drivers/ti/uart/16550_console.S
```diff
@@ -0,0 +1,204 @@
+/*
```
achingupta: Can you not reuse the top level console API from `drivers/console/console.S` and implement the core functions here i.e. similar to the split between `console.S` and `drivers/arm/pl011/pl011_console.S`?

plat/nvidia/tegra/common/tegra_gic.c
```diff
@@ -0,0 +1,95 @@
...
+	gicc_write_ctlr(gicc_base, ENABLE_GRP0 | ENABLE_GRP1);
+	gicc_write_pmr(gicc_base, GIC_PRI_MASK);
+}
+
+/*******************************************************************************
+ * Global gic distributor setup which will be done by the primary cpu after a
+ * cold boot. It marks out the non secure SPIs, PPIs & SGIs and enables them.
+ * It then enables the secure GIC distributor interface.
+ ******************************************************************************/
+static void gic_distif_setup(unsigned int gicd_base)
```
achingupta: Need a clarification here! All interrupts are being configured as non-secure. Is the secure programming done later elsewhere e.g. in TLK?

plat/nvidia/tegra/common/tegra_gic.c
```diff
@@ -0,0 +1,95 @@
...
+	gicc_write_ctlr(TEGRA_GICC_BASE, val);
+}
+
+/*******************************************************************************
+ * Enable secure interrupts and set the priority mask register to allow all
+ * interrupts to trickle in.
+ ******************************************************************************/
+static void gic_cpuif_setup(unsigned int gicc_base)
+{
+	gicc_write_ctlr(gicc_base, ENABLE_GRP0 | ENABLE_GRP1);
```
achingupta: Is there no need to set the FIQEn bit?

plat/nvidia/tegra/include/tegra_private.h
```diff
@@ -0,0 +1,73 @@
...
+
+/* Declarations for tegra_gic.c */
+void gic_setup(void);
+void gic_cpuif_deactivate(void);
+
+/* Declarations for tegra_security.c */
+void tegra_security_setup(void);
+void tegra_security_setup_videomem(uintptr_t base, uint64_t size);
+
+/* Declarations for tegra_memory.c */
```
achingupta: Dangling declaration?

plat/nvidia/tegra/common/aarch64/tegra_helpers.S
```diff
@@ -0,0 +1,344 @@
...
+	 */
+.macro	cpu_init_common
+
+	/* -------------------------------------------------------
+	 * Enable L2 and CPU ECTLR RW access from non-secure world
+	 * -------------------------------------------------------
+	*/
+	mov	x0, #ACTLR_ACCESS_CTRL
+	msr	actlr_el3, x0
+	msr	actlr_el2, x0
```
achingupta: Asking out of curiosity, why can the ACTLR_EL2 not be programmed in EL2?

plat/nvidia/tegra/common/aarch64/tegra_helpers.S
```diff
@@ -0,0 +1,344 @@
...
+	mov	x3, #3
+	movz	x4, #0x8000, lsl #48
+	msr	CPUACTLR_EL1, x4	/* turn off RCG */
+	isb
+	msr	rmr_el3, x3		/* request warm reset */
+	isb
+	dsb	sy
+	wfi
+
+	/* --------------------------------------------------
```
achingupta: I might being thick here but a `wfi` would result in the clocks being gated. You also have an `ISB` and `DSB SY` to meet ordering requirements. So how can instructions be speculatively executed before the warm reset takes place. Would that not defeat the purpose of the `wfi`

plat/nvidia/tegra/common/aarch64/tegra_helpers.S
```diff
@@ -0,0 +1,344 @@
...
+	mov	x0, xzr
+	msr	oslar_el1, x0		/* os lock stays 0 across warm reset */
+	mov	x3, #3
+	movz	x4, #0x8000, lsl #48
+	msr	CPUACTLR_EL1, x4	/* turn off RCG */
+	isb
+	msr	rmr_el3, x3		/* request warm reset */
+	isb
+	dsb	sy
+	wfi
```
achingupta: You might be better off placing the `wfi` in a loop in case a spurious imp. def. wakeup wreaks havoc.

plat/nvidia/tegra/common/aarch64/tegra_helpers.S
```diff
@@ -0,0 +1,344 @@
...
+func tegra_secure_entrypoint
+
+#if ERRATA_TEGRA_INVALIDATE_BTB_AT_BOOT
+
+	/* -------------------------------------------------------
+	 * Invalidate BTB along with I$ to remove any stale
+	 * entries from the branch predictor array.
+	 * -------------------------------------------------------
+	 */
+	mrs	x0, CPUACTLR_EL1
```
achingupta: Further to Dan's comments in #302, I would like to understand this problem better. This is simply 'cause you have a CPU or possibly platform errata which our CPU framework does not scale well enough to apply a fix. So my questions are :

1.  Is this an A57 and A53 errata? I expect it to be a CPU errata since the platform will only assert the appropriate reset pins (warm or cold). Invalidation of the RAMs will have to be done within the CPU.

2.  Junk in I$ and D$ should not be a problem. The architecture allows a software sequence to perform the neccessary invalidations. Are you worried about junk in the BTB and the fact that they need to be invalidated before the first branch instruction is encountered?

plat/nvidia/tegra/common/aarch64/tegra_helpers.S
```diff
@@ -0,0 +1,344 @@
...
+	 * Extract OSLK bit and check if it is '1'. This
+	 * bit remains '0' for A53 on warm-resets. If '1',
+	 * turn off regional clock gating and request warm
+	 * reset.
+	 * -----------------------------------------------
+	 */
+	mrs	x0, oslsr_el1
+	and	x0, x0, #2
+	mrs	x1, mpidr_el1
+	bics	xzr, x0, x1, lsr #7	/* 0 = slow cluster or warm reset */
```
achingupta: Further to your explanation in #302: As per the ARMv8 ARM `OSLSR_EL1.OSLK` is set to 1 after a cold reset. It seems you are setting it to 0 prior to a warm reset. Is the logic the other way around?

plat/nvidia/tegra/common/tegra_gic.c
```diff
@@ -0,0 +1,95 @@
...
+	num_ints = gicd_read_typer(gicd_base) & IT_LINES_NO_MASK;
+	num_ints++;
+	for (ctr = 0; ctr < num_ints; ctr++)
+		gicd_write_igroupr(gicd_base, ctr << IGROUPR_SHIFT, ~0);
+
+	/* enable distributor */
+	gicd_write_ctlr(gicd_base, ENABLE_GRP0 | ENABLE_GRP1);
+}
+
+void gic_setup(void)
```
achingupta: Could you prefix these functions with "tegra_" to avoid namespace pollution. That would be inline with the "arm_" convention we follow.

plat/nvidia/tegra/common/tegra_gic.c
```diff
@@ -0,0 +1,95 @@
+/*
```
achingupta: AFAICS, you are using the raw GIC accessors in `gic_v2.c` which take the base address as an argument. This allows you to bypass the ARM GIC driver. However, in `runtime_exceptions.S` we have calls to some platform provided GIC functions e.g. plat_ic_get_pending_interrupt_type(). The weak definitions of these functions in `plat/common/plat_gic.c` fall into the corresponding functions in the ARM GIC driver e.g. `arm_gic_get_pending_interrupt_type()`. Since the ARM GIC driver is not initialised on the TEGRA port, the base addresses in this routines will be NULL. This will lead to trouble on both the release and debug builds of BL3_1 when interrupts are routed to EL3. 

Are you not seeing issues with the TLKD or am I missing something?

plat/nvidia/tegra/common/tegra_pm.c
```diff
@@ -0,0 +1,348 @@
...
+
+	/*
+	 * Reset hardware settings.
+	 */
+	tegra_prepare_cpu_on_finish(read_mpidr());
+
+	/*
+	 * EL2 is unused for now, so help the context mgmt library setup
+	 * EL2 for us. It uses the SCR_HCE_BIT to decide if we want to
+	 * use EL2 mode at all. Disabling the bit here lets them setup
```
achingupta: Disabling the SCR_EL3.HCE bits will prevent entry into EL2 from a lower exception level. I cannot see how EL2 software can modify things later. Could you please clarify.

plat/nvidia/tegra/common/tegra_pm.c
```diff
@@ -0,0 +1,348 @@
...
+	 * EL2 with defaults, which the actual EL2 software can later
+	 * modify if required.
+	 */
+	ctx = cm_get_context(NON_SECURE);
```
achingupta: Am I correct in understanding that you do not want to use EL2 for the time being? If so, then the correct way of ensuring this would be by setting the EL in the SPSR for BL33 in the `bl33_ep_info` in `bl31_early_platform_setup()` to EL2. All subsequent CPUs will be brought up with the same settings. Is there any reason why this cannot be done?

vwadekar:
Achin, I would request you to upstream this support - once we are ready, at least early next week. If the re-org is done before that, then I will have to rebase anyways. But if not, then lets not wait. This patch is really needed for our consumers as it allows everybody to work from a central place. I am guessing the re-org is going to set me back by at least one week as I will have re-write parts of the NVIDIA port. If we add the review times, then at least 2 weeks after the re-org until the code reaches upstream. It would help if we do this in phases - basic code in Phase 1 and then rework NV support in Phase 2.

drivers/ti/uart/16550_console.S
vwadekar: Sure, will do

include/lib/aarch64/cortex_a_common.h
vwadekar: Done, thanks.

plat/nvidia/tegra/common/tegra_bl31_setup.c
vwadekar: arm_configure_mmu_elX() resides under plat/arm, so is not accessible for plat/nvidia. Dan suggested that I should not rely on the code in plat/arm and should make plat/nvidia independant.

plat/nvidia/tegra/common/tegra_gic.c
vwadekar: TLK or the EL3 software never handles any interrupts. All interrupts are marked as NS and hence passed on to the NS world. TLK does not even need a timer for that matter hence it functions with absolutely no interrupts.

plat/nvidia/tegra/common/tegra_gic.c
vwadekar: Nope

plat/nvidia/tegra/include/tegra_private.h
vwadekar: Thanks for pointing out.

plat/nvidia/tegra/common/aarch64/tegra_helpers.S
vwadekar: I think that makes sense. But this code has been in our downstream repo and has completed weeks of stability cycles so not comfortable removing it. But will make a note of this and discuss with oru team internally. Is this going to be a deal breaker? Or we could merge the patch with this change?

plat/nvidia/tegra/common/aarch64/tegra_helpers.S
vwadekar: Done

plat/nvidia/tegra/common/aarch64/tegra_helpers.S
vwadekar: 1. Yes, this an A57 and A53 errata. We have followed this up with ARM for several weeks and finally come up with this sequence.

2. We have seen that sometimes when the CPU boots up, due to junk in BTB the jumps are invalid and arbitrary. We have seen the CPU end up in ocde paths it shouldnt have. Hence we have to execute this as soon as the CPU boots.

3. We are using warm reset to get to a clean state as that seemed like the safest option to ensure we start with a "clean" slate.

plat/nvidia/tegra/common/tegra_gic.c
vwadekar: Right now TLKD does not expect any interrupts.

plat/nvidia/tegra/common/tegra_pm.c
vwadekar: Thanks. I will try this.

plat/nvidia/tegra/common/aarch64/tegra_helpers.S
vwadekar: True. If the bit is '0', then its a warm reset and we move ahead. If it's '1' we turn off regional clock gating and issue a warm reset.

vwadekar:
#306 is the new push request with fixes.

plat/nvidia/tegra/common/aarch64/tegra_helpers.S
achingupta: It surely is not a deal breaker. The patch can be merged. I just wanted to flag it.

plat/nvidia/tegra/common/aarch64/tegra_helpers.S
achingupta: Thanks Varun. I will follow it up with the teams internally.

plat/nvidia/tegra/common/aarch64/tegra_helpers.S
achingupta: Thanks!

plat/nvidia/tegra/common/tegra_gic.c
achingupta: Thanks. I understand now.

achingupta:
Sure Varun! The phase 1 of tf-issues/#304 is currently under review internally. I do not expect it to be upstream before the middle of next week. It makes sense to allow you to make progress and incorporate the changes introduced by tf-issues/#304 in parallel. I will review your latest pull request today. I would expect it to be upstreamed early next week.

plat/nvidia/tegra/common/tegra_bl31_setup.c
achingupta: Yes! that makes sense. I was evaluating the possibility of making the ARM definition more usable by other platforms. It is not related to this patch set.

vwadekar:
Thanks Achin. Appreciate the help!

plat/nvidia/tegra/common/tegra_gic.c
achingupta: Okie. I am worried about certain secure interrupts e.g. the Secure Physical Timer interrupt being marked as non-secure. Just a nit for future reference.

PR #306, opened 2015-05-29
Tegra soc support v3

vwadekar:
T210 is the latest chip in the Tegra family of SoCs from NVIDIA. It is an
ARM v8 dual-cluster (A57/A53) SoC, with any one of the clusters being active
at a given point in time.

This PR adds support to boot the Trusted Firmware on T210 SoCs. The patch
also adds support to boot secondary CPUs, enter/exit core power states for
all CPUs in the slow/fast clusters. The support to switch between clusters
is still not available in this patch and would be available later.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -212,6 +212,7 @@ INCLUDES		+=	-Iinclude/bl31			\
 				-Iinclude/drivers		\
 				-Iinclude/drivers/arm		\
 				-Iinclude/drivers/io		\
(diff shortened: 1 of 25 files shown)


vwadekar:
Fixed comments from #304

danh-arm:

achingupta:
test this please

vwadekar:
New pull request - #308

PR #311, opened 2015-06-05
Run a Trusted OS during boot on Tegra SoCs

vwadekar:
This patch adds support to run a Trusted OS during boot time. The
previous stage bootloader passes the entry point information in
the 'bl32_ep_info' structure.

The build system expects the dispatcher to be passed as an input
parameter using the 'SPD=<dispatcher>' option. This information
has also been added to the Tegra documentation's 'build' section.

Signed-off-by: Varun Wadekar <vwadekar@nvidia.com>

PR #313, opened 2015-06-09
Boot Trusted OS' on Tegra SoCs

vwadekar:
This patch adds support to run a Trusted OS during boot time. The
previous stage bootloader passes the entry point information in
the 'bl32_ep_info' structure, which is passed over to the SPD.

The build system expects the dispatcher to be passed as an input
parameter using the 'SPD=<dispatcher>' option. The Tegra docs have
also been updated with this information.

### DIFF
--- a/docs/plat/nvidia-tegra.md
+++ b/docs/plat/nvidia-tegra.md
@@ -15,7 +15,16 @@ Directory structure
 * plat/nvidia/tegra/common - Common code for all Tegra SoCs
 * plat/nvidia/tegra/soc/txxx - Chip specific code
 
+Trusted OS dispatcher
+=====================
+Tegra supports multiple Trusted OS', Trusted Little Kernel (TLK) being one of
+them. In order to include the 'tlkd' dispatcher in the image, pass 'SPD=tlkd'
+on the command line while preparing a bl31 image. This allows other Trusted OS
+vendors to use the upstream code and include their dispatchers in the image
+without changing any makefiles.
+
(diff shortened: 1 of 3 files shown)


vwadekar:
Fixed comments from #311

danh-arm:

vwadekar:
Achin/Dan?

achingupta:
Hi Varun. 

It might make sense to move the `tlk-dispatcher.md` to `plat\tegra`? Also, you might want to update the following text from this file. There is now support for the Tegra platform and the command line specified here will still not build TLKD. Let me know if you want to do this in a separate PR.

```
TLK dispatcher adds support for NVIDIA's Trusted Little Kernel (TLK) to work with the Trusted Firmware. TLK-D can be compiled by including it in the platform's makefile. TLK is primarily meant to work with Tegra SoCs, so until Trusted Firmware starts supporting Tegra, the dispatcher code can only be compiled for other platforms.

<snip>

Once a BL32 is ready, TLKD can be compiled using the following command:

CROSS_COMPILE=/bin/aarch64-none-elf- make NEED_BL1=0 NEED_BL2=0 BL32= PLAT= all _
```
Thanks,
Achin

achingupta:
test this please

vwadekar:
It make sense to update the TLKD documentation. As part of #302, Dan suggested that we should move the TLKd docs to docs/spd instead. Frankly, docs/spd seems like a better location. I can make the change and push the TLKD doc update as part of this commit.

vwadekar:
New PR - #317

PR #315, opened 2015-06-10
Authentication Framework

jcastillo-arm:


achingupta:
@jenswi-linaro , @jbech-linaro .

Hi Both,

We have been working on replacing the prototype implementation of Trusted Board Boot (TBB) with a scalable framework that:

1. Supports different Chains of Trust (CoT).
2. Supports use of different cryptographic algorithms for signature and hash verification.
3. Supports use of proprietary image formats for carrying authentication information i.e. an x509v3 certificate is just one of the formats instead of being "the" format.

The result of that work is contained in this pull request. The earlier design had a monolithic authentication module that incorporated:

1. An understanding of the CoT. Furthermore, only one type of CoT was supported.
2. The implementation of the cryptographic libraries and certificate parsing logic.

The present design takes a modular approach as follows:

1. The CoT is represented as a data structure. This structure specifies the images involved in the CoT, their dependencies, how they should loaded, parsed and authenticated.
2. The cryptography and image parsing libraries are implemented under separate modules.
3.  The authentication module uses the CoT description, the cryptography and image parsing modules to perform TBB through well defined interfaces. 

The last commit adds some rudimentary documentation (subject to change as it is by no means final) to help in understanding the changes better. Please let us know if you would like to provide some feedback. 

cheers,
Achin

jbech-linaro:
@achingupta , we will try to find some time to review this.

achingupta:
@jbech-linaro , Thanks. We will merge this PR today in any case as its been around for a while and some internal development depends upon it. Your feedback is still most welcome. Any changes as a result would need to go out as patches on top of this PR.

jbech-linaro:
@achingupta , I see, sorry for being late. Anyhow, I'll have a look at it and I'll provide feedback either as patch suggestions on top of this PR or in another way. (I just added one tiny comment, but I won't add more comments to this pull request since I know you are about to merge it).

achingupta:
@jbech-linaro , Thanks again. The documentation in this PR will definitely undergo changes later. We will incorporate the above feedback when that is done.

PR #316, opened 2015-06-11
Reserve a Video Memory aperture in DRAM memory

vwadekar:
This patch adds support to reserve a memory carveout region in the
DRAM on Tegra SoCs. The memory controller provides specific registers
to specify the aperture's base and size. This aperture can also be
changed dynamically in order to re-size the memory available for
DRM video playback. In case of the new aperture not overlapping
the previous one, the previous aperture has to be cleared before
setting up the new one. This means we do not "leak" any video data
to the NS world.


danh-arm:

vwadekar:
New PR - #318

PR #318, opened 2015-06-12
Reserve a Video Memory aperture in DRAM memory

vwadekar:
This patch adds support to reserve a memory carveout region in the
DRAM on Tegra SoCs. The memory controller provides specific registers
to specify the aperture's base and size. This aperture can also be
changed dynamically in order to re-size the memory available for
DRM video playback. In case of the new aperture not overlapping
the previous one, the previous aperture has to be cleared before
setting up the new one. This means we do not "leak" any video data
to the NS world.

### DIFF
--- a/plat/nvidia/tegra/common/tegra_bl31_setup.c
+++ b/plat/nvidia/tegra/common/tegra_bl31_setup.c
@@ -37,6 +37,7 @@
 #include <cortex_a57.h>
 #include <cortex_a53.h>
 #include <debug.h>
+#include <errno.h>
 #include <memctrl.h>
 #include <mmio.h>
 #include <platform.h>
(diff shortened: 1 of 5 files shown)


vwadekar:
Fixed PR #316

danh-arm:

plat/nvidia/tegra/common/tegra_bl31_setup.c
```diff
@@ -226,3 +227,35 @@ void bl31_plat_arch_setup(void)
 	/* enable the MMU */
 	enable_mmu_el3(0);
 }
+
+/*******************************************************************************
+ * Check if the given NS DRAM range is valid
+ ******************************************************************************/
+int bl31_check_ns_address(uint64_t base, uint64_t size_in_bytes)
+{
+	uint64_t end = base + size_in_bytes;
```
athoelke: If size_in_bytes is close to the maximum UINT64 value then this calculation will wrap with the result of end < base. This will break various of the bounds checks. The "base == end" check in line 242 should check for (base >= end) to also detect this invalid parameter.

plat/nvidia/tegra/common/tegra_bl31_setup.c
```diff
@@ -226,3 +227,35 @@ void bl31_plat_arch_setup(void)
...
+	}
+
+	/*
+	 * TZDRAM aperture contains the BL31 and BL32 images, so we need
+	 * to check if the NS DRAM range overlaps the TZDRAM aperture.
+	 */
+	if (((base >= tegra_bl31_phys_base) && (end <= TZDRAM_END)) ||
+	    ((base <= tegra_bl31_phys_base) && (end >= TZDRAM_END)) ||
+	    ((base >= tegra_bl31_phys_base) && (base <= TZDRAM_END)) ||
+	    ((end >= tegra_bl31_phys_base) && (end <= TZDRAM_END))) {
```
athoelke: This condition is more complex than required. An overlap is not present if:

    `(base >= TZDRAM_END) || (end <= tegra_bl31_phys_base)`

either the region is entirely after the TZDRAM aperture or entirely before it. In any other condition there must be an overlap. So the overlap condition could be written as

    `((base < TZDRAM_END) && (end > tegra_l31_phys_base))`

plat/nvidia/tegra/common/tegra_sip_calls.c
```diff
@@ -0,0 +1,105 @@
...
+		if (err)
+			SMC_RET1(handle, err);
+
+		/*
+		 * Check if Video Memory is aligned to 1MB.
+		 */
+		if (x1 & 0xFFFFF) {
+			ERROR("Unaligned Video Memory base address!\n");
+			SMC_RET1(handle, -ENOTSUP);
+		}
```
athoelke: There is no similar check for `x2`, the size of the region. Later this is stored in MB (>> 20) which will round down. So the stored region size might be smaller than that requested. If alignment is required for this parameter then it should be checked as well. Otherwise the parameter should be rounded before use in the bounds checking and reservation functions.

plat/nvidia/tegra/include/tegra_private.h
```diff
@@ -34,6 +34,12 @@
 #include <xlat_tables.h>
 #include <platform_def.h>
 
+/*******************************************************************************
+ * Tegra DRAM memory base address
+ ******************************************************************************/
+#define TEGRA_DRAM_BASE		0x80000000
+#define TEGRA_DRAM_END		0x27FFFFFFF
```
athoelke: This is defining the end as the last byte - in the region bounds checking code the region end is calculated as the base plus the size - i.e. the first byte after the region. This inconsistency might result in off-by-one boundary calculations.

vwadekar:
Fixed with #319

PR #320, opened 2015-06-17
Add delay timer API v10

danh-arm:
Add a simple generic delay timer driver API, an SP804 dual timer instantiation of this API, and initialize the driver in the FVP BSP.


omnium21:
Thanks for doing this for me Dan.  It looks good to me.

achingupta:
Is this driver expected to be used only as a BL2 driver? SP804 is an untrusted peripheral. AFAIK, Linux uses it as the always on timer. This PR configures it as a free running counter. There will be a conflict if BL3-1 tries to use it as a part of runtime services? Ditto for any normal world image e.g. UEFI runtime. Also, if somehow only BL3-1 were to use it, then would it be safe for the secure world to rely on this timer to correctly keep track of time?

Is there a reason why the Secure frame in the system level implementation of the generic timer or the Generic counter itself cannot be used for the same purpose? I have not followed the earlier discussions so please let me know.

danh-arm:
Achin: Yes, the SP804 delay timer driver should only be used by secure world boot code (BL1 and BL2) as it's later used by Linux (and maybe UEFI). I'll update the commit message to make this clear.

You're also right that it would be good to implement a generic timer version of this driver and get the platform port to use this instead. However, I didn't want to hold up this PR any further and there would be sharing issues with that driver as well (e.g. both BL3-1 and BL3-2 might want to use the secure frame of the generic timer).

PR #321, opened 2015-06-19
Add missing features to the Tegra GIC driver

vwadekar:
In order to handle secure/non-secure interrupts, overload the plat_ic_*
functions and copy GIC helper functions from arm_gic.c. Use arm_gic.c
as the reference to add Tegra's GIC helper functions.

Now that Tegra has its own GIC implementation, we have no use for
plat_gic.c and arm_gic.c files.

### DIFF
--- a/plat/nvidia/tegra/common/tegra_common.mk
+++ b/plat/nvidia/tegra/common/tegra_common.mk
@@ -43,14 +43,12 @@ PLAT_BL_COMMON_SOURCES	:=	lib/aarch64/xlat_tables.c			\
 
 COMMON_DIR		:=	plat/nvidia/tegra/common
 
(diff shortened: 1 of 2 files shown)


danh-arm:

vwadekar:
Dan?

danh-arm:
Note for 3rd parties: The ARM GIC driver in `drivers/arm/gic/` contains support for initialization of GICv3 register state in mixed GICv2/GICv3 systems. In GICv2-only systems (like Tegra), some of this code is unnecessary and potentially unwanted. Future enhancements to the ARM GIC driver will provide more optimised support for GICv2 only systems, which will potentially allow re-alignment between the ARM and Tegra GIC drivers.

vwadekar:
Thanks Dan.

PR #324, opened 2015-06-22
PSCI: Add SYSTEM_SUSPEND API support

soby-mathew:
This patch adds support for SYSTEM_SUSPEND API as mentioned in the PSCI 1.0
specification. This API, on being invoked on the last running core on a
supported platform, will put the system into a low power mode with memory
retention.

The psci_afflvl_suspend() internal API has been reused as most of the actions
to suspend a system are the same as invoking the PSCI CPU_SUSPEND API with the
target affinity level as 'system'. This API needs the 'power state' parameter
for the target low power state. This parameter is not passed by the caller of
the SYSTEM_SUSPEND API. Hence, the platform needs to implement the
get_sys_suspend_power_state() platform function to provide this information.
Also, the platform also needs to add support for suspending the system to the
existing 'plat_pm_ops' functions: affinst_suspend() and
affinst_suspend_finish().

Change-Id: Ib6bf10809cb4e9b92f463755608889aedd83cef5

### DIFF
--- a/docs/firmware-design.md
+++ b/docs/firmware-design.md
@@ -759,7 +759,7 @@ required support.
 |`CPU_FREEZE`           | No      |                                           |
 |`CPU_DEFAULT_SUSPEND`  | No      |                                           |
 |`CPU_HW_STATE`         | No      |                                           |
-|`SYSTEM_SUSPEND`       | No      |                                           |
+|`SYSTEM_SUSPEND`       | Yes*    |                                           |
 |`PSCI_SET_SUSPEND_MODE`| No      |                                           |
 |`PSCI_STAT_RESIDENCY`  | No      |                                           |
 |`PSCI_STAT_COUNT`      | No      |                                           |
--- a/docs/porting-guide.md
+++ b/docs/porting-guide.md
@@ -1211,8 +1211,8 @@ affinity level 0 (CPU), the platform port should power down affinity level 1
 #### plat_pm_ops.affinst_suspend()
 
 Perform the platform specific setup to power off an affinity instance of the
-calling CPU. It is called by the PSCI `CPU_SUSPEND` API
-implementation.
+calling CPU. It is called by the PSCI `CPU_SUSPEND` API and `SYSTEM_SUSPEND`
+API implementation
 
 The `affinity level` (second argument) and `state` (third argument) have a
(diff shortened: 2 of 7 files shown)


vwadekar:
What's the plan? When are we planning to merge this change?

danh-arm:
retest this please

danh-arm:
Sorry about the delay on this. Testing of this feature on ARM development platforms is currently limited, and we wanted to give enough opportunity for others to test and provide feedback.

vwadekar:
Thanks Dan.  Will push a patch for Tegra support.

vwadekar:
Dan, could you please merge back into master branch? Thanks.

danh-arm:
There's no need to remind me, Varun. It will happen in due course!

vwadekar:
In general could you summarise the policy for merges from integration to
master? Unless the changes are in master, they don't reach a lot of people.

People do not always follow the integration branch. So no point in keeping
changes pending in the integration branch. If there's a general rule, I
would let our consumers know that they have to wait for X more days even
though the changes have been merged.

danh-arm:
It depends on how many concurrent Pull Requests (PRs) are being processed at the same time. In simple cases like this where all potential regressions have already been tested, the delay will be < 1 day. If we're trying to merge several things over the course of a few days, it might take up to a week. Typically, I'd say 1-2 days.

The worst case would be if we're trying to make a release while also receiving PRs that will not be merged into the release. In this case, we would merge the PRs onto integration, then rebase integration following the release, then catch up master a day or two later. This whole process could take up 2 weeks. We would inform the PR owner if this were going to happen.

It is OK to generate a pull request based on content that is only available on integration or in another PR. For example, you could have generated your PR for SYSTEM_SUSPEND support in the NVidia port before this PR was merged. There is a risk that the dependent PR will change or that integration will be rewound if there are integration problems. This would require the PR to be updated but that should be a manageable process.

Regards

Dan.

vwadekar:
Thanks. So as a rule of thumb, I would ask for a tentative ETA, for the
merge back to master, for all future PRs to avoid any surprises.

danh-arm:
You can ask, but I can't promise we'll always be so responsive!

vwadekar:
I dont mind pinging till I get some response. Thanks for all the help.

PR #331, opened 2015-07-09
Tegra misc fixes v1

vwadekar:
- Deassert CPU reset signals during power on
- Fix the delay loop used during SC7 exit
- Exclude coherent memory region from memory map


danh-arm:

vwadekar:
New pull request - #333

PR #338, opened 2015-07-19
Handling FIQ in EL3

virtualopensystems-plalov:
The following series implements secure interrupt handling to catch
all FIQs in EL3 regardless the current exception layer which is executing.
The handling in EL3 helps to target some of the interrupts to EL3,
while the rest are routed in S-EL1.
The routing to S-EL1 is feasible, since the EL3 is responsible to acknowledge
secure interrupts and to pass the interrupt ID to S-EL1. Then S-EL1 is
responsible to process the end of interrupt.

### DIFF
--- a/acknowledgements.md
+++ b/acknowledgements.md
@@ -7,5 +7,7 @@ Linaro Limited
 
 NVIDIA Corporation
 
+Virtual Open Systems
+
 Individuals
 -----------
--- a/bl31/interrupt_mgmt.c
+++ b/bl31/interrupt_mgmt.c
@@ -72,10 +72,8 @@ static intr_type_desc_t intr_type_descs[MAX_INTR_TYPES];
  ******************************************************************************/
 static int32_t validate_interrupt_type(uint32_t type)
 {
-	if (type == INTR_TYPE_EL3)
-		return -ENOTSUP;
-
-	if (type != INTR_TYPE_S_EL1 && type != INTR_TYPE_NS)
+	if (type != INTR_TYPE_EL3 && type != INTR_TYPE_S_EL1
+		&& type != INTR_TYPE_NS)
 		return -EINVAL;
 
 	return 0;
@@ -88,6 +86,9 @@ static int32_t validate_routing_model(uint32_t type, uint32_t flags)
 {
 	flags >>= INTR_RM_FLAGS_SHIFT;
 	flags &= INTR_RM_FLAGS_MASK;
+
+	if (type == INTR_TYPE_EL3)
+		return validate_el3_interrupt_rm(flags);
 
 	if (type == INTR_TYPE_S_EL1)
 		return validate_sel1_interrupt_rm(flags);
--- a/bl32/tsp/aarch64/tsp_entrypoint.S
+++ b/bl32/tsp/aarch64/tsp_entrypoint.S
@@ -353,7 +353,7 @@ func	tsp_fiq_entry
 	 * fiq handler is an error.
 	 * ---------------------------------------------
 	 */
-	save_eret_context x2 x3
+	save_eret_context x3 x4
 	bl	tsp_update_sync_fiq_stats
 	bl	tsp_fiq_handler
 	cbnz	x0, tsp_fiq_entry_panic
--- a/bl32/tsp/aarch64/tsp_exceptions.S
+++ b/bl32/tsp/aarch64/tsp_exceptions.S
@@ -32,7 +32,7 @@
 #include <arch.h>
 #include <tsp.h>
 #include <asm_macros.S>
-
+#include "../tsp_private.h"
 
 	/* ----------------------------------------------------
 	 * The caller-saved registers x0-x18 and LR are saved
@@ -139,6 +139,8 @@ fiq_sp_elx:
 	msr	daifclr, #DAIF_ABT_BIT
 
 	save_caller_regs_and_lr
+	/* Tell 'tsp_fiq_handler' that FIQ is caught in S-EL3 */
+	mov x0, #TSP_SEL1_EXCEPTION
 	bl	tsp_fiq_handler
 	cbz	x0, fiq_sp_elx_done
 
--- a/bl32/tsp/tsp_interrupt.c
+++ b/bl32/tsp/tsp_interrupt.c
@@ -47,7 +47,9 @@
  * The 'elr_el3' parameter contains the address of the instruction in normal
  * world where this FIQ was generated.
  ******************************************************************************/
-void tsp_update_sync_fiq_stats(uint32_t type, uint64_t elr_el3)
+const uint32_t tsp_update_sync_fiq_stats(uint32_t type,
+		uint64_t elr_el3,
+		const uint32_t interrupt_id)
 {
 	uint64_t mpidr = read_mpidr();
 	uint32_t linear_id = platform_get_core_pos(mpidr);
@@ -66,6 +68,7 @@ void tsp_update_sync_fiq_stats(uint32_t type, uint64_t elr_el3)
 		tsp_stats[linear_id].sync_fiq_ret_count);
 	spin_unlock(&console_lock);
 #endif
+	return interrupt_id;
 }
 
 /*******************************************************************************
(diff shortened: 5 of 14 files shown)


danh-arm:

danh-arm:
It's not clear to me what you want us to do with this PR. You seemed to acknowledge in https://github.com/ARM-software/tf-issues/issues/313 that you can achieve your use-case using the existing framework. We don't want to make fundamental changes to the framework without resolving the race conditions Achin mentions and we're not in a position to mandate an EL3 <-> S-EL1 interface that meets everybody's needs.

kchappuis-virtualopensystems:
The use case we want to address is to access Secure Timer from EL3 only. Currently in ATF, with FIQs caught in S-EL1, the secure payload can't read/write the Secure Timer register when Secure Timer interrupts occur.

For this reason, we have proposed a patch to extend ATF to always trap FIQs in EL3, where secure monitor will decide to redirect or not the FIQ in S-EL1, according to the FIQ ID.

We understand from Achin's comment, that GICV3 would solve the same problem by setting:
    GROUP0 => FIQ in EL3
    GROUP1-S => Interrupt in Secure world
    GROUP1-NS => Interrupt in Normal world

However, some available platforms (e.g., ARM JUNO board) don't support GICV3 although they provide a Secure Timer. That's why we propose to share Virtual Open Systems solution proposal with the ATF community, since other developers might have similar requirements.

In the meantime, we are actively working on this implementation to further improve it.

athoelke:
Ok - the detail for the use case you have is appreciated. The secure timer does provide a reason to try and solve the only-one-secure-interrupt-group-in-GICv2 challenge. That does not make solutions easy or generally usable.

There is a control bit in the `SCR_EL3` register, `ST` (bit 11) which enables AArch64 Secure-EL1 to access the secure physical timer registers directly without trapping to EL3. So for TEEs that execute in AArch64, this is the preferred way to provide access to the timer if EL3 does not use the timer itself.

If Secure-EL1 is AArch32, there is no system register that maps to the secure physical timer, and the `CNTP_CTL` AArch32 register maps to the non-secure `CNTP_CTL_EL1` register. For such a system, there are several design approaches (none of them as neat as the AArch64 approach):

1. Save and restore the physical timer state when switching between secure and non-secure worlds. The Trusted OS should continue use `CNTP_*` registers to manage the timer, but the interrupt ID will be different; secure interrupt handling remains in Secure-EL1. Although this is what was envisaged by the ARMv8-A designers, leaving the secure physical timer for use by EL3, this does depend on having GICv3 to be a nice solution (proxy the 'other world' timer using the secure timer, and just switch the non-secure interrupt group from 1-NS to 1-S). With GICv2, the timer swapping is more complex - something we have not [yet] tried to implement.

2. Provide a set of 'fast SMC' calls in ARM Trusted Firmware that access the secure physical timer registers, which only allow use from AArch32 Secure-EL1 - either as raw register updates, or a set of timer services. Then modify the Trusted OS to use these SMC calls when handling the timer interrupts instead of using the `CNTP_*` registers directly; secure interrupt handling remains in Secure-EL1.

3. In a GICv2 system, introduce a acknowledged-interrupt handover mechanism between EL3 and Secure-EL1, acknowledging secure interrupts in EL3 and handling them there if "owned" by some EL3 software otherwise transferring the active interrupt to Secure-EL1. This required modifications to the Trusted OS to not acknowledge interrupts locally. In addition, either the Trusted OS must not take secure interrupts locally (always routing them to EL3), or an additional handover mechanism is required to transfer acknowledged interrupts from Secure-EL1 to EL3. The former may be unacceptable for some TEE designs, the latter requires some very careful thinking about the interactions of trap and mask bits in the interrupt handling flows.

Here you have implemented option 3. I think it has some value as a demonstration of how it can be made to work for a given Secure payload and dispatcher (the changes in TSP/TSPD). Also it provides an interesting comparison to the two models that can be built at present which demonstrate receiving Group 1 interrupts either first at Secure-EL1 before forwarding via EL3, or trapping straight to EL3. The generic code changes to the interrupt model are not needed though, as the same effect can be had by TSPD registering for `INTR_TYPE_S_EL1` as it does already.

In the medium term, I expect that Trusted OSes will migrate to AArch64 and we will see direct use of the secure physical timer in the OS. On systems where EL3 is using the timer, such access from the Trusted OS will be trapped to EL3 to provide timer services; or explicit SMC calls could be provided (#2 above); or the timer switching approach during world-switch will be used.

virtualopensystems-kchappuis:
Thanks for the comments.

In our use case EL3 uses exclusively the Secure Timer, in fact, in order to prevent any access to Secure Timer registers from AARCH64 S-EL1, we set SCR_EL3.ST to 0.
Moreover, as EL3 is the highest exception level, we prefer to catch FIQs there and decide, according to the FIQ ID, whether to transmit the interrupt to S-EL1 or not.

Your suggested use case with a S-EL1 AARCH32 application is very interesting and strengthen the interest to catch FIQs in EL3.

In addition, if I understand well, you might expect some modifications in our patch about the interrupt model.
I propose to associate a new fiq handler (tspd_el3_interrupt_handler( ) for example) for the interrupt type INTR_TYPE_EL3, which will not impact to the current implementation of the function tspd_sel1_interrupt_handler( ). As for TSP, we will use ifdef for all the FIQ EL3 handling related instructions.

In conclusion, the goal of this proposal is to minimize the impact of FIQ handling EL3 on the current ATF version while adding the FIQ ID propagation only if the user compiles ATF with the flag TSPD_ROUTE_FIQ_TO_EL3.

danh-arm:
I'm still trying to assess the validity of your use-case. If EL3 has exclusive use of the secure timer, how are you proposing to support Trusted OS that wish to have timer services? Andrew mentions 3 options in his last response. He also mentions that several Trusted OS will find it unacceptable to not handle secure interrupts locally (due to the additional latency in going to EL3 for other secure interrupts). Perhaps in your use-case you do not intend to support a full Trusted OS environment, only a limited S-EL1 application environment?

Just because EL3 is the highest exception level, it doesn't necessarily follow that FIQs should always be handled there first. A bigger consideration is the EL for which secure interrupts should have the lowest latency.

Anyway, if we can be convinced of the validity of your use-case, I can envisage changes to the TSPD/TSP to demonstrate this. However, as Andrew previously stated, this can be achieved without introducing support for `INTR_TYPE_EL3` or a new interrupt handler; you can just continue to allow the TSPD to register for `INTR_TYPE_S_EL1`. Adding support for `INTR_TYPE_EL3` does not minimize the impact on the current TSPD/TSP secure interrupt handling; it just adds to the potential number of code paths. You will need the flag `TSPD_ROUTE_FIQ_TO_EL3` to change the routing behaviour in the TSPD/TSP, though.

virtualopensystems-kchappuis:
In our use-case, EL3 has exclusive use of the Secure timer in order to manage the Normal world and Secure world execution. Our main goal is to improve the system robustness to allow EL3 to execute an action when a specific FIQ occurs. 
We described the Secure Timer use case because EL3 could use exclusively it but we could imagine other use-cases where FIQ will be used in EL3. We think that we are heading in the right way because for example, with GICv3 ARM designed an interrupt controller which allows to define interrupt exclusively for the EL3.
However, as we already mentioned, some available platforms (e.g., ARM JUNO board) don't support GICV3 so we have to propose a software solution to manage FIQs that we want dedicated to EL3 with the current GIC version.

Yes I agree that this software solution, which traps all FIQs in EL3, adds latency because we execute some instructions in EL3 before to propagate the FIQ in S-EL1. However the overhead will not be very important (Save gp registers, EL3 fiq handler => read FIQ ID / checks FIQ propagation or not to S-EL1, Restore gp registers).

Additionally, please consider that in many applications the secure payload is running a legacy OS (32-bits). This is our case and as the Secure Timer cannot be accessed by a AARCH32 S-EL1 application, we propose to use an external timer to support timer services (e.g SP804) for OS and use Secure Timer for EL3.

Finally, I updated the patch according to your technical comments in order to achieve the FIQ EL3 management directly with INTR_TYPE_S_EL1. Please find the link of the new pull request here: https://github.com/ARM-software/arm-trusted-firmware/pull/345

danh-arm:
You are right that GICv3 is designed to allow for interrupts to be exclusively handled at EL3 (as well as others handled at S-EL1). Unfortunately, as this thread has shown, all solutions for trying to do something similar in a GICv2 world are messy, and the "best" solution is dependent on the specfic environment it's implemented in. The problem we've got is that the TSPD/TSP is supposed to be a demonstrator for typical SP/SPDs and we're not convinced your changes are a good demonstrator, even in your specific environment. Of course there's nothing stopping you implementing this in your own SPD/SP.

More specifically, we don't think a typical Trusted OS will find this an acceptable solution, not only because of the additional interrupt latency, but also because the Trusted OS will be unable to mask interrupts in order to implement things like critical sections. Also, we don't think the current solution resolves all the potential race conditions that Achin raises in https://github.com/ARM-software/tf-issues/issues/313#issuecomment-121251906 and it will be difficult to prove that it does.

A simpler and easier-to-verify solution would be to provide an EL3 "Fast SMC" interface to S-EL1 for the services that S-EL1 does not have access to (Andrew's solution 2 above). In the specific case of the secure timer, I can see a useful contribution being provision of an example such interface in the TSPD/TSP. However, that probably depends on porting the TSP to AArch32, which is a non-trivial piece of work (see https://github.com/ARM-software/tf-issues/issues/185).

virtualopensystems-kchappuis:
We understand that the solution to exclusively handle interrupts in EL3 is dependent on the specific environment. However our goal is to minimize the code changes in the Secure Payload application in order to be generic, and therefore we don't want to implement the handler in S-EL1 to return in EL3 according to the ID.

We agree that the Trusted OS will be unable to mask interrupts in order to implement things like critical sections. However we could use a solution where the Secure Payload application disables FIQs by writing the GICD_CTRL register during this section.
Correct me if I'm wrong, but the same issue will be met with GICv3, because the Secure Payload application will work with an IRQ handler and it will not be possible to mask FIQ interrupts as they will be trapped in EL3.

About the AARCH32 TSP application aiming to implement an EL3 "Fast SMC" interface, we have a working 32 bit FreeRTOS port on FastModel/JUNO with ATF (compiled independently and added in the BL32 layer later). So if you are interested we could share it with the community to provide a SP AARCH32 application that could be used to implement the EL3 "Fast SMC".

danh-arm:
> However our goal is to minimize the code changes in the Secure Payload application in order to be generic, and therefore we don't want to implement the handler in S-EL1 to return in EL3 according to the ID.

I think code changes to the SP are inevitable in any of the proposed solutions so it's more a question of which solution is best architecturally.

> However we could use a solution where the Secure Payload application disables FIQs by writing the GICD_CTRL register during this section.

Hmm, splitting the responsibility for GIC management between EL3 and S-EL1 like this also sounds fragile.

> Correct me if I'm wrong, but the same issue will be met with GICv3, because the Secure Payload application will work with an IRQ handler and it will not be possible to mask FIQ interrupts as they will be trapped in EL3.

GICv3 introduces the concept of 3 interrupt groups, which can be independently masked. Group 0 interrupts will (usually) be trapped in EL3 (as FIQs). Secure Group 1 interrupts will (usually) be trapped in Secure-EL1 (as IRQs) while executing in secure state. The latter are the ones that the Trusted OS would need to mask in the critical section situation (yes, this is confusing as they are trapped as FIQs in GICv2). Some Trusted OS also would also mask Non-Secure Group 1 interrupts to prevent the normal world executing during the critical section.

> ... we have a working 32 bit FreeRTOS port on FastModel/JUNO with ATF (compiled independently and added in the BL32 layer later). So if you are interested we could share it with the community to provide a SP AARCH32 application that could be used to implement the EL3 "Fast SMC". 

Sounds interesting. Yes, you could share this but it's probably best to do this in a separate GitHub project as the license may not be compatible.

I'm not sure what you mean by the SP AArch32 app implementing the EL3 "Fast SMC"; these SMCs would be implemented by the EL3 SPD.

virtualopensystems-kchappuis:
With GICv3 how could you independently mask each interrupt group? with a GICv3 register? Because in SPSR register you can only mask FIQ and/or IRQ.
So, are you considering that the trusted OS execution could be preempted by Group 0 interrupts even if the execution of critical sections?

Sure, we understand that the SMC interface has to be implemented by the EL3 SPD. However we haven't implemented this Secure Timer service example on our side (SMC Secure Timer service in EL3 spd + FreeRTOS service call).

Related to the FreeRTOS payload, we will think about it. One option could be to share the patch directly with the FreeRTOS community.

danh-arm:
I'm haven't looked that much into GICv3 yet but yes, there's a register for independently disabling Group 1 interrupts (for the current security state)  - ICC_IGRPEN1_EL1. And yes, I can envisage the Trusted OS on some systems being pre-empted by Group 0 interrupts, but that should be reserved for very specific purposes (e.g. a critical RAS event).

danh-arm:

danh-arm:
Superseded by #345

PR #339, opened 2015-07-21
Tegra denver plat support v1

vwadekar:
Support for Tegra's "Project Denver" CPU platforms

### DIFF
--- a/plat/nvidia/tegra/common/tegra_sip_calls.c
+++ b/plat/nvidia/tegra/common/tegra_sip_calls.c
@@ -38,7 +38,14 @@
 #include <runtime_svc.h>
 #include <tegra_private.h>
 
+#define NS_SWITCH_AARCH32	1
+#define SCR_RW_BITPOS		10
+
+/*******************************************************************************
+ * Tegra SiP SMCs
+ ******************************************************************************/
 #define TEGRA_SIP_NEW_VIDEOMEM_REGION		0x82000003
+#define TEGRA_SIP_AARCH_SWITCH			0x82000004
 
 /*******************************************************************************
  * This function is responsible for handling all SiP calls from the NS world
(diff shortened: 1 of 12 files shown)


danh-arm:

plat/nvidia/tegra/common/tegra_sip_calls.c
```diff
@@ -85,6 +95,24 @@ uint64_t tegra_sip_handler(uint32_t smc_fid,
 
 		SMC_RET1(handle, 0);
 
+	case TEGRA_SIP_AARCH_SWITCH:
```
athoelke: This is a SMC32 call - based on the smc_fid value. Thus you should clean the high bits of the x1/x2 parameters
`x1 = (uint32_t)x1;`
etc. before using them. In the case that a call came from AArch32, it is not guaranteed that the high bits are zero, which this code appears to depend on.

I notice that this also affects the VIDEOMEM_REGION function, but I can't add a comment there in this commit diff.

 I presume you already realise that these two SIP service calls cannot be passed addresses above 4GB as they are 32-bit functions.

plat/nvidia/tegra/common/tegra_sip_calls.c
```diff
@@ -38,7 +38,14 @@
 #include <runtime_svc.h>
 #include <tegra_private.h>
 
+#define NS_SWITCH_AARCH32	1
+#define SCR_RW_BITPOS		10
```
athoelke: You could define this in terms of the SCR_RW_BIT definition by using a GCC builtin function:
`#define SCR_RW_BITPOS __builtin_ctz(SCR_RW_BIT)`
This is the technique used in the GIC drivers to implement the `interrupt_type_to_line()` functionality. (I'm actually wondering if we should have a less obscure macro to do this?

vwadekar:
New PR - #340

PR #342, opened 2015-07-24
tlkd: delete 'NEED_BL32' build variable

vwadekar:
Remove the 'NEED_BL32' flag from the makefile. TLK compiles using a
completely different build system and is present on the device as a
binary blob. The NEED_BL32 flag does not influence the TLK load/boot
sequence at all. Moreover, it expects that TLK binary be present on
the host before we can compile BL31 support for Tegra.

This patch removes the flag from the makefile and thus decouples both
the build systems.

Tested by booting TLK without the NEED_BL32 flag on Tegra.


danh-arm:

danh-arm:
test this please

vwadekar:
Thanks Dan for the fast review.

PR #343, opened 2015-07-27
Tegra210: enable WRAP to INCR burst type conversions

vwadekar:
The Memory Select Switch Controller routes any CPU transactions to
the appropriate slave depending on the transaction address. During
system suspend, it loses all config settings and hence the CPU has
to restore them during resume.

This patch restores the controller's settings for enabling WRAP to
INCR burst type conversions on the master ports, for any incoming
requests from the AXI slave ports.

Tested by performing multiple system suspend cycles.

Signed-off-by: Varun Wadekar <vwadekar@nvidia.com>

PR #344, opened 2015-07-27
Tegra210: enable WRAP to INCR burst type conversions

vwadekar:
The Memory Select Switch Controller routes any CPU transactions to
the appropriate slave depending on the transaction address. During
system suspend, it loses all config settings and hence the CPU has
to restore them during resume.

This patch restores the controller's settings for enabling WRAP to
INCR burst type conversions on the master ports, for any incoming
requests from the AXI slave ports.

Tested by performing multiple system suspend cycles.


vwadekar:
Previous PR - #343

danh-arm:

vwadekar:
Dan?

vwadekar:
Thanks

PR #345, opened 2015-07-29
Add functionality to handle all Secure Interrupts in EL3

virtualopensystems-kchappuis:
The patch adds a new functionality to handle all secure interrupts in EL3. This functionality could be enabled at the compilation time by setting the FLAG TSPD_ROUTE_FIQ_TO_EL3 to 1. The goal of this change is to dedicate part of the secure interrupts to EL3 only. Such interrupt could be the system secure timer for example.

If TSPD_ROUTE_FIQ_TO_EL3 is set to 0, the patch has no impact and the FIQ handling process is directly done in the Secure payload application (S-EL1).

If TSPD_ROUTE_FIQ_TO_EL3 is set to 1:

- In EL3, the handler 'tspd_secure_interrupt_handler', implemented in the 'INTR_TYPE_S_EL1' type, is updated, to manage the case of the secure interrupt occurs during the Secure world execution, to propagate the FIQ ID to the Secure payload by passing the ID as third argument (register: X2).

- In TSP, the handler 'tsp_fiq_handler' is modified to avoid double acknowledge. Therefore the FIQ acknowledgement step is bypassed in S-EL1.

Signed-off-by: Kevin Chappuis <k.chappuis@virtualopensystems.com>
Signed-off-by: Petar Lalov <p.lalov@virtualopensystems.com>

### DIFF
--- a/acknowledgements.md
+++ b/acknowledgements.md
@@ -7,5 +7,7 @@ Linaro Limited
 
 NVIDIA Corporation
 
+Virtual Open Systems
+
 Individuals
 -----------
--- a/bl32/tsp/aarch64/tsp_entrypoint.S
+++ b/bl32/tsp/aarch64/tsp_entrypoint.S
@@ -336,9 +336,9 @@ endfunc tsp_cpu_suspend_entry
 	 */
 func	tsp_fiq_entry
 #if DEBUG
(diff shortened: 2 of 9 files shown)


danh-arm:

danh-arm:

danh-arm:
Seems to be superseded by #363.

PR #351, opened 2015-08-04
Some minor fixes to interrupt-framework-design.md

davwan01:
This patch fixes a pair of typos. The security state had been described
as non-secure where it should have been secure.


danh-arm:

PR #346, opened 2015-07-30
Introduce the new framework for PSCIv1.0

soby-mathew:
This patch stack introduces the new framework to support
the power domain topology and composite power state as specified in
PSCI v1.0 specification better. As a result of these framework changes,
the platform API has changed as well. A platform compatibility layer is
implemented to ensure that the existing platform ports using the old
platform API will continue to work with minimal changes. Care is also taken
to enable the SPDs to continue to work even if the compatibility layer is
disabled.

PR #354, opened 2015-08-07
New framework to support PSCI v1.0 optional features

soby-mathew:


soby-mathew:
Previous PR- https://github.com/ARM-software/arm-trusted-firmware/pull/346

soby-mathew:
Abandoned

PR #355, opened 2015-08-07
New framework to support PSCI v1.0 optional features

soby-mathew:


soby-mathew:
Previous PR- #346

vwadekar:
Will test this once with the Tegra port before leaving my comments.

vwadekar:
@achingupta 

Verified on Tegra T210 platforms. No issues.

BTW, how are we planning to get this PR into master branch? This PR by itself would break the Tegra port. The best way would be to wait until Tegra changes are in the integration branch and then merge to master.

danh-arm:
@vwadekar 

Hi Varun

The idea is that this PR should not break the Tegra port because the compatibility layer is enabled by default. Therefore this can be merged before your Tegra changes that use the new platform porting APIs. If the Tegra port _is_ broken by this PR, then please let us know ASAP with details.

Regards

Dan.

vwadekar:
@danh-arm 

Hi Dan,

The Tegra port is indeed broken due to this PR as some of the macros in platform_def.h are not properly defined. I think we need to add one more commit to this PR with the following fixes to platform_def.h -

#define PLATFORM_MAX_AFFLVL		MPIDR_AFFLVL1
#define PLATFORM_NUM_AFFS		((PLATFORM_CLUSTER_COUNT * \
					  PLATFORM_CORE_COUNT) + \
					  PLATFORM_CLUSTER_COUNT)

I verified that this PR and the changed macros work fine, which means the compatibility layer is working as expected for Tegra. I'm OK, if we want to update this PR with the new commit.

Thanks.

danh-arm:
@vwadekar

Thanks for the info, Varun. It sounds like the existing Tegra port has a bug, which this patchset is exposing. These definitions in the current Tegra port look strange to me:

```
#define PLATFORM_MAX_AFFLVL	MPIDR_AFFLVL2
#define PLATFORM_CORE_COUNT	PLATFORM_MAX_CPUS_PER_CLUSTER
#define PLATFORM_NUM_AFFS	((PLATFORM_CLUSTER_COUNT * \
PLATFORM_CORE_COUNT) + \
PLATFORM_CLUSTER_COUNT + 1)
```

`PLATFORM_CORE_COUNT` should be the total number of cores, not the number of cores per cluster. I think the above should be:

```
#define PLATFORM_MAX_AFFLVL	MPIDR_AFFLVL1
#define PLATFORM_CORE_COUNT	(PLATFORM_CLUSTER_COUNT * \
								PLATFORM_MAX_CPUS_PER_CLUSTER)
#define PLATFORM_NUM_AFFS		(PLATFORM_CORE_COUNT + \
								PLATFORM_CLUSTER_COUNT)
```

Could you submit a small PR with this fixed? We will merge that before #355. Then you can submit your bigger changes to use the new APIs later.

Regards

Dan.

vwadekar:
@ danh-arm

Yes. Tegra port indeed has issues which this PR is exposing. Will push a
patch to fix the issues.

vwadekar:
@danh-arm 

Thanks for the inputs Dan. Pushed PR #358 to fix the issues with the macros in platform_def.h

achingupta:
Superseded by PR #359. Please refer to the change log in the PR description for an overview of the differences between the two.

PR #359, opened 2015-08-11
New framework to support PSCI v1.0 optional features

achingupta:
This patch-set introduces the new framework to better support the power
domain topology and composite power state as specified in PSCI v1.0
specification. PSCI v1.0 mandatory features are already supported in
ARM Trusted Firmware and this patch-set improves support for the
optional features and reduces complexity for platform ports.

As a result of these framework changes, the platform API has changed
as well. A platform compatibility layer is implemented to ensure that the
existing platform ports using the old platform API will continue to work
with minimal changes. Care is also taken to enable the SPDs to
continue to work even if the compatibility layer is disabled.

Supersedes PR #355 

Changes from #355 :
*  It was not mandatory for old platform ports to populate the `validate_power_state()` hook in the `plat_pm_ops_t` structure. This hook is mandatory in the new platform porting interface in the `plat_psci_ops_t` structure. The compatibility layer was treating this hook as optional even with the new porting interface. This bug has been rectified in this pull request. Also, the old hook is called only if has been exported by the old platform ports.

PR #361, opened 2015-08-14
For sm/psci proto v5

achingupta:

Supersedes PR #359 
Supersedes PR #355

Changes from #359 :

1.  New platform porting APIs are prefixed as "plat_" instead of "platform_" for consistency.
2.  Some code readability improvements in `psci_validate_suspend_req()`.
3.  Removed anonymous enums from PSCI code.
4.  "context_id" parameter in PSCI APIs is now a "u_register_t" for portability.

Changes from #355 :

    It was not mandatory for old platform ports to populate the validate_power_state() hook in the plat_pm_ops_t structure. This hook is mandatory in the new platform porting interface in the plat_psci_ops_t structure. The compatibility layer was treating this hook as optional even with the new porting interface. This bug has been rectified in this pull request. Also, the old hook is called only if has been exported by the old platform ports.


achingupta:
@vwadekar 

FYI, new PR for PSCI changes

vwadekar:
@achingupta 

Thanks Achin. Will test and update shortly.

vwadekar:
@achingupta 

Hello Achin,

Why are the following functions still using psci_plat_pm_ops?

- psci_validate_power_state()
- psci_query_sys_suspend_pwrstate()

IIRC, this was not the case in the previous version. Also, I see psci_plat_pm_ops being still used in multiple files, whereas platform-migration-guilde.md mentions that "the previous 'plat_pm_ops' structure is deprecated". Am I missing something?

achingupta:
Hi Varun,

On 17/08/15 06:58, varunw-nvidia wrote:> @achingupta <https://github.com/achingupta>
> 
> Hello Achin,
> 
> Why are the following functions still using psci_plat_pm_ops?
> 
>   * psci_validate_power_state()
>   * psci_query_sys_suspend_pwrstate()
> 
> IIRC, this was not the case in the previous version. Also, I see 
> psci_plat_pm_ops being still used in multiple files, whereas 
> platform-migration-guilde.md mentions that "the previous 'plat_pm_ops' structure 
> is deprecated". Am I missing something?

It is the `psci_plat_pm_ops` which is using these hooks actually :)

Both the new and the old ports have these hooks. `psci_validate_power_state()` was optional in the old porting interface but is critical in the new interface. Please note that `psci_plat_pm_ops` is the new structure with a list of function pointers that the platform should export. `plat_pm_ops` is the old structure which has been deprecated.

I have referenced documentation from PR #355 which elaborates on this:

https://github.com/soby-mathew/arm-trusted-firmware/commit/36a55c315f35a09f8287b93ba4333bfbc2ebc52f#diff-bfe9cf0bda099798fed6f0044b763787R148

Do let us know if you have any other concerns. We intend to merge this PR today.

cheers,
Achin

vwadekar:
@achingupta 

Thanks Achin. I am trying to get System Suspend working with this PR. Although, it was working with the previous PR, now I am not hitting the code in my platform's system suspend handler. It might be due to incorrect settings from the get_sys_suspend_power_state() handler. FWIW, I had not implemented this callback previously, but was able to enter System Suspend.

vwadekar:
@achingupta 

Hello Achin,

I have figured out the solution to get SYSTEM SUSPEND working with this PR. We can go ahead with the merge.

Thanks.

achingupta:
@vwadekar 

Thanks for letting us know.

PR #363, opened 2015-08-17
Add a dispatcher for 32-bit Secure Payload

virtualopensystems-kchappuis:
This patch adds a new dispatcher based on the previous TSP aarch64 dispatcher in order to support 32-bit Secure Payload in ARM Trusted Firmware. A 32-bits Secure payload example could be download here: "http://interactive.freertos.org/forums/ 135282-Any-other-files-Any-other-manufacturers-Any-other-business" in the topic "FreeRTOS v8.2.2 port (AARCH32) for ARMv8 platform (ARM FastModel virtual platform and ARM JUNO Development Platform) using the GCC ARM compiler (arm-none-eabi-)".
As PSCI can suspend the primary CPU, in which the Secure Payload is not migratable while executing, a compilation flag has been added in order to prevent this case.

The idea is to add a new dispatcher directory to implement it. Since the architecture is based on the previous TSP dispatcher for 64-bit Secure Payloads, we use the same approach as previous dispatchers.


danh-arm:

danh-arm:

danh-arm:
I'm confused about the purpose of this Pull Request. First, the existing TSPD already supports 32-bit SPs (although the TSP has not yet been ported to AArch32 to demonstrate this). Second, the dispatcher in this PR seems to be fundamentally the same as the modified TSPDs in #338 and #345, which we've already said we do not believe is a good demonstrator for how to solve the "only-one-secure-interrupt-group-in-GICv2" problem. The same arguments apply here so we do not intend to merge this functionality.

virtualopensystems-kchappuis:
Hello, during our last conversation, we talked about adding a Secure Timer service in EL3 for a S-EL1 aarch32 application which by default does not have direct access to.  However, as you mentioned, the TSP has not yet been ported to AARCH32 and it is not a trivial piece of work. So the idea was to share our 32-bit FreeRTOS port that is now done and also published in the FreeRTOS contribution community.

The dispatcher in this PR, it is not fundamentally the same as the modified TSPDs in #345 because in this dispatcher we don't trap all FIQs in EL3. The FIQ management is the same as the previous TSPD dispatcher. So the goal of this PR it is not to demonstrate how to solve the "only-one-secure-interrupt-group-in-GICv2" problem but it's to demonstrate the Secure Timer service in EL3 for 32-bit Secure Payload by executing our FreeRTOS port solution in BL3-2 layer.

danh-arm:
Sorry, I had not looked at this dispatcher in detail when I made my last comment.

I can see that adding dispatcher support for accessing the secure timer is useful, but:
*   Ideally, the additional functionality would be added to the existing TSPD. Comparing your dispatcher with the TSPD reveals:
    *   The current TSPD hardcodes the SPSR of the TSP to be AArch64. This could easily be made configurable at build time using a flag (which actually you've already added but for a different purpose).
    *   Your FreeRTOS SP is UP non-migrateable and you've removed all the PM functionality from the TSPD in your dispatcher. Potentially the SP migrate type could also be made a build option if your FreeRTOS cannot be made MP.
    *   Some of the existing TSPD SMCs have been removed. This seems unnecessary and make this dispatcher not compliant to the SMC calling convention.
    *   The `tspd_sel1_interrupt_handler()` implementation restores some additional context before transitioning to the TSP. It's possible there might need to be some more generic change needed here in the context management code. I don't fully understand what's going on here yet; some elaboration would be good.
*   If your dispatcher really does need to be separate from the TSPD, then it needs to have a completely different name to indicate that it's specific to your FreeRTOS SP.
*   You can't hack the PSCI code to optionally not handle CPU_SUSPEND on the primary CPU. That makes the implementation not PSCI spec compliant. I guess you just need to fix your FreeRTOS SP.
*   Any new SP build flag(s) should be added to the SPD makefile, not the top level makefile.
*   The timer interface looks like it needs a bit of work to separate initialization from starting the timer. Some comment on the initialization state would be good too (e.g. one-shot or periodic).

danh-arm:
Also, this PR targets the master branch instead of the integration branch.

virtualopensystems-kchappuis:
I agree, this functionality can be directly added in the current TSPD so if you prefer this solution it can certainly be implemented directly there.
Thanks for your comments, I will consider it in order to improve the patch then I will come back to you.

athoelke:
I agree with @danh-arm regarding the CPU_SUSPEND changes - we cannot allow a spec-breaking change into the upstream code (even under a build option). PSCI is very clear that a UP secure payload is enttitled to deny a CPU_OFF request but cannot stop a CPU_SUSPEND request.

I suggest that you carry that patch yourself (and/or provide in a separate branch in your own fork for others to use) in order to allow your test RTOS payload to work until you have implemented enough of the spd_pm hooks to inform the PSCI framework that this is a non-migrating UP payload, prevent CPU_OFF on the primary CPU and handle suspend/resume of the secure payload.

The additional register context in the interrupt handler might not be necessary. The entrypoint in the payload for handling fiq's coming in from normal world cannot be the normal FIQ exception vector in Secure-EL1 - see the TSP implementation of this (in AArch64) as an example. In the current TSP, the previous spsr/elr value for Secure-EL1 is not needed to handle the fiq (in the TSP) as the completion of the fiq is signalled via an SMC to the TSPD and the "current" state of the secure payload is not used or modified during the FIQ handling.

Some secure payloads might do more processing in response to an interrupt, possibly wanting to save the current execution context and switch to an alternate task - in which case this interrupted context information is required and the associated SPD would have to do more work during interrupt handling to provide that information. However, if that is the case for FreeRTOS then is it possible to use the architectural registers for that information: in the SPD copy the saved secure copies of `SPSR_EL3` and `ELR_EL3` (the previous Secure-EL1 context) into `SPSR_FIQ` and `LR_FIQ`. The former is `CTX_SPSR_FIQ` in the EL1 system register context structure (`sysregs_ctx`), the latter is aliased to `X30` in the GP register context. Of course, whether that is viable depends on the implementation of this functionality in FreeRTOS.

This is just one example of how the details of the interaction between the SPD and SP are very tightly coupled - and depend heavily on the approach to task management, MP capability, interrupt support, nesting and re-entrancy, FP usage etc. in the Trusted OS. Trusted OS designers make different decisions about the level of capability required and complexity that can be managed - and the best implementation of the SP/SPD interaction can be highly Trusted OS specific.

danh-arm:
After some offline discussion, we agreed that this demonstration code is better placed in a separate GitHub fork, rather than the upstream master branch. Therefore closing this PR.

PR #368, opened 2015-08-20
TBB: abort boot if BL3-2 cannot be authenticated

jcastillo-arm:
BL3-2 image (Secure Payload) is optional. If the image cannot be
loaded a warning message is printed and the boot process continues.
According to the TBBR document, this behaviour should not apply in
case of an authentication error, where the boot process should be
aborted.

This patch modifies the load_auth_image() function to distinguish
between a load error and an authentication error. The caller uses
the return value to abort the boot process or continue.

In case of authentication error, the memory region used to store
the image is wiped clean.

Change-Id: I534391d526d514b2a85981c3dda00de67e0e7992

### DIFF
--- a/bl2/bl2_main.c
+++ b/bl2/bl2_main.c
@@ -238,8 +238,14 @@ void bl2_main(void)
 	}
 
 	e = load_bl32(bl2_to_bl31_params);
-	if (e)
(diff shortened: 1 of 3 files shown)


jcastillo-arm:
retest this please

PR #369, opened 2015-08-21
Tegra: wait for 512 timer ticks before retention entry

vwadekar:
This patch programs the CPUECTLR_EL1 and L2ECTLR_EL1 registers,
so that the core waits for 512 generic timer CNTVALUEB ticks before
entering retention state, after executing a WFI instruction.

This functionality is configurable and can be enabled for platforms
by setting the 'ENABLE_PROCESSOR_RETENTION' flag.


danh-arm:

danh-arm:

include/lib/cpus/aarch64/cortex_a53.h
```diff
@@ -41,6 +41,15 @@
 
 #define CPUECTLR_SMP_BIT		(1 << 6)
```
danh-arm: It's good practice to define the shift to the retention control bits, even though it is zero in this case.

Also, it would be good to add defines for the Advanced SIMD and Floating-point retention control bits for completeness.

include/lib/cpus/aarch64/cortex_a53.h
```diff
@@ -41,6 +41,15 @@
...
 
+#define CPU_RETENTION_ENTRY_2_TICKS	0x1
+#define CPU_RETENTION_ENTRY_8_TICKS	0x2
+#define CPU_RETENTION_ENTRY_32_TICKS	0x3
+#define CPU_RETENTION_ENTRY_64_TICKS	0x4
+#define CPU_RETENTION_ENTRY_128_TICKS	0x5
+#define CPU_RETENTION_ENTRY_256_TICKS	0x6
+#define CPU_RETENTION_ENTRY_512_TICKS	0x7
```
danh-arm: These tick defines could be named a little more generically (e.g. RETENTION_ENTRY_TICKS_XXX) and shared with the L2 and Advanced SIMD and Floating-point retention control.

include/lib/cpus/aarch64/cortex_a57.h
```diff
@@ -44,6 +44,15 @@
...
 
+#define CPU_RETENTION_ENTRY_2_TICKS	0x1
+#define CPU_RETENTION_ENTRY_8_TICKS	0x2
+#define CPU_RETENTION_ENTRY_32_TICKS	0x3
+#define CPU_RETENTION_ENTRY_64_TICKS	0x4
+#define CPU_RETENTION_ENTRY_128_TICKS	0x5
+#define CPU_RETENTION_ENTRY_256_TICKS	0x6
+#define CPU_RETENTION_ENTRY_512_TICKS	0x7
+#define CPU_RETENTION_ENTRY_MASK	CPU_RETENTION_ENTRY_512_TICKS
+
```
danh-arm: Same comments as in cortex_a53.h

plat/nvidia/tegra/common/aarch64/tegra_helpers.S
```diff
@@ -57,6 +57,23 @@
...
+	 * ---------------------------
+	*/
+	mrs	x0, L2ECTLR_EL1
+	bic	x0, x0, #L2_RETENTION_ENTRY_MASK
+	orr	x0, x0, #L2_RETENTION_ENTRY_512_TICKS
+	msr	L2ECTLR_EL1, x0
+
+	mrs	x0, CPUECTLR_EL1
+	bic	x0, x0, #CPU_RETENTION_ENTRY_MASK
+	orr	x0, x0, #CPU_RETENTION_ENTRY_512_TICKS
```
danh-arm: It's good practice to use the shift value to the control bits, even if it's zero in this case.

danh-arm:
I'd prefer the changes to the cpu library code to be in a separate patch to the tegra port changes.

include/lib/cpus/aarch64/cortex_a53.h
vwadekar: > Also, it would be good to add defines for the Advanced SIMD and Floating-point retention control bits for completeness.

Where can I get a description of these values? I checked the "Advance SIMD and Floation Point" section in the A57r1p3 TRM.

danh-arm: I beg your pardon, the Advanced SIMD and Floating-point retention control bits only apply to A53, not A57.

include/lib/cpus/aarch64/cortex_a53.h
vwadekar: Is it ok to just define the RETENTION_ENTRY_TICKS_xxx macros and use them directly in the code? We can have the CPUECTLR and L2ECTLR masks though.

vwadekar:
New PR - #370

include/lib/cpus/aarch64/cortex_a53.h
danh-arm: It's OK to use the RETENTION_ENTRY_TICKS_xxx macros in the code as long as they are shifted correctly (even though the shift always happens to be zero). I see this is what you've done in the latest PR.

PR #373, opened 2015-08-26
Tlkd pm handlers v1

vwadekar:
This PR modifies the PSCI service to send current suspend levels to the dispatcher's suspend handler. This would allow payloads to take different actions for CPU_SUSPEND/SYSTEM_SUSPEND.

This would allow TLK to save/restore certain hardware registers across SYSTEM_SUSPEND. This save/restore is not required for CPU_SUSPEND as the hardware blocks lose their settings during SYSTEM_SUSPEND.

### DIFF
--- a/include/bl31/services/psci.h
+++ b/include/bl31/services/psci.h
@@ -286,7 +286,7 @@ typedef struct plat_psci_ops {
 typedef struct spd_pm_ops {
 	void (*svc_on)(uint64_t target_cpu);
 	int32_t (*svc_off)(uint64_t __unused);
-	void (*svc_suspend)(uint64_t __unused);
+	void (*svc_suspend)(uint64_t suspend_level);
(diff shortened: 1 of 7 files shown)


danh-arm:

danh-arm:

vwadekar:
@danh-arm @achingupta 

Hello,

This change modifies the common PSCI layer to allow TLK to receive SYSTEM_SUSPEND events. Since we were already passing one input parameter to the suspend handlers, I thought it would be OK to pass the actual suspend level instead. The PSCI spec is not clear if suspend handlers for EL1 payloads should be able to distinguish between CPU_SUSPEND and SYSTEM_SUSPEND. I hope that I dont end up moving away from the spec with this code.

Thoughts?

-Varun

vwadekar:
New PR - #376

PR #376, opened 2015-08-26
Tlkd pm handlers v2

vwadekar:
This PR modifies the PSCI service to send current suspend levels to the dispatcher's suspend handler. This would allow payloads to take different actions for CPU_SUSPEND/SYSTEM_SUSPEND.

This would allow TLK to save/restore certain hardware registers across SYSTEM_SUSPEND. This save/restore is not required for CPU_SUSPEND as the hardware blocks lose their settings only during SYSTEM_SUSPEND.

Previous PR - #373

PR #377, opened 2015-08-27
Tlkd pm handlers v3

vwadekar:
This PR modifies the PSCI service to send current suspend levels to the dispatcher's suspend handler. This would allow payloads to take different actions for CPU_SUSPEND/SYSTEM_SUSPEND.

This would allow TLK to save/restore certain hardware registers across SYSTEM_SUSPEND. This save/restore is not required for CPU_SUSPEND as the hardware blocks lose their settings only during SYSTEM_SUSPEND.

Previous PRs - #373, #376

### DIFF
--- a/include/bl31/services/psci.h
+++ b/include/bl31/services/psci.h
@@ -286,7 +286,7 @@ typedef struct plat_psci_ops {
 typedef struct spd_pm_ops {
 	void (*svc_on)(uint64_t target_cpu);
 	int32_t (*svc_off)(uint64_t __unused);
-	void (*svc_suspend)(uint64_t __unused);
+	void (*svc_suspend)(uint64_t suspend_level);
 	void (*svc_on_finish)(uint64_t __unused);
 	void (*svc_suspend_finish)(uint64_t suspend_level);
 	int32_t (*svc_migrate)(uint64_t from_cpu, uint64_t to_cpu);
--- a/include/bl32/payloads/tlk.h
+++ b/include/bl32/payloads/tlk.h
@@ -43,6 +43,9 @@
 #define TLK_REGISTER_LOGBUF	TLK_TOS_STD_FID(0x1)
 #define TLK_REGISTER_REQBUF	TLK_TOS_STD_FID(0x2)
 #define TLK_RESUME_FID		TLK_TOS_STD_FID(0x100)
+#define TLK_SYSTEM_SUSPEND	TLK_TOS_STD_FID(0xE001)
+#define TLK_SYSTEM_RESUME	TLK_TOS_STD_FID(0xE002)
+#define TLK_SYSTEM_OFF		TLK_TOS_STD_FID(0xE003)
 
 /*
  * SMC function IDs that TLK uses to signal various forms of completions
@@ -52,6 +55,9 @@
 #define TLK_PREEMPTED		(0x32000002 | (1 << 31))
 #define TLK_ENTRY_DONE		(0x32000003 | (1 << 31))
 #define TLK_VA_TRANSLATE	(0x32000004 | (1 << 31))
+#define TLK_SUSPEND_DONE	(0x32000005 | (1 << 31))
+#define TLK_RESUME_DONE		(0x32000006 | (1 << 31))
+#define TLK_SYSTEM_OFF_DONE	(0x32000007 | (1 << 31))
 
 /*
  * Trusted Application specific function IDs
--- a/services/spd/opteed/opteed_pm.c
+++ b/services/spd/opteed/opteed_pm.c
@@ -81,7 +81,7 @@ static int32_t opteed_cpu_off_handler(uint64_t unused)
  * This cpu is being suspended. S-EL1 state must have been saved in the
  * resident cpu (mpidr format) if it is a UP/UP migratable OPTEE.
  ******************************************************************************/
-static void opteed_cpu_suspend_handler(uint64_t unused)
+static void opteed_cpu_suspend_handler(uint64_t suspend_level)
 {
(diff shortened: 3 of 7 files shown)


danh-arm:

danh-arm:

vwadekar:
@soby-mathew @athoelke 

Hello,

Please let me know if you have any further comments.

Thanks.

achingupta:
FWIU, the requirement is to tell the TLKD which is the highest power domain that is entering a power down state after PSCI state coordination. If the system power domain could enter a power down state then TLK will save and restore some HW registers. 

If the above makes sense then I don't think the `end_pwrlevel` variable in `psci_suspend_to_pwrdown_start()` conveys that information. It only indicates what the highest requested power level is. The highest power level that could actually be powered down might be much lower. This level is determined in `psci_do_state_coordination()` but never exposed outside this function directly. Instead the `state_info` parameter transforms from carrying the requested states to carrying the target states. The only way to find this "level" is by parsing the returned `state_info` using the `psci_find_max_off_lvl()` function in `psci_common.c`.

I think this works for TLKD/TLK since the SYSTEM_SUSPEND request will always land on the last CPU in the system and `end_pwrlvl` will be the highest power level that will be powered off. It will be a cause of redundancy in case CPU_SUSPEND is being used by multiple CPUs to enter the deepest power state. Ideally, the information that should be passed to a SPD is the highest power level that could enter a power down state. I am assuming that most SPDs do not care about retention states.

I don't think the semantics of 'suspend_level' being passed to the SPD are quite clear. I will have a chat internally and get back. Do shout if I am getting this wrong.

vwadekar:
@achingupta 

Hello Achin,

Your understanding is correct. TLK only requires the SYSTEM_SUSPEND notification for now. Do let me know how I should proceed to get the highest power level that will be powered off to solve this problem.

Thanks.

achingupta:
@vwadekar 

Hi Varun,

We are open to suggestions from your end. Give us until the end of next week to sort this out.

thanks,
Achin

achingupta:
Just thinking aloud here. We should just change the semantics of the `suspend_level` parameter being passed to the `svc_suspend_finish()` and the `svc_suspend()` hooks such that it refers:

1.  to the highest power level which will be targetted to enter a power down state in the latter case and
2.  the highest power level which is emerging from a power down state in the former case.

I don't think TOSes are using this parameter in `svc_suspend_finish()` even now (excluding TLK). This seems to be the case with OPTEE but I need to check. More on this early next week! I missed your email about the urgency of this PR earlier.

vwadekar:
@achingupta 

Hello Achin,

Thanks for taking time to review this PR. Appreciate the help.

I was looking at 'psci_do_state_coordination()' and see that we call a platform's 'plat_get_target_pwr_state()'. Do you think we can use this platform API to get the actual targeted power state i.e. SYSTEM_SUSPEND in our case? Although Tegra uses the common implementation now, for this we will have to provide a handler for Tegra and then return "proper" states.

I understand this might not be an ideal solution and you might want to come up with another solution which is cleaner. Do let me know you thoughts.

Thanks.

achingupta:
Hi Varun,

On Mon, Sep 07, 2015 at 12:58:42PM +0100, varunw-nvidia wrote:
> @achingupta<https://github.com/achingupta>
>
> Hello Achin,
>
> Thanks for taking time to review this PR. Appreciate the help.

No worries! We try our best.

>
> I was looking at 'psci_do_state_coordination()' and see that we call a platform's 'plat_get_target_pwr_state()'. Do you think we can use this platform API to get the actual targeted power state i.e. SYSTEM_SUSPEND in our case? Although Tegra uses the common imple\
mentation now, for this we will have to provide a handler for Tegra and then return "proper" states.

`plat_get_target_pwr_state()` takes the requested power states for a particular
power level as an input, performs state coordination and returns the power state
that the power domain at that level should enter.

I guess what you are suggesting is that Tegra overrides the default
implementation of this function in `plat/common/aarch64/plat_psci_common.c`. The
Tegra implementation apart from performing state coordination will stash the
target power state for the system level. TLKD will access the stashed state
through a private API. Am I getting this right? You will have to invalidate the
stashed value when the system exits the low power state.

This approach should work but I think it will be simpler to just change the
semantics of the `suspend_level` parameter to the SPD PM handlers. I did have a
chat internally about my proposal and we have agreement.

>
> I understand this might not be an ideal solution and you might want to come up with another solution which is cleaner. Do let me know you thoughts.

I will be pushing out a PR which implements my proposal tomorrow. I would expect
it to be merged in a couple of days if not sooner. You would need to just rebase
the TLKD specific patch on top of this PR. Does that work for you? Do let us
know either way.

Thanks,
Achin

>
> Thanks.
>
> �
> Reply to this email directly or view it on GitHub<https://github.com/ARM-software/arm-trusted-firmware/pull/377#issuecomment-138279453>.

-- IMPORTANT NOTICE: The contents of this email and any attachments are confidential and may also be privileged. If you are not the intended recipient, please notify the sender immediately and do not disclose the contents to any other person, use it for any purpose, or store or copy the information in any medium.  Thank you.

ARM Limited, Registered office 110 Fulbourn Road, Cambridge CB1 9NJ, Registered in England & Wales, Company No:  2557590
ARM Holdings plc, Registered office 110 Fulbourn Road, Cambridge CB1 9NJ, Registered in England & Wales, Company No:  2548782

achingupta:
@vwadekar @soby-mathew @athoelke @danh-arm 

https://github.com/ARM-software/arm-trusted-firmware/pull/386 implements what I have described above. Please comment.

vwadekar:
New PR - #396

PR #380, opened 2015-09-01
Ensure BL2 security state is secure

vikramkanigiri:
BL2 loads secure runtime code(BL3-1, BL3-2) and hence it has to
run in secure world otherwise BL3-1/BL3-2 have to execute from
non-secure memory. Hence, This patch removes the change_security_state()
call in bl1_run_bl2() and replaces it with an assert to confirm
the BL2 as secure.

Fixes ARM-software/tf-issues#314

Change-Id: I611b83f5c4090e58a76a2e950b0d797b46df3c29

### DIFF
--- a/bl1/bl1_main.c
+++ b/bl1/bl1_main.c
@@ -44,13 +44,13 @@
  ******************************************************************************/
 static void __dead2 bl1_run_bl2(entry_point_info_t *bl2_ep)
 {
+	/* Check bl2 security state is expected as secure */
+	assert(GET_SECURITY_STATE(bl2_ep->h.attr) == SECURE);
+
 	bl1_arch_next_el_setup();
 
 	/* Tell next EL what we want done */
 	bl2_ep->args.arg0 = RUN_IMAGE;
-
-	if (GET_SECURITY_STATE(bl2_ep->h.attr) == NON_SECURE)
-		change_security_state(GET_SECURITY_STATE(bl2_ep->h.attr));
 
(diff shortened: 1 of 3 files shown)


bl1/bl1_main.c
```diff
@@ -44,14 +44,14 @@
...
+	/* Check bl2 security state is expected as secure */
+	assert(GET_SECURITY_STATE(bl2_ep->h.attr) == SECURE);
+
 	bl1_arch_next_el_setup();
 
 	/* Tell next EL what we want done */
 	bl2_ep->args.arg0 = RUN_IMAGE;
 
-	if (GET_SECURITY_STATE(bl2_ep->h.attr) == NON_SECURE)
-		change_security_state(GET_SECURITY_STATE(bl2_ep->h.attr));
```
danh-arm: I'm in my earlier comment [here](https://github.com/ARM-software/tf-issues/issues/314#issuecomment-124031548), I suggested that the `change_security_state` call should be left in, since future "firmware update" functionality may change the `SCR_NS_BIT`. However, after discussion with the "firmware update" author, we agreed that it's the responsibility of the "firmware update" code to ensure the `SCR_NS_BIT` is cleared by the time this code is reached.

However, this `bl1_run_bl2` function should have an additional assert that the `SCR_NS_BIT` is clear.

vikramkanigiri:
Superseded by https://github.com/ARM-software/arm-trusted-firmware/pull/383

PR #385, opened 2015-09-07
Re-design bakery lock memory allocation and algorithm

vikramkanigiri:
This patch unifies the bakery lock api's across coherent and normal
memory implementation of locks by using same data type `bakery_lock_t`
and similar arguments to functions.

A separate section `bakery_lock` has been created and used to allocate
memory for bakery locks using `DEFINE_BAKERY_LOCK`. When locks are
allocated in normal memory, each lock for a core has to spread
across multiple cache lines. By using the total size allocated in a
separate cache line for a single core at compile time, the memory for
other core locks is allocated at link time by multiplying the single
core locks size with (PLATFORM_CORE_COUNT - 1). The normal memory lock
algorithm now uses lock address instead of the `id` in the per_cpu_data.
For locks allocated in coherent memory, it moves locks from
tzfw_coherent_memory to bakery_lock section.

The bakery locks are allocated as part of bss or in coherent memory
depending on usage of coherent memory. Both these regions are
initialised to zero as part of run_time_init before locks are used.
Hence, bakery_lock_init() is made an empty function as the lock memory
is already initialised to zero.

The above design lead to the removal of `platform_cpu_data` usage
as bakery locks for power controller in ARM platforms and also
moved psci bakery locks from non_cpu_power_pd_node to psci_locks.

NOTE: The bakery lock API when USE_COHERENT_MEM is not set has changed.
This is a breaking change for all platform ports that allocate bakery
locks in normal memory. The changes made to the Juno and FVP ports can
be used as a reference to update such platform ports.

Change-Id: Ic3751c0066b8032dcbf9d88f1d4dc73d15f61d8b

### DIFF
--- a/bl31/bl31.ld.S
+++ b/bl31/bl31.ld.S
@@ -101,10 +101,31 @@ SECTIONS
      * The .bss section gets initialised to 0 at runtime.
      * Its base address must be 16-byte aligned.
      */
-    .bss : ALIGN(16) {
+    .bss (NOLOAD) : ALIGN(16) {
         __BSS_START__ = .;
         *(.bss*)
         *(COMMON)
+#if !USE_COHERENT_MEM
+        /*
+         * Bakery locks are stored in normal .bss memory
+         *
+         * Each lock's data is spread across multiple cache lines, one per CPU,
+         * but multiple locks can share the same cache line.
+         * The compiler will allocate enough memory for one CPU's bakery locks,
+         * the remaining cache lines are allocated by the linker script
+         */
+        . = ALIGN(CACHE_WRITEBACK_GRANULE);
+        __BAKERY_LOCK_START__ = .;
+        *(bakery_lock)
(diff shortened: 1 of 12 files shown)


danh-arm:
I would prefer this to be split into 3 patches. The 1st one with the generic changes, followed by the ARM platform ports changes, followed by the MTK platform port changes. As far as I understand, the the first patch should not break the platform ports when USE_COHERENT_MEM is not set, and splitting the patches will help demonstrate this.

danh-arm:
nit: Please put the important "Note" at the end of the commit message in capital letters.

danh-arm:
New optional platform porting constant `PLAT_PERCPU_BAKERY_LOCK_SIZE` is not defined in `porting_guide.md`.

achingupta:
Please close a stale PR when a new one supersedes it.

Superseded by #390

PR #386, opened 2015-09-07

achingupta:
to the SPD in the "suspend_level" parameter of the svc_suspend() hook.

patch does this by changing the semantics of the SPD svc_suspend_finish() hook's
"suspend_level" parameter.


achingupta:
Superseded by #387

PR #387, opened 2015-09-10

achingupta:
to the SPD in the "max_off_pwrlvl" parameter of the svc_suspend() hook.

patch does this by changing the semantics of the parameter passed to an SPD's
svc_suspend_finish() hook. The name of the parameter has been changed from
"suspend_level" to "max_off_pwrlvl" as well. Same changes have been made to the
parameter passed to the tsp_cpu_resume_main() function.

Supersedes #386 

Changelog:
*  Made parameter names more consistent. References to `suspend_level` have been replaced by `max_off_pwrlvl`

### DIFF
--- a/bl32/tsp/tsp_main.c
+++ b/bl32/tsp/tsp_main.c
@@ -251,7 +251,7 @@ tsp_args_t *tsp_cpu_suspend_main(uint64_t arg0,
  * cpu's architectural state has been restored after wakeup from an earlier psci
  * cpu_suspend request.
  ******************************************************************************/
-tsp_args_t *tsp_cpu_resume_main(uint64_t suspend_level,
+tsp_args_t *tsp_cpu_resume_main(uint64_t max_off_pwrlvl,
 			      uint64_t arg1,
 			      uint64_t arg2,
 			      uint64_t arg3,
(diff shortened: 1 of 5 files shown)


achingupta:
@jenswi-linaro @jbech-linaro 

Hi Both,

This PR changes the semantics of the `suspend_level` parameter passed to the OPTEED PM handler when the core resumes from a low power state. Instead of being the highest power level that was requested to be placed in a low power state, the parameter now holds the highest power level that was actually placed in a power down state. The latter parameter is determined in the PSCI code after performing state coordination using the former parameter as one of the inputs.

Similarly, the SPD suspend PM hook's unused parameter now carries the highest power level that will actually be powered down.

I had a quick look at the OPTEE code and it does not seem that this PR will break anything. It would be great if you could have a look yourself as well.

Thanks,
Achin

jbech-linaro:
Hi @achingupta , we're not really doing anything on that call, so it will be fine.

`Acked-by: Joakim Bech <joakim.bech@linaro.org>`

achingupta:
Thanks Joakim.  

Superseded by #388. Added a minor tweak to the commit message.

PR #388, opened 2015-09-10

achingupta:
to the SPD in the "max_off_pwrlvl" parameter of the svc_suspend() hook.

patch does this by changing the semantics of the parameter passed to an SPD's
svc_suspend_finish() hook. The name of the parameter has been changed from
"suspend_level" to "max_off_pwrlvl" as well. Same changes have been made to the
parameter passed to the tsp_cpu_resume_main() function.

NOTE: THIS PATCH CHANGES THE SEMANTICS OF THE EXISTING "svc_suspend_finish()"
      API BETWEEN THE PSCI AND SPD/SP IMPLEMENTATIONS. THE LATTER MIGHT NEED
      UPDATES TO ENSURE CORRECT BEHAVIOUR.

Supersedes #387 

Changelog:
*  Edited commit message to warn SPD/SP implementations about the change in semantics of an API.

### DIFF
--- a/bl32/tsp/tsp_main.c
+++ b/bl32/tsp/tsp_main.c
@@ -251,7 +251,7 @@ tsp_args_t *tsp_cpu_suspend_main(uint64_t arg0,
  * cpu's architectural state has been restored after wakeup from an earlier psci
  * cpu_suspend request.
  ******************************************************************************/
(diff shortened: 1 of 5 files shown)


achingupta:
test this please

achingupta:
retest this please

achingupta:
retest this please

PR #392, opened 2015-09-14
mt8173: support big/Little cluster power off for CPU idle(C2 state)

mtk09422:
1. add power control for both big and Little cluster in MCDI
2. fix incorrect PCM_HOTPLUG_VALID_MASK in spm_hotplug.c

Change-Id: Ib06b578ba0a164855aba3b4d4b8af4d520c42707
Signed-off-by: Weiyi Lu <weiyi.lu@mediatek.com>


achingupta:
test this please

danh-arm:

danh-arm:

danh-arm:
test this please

athoelke:
There are a couple of things happening in this patch:
1. Obvious cluster-power-down changes: updates to the firmware codes to support the C2 state, and movement of some of the system state changes to properly manage the system for CPU, Cluster or System power down scenarios
2. Some effort to control/manage a race condition between CPU_ON and CPU_SUSPEND, when the CPU_ON is targeting a CPU in a cluster where the last active core is about to suspend to C2 level.

The CPU_ON vs. CLUSTER_OFF/SUSPEND race is an interesting case. The PSCI framework has to assume that the platform port, or the platform power controller can resolve this race by serialising the underlying requests. Either the cluster does indeed power down, and the CPU_ON request is only processed by the power controller after this; or the cluster-power-down request is relaxed to a cpu-power-down by the power controller because a new CPU has started executing (CPU_ON called but the on_finisher has not yet run).

In one of our platforms, we discovered that the behaviour of the power controller required that we reset some control bits on every CPU wake-up, for both suspend_finish() and on_finish() to handle the possibility that this race occurred.

In this code, it is not obvious that the implementation of the `is_on_pending[]` changes will properly resolve a race here, although that does depend on exactly what shared state is not being correctly updated without these changes. For example, the use of `spm_lock` to set and clear the `is_on_pending[]` flags has no effect of the atomicity of the flag values with respect to the `is_in_on_pending()` function. In addition, if the `afflvl` downgrade logic in line 377 of `plat_pm.c` occurs, the following code will just do the `AFFLVL0` power-down; but the `plat_affinst_suspend_finish()` will still do the `AFFLVL1` power-up logic as this is the value passed by the framework. In this case, you code probably does not break anything, but the asymmetry is a risk for future code changes.

Is it possible for you to describe the race that this data structure and logic is trying to resolve?

WeiyiLu:
Actually there is a limitation of our power control HW SPM. It could manage the cluster power but unable to control the CCI coherency, so CCI coherency could only be controlled by platform port.
We indeed spend some time investigating the condition you mentioned. When the CPU_ON is targeting a CPU in a cluster where the last active core is about to suspend to C2 level, cluster power control could be resolved by SPM serializing the underlying requests but CCI coherency control cannot be guaranteed at the same time. Therefore it would be a risky operation of CCI coherency with cluster power control under such condition.

athoelke:
That sounds very similar to the limitations on the ARM FVPs and Juno platform. In both cases the power controller will ensure that the cluster is not powered off if one of the CPUs is being brought out of reset - but the power controller does not program CCI coherency. This part of the race is handled correctly by the PSCI framework and platform port without requiring additional locks and tracking of on-pending. Of course, it is possible that some detail of the hardware is different enough to require a different solution for your platform.

Let me try and describe what should happen. Let **A** be the `CPU_ON` request, **B** be the CPU_SUSPEND (to cluster level), and **C** the new CPU powering on through `afflvl_on_finish()`.

1. If **A** and **C** happen before **B** then the `afflvl_suspend()` call in **B** will only request cpu power down as a new CPU is running.
2. If **A** and **C** happen after **B** then the `afflvl_suspend()` call in **B** will request cluster power down and the cluster will really power off. **A** will cause the cluster to powered back on and **C** will call `afflvl_on_finisher()` to do cluster and cpu power on actions.
3. If **A** happens before **B** which happens before **C**. Then **A** will cause the power controller to power on the new CPU, but not the cluster because it is running. The PSCI framework will not detect this at **B**, and so `afflvl_suspend()` is called to power down the cluster. This request is made to the power controller, and the cluster is taken out of coherency. When **B** completes by releasing the lock on the cluster affinity-node/power-domain-node, then **C** the powering-on CPU _on finisher_ can run after acquiring this lock. At this point, the PSCI framework will detect that the cluster had been powered off (from a firmware perspective) and so the `afflvl_on_finisher()` is invoked to do cluster and cpu power on actions. I.e. the CCI coherency should be enabled at this point.

So the tricky case you describe is #3 above, and the framework solves this by (a) ensuring that **B** completes before **C** can run because of the cluster-level lock in the PSCI framework, and (b) the level of power on passed to `afflvl_on_finisher()` in **C** is not based on the cluster state at **A**, but on the PSCI state of the cluster when **C** happens. This should ensure that CCI is re-enabled in the on_finisher() if the cpu_suspend() call disabled it.

Look at the Juno and FVP power management code for cpu_suspend and on_finish to see how this is handled.

PR #394, opened 2015-09-14
Support for ARM CoreLink CCN interconnects

achingupta:
This PR adds support in ARM TF to run on platforms that have system caches e.g. platforms that use CCN IP. It further adds a generic driver to configure any variant of the CCN IP.


achingupta:
@vwadekar 

Hi Varun,

The second patch in this stack performs cache maintenance on the video carve-out region on Tegra in the presence of system (L3) caches. Tegra might not support system caches just yet, but could you please have a look and let us know whether it still works and makes sense to you.

thanks,
Achin

PR #395, opened 2015-09-15
Bug fix in the SP804 dual timer driver

sandrine-bailleux:
The generic delay timer driver expects a pointer to a timer_ops_t
structure containing the specific timer driver information. It
doesn't make a copy of the structure, instead it just keeps the
pointer. Therefore, this pointer must remain valid over time.

The SP804 driver doesn't satisfy this requirement. The
sp804_timer_init() macro creates a temporary instanciation of the
timer_ops_t structure on the fly and passes it to the generic
delay timer. When this temporary instanciation gets deallocated,
the generic delay timer is left with a pointer to invalid data.

This patch fixes this bug by statically allocating the SP804
timer_ops_t structure in the SP804 driver. As a result, the SP804
driver init function no longer expects the timer_ops_t structure in
argument, instead it receives the clock multiplier and divider as
arguments and populates its internal timer_ops_t structure itself.

### DIFF
--- a/drivers/arm/sp804/sp804_delay_timer.c
+++ b/drivers/arm/sp804/sp804_delay_timer.c
@@ -33,6 +33,7 @@
 #include <mmio.h>
 
 uintptr_t sp804_base_addr;
+static timer_ops_t sp804_ops;
 
 #define SP804_TIMER1_LOAD	(sp804_base_addr + 0x000)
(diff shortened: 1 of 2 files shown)


sandrine-bailleux:
Hi Dan,

Thanks for the review comments.
I have modified the patch according to your suggestion and created a new pull request: #399 

Thanks,
Sandrine

PR #396, opened 2015-09-24
Send power management events to the Trusted OS (TLK)

vwadekar:
This patch adds PM handlers to TLKD for the system suspend/resume and
system poweroff/reset cases. TLK expects all SMCs through a single
handler, which then fork out into multiple handlers depending on the
SMC. We tap into the same single entrypoint by restoring the S-EL1
context before passing the PM event via register 'x0'. On completion
of the PM event, TLK sends a completion SMC and TLKD then moves on
with the PM process.


vwadekar:
Previous PR - #377

danh-arm:

danh-arm:

vwadekar:
@danh-arm  @achingupta 

Any feedback?

achingupta:
@vwadekar 

Looks fine apart from the minor nit above.

vwadekar:
New PR - #400

PR #390, opened 2015-09-11
Re-design bakery lock allocation and algorithm

vikramkanigiri:
This request contains following changes:

1. Re-design of bakery lock memory allocation and algorithm by unifying macros, apis across coherent memory and normal memory usage.
2. Updates ARM platform ports to use the new bakery lock apis
3. Update mediatek port to use the new bakery lock macros.

PR #398, opened 2015-09-25
Fix relocation of __PERCPU_BAKERY_LOCK_SIZE__ in PR #390

achingupta:
When a platform port does not define PLAT_PERCPU_BAKERY_LOCK_SIZE, the total
memory that should be allocated per-cpu to accommodate all bakery locks is
calculated by the linker in bl31.ld.S. The linker stores this value in the
__PERCPU_BAKERY_LOCK_SIZE__ linker symbol. The runtime value of this symbol is
different from the link time value as the symbol is relocated into the current
section (.bss). This patch fixes this issue by marking the symbol as ABSOLUTE
which allows it to retain its correct value even at runtime.

The description of PLAT_PERCPU_BAKERY_LOCK_SIZE in the porting-guide.md has been
made clearer as well.

Change-Id: Ia0cfd42f51deaf739d792297e60cad5c6e6e610b

### DIFF
--- a/bl31/bl31.ld.S
+++ b/bl31/bl31.ld.S
@@ -124,7 +124,7 @@ SECTIONS
         __BAKERY_LOCK_START__ = .;
         *(bakery_lock)
         . = ALIGN(CACHE_WRITEBACK_GRANULE);
-        __PERCPU_BAKERY_LOCK_SIZE__ = . - __BAKERY_LOCK_START__;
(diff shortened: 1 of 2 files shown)


vwadekar:
Tried it out on Tegra and do not see the previous crash during CPU boot. Thanks for the quick fix.

Acked by: Varun Wadekar <vwadekar@nvidia.com>

danh-arm:
Looks good to me also

PR #403, opened 2015-10-08

vwadekar:


danh-arm:

danh-arm:

danh-arm:
test this please

athoelke:
The difference in those switch cases is that the original RESUME checked that the CPU is in a preempted-SMC state and bailed if not; and the new RESUME checks that the CPU has no ongoing preempted-SMC (which is the expected behaviour for issuing new requests). I would expect that those condition checks need to be conditional on the original `smc_id`?

vwadekar:
New PR - #404

PR #404, opened 2015-10-09

vwadekar:

### DIFF
--- a/services/spd/tlkd/tlkd_main.c
+++ b/services/spd/tlkd/tlkd_main.c
@@ -199,41 +199,6 @@ uint64_t tlkd_smc_handler(uint32_t smc_fid,
 		SMC_RET1(ns_cpu_context, x1);
 
 	/*
-	 * Request from non secure world to resume the preempted
-	 * Standard SMC call.
-	 */
-	case TLK_RESUME_FID:
-
-		/* RESUME should be invoked only by normal world */
-		if (!ns)
-			SMC_RET1(handle, SMC_UNK);
-
-		/*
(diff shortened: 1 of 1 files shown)


vwadekar:
Previous PR - #403

danh-arm:

danh-arm:

danh-arm:
test this please

danh-arm:
Fair enough, Varun. It's your choice in the end. Merging...

vwadekar:
Thanks Dan

athoelke:
It is not a nit, it is a **bug**. As currently written it is not possible to RESUME after a preempted SMC call as the `if` condition is false because the second expression is false, so the `else if` condition will be tested and prove to be true resulting in a SMC_UNK response. Dan's code is a correct implementation of what is needed.

However, the rules of C mean that Dan's initial concern is invalid - the expression to the left of a binary logical operator (&& and ||) will always be evaluated before any part of the expression on the right; and the expression on the right will never be evaluated if the result is fully determined by the expression on the left. So expressions like `(p && p->member)` are guaranteed to be safe in the face of `p==NULL`.

danh-arm:
Argh, I really shouldn't merge code on a Friday afternoon. I will revert the integration branch to the master branch.
Varun - you'll need to raise a new PR. Can you confirm this is tested next time as it looks like both this version and the last will never work?

vwadekar:
We have a test which was consistently failing until we fixed it with this
change. The test passes with this as well as the last version.

vwadekar:
Pushed new PR - #405.

PR #405, opened 2015-10-09

vwadekar:

### DIFF
--- a/services/spd/tlkd/tlkd_main.c
+++ b/services/spd/tlkd/tlkd_main.c
@@ -199,41 +199,6 @@ uint64_t tlkd_smc_handler(uint32_t smc_fid,
 		SMC_RET1(ns_cpu_context, x1);
 
 	/*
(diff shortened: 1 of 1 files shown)


danh-arm:

danh-arm:

vwadekar:
The testapp for stress testing the secure storage scenario passes on our reference board with this change.

vwadekar:
@danh-arm @athoelke 

Any feedback?

danh-arm:
I would still question how robust the test case is for using the TLK_RESUME_FID SMC, since the previous fix versions passed even though it was not possible to RESUME after a preempted SMC call. I'm still struggling to understand the conditions under which the existing code would fail, while the previous fix versions would pass.

Having said that, the code looks fine to me and I don't really want to get too involved in others' test setups. Therefore merging.

vwadekar:
Thanks Dan. I agree that the test case needs to be more robust. We are
working on it.

PR #417, opened 2015-10-27
Makefile rework and bug fix #2

jcastillo-arm:


danh-arm:
This supersedes #412 by placing the new helper make files in a more suitable directory.

PR #424, opened 2015-11-06
IMF: postpone SCR_EL3 update if context is not initialized

jcastillo-arm:
The set_routing_model() function in the Interrupt Management
Framework calls the context management library to update the
SCR_EL3 register. This context management library assumes that
the context has been previously initialized. Consequently, if
a Secure Payload Dispatcher (SPD) tries to set the routing model
before initializing the context, the system will fail (in debug
mode, an assertion will be raised).

This patch fixes the issue by checking if the context has been
initialized before updating SCR_EL3. If a valid context is not
available, SCR_EL3 update will be done when the SPD calls the
context initialization function. This function will call
get_scr_el3_from_routing_model() to obtain the SCR_EL3 value.
If the SPD does not call the context initialization function
then it is SPD's responsibility to obtain SCR_EL3 from the IMF
and update the context accordingly.

Fixes ARM-software/tf-issues#327

Change-Id: Ic2f1c6e899e578a8db858ec43747c63a8539c16f


jcastillo-arm:
retest this please

PR #426, opened 2015-11-10
Tegra: introduce per-soc system reset handler

vwadekar:
This patch adds a per-soc system reset handler for Tegra chips. The
handler gets executed before the actual system resets. This allows
for custom handling of the system reset sequence on each SoC.

Signed-off-by: Varun Wadekar <vwadekar@nvidia.com>

PR #428, opened 2015-11-10
Tegra: introduce per-soc system reset handler

vwadekar:
This patch adds a per-soc system reset handler for Tegra chips. The
handler gets executed before the actual system resets. This allows
for custom handling of the system reset sequence on each SoC.

### DIFF
--- a/plat/nvidia/tegra/common/tegra_pm.c
+++ b/plat/nvidia/tegra/common/tegra_pm.c
@@ -55,6 +55,7 @@ static int system_suspended;
 #pragma weak tegra_soc_prepare_cpu_on
 #pragma weak tegra_soc_prepare_cpu_off
 #pragma weak tegra_soc_prepare_cpu_on_finish
+#pragma weak tegra_soc_prepare_system_reset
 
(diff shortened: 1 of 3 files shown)


vwadekar:
Previous PR - #426

danh-arm:

danh-arm:

achingupta:
test this please

achingupta:
retest this please

achingupta:
@vwadekar 

Hi Varun,

The changes are Tegra specific so seem fine to me. Just one comment.

The SoC reset handlers for both T210 and T132 seem to be exactly the same. Even the constants that they depend upon have the same values. Is it not worth moving them to common code? 

I will wait for the CI job to finish. It seems that some github magic prevented it from triggering last night. 

Thanks,
Achin

vwadekar:
Hello Achin,

Even though the code is same for the current chips, it will change in the
future chips. In order to avoid porting one more function for future chips
I designed it this way. So newer chips start with the weak implementation
to start with and then we add details as we move along.

PR #430, opened 2015-11-17
Fix build error when `BL32` is not defined

jcastillo-arm:
If an SPD wants to use a prebuilt binary as BL32 image (for example,
the OPTEE Dispatcher), it must point the `BL32` variable to the
image file. This dependency should apply only to the `fip` target.
However, it also applies to the `all` target at the moment. If the
user tries to build all individual TF images using `make all`
without setting BL32, the build fails. The following command will
throw the error:

    make CROSS_COMPILE=aarch64-linux-gnu- SPD=opteed all
    ...
    ...
    aarch64-linux-gnu-gcc: fatal error: no input files
    compilation terminated.
    make: *** [build/fvp/release/bl32/bl32.ld] Error 1

The reason is that the build system checks if BL32 is defined, and
if it is not, it will try to build BL32 from source. If the SPD
makefile does not provide support for that (as is the case of the
OPTEE Dispatcher, since OPTEE is provided as an external binary),
the build will fail.

This patch fixes the issue by checking if `BL32_SOURCES` has been
defined by the SPD before attempting to build BL32 from source.
If neither `BL32` nor `BL32_SOURCES` is defined when building the
FIP, a warning message will be printed and the process aborted.

Fixes ARM-software/tf-issues#333

Change-Id: I5e801ad333103ed9b042e5c4757424c8df2ff6e4

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -240,10 +240,15 @@ ifneq (${SPD},none)
         $(info Including ${SPD_MAKE})
         include ${SPD_MAKE}
 
-        # If there's BL3-2 companion for the chosen SPD, and the SPD wants to build the
-        # BL3-2 from source, we expect that the SPD's Makefile would set NEED_BL32
-        # variable to "yes". In case the BL3-2 is a binary which needs to be included in
-        # fip, then the NEED_BL32 needs to be set and BL3-2 would need to point to the bin.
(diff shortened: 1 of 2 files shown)


jcastillo-arm:
retest this please

PR #443, opened 2015-11-26
Add support to boot EL3 payloads and only a single CPU at cold reset

achingupta:
This PR is a combination of two features:

*  Support to load and execute a binary file in EL3 instead of the BL3-1 image
*  Optimize the code footprint in case only a single CPU is released from reset at cold boot on a given platform


achingupta:
test this please

achingupta:
test this please

PR #447, opened 2015-12-02
Mediatek bl31

mtk09422:
Add plat parameter structures and memory configuration for Mediatek BL31

previous PR: https://github.com/ARM-software/arm-trusted-firmware/pull/436

PR #448, opened 2015-12-02
Spm

mtk09422:
Update SPM/DCM related code for power control logic

previous PR: https://github.com/ARM-software/arm-trusted-firmware/pull/437


danh-arm:

danh-arm:

danh-arm:
test this please

danh-arm:
Contains same build error as #447

jimmyh-mtk:
fix build error in #447, could you trigger the test build again? Thanks.

danh-arm:
retest this please

WeiyiLu:
Thanks for the information, it's indeed unnecessary to trace the on_pending state. We will submit a new commit for related code removal.

jimmyh-mtk:
We are currently discussing with Julius (@Google) on whether the regulator control code should be in BL31 or not. The mmu patch is needed for that. If it turns out that we shouldn't put those code in BL31, we then will need to update some code for this. We will not update the patches (mmu and regulator control related code) until we have a conclusion on this topic.

PR #441, opened 2015-11-26
Modify TSP and ARM standard platforms for new GIC drivers

soby-mathew:
This patch stack adds support for GICv3 interrupt model to Test Secure Payload
(TSP) and its dispatcher (TSPD). It also migrates the ARM Standard Platform
layer to the new GIC drivers. The support for EL3 interrupt type is added in the
Interrupt management framework as well.

PR #451, opened 2015-12-04
Modify TSP and ARM standard platforms for new GIC drivers v1

soby-mathew:
This patch stack adds support for GICv3 interrupt model to Test Secure Payload
(TSP) and its dispatcher (TSPD). It also migrates the ARM Standard Platform
layer to the new GIC drivers. The support for EL3 interrupt type is added in the
Interrupt management framework as well.


soby-mathew:
Previous PR #441

soby-mathew:
Thanks Dan for the comments. Issues fixed in the new PR #456.

PR #454, opened 2015-12-08
Add CCI-400 specific driver to deprecated driver list

yatharth-arm:
Add compile time `__warn_deprecated` flag to public api's in CCI-400
specific driver so that user is aware of the driver being deprecated.
Similarly, it also adds an error message when `ERROR_DEPRECATED` is set
to prevent succesful compilation if CCI-400 specific driver is used.

Change-Id: Id7e61a560262abc01cbbd432ca85b9bf448a194d


danh-arm:

danh-arm:

PR #459, opened 2015-12-09
Introduce the ARM TF reset design document

sandrine-bailleux:
This patch introduces a new document presenting the ARM Trusted
Firmware Reset Design. It shows the reset code flow, lists the
different build options that affect it, in which case to use them
and what their exact effect is.


sandrine-bailleux:
Hi Achin,

I have addressed your comments in this new pull request: #471 

Cheers,
Sandrine

PR #453, opened 2015-12-08
Firmware Update patch stack

yatharth-arm:

PR #461, opened 2015-12-09
Include psci.h from tegra platform header

yatharth-arm:
The `plat/nvidia/tegra/include/tegra_private.h` file uses resources
from psci.h (for example, psci_power_state_t) but does not explicitly
include psci.h. This does not currently cause a problem since psci.h
is indirectly included via other headers. However, this may not be
the case in future.

This patch explicitly includes psci.h from tegra_private.h

Change-Id: Ia991147898dbd117c1d3496a95850995a5554c05

### DIFF
--- a/plat/nvidia/tegra/include/tegra_private.h
+++ b/plat/nvidia/tegra/include/tegra_private.h
@@ -33,6 +33,7 @@
 
 #include <arch.h>
 #include <platform_def.h>
+#include <psci.h>
 #include <xlat_tables.h>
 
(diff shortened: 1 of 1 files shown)


danh-arm:

danh-arm:
@vwadekar
Hi Varun

We need to include this Tegra header fix before we merge #453. Let us know if you have any issue with this.

Regards

Dan.

vwadekar:
Thanks Dan. This change looks good. No issues.

PR #476, opened 2015-12-29
platform.h: Include common_def.h

sorenb-xlnx:
APIs in platform.h are marked as deprecated using the macro
'__warn_deprecated' which is defined in the header common_def.h.
Include that header from platform.h to avoid compilation errors like:
  In file included from plat/common/aarch64/plat_psci_common.c:33:0:
  include/plat/common/platform.h: In function 'platform_get_core_pos':
  include/plat/common/platform.h:276:57: error: expected declaration specifiers before '__warn_deprecated'
   unsigned int platform_get_core_pos(unsigned long mpidr) __warn_deprecated;
                                                           ^

Fixes ARM-software/tf-issues#340

Signed-off-by: Soren Brinkmann <soren.brinkmann@xilinx.com>

### DIFF
--- a/include/plat/common/platform.h
+++ b/include/plat/common/platform.h
@@ -31,6 +31,7 @@
 #ifndef __PLATFORM_H__
 #define __PLATFORM_H__
 
+#include <common_def.h>
 #include <psci.h>
 #include <stdint.h>
 #include <types.h>


danh-arm:

danh-arm:

achingupta:
I don't think the header should be included in `platform.h`. `common_def.h` is expected to be optionally included. [This](https://github.com/ARM-software/arm-trusted-firmware/blob/master/docs/porting-guide.md#file--platform_defh-mandatory) says that a platform port can optionally include this file for typical values of common constants. This is what the ARM FVP/Juno and Nvidia ports do as well. It would not be correct to include an optional file in a mandatory one. 

There is however a slight problem in the way this file has been included in the current code base. It is included by `context_mgmt.h`. This violates the optional nature of this file since it is being included in generic code but I see this as a separate problem.

sorenb-xlnx:
So, what do you suggest? Should *__warn_deprecated* move into platform.h or some other generic header?

achingupta:
I think this definition should be in a generic header but would like to understand the rationale behind putting it in `common_def.h` in the first place. This is a simple macro which should never need to be overridden. This will have to wait until next week as almost everyone is on holiday. I hope you will be able to make progress in the meantime?

sorenb-xlnx:
Sure, I have the "fix" I proposed here.

danh-arm:
I agree common_def.h is the wrong place for this definition. It's a bit odd that this definition exists at all since elsewhere in the codebase we use `__attribute__ ((XXX))` directly, which implicitly locks us to GCC-like compilers.

I think to fix this properly we should create a new `include/common/compiler_defs.h` to supplement `include/stdlib/sys/cdefs.h`. Also, we should remove direct usage of `__attribute__ ((XXX))` and either use the defintions in cdefs.h (e.g. __unused) or define new ones. This will help us be more portable for other compilers.

danh-arm:
Also, for future reference, this PR incorrectly targets the master branch instead of the integration branch. See https://github.com/ARM-software/arm-trusted-firmware/blob/master/contributing.md.

sorenb-xlnx:
OK. I'll close this PR then and we can move further discussions to the corresponding issue or new PRs.

PR #477, opened 2016-01-04
Add ret to return from console_uninit() function

mtk09422:
The 'ret' was missing in console_uninit() implementation, so the
program doesn't return from console_uninit(). Instead, it keeps
executing the following instructions which is not expected.

Change-Id: I810684f37f61c41c6f95a3bb36914d0765da8571
Signed-off-by: Jimmy Huang <jimmy.huang@mediatek.com>

### DIFF
--- a/drivers/console/console.S
+++ b/drivers/console/console.S
@@ -79,6 +79,7 @@ func console_uninit
 	mov	x0, #0
 	adrp	x3, console_base
 	str	x0, [x3, :lo12:console_base]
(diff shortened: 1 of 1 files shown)


danh-arm:

danh-arm:

danh-arm:
test this please

danh-arm:
Thanks for this fix; we should have noticed that one earlier.

PR #495, opened 2016-01-20
ARM plat: add build option to unlock access to non-secure timer

jcastillo-arm:
Currently, Trusted Firmware on ARM platforms unlocks access to the
timer frame registers that will be used by the Non-Secure world. This
unlock operation should be done by the Non-Secure software itself,
instead of relying on secure firmware settings.

This patch adds a new ARM specific build option 'ARM_CONFIG_CNTACR'
to unlock access to the timer frame by setting the corresponding
bits in the CNTACR<N> register. The frame id <N> is defined by
'PLAT_ARM_NSTIMER_FRAME_ID'. Default value is true (unlock timer
access).

Documentation updated accordingly.

Fixes ARM-software/tf-issues#170

Change-Id: Id9d606efd781e43bc581868cd2e5f9c8905bdbf6

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -456,6 +456,12 @@ map is explained in the [Firmware Design].
     Trusted Watchdog may be disabled at build time for testing or development
     purposes.
 
+*   `ARM_CONFIG_CNTACR`: boolean option to unlock access to the CNTBase<N>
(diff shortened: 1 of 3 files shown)


jcastillo-arm:
retest this please

PR #496, opened 2016-01-20
cdefs.h: Use __deprecated__ instead of deprecated

sorenb-xlnx:
Use the form with underscores to define the '__deprecated' macro to avoid
collisions with potentially defined macros, as suggested in gcc docs
(https://gcc.gnu.org/onlinedocs/gcc/Attribute-Syntax.html#Attribute-Syntax).

Signed-off-by: Soren Brinkmann <soren.brinkmann@xilinx.com>

### DIFF
--- a/include/stdlib/sys/cdefs.h
+++ b/include/stdlib/sys/cdefs.h
@@ -283,7 +283,7 @@
 
 #if __GNUC_PREREQ__(3, 1)
 #define	__noinline	__attribute__ ((__noinline__))
-#define	__deprecated	__attribute__ ((deprecated))
+#define	__deprecated	__attribute__ ((__deprecated__))
 #else
 #define	__noinline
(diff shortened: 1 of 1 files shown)


danh-arm:

danh-arm:

danh-arm:
test this please

danh-arm:
@sorenb-xlnx One other comment: If you like, you may add Xilinx to the list of contributors in the https://github.com/ARM-software/arm-trusted-firmware/blob/master/acknowledgements.md file in your next contribution. This is optional.

sorenb-xlnx:
Yes, I was planning to include such a change in our PR for ZynqMP
support. But that will take a little longer. I have merged v1.2 and also
seem to have the migration to the new platform API working. I hope to
get a PR prepared during the next few months.

	Sören

PR #502, opened 2016-01-28
build system: allow overriding the build's timestamp

pgeorgi:
This allows reproducible builds (same source and same compiler produce
bit-identical results) and also allows coordinating the timestamp across
multiple projects, eg. with another firmware.

### DIFF
--- a/make_helpers/build_macros.mk
+++ b/make_helpers/build_macros.mk
@@ -290,6 +290,11 @@ define MAKE_TOOL_ARGS
         $(if $(3),$(eval $(call FIP_ADD_PAYLOAD,$(2),--$(3),bl$(1))))
 endef
 
+# Allow overriding the timestamp, for example for reproducible builds, or to
+# synchronize timestamps across multiple projects.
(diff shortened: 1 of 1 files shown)


danh-arm:

danh-arm:

danh-arm:
Hi Patrick

Thanks for your contribution but we need your company to complete a CLA before we can accept any patches. See https://github.com/ARM-software/arm-trusted-firmware/blob/master/contributing.md.

Regards

Dan.

stefanreinauer:
You should have the CLA now. Please ping me (reinauer@google.com) if sth is missing.

danh-arm:
The target branch for all PRs should be "integration", not "master". You'll need to raise another PR to address this.

pgeorgi:
I'll reopen a new PR to get onto integration and fix the white spaces.
As for the extra quotes, I was wondering which version of BUILD_MESSAGE_TIMESTAMP looks less weird, but it was a coin toss for me. will go with your version, since that seems to be the more intuitive (sample size n=2).

PR #492, opened 2016-01-17
Support for Rockchip's RK3368 SoCs

rkchrome:
RK3368 is one of the Rockchip family SoCs, which is an
Octa-core ARM Cortex-A53 SoC.

This patch adds support to boot the Trusted Firmware on RK3368
SoCs, and adds support to boot secondary CPUs, enter/exit core
power states for all CPUs in the slow/fast clusters.

This is the initialize version for rockchip SoCs.(RK3368 and next SoCs)
* Support arm gicv2 & gicv3.
* Boot up 8 cores.
* Add a generic UART driver.
* Add generic CPU helper functions
* Support suspend/resume
* Add system_off & system_reset implementation
* Add delay timer platform implementation

Change-Id: I704bb3532d65e8c70dbd99b512c5e6e440ea6f43
Signed-off-by: xxx <xxx@rock-chips.com>
Signed-off-by: Caesar Wang <wxt@rock-chips.com>

PR #505, opened 2016-02-01
Support for Rockchip's RK3368 SoCs

rkchrome:
RK3368 is one of the Rockchip family SoCs, which is an
Octa-core ARM Cortex-A53 SoC.

This patch adds support to boot the Trusted Firmware on RK3368

This is the initialize version for rockchip SoCs.(RK3368 and next SoCs)
* Support arm gicv2 & gicv3.
* Boot up 8 cores.
* Add a generic UART driver.
* Support suspend/resume.

### DIFF
--- a/plat/rockchip/common/arm_gicv2.c
+++ b/plat/rockchip/common/arm_gicv2.c
@@ -1,0 +1,101 @@
+/*
+ * Copyright (c) 2016, ARM Limited and Contributors. All rights reserved.
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * Neither the name of ARM nor the names of its contributors may be used
+ * to endorse or promote products derived from this software without specific
+ * prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <bl_common.h>
+#include <gicv2.h>
+#include <platform_def.h>
+
+/******************************************************************************
+ * The following functions are defined as weak to allow a platform to override
+ * the way the GICv2 driver is initialised and used.
+ *****************************************************************************/
+#pragma weak plat_rockchip_gic_driver_init
+#pragma weak plat_rockchip_gic_init
+#pragma weak plat_rockchip_gic_cpuif_enable
+#pragma weak plat_rockchip_gic_cpuif_disable
+#pragma weak plat_rockchip_gic_pcpu_init
+
+/******************************************************************************
+ * On a GICv2 system, the Group 1 secure interrupts are treated as Group 0
+ * interrupts.
+ *****************************************************************************/
+const unsigned int g0_interrupt_array[] = {
+	PLAT_RK_G1S_IRQS,
+};
+
+/*
+ * Ideally `arm_gic_data` structure definition should be a `const` but it is
+ * kept as modifiable for overwriting with different GICD and GICC base when
+ * running on FVP with VE memory map.
+ */
+gicv2_driver_data_t arm_gic_data = {
+	.gicd_base = PLAT_RK_GICD_BASE,
+	.gicc_base = PLAT_RK_GICC_BASE,
+	.g0_interrupt_num = ARRAY_SIZE(g0_interrupt_array),
+	.g0_interrupt_array = g0_interrupt_array,
+};
+
+/******************************************************************************
+ * ARM common helper to initialize the GICv2 only driver.
+ *****************************************************************************/
+void plat_rockchip_gic_driver_init(void)
+{
+	gicv2_driver_init(&arm_gic_data);
+}
+
+void plat_rockchip_gic_init(void)
+{
+	gicv2_distif_init();
+	gicv2_pcpu_distif_init();
+	gicv2_cpuif_enable();
+}
+
+/******************************************************************************
+ * ARM common helper to enable the GICv2 CPU interface
+ *****************************************************************************/
+void plat_rockchip_gic_cpuif_enable(void)
+{
+	gicv2_cpuif_enable();
+}
+
+/******************************************************************************
+ * ARM common helper to disable the GICv2 CPU interface
+ *****************************************************************************/
+void plat_rockchip_gic_cpuif_disable(void)
+{
+	gicv2_cpuif_disable();
+}
+
+/******************************************************************************
+ * ARM common helper to initialize the per cpu distributor interface in GICv2
+ *****************************************************************************/
+void plat_rockchip_gic_pcpu_init(void)
+{
+	gicv2_pcpu_distif_init();
+}
--- a/plat/rockchip/common/arm_gicv3.c
+++ b/plat/rockchip/common/arm_gicv3.c
@@ -1,0 +1,115 @@
+/*
+ * Copyright (c) 2016, ARM Limited and Contributors. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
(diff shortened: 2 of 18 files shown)


danh-arm:

danh-arm:

rkchrome:
As the previous version on https://github.com/ARM-software/arm-trusted-firmware/pull/492

plat/rockchip/common/arm_gicv2.c
```diff
@@ -0,0 +1,101 @@
...
+
+/******************************************************************************
+ * The following functions are defined as weak to allow a platform to override
+ * the way the GICv2 driver is initialised and used.
+ *****************************************************************************/
+#pragma weak plat_rockchip_gic_driver_init
+#pragma weak plat_rockchip_gic_init
+#pragma weak plat_rockchip_gic_cpuif_enable
+#pragma weak plat_rockchip_gic_cpuif_disable
+#pragma weak plat_rockchip_gic_pcpu_init
```
danh-arm: minor: You've renamed these functions to plat_rockchip_xxx but you still use arm_ for arm_gic_data and the filename

plat/rockchip/rk3368/drivers/pmu/pmu.c
```diff
@@ -0,0 +1,247 @@
...
+static struct psram_data_t *psram_sleep_cfg =
+	(struct psram_data_t *)PSRAM_DT_BASE;
+
+/*****************************************************************************
+ * sram only surpport 32-bits access
+ ******************************************************************************/
+static void
+u32_align_cpy(unsigned int *dst, const unsigned int *scr, size_t len)
+{
+	uint32_t i;
```
danh-arm: minor: Strictly speaking, it would be better if `scr` was `uint32_t` (as it must be exactly 32-bits) and `i` were `unsigned int`

plat/rockchip/rk3368/drivers/pmu/pmu.c
```diff
@@ -0,0 +1,247 @@
...
+/*
+ * The struct is used in pmu_cpus_on.S which
+ * gets the data of the struct by the following index
+ * #define PSRAM_DT_SP		0x0
+ * #define PSRAM_DT_DDR_FUNC	0x8
+ * #define PSRAM_DT_DDR_DATA	0x10
+ * #define PSRAM_DT_DDRFLAG	0x18
+ * #define PSRAM_DT_SYS_MODE	0x1c
+ * #define PSRAM_DT_MPIDR	0x20
+ */
```
danh-arm: It would improve robustness if you added compile time asserts that these defines are correct. For example:
```
CASSERT(PSRAM_DT_SIZE == sizeof(psram_data_t), assert_psram_dt_size_mismatch);
CASSERT(PSRAM_DT_MPIDR == __builtin_offsetof(psram_data_t, boot_mpidr), assert_psram_dt_mpidr_offset_mistmatch);
...
```

plat/rockchip/rk3368/drivers/pmu/pmu.c
```diff
@@ -0,0 +1,247 @@
...
+	 */
+	if (pd_state == pmu_pd_on) {
+		cpus_id_power_domain(cluster, cpu, pmu_pd_off, CKECK_WFI_MSK);
+
+		cpuon_id = (cluster * PLATFORM_CLUSTER0_CORE_COUNT) + cpu;
+		assert(cpuson_flags[cpuon_id] == 0);
+		cpuson_flags[cpuon_id] = PMU_CPU_HOTPLUG;
+		cpuson_entry_point[cpuon_id] = entrypoint;
+		mmio_write_32(SGRF_BASE + SGRF_SOC_CON(1 + cluster),
+			      (PMUSRAM_BASE >> 16) | 0xffff0000);
```
danh-arm: minor: It would be a bit clearer if you didn't use magic masks/offsets and/or added a comment saying explaining what you're writing.

plat/rockchip/rk3368/drivers/pmu/pmu.c
```diff
@@ -0,0 +1,247 @@
...
+		dsb();
+	}
+
+	cpus_id_power_domain(cluster, cpu, pd_state, CKECK_WFI_MSK);
+
+	if (pd_state == pmu_pd_on) {
+		mmio_write_32(SGRF_BASE + SGRF_SOC_CON(1 + cluster),
+			      0xffff0000 | 0xffff);
```
danh-arm: ditto

plat/rockchip/rk3368/drivers/pmu/pmu.c
```diff
@@ -0,0 +1,247 @@
...
+
+	/*
+	 * pmu sram code and data prepare
+	 */
+	sram_dst = (uint32_t *)PMUSRAM_BASE;
+	sram_src = (uint32_t *)&pmu_cpuson_entrypoint_start;
+	sram_size = (uint32_t *)&pmu_cpuson_entrypoint_end -
+		    (uint32_t *)sram_src;
+
+	u32_align_cpy(sram_dst, sram_src, sram_size);
```
danh-arm: I understand now that pmu_cpuson_entrypoint must be executed from PMUSRAM, but this is still a potential security issue. Is there some way to prevent the normal world from mapping in this PMUSRAM (or the memory that BL31 executes in)? It would be even better if the hardware made this memory is only accessible to the secure world.

danh-arm: Actually, I've just seen this comment from the previous version:
```
1、The cpu reset or warm boot address in this soc needs aligned 64K bytes. so it can not be pointed to the entrypoit function in psci freely.
2、If the system resumes from system suspend，the boot code needs to be executed in pmusram in which we have the ddr initiated because of The DDR Controller power down.
So we program this position-independent code.
```
Regarding 1), it is possible to align code to 64KB by putting it in its own section and specifying the alignment in the linker script. We don't currently provide support for platform defined sections so you would need to modify the generic BL31 linker script. If you go this route then we should think more about providing linker script platform hooks to support your needs.

Regarding 2), this implies that your BL31 code is running from DDR. Is that correct? I was previously assuming that the memory you use for BL31 (starting at TZRAM_BASE) is on-chip RAM, or at least RAM that is available immediately after cold boot and when resuming from system suspend. If this is not the case, then it explains your workaround. Another alternative might be to run all of BL31 from PSURAM, but I'm not sure if it's big enough (seems to be 64KB).

plat/rockchip/rk3368/plat_delay_timer.c
```diff
@@ -0,0 +1,54 @@
...
+static const timer_ops_t plat_timer_ops = {
+	.get_timer_value	= plat_get_timer_value,
+	.clk_mult		= 1,
+	.clk_div		= SYS_COUNTER_FREQ_IN_MHZ,
+};
+
+void plat_delay_timer_init(void)
+{
+	timer_init(&plat_timer_ops);
+}
```
danh-arm: To reply to previous comment here:

> I think the generic delay timer implementation is indeed needed.
So, will you upstream for generic implementation?
I guess that's also need to consider the code on tegra/mtk platform, right?. 

We (ARM) could do it, but so could you. Given that you already have something, it seems easiest to use this as a starting point.

In any case, this does not need to delay upstreaming of your platform. We can make move this driver to a more generic location later.

plat/rockchip/common/arm_gicv3.c
```diff
@@ -0,0 +1,115 @@
+/*
```
danh-arm: There's no need to upstream this file - it is not used.

plat/rockchip/rk3368/platform.mk
```diff
@@ -0,0 +1,67 @@
...
+PLAT_INCLUDES		:=	-I${RK_PLAT_SOC}/				\
+				-I${RK_PLAT_SOC}/drivers/pmu/			\
+				-I${RK_PLAT_SOC}/drivers/soc/			\
+				-I${RK_PLAT_SOC}/include/
+
+RK_GIC_SOURCES         :=	drivers/arm/gic/common/gic_common.c		\
+				drivers/arm/gic/v2/gicv2_main.c			\
+				drivers/arm/gic/v2/gicv2_helpers.c		\
+				plat/common/plat_gicv2.c			\
+				${RK_PLAT}/common/arm_gicv2.c
```
danh-arm: You said in the last version that you are not integrating a Secure-EL1 Payload. Also you are not calling the GIC initialization code. It seems to me you can remove all GIC and interrupt handling code from your port.

danh-arm:
The commit and PR messages looks like they need updating. For example:

"Support arm gicv2 & gicv3." gicv3 has been removed and it seems to me the gicv2 stuff can be too.
"Add a generic UART driver." You're now using the already upstream one.
"Support suspend/resume." It looks like CPU suspend is not supported yet. Do you mean system suspend?
"Add delay timer platform implementation.". It's actually a generic timer implementation of the delay timer API that happens to be in the platform port.

plat/rockchip/rk3368/plat_pm.c
```diff
@@ -0,0 +1,271 @@
...
+{
+	wfi();
+	ERROR("RockChip System Off: operation not handled.\n");
+	panic();
+}
+
+static void __dead2 rockchip_system_reset(void)
+{
+	wfi();
+	ERROR("RockChip System Reset: operation not handled.\n");
```
mmind: System reset could be done via the CRU-based reset, like in the kernel. Or is something speaking against this?

plat/rockchip/common/arm_gicv2.c
rkchrome: Done

plat/rockchip/rk3368/drivers/pmu/pmu.c
rkchrome: Done

plat/rockchip/rk3368/drivers/pmu/pmu.c
rkchrome: Done

plat/rockchip/rk3368/drivers/pmu/pmu.c
rkchrome: Done

plat/rockchip/rk3368/drivers/pmu/pmu.c
rkchrome: Done

plat/rockchip/rk3368/drivers/pmu/pmu.c
rkchrome: 1. The memory for Bl31 ( a regions in ddr and PMUSRAM) is accessible to the secure world by hardware made.
2. For regarding 1),We expect that you will provide linker script platform hooks to us, It will make the work easier.
3. For regarding 2),All of Bl31 runs from ddr, we have not enough on-chip ram for bl31, PMUSURAM only is 4K byte for cold boot and ddr controlling resume.

plat/rockchip/common/arm_gicv3.c
rkchrome: Will be used by next SoCs

plat/rockchip/rk3368/platform.mk
rkchrome: I think that system needs to config all interrupt in group1 since the reset value is in group0.

rkchrome: We think that system needs to config all interrupt in group1 since the reset value is in group0.

plat/rockchip/rk3368/plat_pm.c
rkchrome: CPU reset/suspend/resume are completed in the newer pull.

rkchrome:
Anyway, the new pull patch at https://github.com/ARM-software/arm-trusted-firmware/pull/536

PR #506, opened 2016-02-01
Upstream Mediatek MT6795 platform code

coboy-chen:
Change-Id: I4dd4398e83f61280e4bf1b9c32892a49edac7a83


coboy-chen:
This patch is mainly used to support for MediaTek MT6795 chipset and has already passed internally review.

danh-arm:

danh-arm:

danh-arm:
Could do with a bit more information in the commit message, e.g. a high level description of the SoC features and what software features this initial port supports.

plat/mediatek/common/custom/oem_svc.c
```diff
@@ -0,0 +1,130 @@
...
+#include <stdint.h>
+#include <uuid.h>
+
+/* OEM Service UUID */
+DEFINE_SVC_UUID(oem_svc_uid,
+		0xb943add0, 0x069d, 0x11e4, 0x91, 0x91,
+		0x08, 0x00, 0x20, 0x0c, 0x9a, 0x66);
+
+
+/* Setup Standard Services */
```
danh-arm: minor: Setup OEM services

plat/mediatek/common/custom/oem_svc.c
```diff
@@ -0,0 +1,130 @@
...
+		rc = SMC_UNK;
+		WARN("Unimplemented OEM Call: 0x%x\n", smc_fid);
+	}
+
+	SMC_RET1(handle, rc);
+}
+
+/*
+ * Top-level Standard Service SMC handler. This handler will in turn dispatch
+ * calls to related SMC handler
```
danh-arm: Minor: Top-level OEM Service SMC handler...

plat/mediatek/common/custom/oem_svc.c
```diff
@@ -0,0 +1,130 @@
...
+	if (is_oem_fid(smc_fid)) {
+		return oem_smc_handler(smc_fid, x1, x2, x3, x4, cookie,
+				handle, flags);
+	}
+
+	switch (smc_fid) {
+	case OEM_SVC_CALL_COUNT:
+		/*
+		 * Return the number of Standard Service Calls. PSCI is the only
+		 * standard service implemented; so return number of PSCI calls
```
danh-arm: minor: Not PSCI calls, OEM service calls.

plat/mediatek/common/mtk_sip_svc.h
```diff
@@ -62,5 +100,6 @@ enum {
  * Return MTK_SIP_E_SUCCESS on success, and MTK_SIP_E_INVALID_PARAM on failure.
  */
 uint64_t mt_sip_set_authorized_sreg(uint32_t sreg, uint32_t val);
+void save_kernel_info(uint64_t pc, uint64_t r0, uint64_t r1, uint64_t k32_64);
```
danh-arm: Minor: Seems odd to declare this in a common header then define it in the platform specific code.

plat/mediatek/mt6795/aarch64/platform_common.c
```diff
@@ -0,0 +1,148 @@
...
+
+/* Table of regions to map using the MMU.  */
+const mmap_region_t plat_mmap[] = {
+	/* for TF text, RO, RW */
+	MAP_REGION_FLAT(TZRAM_BASE, TZRAM_SIZE + TZRAM2_SIZE,
+			MT_MEMORY | MT_RW | MT_SECURE),
+	MAP_REGION_FLAT(MTK_DEV_RNG0_BASE, MTK_DEV_RNG0_SIZE,
+			MT_DEVICE | MT_RW | MT_SECURE),
+	MAP_REGION_FLAT(MTK_DEV_RNG1_BASE, MTK_DEV_RNG1_SIZE,
+			MT_DEVICE | MT_RW | MT_SECURE),
```
danh-arm: Does BL31 really need to map in these RNGs? They don't seem to be used.

plat/mediatek/mt6795/bl31_plat_setup.c
```diff
@@ -0,0 +1,483 @@
...
+	/*
+	 * Tell BL3-1 where the non-trusted software image
+	 * is located and the entry state information
+	 */
+	/* BL33_START_ADDRESS */
+	bl33_image_ep_info.pc = pmtk_bl_param_t->bl33_start_addr;
+	bl33_image_ep_info.spsr = plat_get_spsr_for_bl33_entry();
+	bl33_image_ep_info.args.arg4 =  pmtk_bl_param_t->bootarg_loc;
+	bl33_image_ep_info.args.arg5 =  pmtk_bl_param_t->bootarg_size;
+	SET_SECURITY_STATE(bl33_image_ep_info.h.attr, NON_SECURE);
```
danh-arm: So you have a mechanism to initialize bl33_image_ep_info with information passed from your preloader. But when it comes passing control to BL33, the BL33 image ep info is initialized again with information provided by a SiP SMC. Is there any point retaining the initialization here?

plat/mediatek/mt6795/drivers/rtc/rtc.c
```diff
@@ -0,0 +1,103 @@
+/*
```
danh-arm: Could this driver be shared with mt8173 (and placed in plat/mediatek/common)?

plat/mediatek/mt6795/drivers/timer/mt_cpuxgpt.c
```diff
@@ -0,0 +1,129 @@
+/*
```
danh-arm: Could this driver be shared with mt8173 (and placed in plat/mediatek/common)?

plat/mediatek/mt6795/drivers/uart/8250_console.S
```diff
@@ -0,0 +1,174 @@
+/*
```
danh-arm: Could this driver be shared with mt8173 (and placed in plat/mediatek/common)?

plat/mediatek/mt6795/include/mcucfg.h
```diff
@@ -0,0 +1,180 @@
+/*
```
danh-arm: Could this header be shared with mt8173 (and placed in plat/mediatek/common)?

plat/mediatek/mt6795/include/power_tracer.h
```diff
@@ -0,0 +1,43 @@
+/*
```
danh-arm: Could headers like this be shared with mt8173 (and placed in plat/mediatek/common)?

plat/mediatek/mt6795/plat_pm.c
```diff
@@ -0,0 +1,516 @@
+/*
```
danh-arm: This looks very similar to the mt8173 implementation

plat/mediatek/common/mtk_sip_svc.c
```diff
@@ -56,11 +72,9 @@ static uint64_t mediatek_sip_handler(uint32_t smc_fid,
 		SMC_RET1(handle, ret);
 
 	default:
-		ERROR("%s: unhandled SMC (0x%x)\n", __func__, smc_fid);
-		break;
+		return mediatek_plat_sip_handler(smc_fid, x1, x2, x3, x4,
+			cookie, handle, flags);
```
danh-arm: It seems strange to me to provide a hook here from plat/mediatek/common code to the platform specific code, but then not share any other code with the other upstream MTK platform. Not even the mt_sip_set_authorized_sreg() implementation. You might as well provide separate SiP service code (or better, do more code sharing).

plat/mediatek/mt6795/bl31.ld.S
```diff
@@ -0,0 +1,198 @@
+/*
```
danh-arm: I can see why you've needed to fork the linker script but at some point it would be good to understand if there are any platform hooks we could provide in the generic linker script to help you do what you want without duplicating this whole file.

plat/mediatek/mt6795/drivers/pmic/pmic_wrap_init.c
```diff
@@ -0,0 +1,188 @@
+/*
```
danh-arm: Could this driver be shared with mt8173 (and placed in plat/mediatek/common)?

plat/mediatek/mt6795/bl31_plat_setup.c
danh-arm: Perhaps it's to retain support for booting into BL33 when there is no BL32?

plat/mediatek/mt6795/aarch64/platform_common.c
```diff
@@ -0,0 +1,148 @@
...
+}
+
+uint32_t plat_get_spsr_for_bl32_entry(void)
+{
+	/*
+	 * The Secure Payload Dispatcher service is responsible for
+	 * setting the SPSR prior to entry into the BL3-2 image.
+	 */
+	return 0;
+}
```
danh-arm: So do you require the presence of an SPD or is this optional? Is there a plan to upstream that? The boot sequence in the presence of an SPD is a bit hard to understand.

plat/mediatek/mt6795/plat_sip_svc.c
```diff
@@ -0,0 +1,122 @@
...
+		break;
+#endif
+	case MTK_SIP_KERNEL_WDT_AARCH32:
+	case MTK_SIP_KERNEL_WDT_AARCH64:
+		wdt_kernel_cb_addr = x1;
+		INFO("MTK_SIP_KERNEL_WDT : 0x%lx\n",  wdt_kernel_cb_addr);
+		INFO("teearg->atf_aee_debug_buf_start : 0x%x\n",
+					 teearg->atf_aee_debug_buf_start);
+		rc = teearg->atf_aee_debug_buf_start;
+		break;
```
danh-arm: It looks to me like all these SiP calls are intended for use by Secure-EL1 Payload, in which case you should return an error for non-secure callers (like you do for MTK_SIP_TBASE_HWUID_AARCH32).

plat/mediatek/mt6795/plat_sip_svc.c
```diff
@@ -0,0 +1,122 @@
...
+	default:
+		rc = SMC_UNK;
+		console_init(gteearg.atf_log_port,
+			MT6795_UART_CLOCK, MT6795_BAUDRATE);
+		ERROR("%s: unhandled SMC (0x%x)\n", __func__, smc_fid);
+		console_uninit();
+		assert(0);
```
danh-arm: So you're saying it's a programming error to reach here, but you are not preventing a non-secure caller from executing this code path. It's probably better not to assert in this case, even for secure callers.

plat/mediatek/mt6795/plat_topology.c
```diff
@@ -0,0 +1,57 @@
+/*
```
danh-arm: Consider migrating to the new PSCI platform porting interface. See docs/platform-migration-guide.md.

plat/mediatek/mt6795/platform.mk
```diff
@@ -0,0 +1,105 @@
...
+				${OEMS_INCLUDES}
+
+PLAT_BL_COMMON_SOURCES	:=	lib/aarch64/xlat_tables.c			\
+				plat/common/aarch64/plat_common.c		\
+				plat/common/plat_gic.c
+
+BL31_SOURCES		+=	drivers/arm/cci/cci.c				\
+				drivers/arm/gic/arm_gic.c			\
+				drivers/arm/gic/gic_v2.c \
+				drivers/arm/gic/gic_v3.c \
```
danh-arm: This GIC driver is deprecated. Consider migrating to the new GICv2 optimised driver (drivers/arm/gic/v2/*)

plat/mediatek/mt6795/platform.mk
```diff
@@ -0,0 +1,105 @@
...
+				drivers/console/console.S			\
+				drivers/delay_timer/delay_timer.c		\
+				lib/cpus/aarch64/aem_generic.S			\
+				lib/cpus/aarch64/cortex_a53.S			\
+				lib/cpus/aarch64/cortex_a57.S			\
+				lib/cpus/aarch64/cortex_a72.S			\
```
danh-arm: Do you really have 4 types of CPU in your system? Can probably remove some of these CPU specific libraries (at least aem_generic.S).

plat/mediatek/mt6795/aarch64/plat_helpers.S
```diff
@@ -0,0 +1,168 @@
...
+	 * void plat_secondary_cold_boot_setup (void);
+	 *
+	 * This function performs any platform specific actions
+	 * needed for a secondary cpu after a cold reset e.g
+	 * mark the cpu's presence, mechanism to place it in a
+	 * holding pen etc.
+	 * -----------------------------------------------------
+	 */
+func plat_secondary_cold_boot_setup
+	/* MT6795 Oak does not do cold boot for secondary CPU */
```
danh-arm: You might want to remove the Oak reference

plat/mediatek/common/custom/oem_svc.c
```diff
@@ -0,0 +1,130 @@
...
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+#include <arch.h>
```
jcastillo-arm: minor: sort headers alphabetically

plat/mediatek/common/custom/oem_svc.c
```diff
@@ -0,0 +1,130 @@
...
+{
+	/*
+	 * Invoke related module setup from here
+	 */
+
+	return 0;
+}
+
+/*******************************************************************************
+ * OEM top level handler for servicing SMCs.
```
jcastillo-arm: minor: comment needs update (top level handler is oem_svc_smc_handler)

plat/mediatek/common/custom/oem_svc.c
```diff
@@ -0,0 +1,130 @@
...
+		/* Return the version of current implementation */
+		SMC_RET2(handle, OEM_VERSION_MAJOR, OEM_VERSION_MINOR);
+
+	default:
+		WARN("Unimplemented OEM Service Call: 0x%x\n", smc_fid);
+		SMC_RET1(handle, SMC_UNK);
+	}
+}
+
+/* Register Standard Service Calls as runtime service */
```
jcastillo-arm: minor: Register OEM Service Calls

plat/mediatek/common/mtk_sip_svc.h
```diff
@@ -33,20 +33,58 @@
...
+
+struct kernel_info {
+	uint64_t pc;
+	uint64_t r0;
+	uint64_t r1;
+	uint64_t r2;
+	uint64_t k32_64;
+};
+
+typedef struct mtk_bl_param {
```
jcastillo-arm: This structure is not related to SiP services. It should be moved to a common file.

plat/mediatek/mt6795/aarch64/plat_helpers.S
```diff
@@ -0,0 +1,168 @@
...
+	.globl	plat_crash_console_putc
+	.globl	platform_mem_init
+
+
+	.macro crash_ram_log
+	 /*
+	 * Check teearg->atf_log_buf_size.
+	 * Exit if atf_log_buf_size equals 0
+	 */
+	adr	x2, ptr_atf_crash_flag
```
jcastillo-arm: ptr_atf_crash_flag: label not defined

plat/mediatek/mt6795/aarch64/plat_helpers.S
```diff
@@ -0,0 +1,168 @@
...
+1:
+	adr	x2, ptr_atf_crash_flag
+	ldr	x2, [x2]
+	mov_imm x1, 0xdead1abf
+	/* p_atf_log_ctrl->atf_crash_flag = 0xdead1abf */
+	str	w1, [x2]
+	/* can't use w3 return addr, w4, start of buffer addr */
+	ldr	w2, [x2]
+	cmp	w2, w1
+	b.ne	1b
```
jcastillo-arm: Question: why a Read-After-Write loop is required to set the magic number?

plat/mediatek/mt6795/aarch64/plat_helpers.S
```diff
@@ -0,0 +1,168 @@
...
+	/* exit if ptr_atf_crash_flag equals NULL */
+	cbz x2, exit_putc
+
+	/*
+	 * set atf crash magic number
+	 */
+1:
+	adr	x2, ptr_atf_crash_flag
+	ldr	x2, [x2]
+	mov_imm x1, 0xdead1abf
```
jcastillo-arm: minor: I think these three sentences can be moved out of the loop

plat/mediatek/mt6795/aarch64/plat_helpers.S
```diff
@@ -0,0 +1,168 @@
...
+endfunc platform_is_primary_cpu
+
+	/* ---------------------------------------------
+	 * int plat_crash_console_init(void)
+	 * Function to initialize the crash console
+	 * without a C Runtime to print crash report.
+	 * Clobber list : x0, x1, x2
+	 * ---------------------------------------------
+	 */
+func plat_crash_console_init
```
jcastillo-arm: From Porting Guide: "It should only use the general purpose registers x0 to x2 to do the
initialization and returns 1 on success."

plat/mediatek/mt6795/aarch64/plat_helpers.S
```diff
@@ -0,0 +1,168 @@
...
+
+	/* --------------------------------------------------------
+	 * void platform_mem_init (void);
+	 *
+	 * Any memory init, relocation to be done before the
+	 * platform boots. Called very early in the boot process.
+	 * --------------------------------------------------------
+	 */
+func platform_mem_init
+	mov	x0, #0
```
jcastillo-arm: minor: platform_mem_init() does not return a value

plat/mediatek/mt6795/aarch64/platform_common.c
```diff
@@ -0,0 +1,148 @@
+/*
```
jcastillo-arm: The filename suggests a common file to Mediatek platforms, but there is one for each platform and part of the code is duplicated. Could this file be placed in plat/mediatek/common and shared with mt8173?

plat/mediatek/mt6795/bl31_plat_setup.c
```diff
@@ -0,0 +1,483 @@
...
+ * parameters passed by the calling EL (S-EL1 in BL2 & S-EL3 in BL1) before they
+ * are lost (potentially). This needs to be done before the MMU is initialized
+ * so that the memory layout can be used while creating page tables.
+ * BL2 has flushed this information to memory, so we are guaranteed to pick up
+ * good data.
+ ******************************************************************************/
+void bl31_early_platform_setup(bl31_params_t *from_bl2,
+						 void *plat_params_from_bl2)
+{
+	mtk_bl_param_t *pmtk_bl_param_t = (mtk_bl_param_t *)from_bl2;
```
jcastillo-arm: minor: suffix '_t' is usually for datatypes. Suggest '*pmtk_bl_param' for the pointer.

plat/mediatek/mt6795/bl31_plat_setup.c
```diff
@@ -0,0 +1,483 @@
...
+	/* Initialize the gic cpu and distributor interfaces */
+	plat_mt_gic_init();
+	arm_gic_setup();
+
+	/* Topologies are best known to the platform. */
+	mt_setup_topology();
+}
+/*******************************************************************************
+ * Perform the very early platform specific architectural setup here. At the
+ * moment this is only intializes the mmu in a quick and dirty way.
```
jcastillo-arm: minor: comment needs update (the function does more than that).

plat/mediatek/mt6795/bl31_plat_setup.c
```diff
@@ -0,0 +1,483 @@
...
+		 * add TZRAM_BASE to memory map
+		 * then set RO and COHERENT to different attribute
+		 */
+		plat_configure_mmu_el3(
+			BL31_RO_BASE & ~(PAGE_SIZE_MASK),
+			((TZRAM_SIZE & ~(PAGE_SIZE_MASK)) + PAGE_SIZE),
+			(BL31_RO_BASE & ~(PAGE_SIZE_MASK)),
+			BL31_RO_LIMIT,
+			BL31_COHERENT_RAM_BASE,
+			BL31_COHERENT_RAM_LIMIT);
```
jcastillo-arm: minor: tab alignment

plat/mediatek/mt6795/bl31_plat_setup.c
```diff
@@ -0,0 +1,483 @@
...
+
+	console_init(teearg->atf_log_port, MT6795_UART_CLOCK, MT6795_BAUDRATE);
+	memcpy((void *)&gteearg, (void *)teearg, sizeof(atf_arg_t));
+
+	/* init system counter in ATF */
+	setup_syscnt();
+
+	VERBOSE("bl31_setup\n");
+
+	assert(from_bl2 != NULL);
```
jcastillo-arm: Assert should be moved before line 156 (and ideally at the beginning of the function)

plat/mediatek/mt6795/bl31_plat_setup.c
```diff
@@ -0,0 +1,483 @@
...
+	/* Topologies are best known to the platform. */
+	mt_setup_topology();
+}
+/*******************************************************************************
+ * Perform the very early platform specific architectural setup here. At the
+ * moment this is only intializes the mmu in a quick and dirty way.
+ ******************************************************************************/
+void bl31_plat_arch_setup(void)
+{
+	atf_arg_t_ptr teearg = &gteearg;
```
jcastillo-arm: minor: can use gteearg directly instead of dereferencing a pointer.

plat/mediatek/mt6795/plat_sip_svc.c
```diff
@@ -0,0 +1,122 @@
...
+				uint64_t x2,
+				uint64_t x3,
+				uint64_t x4,
+				void *cookie,
+				void *handle,
+				uint64_t flags)
+{
+	uint64_t rc = 0;
+	uint32_t ns;
+	atf_arg_t_ptr teearg = &gteearg;
```
jcastillo-arm: minor: can use gteearg directly instead of dereferencing a pointer (as used in line 115)

plat/mediatek/mt6795/bl31_plat_setup.c
```diff
@@ -0,0 +1,483 @@
...
+	bl33_image_ep_info.spsr = plat_get_spsr_for_bl33_entry();
+	bl33_image_ep_info.args.arg4 =  pmtk_bl_param_t->bootarg_loc;
+	bl33_image_ep_info.args.arg5 =  pmtk_bl_param_t->bootarg_size;
+	SET_SECURITY_STATE(bl33_image_ep_info.h.attr, NON_SECURE);
+}
+/*******************************************************************************
+ * Perform any BL3-1 platform setup code
+ ******************************************************************************/
+
+const unsigned int mt_irq_sec_array[] = {
```
jcastillo-arm: minor: static

plat/mediatek/mt6795/bl31_plat_setup.c
```diff
@@ -0,0 +1,483 @@
...
+			k_info.r1 = r1;
+	}
+}
+
+void set_kernel_k32_64(uint64_t k32_64)
+{
+	k_info.k32_64 = k32_64;
+}
+
+uint64_t get_kernel_info_pc(void)
```
jcastillo-arm: minor: static

plat/mediatek/mt6795/bl31_plat_setup.c
```diff
@@ -0,0 +1,483 @@
...
+{
+	return k_info.pc;
+}
+
+uint64_t get_kernel_info_r0(void)
```
jcastillo-arm: minor: static

plat/mediatek/mt6795/bl31_plat_setup.c
```diff
@@ -0,0 +1,483 @@
...
+{
+	return k_info.r0;
+}
+
+uint64_t get_kernel_info_r1(void)
```
jcastillo-arm: minor: static

plat/mediatek/mt6795/bl31_plat_setup.c
```diff
@@ -0,0 +1,483 @@
...
+{
+	return k_info.r1;
+}
+uint64_t get_kernel_info_r2(void)
```
jcastillo-arm: minor: static

plat/mediatek/mt6795/bl31_plat_setup.c
```diff
@@ -0,0 +1,483 @@
...
+{
+	return k_info.r2;
+}
+
+
+entry_point_info_t *bl31_plat_get_next_kernel64_ep_info(void)
```
jcastillo-arm: minor: static

plat/mediatek/mt6795/bl31_plat_setup.c
```diff
@@ -0,0 +1,483 @@
...
+	SET_SECURITY_STATE(next_image_info->h.attr, NON_SECURE);
+
+	/* None of the images on this platform can have 0x0 as the entrypoint */
+	if (next_image_info->pc)
+		return next_image_info;
+	else
+		return NULL;
+}
+
+entry_point_info_t *bl31_plat_get_next_kernel32_ep_info(void)
```
jcastillo-arm: minor: static

plat/mediatek/common/custom/oem_svc.c
leon-chen-mtk: We will fix these comments.

plat/mediatek/common/custom/oem_svc.c
leon-chen-mtk: We will fix these comments.

plat/mediatek/common/custom/oem_svc.c
leon-chen-mtk: We will fix these comments.

plat/mediatek/common/mtk_sip_svc.h
leon-chen-mtk: We will move this function to mediatek/common/platform_common.c

plat/mediatek/mt6795/aarch64/platform_common.c
leon-chen-mtk: These regions are used.
GIC, clock controller, power management related function are located in MTK_DEV_RNG0 range.
Peripheral controllers including UART, SPI are located in MTK_DEV_RNG1 range.

plat/mediatek/mt6795/bl31_plat_setup.c
leon-chen-mtk: So far MTK firmware boot to Linux kernel is:
Without Secure-OS
MTK BootRom->MTK Preloader->ARM-TF->LK and load linux boot image->SMC back to ARM-TF->Linux kernel

With Secure-OS
MTK BootRom->MTK Preloader->ARM-TF->Enter BL32 init->SMC back to ARM-TF->LK and load linux boot image->SMC back to ARM-TF->Linux kernel

This upstream does not include BL32 part.
As above description, bl33_image_ep_info struct will be used as LK and Linux kernel entry point information container. It is why bl33_image_ep_info is filled again in SiP SMC.

So we just re-use bl33_image_ep_info for BL33 Linux kernel entry information. Do you have any suggestion?

plat/mediatek/mt6795/drivers/rtc/rtc.c
leon-chen-mtk: We will do driver code sharing as many as possbile with other platform team

plat/mediatek/mt6795/drivers/timer/mt_cpuxgpt.c
leon-chen-mtk: We will do driver code sharing as many as possbile with other platform team

plat/mediatek/mt6795/drivers/uart/8250_console.S
leon-chen-mtk: We will do driver code sharing as many as possbile with other platform team

plat/mediatek/mt6795/include/mcucfg.h
leon-chen-mtk: We will do driver code sharing as many as possbile with other platform team

plat/mediatek/mt6795/include/power_tracer.h
leon-chen-mtk: We will do driver code sharing as many as possbile with other platform team

plat/mediatek/mt6795/plat_pm.c
leon-chen-mtk: We will do driver code sharing as many as possbile with other platform team

plat/mediatek/common/mtk_sip_svc.c
leon-chen-mtk: We will do more code sharing with other platform team. We move common functions  to common folder and define these functions as weak functions. If the same functions need to be customized by platform, re-implement these weak functions.

plat/mediatek/mt6795/bl31.ld.S
leon-chen-mtk: So far we put ARM-TF in internal SRAM, but some platform's SRAM space is not enough. We place xlat_table and coherent data in another memory space. In that case, we need to modify bl31 linker script to put data in another memory region.

If ARM can provide a way for platform linker script cusomization, that will be good.

plat/mediatek/mt6795/drivers/pmic/pmic_wrap_init.c
leon-chen-mtk: We will do driver code sharing as many as possbile with other platform team

plat/mediatek/mt6795/aarch64/platform_common.c
leon-chen-mtk: We will remove plat_get_spsr_for_bl32_entry() function. Secure-OS setup function will assign bl32 spsr.

plat/mediatek/mt6795/plat_sip_svc.c
leon-chen-mtk: So far our sip_smc_handler() do not service non-secure callers
/* Determine which security state this SMC originated from */
 ns = is_caller_non_secure(flags);
 if (!ns)
  SMC_RET1(handle, SMC_UNK);

We will remove this limitation, since Secure-OS may need EL3 SiP service. We will return error for non-secure callers using Secure-OS SiP service.

plat/mediatek/mt6795/plat_sip_svc.c
leon-chen-mtk: We will follow SMC calling convention

plat/mediatek/mt6795/plat_topology.c
leon-chen-mtk: We will follow your suggestion

plat/mediatek/mt6795/platform.mk
leon-chen-mtk: We will follow your suggestion

plat/mediatek/mt6795/platform.mk
leon-chen-mtk: We will remove
lib/cpus/aarch64/cortex_a57.S 
lib/cpus/aarch64/cortex_a72.S
lib/cpus/aarch64/aem_generic.S

plat/mediatek/mt6795/aarch64/plat_helpers.S
leon-chen-mtk: We will fix these comments

plat/mediatek/common/custom/oem_svc.c
leon-chen-mtk: We will re-order header files.

plat/mediatek/common/custom/oem_svc.c
leon-chen-mtk: We will fix these comments.

plat/mediatek/common/custom/oem_svc.c
leon-chen-mtk: We will fix these comments.

plat/mediatek/common/mtk_sip_svc.h
leon-chen-mtk: We will move mtk_bl_param to a common header file

plat/mediatek/mt6795/aarch64/plat_helpers.S
leon-chen-mtk: We will remove MTK_RAM_LOG related code

plat/mediatek/mt6795/aarch64/plat_helpers.S
leon-chen-mtk: We will remove MTK_RAM_LOG related code

plat/mediatek/mt6795/aarch64/plat_helpers.S
leon-chen-mtk: We will remove MTK_RAM_LOG related code

plat/mediatek/mt6795/aarch64/plat_helpers.S
leon-chen-mtk: We will remove crash x9 code

plat/mediatek/mt6795/aarch64/plat_helpers.S
leon-chen-mtk: We will remove "mov x0, #0"

plat/mediatek/mt6795/aarch64/platform_common.c
leon-chen-mtk: We will do more code sharing with other platform, put common code in mediatek/common/platform_common.c

plat/mediatek/mt6795/bl31_plat_setup.c
leon-chen-mtk: We will remove _t suffux

plat/mediatek/mt6795/bl31_plat_setup.c
leon-chen-mtk: We will add more comments for this function.

plat/mediatek/mt6795/bl31_plat_setup.c
leon-chen-mtk: We will tab align.

plat/mediatek/mt6795/bl31_plat_setup.c
leon-chen-mtk: We will move assert macro to the beginning of this function

plat/mediatek/mt6795/bl31_plat_setup.c
leon-chen-mtk: We will follow your suggestion

plat/mediatek/mt6795/plat_sip_svc.c
leon-chen-mtk: We will follow your suggestion

plat/mediatek/mt6795/bl31_plat_setup.c
leon-chen-mtk: We will follow your suggestion. Define following data/functions as static.
const unsigned int mt_irq_sec_array[] = {
uint64_t get_kernel_info_pc(void)
uint64_t get_kernel_info_r0(void)
uint64_t get_kernel_info_r1(void)
uint64_t get_kernel_info_r2(void)
entry_point_info_t *bl31_plat_get_next_kernel64_ep_info(void)
entry_point_info_t *bl31_plat_get_next_kernel32_ep_info(void)

plat/mediatek/mt6795/bl31_plat_setup.c
leon-chen-mtk: We will follow your suggestion

plat/mediatek/mt6795/bl31_plat_setup.c
leon-chen-mtk: We will follow your suggestion

plat/mediatek/mt6795/bl31_plat_setup.c
leon-chen-mtk: We will follow your suggestion

plat/mediatek/mt6795/bl31_plat_setup.c
leon-chen-mtk: We will follow your suggestion

plat/mediatek/mt6795/bl31_plat_setup.c
leon-chen-mtk: We will follow your suggestion

plat/mediatek/mt6795/bl31_plat_setup.c
leon-chen-mtk: We will follow your suggestion

plat/mediatek/mt6795/bl31.ld.S
danh-arm: I've added an issue for this: https://github.com/ARM-software/tf-issues/issues/364
But we haven't committed to doing this yet.

plat/mediatek/mt6795/bl31_plat_setup.c
danh-arm: OK, now I understand. This is an unusual boot flow. I presume this is because your "LK and load linux boot image" executes at a different register width than the Linux kernel? Is this a temporary measure until you've ported your linux boot image to AArch64, or do you have a need to support both AArch32 and AArch64 kernels?

This is another area where we might need to consider more architected software support.

plat/mediatek/mt6795/plat_sip_svc.c
```diff
@@ -0,0 +1,122 @@
...
+				SMC_RET1(handle, SMC_UNK);
+		SMC_RET4(handle, teearg->hwuid[0], teearg->hwuid[1],
+				teearg->hwuid[2], teearg->hwuid[3]);
+		break;
+	case MTK_SIP_KERNEL_BOOT_AARCH32:
+		wdt_kernel_cb_addr = 0;
+		console_init(gteearg.atf_log_port,
+			MT6795_UART_CLOCK, MT6795_BAUDRATE);
+		INFO("save kernel info\n");
+		save_kernel_info(x1, x2, x3, x4);
```
danh-arm: x1 is used as the PC for the next non-secure image but you don't do any validation of this argument. You should check that this value lies within an expected range (e.g. not in secure memory).

plat/mediatek/mt6795/plat_sip_svc.c
danh-arm: OK. In general for NS callers, you need to check the args coming from the NS world before using them.

plat/mediatek/mt6795/bl31_plat_setup.c
leon-chen-mtk: Yes, we need to support both AArch32 and AArch64 kernel, so LK(AArch32) SMC back to ARM-TF no matter kernel is 32 bit or 64 bit for boot flow uniformly.

plat/mediatek/mt6795/plat_sip_svc.c
leon-chen-mtk: This EL1 NS dump callback function will be entered after watchdog timeout trap in EL3 and dumping limited EL3 context(EL3 dump code and fiq register code is not upstream), leaving EL3 back to EL1 for dumping kernel context.
Do you concern this callback is a security hole without validating memory address range, e.g, if a hacker can manipulate this pointer for hacking EL1 system, this validation does really matter? Or just for programming error?
If this is a security hole, could you please give us a scenario?

danh-arm: It's a bit hard to say whether this is a security hole without seeing the EL3 dump and fiq register code. In particular, whether the normal world can induce an EL3 dump via some means other than an EL3 watchdog timeout. At the least, it seems like a potential attack vector, which should be closed off in a production environment. Also, why does the normal world need to supply these arguments - surely the secure world should know these base addresses/sizes itself? In general the secure world is not responsible for protecting 1 normal world entiy from another; just for protecting itself from the whole of the normal world.

leon-chen-mtk: Thank you for the advice, but we have more confuse about this potential security hole, please let me clear some misunderstanding.
Below is watchdog timeout in our turn key scenario:
watchdog timeout occur->trap into EL3 fiq handler and determine it is wdt timeout->dump this cpu's EL1 context when wdt occur(general purpose registers, spsr, pc, etc, not EL3 context)->el3_exit->enter wdt_kernel_cb_addr in EL1(kernel)

wdt_kernel_cb_addr is registered while EL1 kernel boot up via SiP SMC call. wdt_kernel_cb_addr is responsible for generating kernel debug information(each kernel process state, interrupts statistics, etc) when wdt timeout occurs.

Q1: In particular, whether the normal world can induce an EL3 dump via some means other than an EL3 watchdog timeout
A1: Sorry that I make you confuse. We do not dump EL3 context when watchdog happen. EL3 dump this cpu's EL1 non-secure context snapshot(when watchdog timeout occurs).

Q2: why does the normal world need to supply these arguments - surely the secure world should know these base addresses/sizes itself?
A2: As mentioned above, if wdt_kernel_cb_addr is not registered from kernel space, EL3 do not know which kernel function should be back when watchdog timeout occur.

We appreciate that you review our code seriously, please clarify our uncertain points, thank you.

plat/mediatek/mt6795/aarch64/platform_common.c
athoelke: Is BL31 mapping all of the secure SRAM (`TZRAM` and `TZRAM2`) as read-write here?

The BL31 memory is mapped specifically in the MMU setup function, and this looks unnecessary. You do need to map memory that is shared with the preloader, boot ROM and/or Trusted OS, but that should be done with a minimal memory mapping.

plat/mediatek/mt6795/plat_sip_svc.c
athoelke: Without any visible code that uses `wdt_kernel_cb_addr` it is not possible to determine if there is a security issue here. The same is true for the `atf_ram_dump_*` variables. The specific concerns/questions here (for each call) are:

1. Is this intended to be called by the Trusted OS (ns=0) or LK/Kernel (ns=1)? - calls from an invalid security state must be rejected
2. Is there a point during boot after which this call is invalid and should be disabled, or are some of the variables "write-once" until next boot? - e.g. the `atf_ram_buf_*` and `wdt_kernel_cb_addr` could reject/ignore the SMC calls if they are non-zero to avoid leaving a potentially exploitable SMC function enabled
3. Are the parameters passed from normal world ever dereferenced in secure state, or just stored in EL3 and later used to initialise non-secure register state? The former require strong validation of the parameter, the latter may only need this sometimes
4. For SMC32/SMC64 APIs, are the top bits of the parameters being properly cleaned for a SMC32 call? This can only be skipped if the parameter is later used to initialise a 32-bit register/variable.

plat/mediatek/mt6795/plat_sip_svc.c
athoelke: Can you ensure that the kernel boot SMC calls are disabled after the kernel has been booted? Otherwise this call provides an interesting attack surface for a kernel exploit to be able to boot a completely new kernel by issuing this SMC...

chunfanmediatek: We will disable useless SMC calls at runtime. (Which is used in booting time only), Thanks

plat/mediatek/mt6795/plat_sip_svc.c
chunfanmediatek: Hi Andrew,
   Thanks for you suggestion. We will check each SMC call base on these 4 criteria.
We will disable useless SMC calls at runtime. (Which is used in booting time only)
About rule 4. Do you mean we should clean the top 32 bits for a smc32 call before we use it?

plat/mediatek/mt6795/aarch64/platform_common.c
chunfanmediatek: Here is some redundant mapping region. We will review the mapping table and remove the useless one. Thanks

plat/mediatek/mt6795/plat_sip_svc.c
athoelke: Yes.

If a SMC32 call comes from AArch32, then the parameter registers will have been populated using R0-R7. The ARMv8-A architecture does not specify exactly what the upper bits of X0-X7 will be if they were last modified in an AArch32 execution state. They might be zero or they might retain the value that was last written from an AArch64 state. It is best to resolve that uncertainty by explicitly truncating and extending the parameter in these cases, e.g.

    x1 = (uint64_t)((uint32_t) x1);   // zero extend a 32-bit unsigned value or address
    x2 = (int64_t)((int32_t) x2);     // sign-extend a 32-bit signed value

chunfanmediatek:
Hi ARM friends, is any update of #506 pull request?

danh-arm:
Hi. The reason this has not been reviewed again is because it wasn't clear that it was ready for review. You added another commit to this PR instead of force pushing a new clean commit, or (even better) raising a new PR with the clean commit. See the FAQ on the wiki for more information:
https://github.com/ARM-software/arm-trusted-firmware/wiki/Trusted-Firmware-Frequently-Asked-Questions

Regarding the changes in 41c93ba, these are a bit hard to review. Some suggestions:

*   Don't make changes to generic code, or if you need to, put them in a separate patch. 
*   Don't merge in upstream changes as part of your patch(es).
*   Rebase to the latest master.
*   You've made major changes to the existing upstream mt8173 port.
     * Can you put all the changes that modify the existing port (including common MediaTek code) into a separate earlier patch, before adding the mt8173 port?
     * Have you got approval from the MediaTek sub-maintainer, Yidi Lin (mtk09422) for these changes? I'd like to see approval from them before merging.

danh-arm:
@ freedomtan, @chunfanmediatek, @coboy-chen 

I'm going to close this PR as it seems to have stalled. Please raise a new one with a clean set of commits if this is still needed.

PR #513, opened 2016-02-03
build system: allow overriding the build's timestamp

pgeorgi:
This allows reproducible builds (same source and same compiler produce
bit-identical results) and also allows coordinating the timestamp across
multiple projects, eg. with another firmware.

[tested both with passing an override and without. both produced an image with the expected string]


danh-arm:

danh-arm:

danh-arm:
test this please

PR #512, opened 2016-02-03
Pl061 gpio v4

hzhuang1:
Implement gpio framework and PL061 gpio driver. ARM-software/tf-issues#351

Changelog:
v4:
1. Clean code with more assert.
2. Remove return flag in set_direction() and set_value().
3. Add PLAT_PL061_MAX_GPIOS define.

v3:
1. Add more comments in gpio driver.
2. Add more assert in both gpio & pl061 driver.
3. Add get_direction() in gpio ops.

v2:
1. Update with the standard copyright text.
2. Replace "uint32_t" by "unsigned int" according to the rules in wiki.

PR #518, opened 2016-02-12
Pl061 gpio v5

hzhuang1:
Implement gpio framework and PL061 gpio driver. ARM-software/tf-issues#351

Changelog:
v5:
1. Move comments after license text.
2. Use CASSERT on PLAT_PL061_MAX_GPIOS.
3. Add assert(ops) in gpio driver.
4. Adjust the comments of PLAT_PL061_MAX_GPIOS in document.

 v4:
 1. Clean code with more assert.
 2. Remove return flag in set_direction() and set_value().
 3. Add PLAT_PL061_MAX_GPIOS define.

v3:
 1. Add more comments in gpio driver.
 2. Add more assert in both gpio & pl061 driver.
 3. Add get_direction() in gpio ops.

v2:
 1. Update with the standard copyright text.
 2. Replace "uint32_t" by "unsigned int" according to the rules in wiki.

### DIFF
--- a/docs/porting-guide.md
+++ b/docs/porting-guide.md
@@ -466,6 +466,18 @@ memory layout implies some image overlaying like in ARM standard platforms.
 
     Defines the maximum address that the TSP's progbits sections can occupy.
 
+If the platform port uses the PL061 GPIO driver, the following constant may
+optionally be defined:
+
+*   **PLAT_PL061_MAX_GPIOS**
+    Maximum number of GPIOs required by the platform. This allows control how
+    much memory is allocated for PL061 GPIO controllers. The default value is
+    32.
+    [For example, define the build flag in platform.mk]:
+    PLAT_PL061_MAX_GPIOS    :=      160
+    $(eval $(call add_define,PLAT_PL061_MAX_GPIOS))
+
+
 ### File : plat_macros.S [mandatory]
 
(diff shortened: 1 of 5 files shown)


danh-arm:

danh-arm:

danh-arm:
You made a good point in the last review here:
https://github.com/ARM-software/arm-trusted-firmware/pull/512#commitcomment-16060984

I'm not sure use of `unsigned int` for array indices is "forbidden" in C but I agree it's a bad idea due to the implicit sign extension rules. There are other places in the codebase we should probably correct. I think we're missing something here in our coding guidelines:
https://github.com/ARM-software/arm-trusted-firmware/wiki/Trusted-Firmware-Coding-Guidelines

I'll try to get something added here. If you have any references to good practice, that would be useful.

danh-arm:
test this please

danh-arm:
I'll merge this after the CI test since my comments are so minor.

danh-arm:
Unfortunately, this patch (through no fault of yours) breaks the mt8173 platform. That platform contains it's only gpio code, which conflict with this. I believe this is due to that platform include `gpio.h` as a system include instead of a user include, which incorrectly picks up your header instead of the local one.

I'll get someone our side to fix this and then we can merge yours on top.

danh-arm:
retest this please

PR #520, opened 2016-02-16
Vk/scp flexibility

vikramkanigiri:
This pull requests adds more configurability for ARM platforms with respect SCP i.e whether to load SCP on some platforms and different base addresses for BOM/SCPI protocol


danh-arm:
retest this please

PR #527, opened 2016-02-18
Remove non-ASCII character from comment

antonio-nino-diaz-arm:
Replaced a long dash in a comment by the ASCII character '-'. Support
for multibyte character in the source character set is not enforced by
the C99 standard. To maximize compatibility with C processing tools
(e.g. compilers or static code analysis tools), they should be removed.

Change-Id: Ie318e380d3b44755109f042a76ebfd2229f42ae3


danh-arm:

danh-arm:

PR #530, opened 2016-02-18
Add support for %p in tf_printf()

antonio-nino-diaz-arm:
This patch adds support for the `%p` format specifier in tf_printf()
following the example of the printf implementation of the stdlib used
in the trusted firmware.

Fixes ARM-software/tf-issues#292

Change-Id: I0b3230c783f735d3e039be25a9405f00023420da


danh-arm:

PR #533, opened 2016-02-22
Fix the inconsistencies in bl1_tbbr_image_descs[]

yatharth-arm:
This patch fixes inconsistencies in bl1_tbbr_image_descs[]
and miscellaneous fixes in Firmware Update code.

Following are the changes:
* As part of the original FWU changes, a `copied_size`
  field was added to `image_info_t`. This was a subtle binary
  compatibility break because it changed the size of the
  `bl31_params_t` struct, which could cause problems if
  somebody used different versions of BL2 or BL31, one with
  the old `image_info_t` and one with the new version.
  This patch put the `copied_size` within the `image_desc_t`.
* EXECUTABLE flag is now stored in `ep_info.h.attr` in place
  of `image_info.h.attr`, associating it to an entrypoint.
* The `image_info.image_base` is only relevant for secure
  images that are copied from non-secure memory into secure
  memory. This patch removes initializing `image_base` for
  non secure images in the bl1_tbbr_image_descs[].
* A new macro `SET_STATIC_PARAM_HEAD` is added for populating
  bl1_tbbr_image_descs[].ep_info/image_info.h members statically.
  The version, image_type and image attributes are now
  populated using this new macro.
* Added PLAT_ARM_NVM_BASE and PLAT_ARM_NVM_SIZE to avoid direct
  usage of V2M_FLASH0_XXX in plat/arm/common/arm_bl1_fwu.c.
* Refactoring of code/macros related to SECURE and EXECUTABLE flags.

NOTE: PLATFORM PORTS THAT RELY ON THE SIZE OF `image_info_t`
      OR USE the "EXECUTABLE" BIT WITHIN `image_info_t.h.attr`
      OR USE THEIR OWN `image_desc_t` ARRAY IN BL1, MAY BE
      BROKEN BY THIS CHANGE. THIS IS CONSIDERED UNLIKELY.

Change-Id: Id4e5989af7bf0ed263d19d3751939da1169b561d

### DIFF
--- a/bl1/bl1_context_mgmt.c
+++ b/bl1/bl1_context_mgmt.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015, ARM Limited and Contributors. All rights reserved.
+ * Copyright (c) 2015-2016, ARM Limited and Contributors. All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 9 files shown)


danh-arm:

PR #539, opened 2016-03-07
Compile stdlib C files individually

antonio-nino-diaz-arm:
All C files of stdlib were included into std.c, which was the file
that the Makefile actually compiled. This is a poor way of compiling
all the files and, while it may work fine most times, it's
discouraged.

In this particular case, each C file included its own headers, which
were later included into std.c. For example, this caused problems
because a duplicated typedef of u_short in both subr_prf.c and
types.h. While that may require an issue on its own, this kind of
problems are avoided if all C files are as independent as possible.

Change-Id: I9a7833fd2933003f19a5d7db921ed8542ea2d04a


danh-arm:

PR #541, opened 2016-03-07
Initialize secondary CPUs during cold boot

antonio-nino-diaz-arm:
The previous reset code in BL1 performed the following steps in
order:

1. Warm/Cold boot detection.
   If it's a warm boot, jump to warm boot entrypoint.

2. Primary/Secondary CPU detection.
   If it's a secondary CPU, jump to plat_secondary_cold_boot_setup(),
   which doesn't return.

3. CPU initialisations (cache, TLB...).

4. Memory and C runtime initialization.

For a secondary CPU, steps 3 and 4 are never reached. This shouldn't
be a problem in most cases, since current implementations of
plat_secondary_cold_boot_setup() either panic or power down the
secondary CPUs.

The main concern is the lack of secondary CPU initialization when
bare metal EL3 payloads are used in case they don't take care of this
initialisation themselves.

This patch moves the detection of primary/secondary CPU after step 3
so that the CPU initialisations are performed per-CPU, while the
memory and the C runtime initialisation are only performed on the
primary CPU. The diagrams used in the ARM Trusted Firmware Reset
Design documentation file have been updated to reflect the new boot
flow.

Platforms ports might be affected by this patch depending on the
behaviour of plat_secondary_cold_boot_setup(), as the state of the
platform when entering this function will be different.

Fixes ARM-software/tf-issues#342

Change-Id: Icbf4a0ee2a3e5b856030064472f9fa6696f2eb9e


danh-arm:

PR #542, opened 2016-03-07
Initialize all translation table entries

sandrine-bailleux-arm:
The current translation table code maps in a series of regions, zeroing
the unmapped table entries before and in between the mapped regions. It
doesn't, however, zero the unmapped entries after the last mapped
region, leaving those entries at whatever value that memory has
initially.

This is bad because those values can look like valid translation table
entries, pointing to valid physical addresses. The CPU is allowed to do
speculative reads from any such addresses. If the addresses point to
device memory, the results can be unpredictable.

This patch zeroes the translation table entries following the last
mapped region, ensuring all table entries are either valid or zero
(invalid).

In addition, it limits the value of ADDR_SPACE_SIZE to those allowed by
the architecture and supported by the current code (see D4.2.5 in the
Architecture Reference Manual). This simplifies this patch a lot and
ensures existing code doesn't do unexpected things.

### DIFF
--- a/lib/aarch64/xlat_tables.c
+++ b/lib/aarch64/xlat_tables.c
@@ -52,7 +52,17 @@
 #define debug_print(...) ((void)0)
 #endif
 
-CASSERT(ADDR_SPACE_SIZE > 0, assert_valid_addr_space_size);
+#define IS_POWER_OF_TWO(x)	(((x) & ((x) - 1)) == 0)
+
+/*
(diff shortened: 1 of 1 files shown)


danh-arm:

danh-arm:

danh-arm:
test this please

danh-arm:
jenkins: retest this please

PR #550, opened 2016-03-15
Remove all non-configurable dead loops

antonio-nino-diaz-arm:
Added a new platform porting function plat_panic_handler, to allow
platforms to handle unexpected error situations. It must be
implemented in assembly as it may be called before the C environment
is initialized. A default implementation is provided, which simply
spins.

Corrected all dead loops in generic code to call this function
instead. This includes the dead loop that occurs at the end of the
call to panic().

All unnecesary wfis from bl32/tsp/aarch64/tsp_exceptions.S have
been removed.

Change-Id: I67cb85f6112fa8e77bd62f5718efcef4173d8134


danh-arm:

PR #552, opened 2016-03-15
Add cache topology info to FVP DTBs

antonio-nino-diaz-arm:
From version 4.0 onwards, the ARM64 Linux kernel expects the device
tree to indicate the cache hierarchy. Failing to provide this
information results in the following warning message to be printed by
the kernel:

    `Unable to detect cache hierarchy from DT for CPU x`

All the FVP device trees provided in the TF source tree have been
modified to add this information.

Fixes ARM-software/tf-issues#325

Change-Id: I0ff888992e602b81a0fe1744a86151d625727511


danh-arm:

PR #553, opened 2016-03-15
std_svc: psci: introduce handler for system suspend

vwadekar:
System Suspend needs to be executed on the last standing CPU and
psci_system_suspend() makes sure of that. The real problem with
the current code is that it acquires locks further into the CPU
suspend sequence even for the system suspend case. This seems
unnecessary and adds to the entry latency times.

This patch introduces a separate handler for the system suspend
case which does not acquire any locks and hence reduces the
entry time into the state.

### DIFF
--- a/services/std_svc/psci/psci_main.c
+++ b/services/std_svc/psci/psci_main.c
@@ -147,7 +147,6 @@ int psci_cpu_suspend(unsigned int power_state,
 	return PSCI_E_SUCCESS;
 }
 
-
 int psci_system_suspend(uintptr_t entrypoint, u_register_t context_id)
 {
 	int rc;
@@ -173,16 +172,14 @@ int psci_system_suspend(uintptr_t entrypoint, u_register_t context_id)
 	assert(is_local_state_off(state_info.pwr_domain_state[PLAT_MAX_PWR_LVL]));
 
 	/*
-	 * Do what is needed to enter the system suspend state. This function
-	 * might return if the power down was abandoned for any reason, e.g.
-	 * arrival of an interrupt
+	 * Do what is needed to enter the system suspend state.
 	 */
-	psci_cpu_suspend_start(&ep,
-			    PLAT_MAX_PWR_LVL,
-			    &state_info,
-			    PSTATE_TYPE_POWERDOWN);
+	psci_system_suspend_start(&ep, &state_info);
 
-	return PSCI_E_SUCCESS;
+	/*
+	 * We should not return here upon exiting system suspend state.
+	 */
+	panic();
 }
 
 int psci_cpu_off(void)
--- a/services/std_svc/psci/psci_private.h
+++ b/services/std_svc/psci/psci_private.h
@@ -213,6 +213,8 @@ void psci_cpu_on_finish(unsigned int cpu_idx,
 int psci_do_cpu_off(unsigned int end_pwrlvl);
 
 /* Private exported functions from psci_suspend.c */
+void psci_system_suspend_start(entry_point_info_t *ep,
+			       psci_power_state_t *state_info);
 void psci_cpu_suspend_start(entry_point_info_t *ep,
 			unsigned int end_pwrlvl,
 			psci_power_state_t *state_info,
--- a/services/std_svc/psci/psci_suspend.c
+++ b/services/std_svc/psci/psci_suspend.c
@@ -111,6 +111,43 @@ static void psci_suspend_to_pwrdown_start(unsigned int end_pwrlvl,
 }
 
 /*******************************************************************************
+ * Top level handler which is called from the last standing CPU when the entire
+ * system wants to suspend its execution. It is assumed that along with
+ * suspending the current cpu power domain, power domains at higher levels until
+ * the target power level will be suspended as well. It coordinates with the
+ * platform to negotiate the target state for each of the power domain level
+ * till the target power domain level. It then performs generic, architectural,
+ * platform setup and state management required to suspend that power domain
+ * level and power domain levels below it.
+ *
+ * All the required parameter checks are performed at the beginning and after
+ * the state transition has been done, no further error is expected and it is
+ * not possible to undo any of the actions taken beyond that point.
+ ******************************************************************************/
+void psci_system_suspend_start(entry_point_info_t *ep,
+			       psci_power_state_t *state_info)
+{
+	/*
+	 * This function is passed the requested state info and
+	 * it returns the negotiated state info for each power level upto
+	 * the end level specified.
+	 */
+	psci_do_state_coordination(PLAT_MAX_PWR_LVL, state_info);
+
+	psci_suspend_to_pwrdown_start(PLAT_MAX_PWR_LVL, ep, state_info);
+
+	/*
+	 * Plat. management: Allow the platform to perform the
+	 * necessary actions to turn off this cpu e.g. set the
+	 * platform defined mailbox with the psci entrypoint,
+	 * program the power controller etc.
+	 */
+	psci_plat_pm_ops->pwr_domain_suspend(state_info);
+
+	psci_power_down_wfi();
+}
+
+/*******************************************************************************
  * Top level handler which is called when a cpu wants to suspend its execution.
  * It is assumed that along with suspending the cpu power domain, power domains
  * at higher levels until the target power level will be suspended as well. It


danh-arm:

vwadekar:
Dan et al, please let me know your comments.

The real problem with Tegra is that we want to run the Trusted Firmware from our on-chip TZRAM. TZRAM on Tegra loses power when we enter System Suspend, which means we have to save the entire BL3-1 code back to TZDRAM before System Suspend. The only place where this can be done is the platform's suspend entry handler. But with the locks being acquired around this function, the lock value is not stored to the TZDRAM.

But while looking at the code it seemed logical to remove the locking for other reasons - entry latency, as well. Let me know if this approach looks reasonable

danh-arm:

achingupta:
This patch does not fly for me from what I have been able to understand. Glad to be corrected!

* Locks are taken even while suspending the system so that we could re-use code. Moreover, since all other CPUs have been turned off there should be no contention. So, I am not quite sure what you mean by the "real problem" in the commit message. The use of bakery locks is slower than using spin locks. ARM TF does not provide a framework to allow the platform provide a more suitable lock implementation e.g. spinlocks if there is a global exclusive monitor. But this is a separate issue.

* Are there any numbers that illustrate the improvement in performance? It looks like that you will be spending far more time transferring BL3-1 context than acquiring locks.

* The major reason for re-using `psci_cpu_suspend_start()` was to conform to the PSCI specification that says that `SYSTEM_SUSPEND` semantics are are equivalent to a `CPU_SUSPEND` to the deepest low-power state. Hence the need for the same steps. This patch is a bit cheeky in that it assumes entry in `psci_system_suspend_start()` is a point of no return. This might be true for Tegra but I cannot see how the same assumption can be made for all other platforms. Hence, the `panic()` at the end of `psci_system_suspend()` is incorrect. 

As you say, the real problem is that the TZRAM powers off during system suspend. AFAIU, what you really want is a guarantee that no global variable is written between `psci_plat_pm_ops->pwr_domain_suspend(state_info)` and `psci_power_down_wfi()`. I don't think expecting this or using `psci_system_suspend_start()` are scalable approaches.

Why not run BL3-1 out of the TZDRAM? There must be enough space since you are saving its live context in any case?

The other option could be to add a platform hook just before `psci_power_down_wfi()` to perform any late actions e.g. like your use case. This has been mooted in the past. The GIC CPU interface could be disabled here instead of in the platform handler to increase the possibility of terminating the suspend request in case of an interrupt. This is just a small example.

danh-arm:
I agree with Achin as to why the current patch doesn't fly.

In general, using on-chip RAM for BL31 memory provides better physical security than DDR, but that benefit is lost if the memory is copied unencrypted to DDR during system suspend. You might want to consider encrypting the memory before copying. Otherwise, as Achin, says, you might as well use DDR for BL31 memory all the time.

I'm open to the idea of a new late platform hook for system suspend if needed. The common code could guarantee no more writes after this is called.

vwadekar:
@danh-arm

The dram memory aperture in question is already protected against non-TZ
accesses by the memory controller on Tegra. So we don't need encryption
before copying to it. And you are right, the onchip memory is indeed way
faster than dram. For some markets where Tegra would be used system suspend
is not a frequent use case so we don't lose the advantage of tzram by
copying to dram.

For the locks, it seems weird to acquire them even though system suspend
always runs on the last standing CPU.

danh-arm:
OK, so you're running BL31 in on-chip RAM instead of DDR for performance reasons, not physical security reasons. Even though both are only accessible to the secure world, DDR is easier to physically probe than on-chip RAM. If you don't want to mitigate that security risk, there's no need to encrypt.

You're right that the implementation doesn't strictly need to acquire the locks in this case. There's a potential optimisation there but:

1.  What functional problem is there by copying to DDR before the locks are acquired (and released)? It seems like it will still resume OK without issue.
2.  What actual performance saving is there in this optimisation? As Achin says, any saving is likely to be dwarfed by the copy operation. Increasing code complexity for speculative performance savings is not usually a good trade-off.

vwadekar:
@danh-arm 

Let's leave TZRAM out of the equation for the moment.

In order to get a better picture I profiled the locked v non-locked version of the code and see a performance gain of 10us during System Suspend entry with the non-locked version. Tried this across multiple System Suspend entry sequences and I usually see a gain of 10-11us.

So, there is a good performance gain here with the non-locked version.

danh-arm:
Hi Varun. Have you measured the time taken for the BL31 copy to DRAM or the overall System Suspend operation? I guess the latter is of the order of 100s of ms. In that context, 10us is insignifcant.

vwadekar:
@danh-arm 

I am leaving TZRAM out of the equation for the moment. This means running BL31 from TZRAM and copy from TZRAM to DRAM is out of scope at the moment.

Having said that, when running BL31 from DRAM I get a performance gain of 10us with the non-locked version.

danh-arm:
Even ignoring the DDR copy, on any platform, not just your platform, a system suspend operation requires the OS to take many actions (e.g. hotplugging all other CPUs and suspending all devices) before the call to SYSTEM_SUSPEND. Saving 10us at the end of this operation is insignificant. This might be more significant during ordinary CPU_SUSPEND operations, which can occur very frequently, but this patch doesn't improve that case.

As Achin mentioned above, a more promising approach might be to provide a framework for platforms to provide their own, more efficent lock implementations, if there are platforms out there that can make use of this (e.g. have a global exclusive monitor).

vwadekar:
@danh-arm 

The 10us might seem insignificant to you, but for our target markets we are looking to save even 1us as there are hard timelines that we have to meet for SC7 entry and exit scenarios. If you dont want to change the behavior for all platforms, then we should at least make this part configurable so that platforms like ours can override the default behavior.

vwadekar:
@ danh-arm

Coming back to the TZRAM issue, what would be a preferable solution? From the comments so far, it seems you guys are open to having a platform handler which gets called after relinquishing all locks (?)

So I am guessing the following code would do -

`if (end_pwrlvl == MPIDR_AFFLVL2)
        psci_plat_pm_ops->pwr_domain_suspend**_late**(state_info);`

achingupta:
On Wed, Mar 16, 2016 at 02:13:23PM -0700, varunw-nvidia wrote:
>    @danh-arm
>
>    The 10us might seem insignificant to you, but for our target markets we
>    are looking to save even 1us as there are hard timelines that we have to
>    meet for SC7 entry and exit scenarios. If you dont want to change the
>    behavior for all platforms, then we should at least make this part
>    configurable so that platforms like ours can override the default
>    behavior.

I still have my doubts about the importance of 10us while performing a suspend
to RAM.  I will check with the kernel folk to be absolutely sure.  Please note
that it is possible to enter system level power states by calling the
CPU_SUSPEND API on all cores.  In this case `psci_cpu_suspend_start()` will need
to acquire locks as usual. It does not seem ideal to have almost identical
implementations of the same function to enable differentiation between Tegra and
all other platforms in the wild.

This is just my opinion!

>
>    -
>    You are receiving this because you commented.
>    Reply to this email directly or view it on GitHub
>
>    Link:
>    https://github.com/ARM-software/arm-trusted-firmware/pull/553#issuecomment-197551592
IMPORTANT NOTICE: The contents of this email and any attachments are confidential and may also be privileged. If you are not the intended recipient, please notify the sender immediately and do not disclose the contents to any other person, use it for any purpose, or store or copy the information in any medium. Thank you.

achingupta:
>    @ danh-arm
>
>    Coming back to the TZRAM issue, what would be a preferable solution? From
>    the comments so far, it seems you guys are open to having a platform
>    handler which gets called after relinquishing all locks (?)
>
>    So I am guessing the following code would do -
>
>    if (end_pwrlvl == MPIDR_AFFLVL2)
>    psci_plat_pm_ops->pwr_domain_suspend**_late**(state_info);

I am thinking on the lines of (does not have to be called this):

 psci_plat_pm_ops->prepare_for_pwrdwn_wfi(state_info)

It will be called unconditionally before the final wfi. This would allow any
late actions to be performed in other APIs as well e.g. CPU_OFF. More thought is
needed.

vwadekar:
@danh-arm 

Dan, your thoughts on Achin's proposal? Once we reach a consensus I will push a new PR for the late suspend handler.

danh-arm:
Hi Varun

Yes, I see the need for a new hook, although I'm not sure about re-using this in the CPU_OFF path; it needs to be crystal clear to the platform under which circumstances this hook is called. Since there are a few details to work through and this is very sensitive code, I'd prefer us to implement this our side. I've created a new tf-issue for it (https://github.com/ARM-software/tf-issues/issues/375). We'll schedule this ASAP.

Regarding the system suspend optimization to not acquire locks, I'm not ruling this out but we're still skeptical our side of its value. We would need some numbers showing that this is a significant proportion of the overall system suspend time. Perhaps you're using system suspend in ways we haven't envisaged? Also, this would need additional work to ensure good code sharing and to avoid impacting the (more crucial) CPU suspend path. If it turns out this is still needed, then create a separate tf-issue and we'll look into this as a lower priority.

Regards

Dan.

vwadekar:
Closing this request as we can live with the solution for ARM-software/tf-issues#375 for now

PR #556, opened 2016-03-17
Support for Rockchip's family SoCs

rkchrome:
This patch adds to support the RK3368 and RK3399 SoCs.

RK3368/RK3399 is one of the Rockchip family SoCs, which is an
multi-cores ARM SoCs.

This patch adds support to boot the Trusted Firmware on RK3368/RK3399

This is the initial version for rockchip SoCs.(RK3368/RK3399 and next SoCs)
* Support arm gicv2 & gicv3.
* Boot up multi-cores CPU.
* Support suspend/resume.

Signed-off-by: Shengfei xu <xsf@rock-chips.com>


danh-arm:

danh-arm:

plat/rockchip/rk3368/aarch64/platform_common.c
```diff
@@ -0,0 +1,127 @@
+/*
```
soby-mathew: This file and the corresponding file in rk3399 is same except for plat_mmap[] array. You could consider sharing this file and just separating the array between the 2 variants.

plat/rockchip/common/plat_pm.c
```diff
@@ -0,0 +1,294 @@
...
+}
+
+/*******************************************************************************
+ * RockChip handlers to shutdown/reboot the system
+ ******************************************************************************/
+static void __dead2 rockchip_system_off(void)
+{
+	assert(rockchip_ops && rockchip_ops->system_off);
+
+	rockchip_ops->system_off();
```
danh-arm: This function in rockchip_ops is not defined for your SoCs. Therefore the assert will always fail. This function is a mandatory PSCI feature so really should be implemented before upstreaming.

plat/rockchip/common/plat_pm.c
```diff
@@ -0,0 +1,294 @@
...
+}
+
+static void __dead2 rockchip_system_reset(void)
+{
+	assert(rockchip_ops && rockchip_ops->sys_gbl_soft_reset);
+
+	rockchip_ops->sys_gbl_soft_reset();
```
danh-arm: Does this cause a platform cold reset (i.e. a hardware power-cycle sequence) as specified in the PSCI spec? If this only does a warm reset of the SoC then it's not enough.

plat/rockchip/rk3399/drivers/pmu/pmu.c
```diff
@@ -0,0 +1,394 @@
...
+	sram_src = (uint32_t *)&pmu_cpuson_entrypoint_start;
+	sram_size = (uint32_t *)&pmu_cpuson_entrypoint_end -
+		    (uint32_t *)sram_src;
+
+	u32_align_cpy(sram_dst, sram_src, sram_size);
+
+	psram_sleep_cfg->sp = PSRAM_DT_BASE;
+}
+
+static inline uint32_t pmu_power_domain_st(uint32_t pd)
```
soby-mathew: This function, pmu_power_domain_ctr and check_cpu_wfie can be made common across the 2 variants the SoC

plat/rockchip/rk3399/drivers/pmu/pmu.c
```diff
@@ -0,0 +1,394 @@
...
+	dsb();
+
+	cpus_power_domain_off(cpu_id, core_pwr_wfi_int);
+
+	return 0;
+}
+
+static int cores_pwr_domain_on_finish(void)
+{
+	uint32_t cpu_id = plat_core_pos_by_mpidr(read_mpidr_el1());
```
soby-mathew: nit : Use plat_my_core_pos() here. Similarly in other places where core position of current CPU is queried.

plat/rockchip/rk3399/drivers/pmu/pmu.c
```diff
@@ -0,0 +1,394 @@
...
+	dsb();
+
+	cpus_power_domain_on(cpu_id);
+
+	return 0;
+}
+
+static int cores_pwr_domain_off(void)
+{
+	uint32_t cpu_id = plat_core_pos_by_mpidr(read_mpidr_el1());
```
soby-mathew: Use plat_my_core_pos() here.

plat/rockchip/rk3399/drivers/pmu/pmu.c
```diff
@@ -0,0 +1,394 @@
...
+{
+	uint32_t cfg_info;
+	uint32_t cpu_pd = PD_CPUL0 + cpu_id;
+	/*
+	  * There are two ways to powering on or off on core.
+	  * 1) Control it power domain into on or off in PMU_PWRDN_CON reg
+	  * 2) Enable the core power manage in PMU_CORE_PM_CON reg,
+	  *     then, if the core enter into wfi, it power domain will be
+	  *     powered off automatically.
+	  */
```
soby-mathew: Please feel free to correct me on my analysis, I am trying to figure out the need for bakery locks here. 

The PSCI state coordination ensures that CPU ON cannot be invoked on a **suspended** core and can only be invoked on a target core which has completed the CPU OFF platform hook execution. The exception in the Rockchip platform is "nonboot_cpus_off()" which turns OFF other cores  from the primary. This is invoked from 2 places:

1.  bl31_platform_setup(). 
When "nonboot_cpus_off()" is invoked from here, there is no race condition because it is cold boot and only the primary is running. Hence there is no need for locks in this case.

2. sys_pwr_domain_suspend() (only for rk3368 platform)
This can be called due to PSCI SYSTEM SUSPEND API or PSCI CPU SUSPEND to system power level. In the former case, the PSCI framework ensures that all secondaries are OFF and in the latter, the PSCI framework executes the platform hooks with "system" level locks held. Hence I don't see a race here as well.   
BTW, the rk3399 platform does not invoke "nonboot_cpus_off()" in this function.

I could be missing something but could you please explain me the scenario where you expect the race to happen to justify the use of locks here?

plat/rockchip/rk3368/aarch64/platform_common.c
rkchrome: Okay ,done.

plat/rockchip/common/plat_pm.c
rkchrome: Thanks for pointing out.

plat/rockchip/common/plat_pm.c
rkchrome: That will reset all of the soc logics，then the system will reboot from maskrom.

plat/rockchip/rk3399/drivers/pmu/pmu.c
rkchrome: Okay, done

plat/rockchip/rk3399/drivers/pmu/pmu.c
rkchrome: Done

plat/rockchip/rk3399/drivers/pmu/pmu.c
rkchrome: Done

plat/rockchip/rk3399/drivers/pmu/pmu.c
rkchrome: As explained above （line 33）：
/*
 * These are wrapper macros to the Coherent Memory Bakery Lock API.
 * The PMU_CORE_PM_CON(cpu_id) reg is used to control cpu to
 * enter into suspend/reusme or off/on mode.
 * Consider the situation that cpu0 is in suspend mode,
 * but cpu1 want to power down cpu0 at the same times.
 * Although the situation do not happen on linux os.
 */
The situation do not happen on linux os. We just want to make sure that if there is a case that operating the reg multiple at same time is safe.
Anyway, we will detele it.

rkchrome:
Thanks Dan anf Soby for reviewing, the new pull request on https://github.com/ARM-software/arm-trusted-firmware/pull/557

plat/rockchip/rk3399/drivers/pmu/pmu.c
```diff
@@ -0,0 +1,394 @@
...
+#include <soc.h>
+#include <pmu.h>
+
+static struct psram_data_t *psram_sleep_cfg =
+	(struct psram_data_t *)PSRAM_DT_BASE;
+
+/*
+ * These are wrapper macros to the Coherent Memory Bakery Lock API.
+ * The PMU_CORE_PM_CON(cpu_id) reg is used to control cpu to
+ * enter into suspend/reusme or off/on mode.
```
afaerber: Typo: s/reusme/resume/

PR #557, opened 2016-03-18
Support for Rockchip's family SoCs

rkchrome:
This patch adds to support the RK3368 and RK3399 SoCs.

RK3368/RK3399 is one of the Rockchip family SoCs, which is an
multi-cores ARM SoCs.

This patch adds support to boot the Trusted Firmware on RK3368/RK3399

This is the initial version for rockchip SoCs.(RK3368/RK3399 and next SoCs)
* Support arm gicv2 & gicv3.
* Boot up multi-cores CPU.
* Support suspend/resume.

Signed-off-by: Shengfei xu <xsf@rock-chips.com>


rkchrome:
The pervious version on https://github.com/ARM-software/arm-trusted-firmware/pull/556

danh-arm:

danh-arm:

plat/rockchip/common/drivers/pmu/pmu_com.h
```diff
@@ -0,0 +1,101 @@
...
+	uint32_t pwrdn_st = mmio_read_32(PMU_BASE + PMU_PWRDN_ST) &  BIT(pd);
+
+	if (pwrdn_st)
+		return pmu_pd_off;
+	else
+		return pmu_pd_on;
+}
+
+static int pmu_power_domain_ctr(uint32_t pd, uint32_t pd_state)
+{
```
soby-mathew: "PMU_BASE + PMU_PWRDN_CON" register seems to be shared for all cores. And you are using a read - modify - write sequence to modify the bit pattern corresponding to each CPU. I believe this function would internally need to use a **single global lock** to protect the consistency of the register when accessed by multiple CPUs.

I could be wrong here but it would be good to discuss before you modify the code.

rkchrome: Sorry for delay reply.

For rk3368, maybe that's a issue. 
In some cases, the linux and atf will likely be shared together in suspend and resume.
For example, the linux on/off the GPU and atf on/off the CPU.

Then...
For rk3399, that is *not* a issue since that's impossible to happen on rk3399.

rkchrome:
See https://github.com/ARM-software/arm-trusted-firmware/pull/568

PR #558, opened 2016-03-18
add MMC driver v1

hzhuang1:
Add MMC driver to fix the issue in below.

https://github.com/ARM-software/tf-issues/issues/376


danh-arm:

danh-arm:

danh-arm:
One high level comment on this: You seem to use the terms MMC and eMMC interchangeably, when one is a subset of the other. Can you use eMMC throughout (including the code) or explicitly have separate MMC and eMMC layers to make it clear which features belong to which.

hzhuang1:
OK. I'll use emmc instead.

hzhuang1:
Split this pull request with two. One is for emmc & io_block driver. And the other one is for dwmmc driver.
https://github.com/ARM-software/arm-trusted-firmware/pull/573
https://github.com/ARM-software/arm-trusted-firmware/pull/574

PR #561, opened 2016-03-22
Enable preloaded BL33 alternative boot flow

antonio-nino-diaz-arm:
Enable alternative boot flow where BL2 does not load BL33 from
non-volatile storage, and BL31 hands execution over to a preloaded
BL33.

The flag used to enable this bootflow is BL33_BASE, which must hold
the entrypoint address of the BL33 image. The User Guide has been
updated with an example of how to use this option with a bootwrapped
kernel.

Change-Id: I48087421a7b0636ac40dca7d457d745129da474f

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -301,6 +301,29 @@ include lib/cpus/cpu-ops.mk
 
 
 ################################################################################
+# Check incompatible options
+################################################################################
+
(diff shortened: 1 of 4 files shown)


danh-arm:

PR #562, opened 2016-03-23
fix compiler warning in psci_set_req_local_pwr_state

sbranden:
add missing assert to range check input parameter.
This fixes a warning emitted by gcc5.1 compiler.

Fixes arm-software/tf-issues#379

Signed-off-by: Mathieu Rondonneau <mathieu.rondonneau@broadcom.com>


danh-arm:

danh-arm:

danh-arm:
Can you confirm if this issue applies to the latest upstream master? I believe this was supposed to be fixed by #515.

danh-arm:
I'm assuming this was already addressed by #515 so this PR is not needed. Please re-open if that is not the case.

PR #563, opened 2016-03-23
Add support for %z in tf_print()

sbranden:
Add support for %z format specific in tf_printf() to support
printing data of type size_t

Fixes ARM-software/tf-issues#380

Signed-off-by Scott Branden <scott.branden@broadcom.com>

### DIFF
--- a/common/tf_printf.c
+++ b/common/tf_printf.c
@@ -69,6 +69,7 @@ static void string_print(const char *str)
  * %ld and %lld - signed 64 bit decimal format
  * %lu and %llu - unsigned 64 bit decimal format
  * %p - pointer format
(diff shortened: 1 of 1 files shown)


danh-arm:

danh-arm:

sbranden:
why do I need to assert the value is 4 or 8?  If sizeof(size_t) == 8 then 64bit is used, otherwise the size will be for 32 bit.

danh-arm:
> why do I need to assert the value is 4 or 8?

You don't - that's why I said ignore the comment. I just mentioned the compile time assert as a stepping stone to future 128-bit compatibility! But that's not important so ignore.

PR #565, opened 2016-03-23
Add BL31 relocation support

sbranden:
Add support to make BL31 relocatable.
This can be enabled by defining BL31_ENABLE_RELOCATION.

Fixes ARM-software/tf-issues#382

Signed-off-by Pramod Kumar pramod.kumar@broadcom.com
Signed-off-by: Mathieu Rondonneau mathieu.rondonneau@broadcom.com
Signed-off-by Scott Branden scott.branden@broadcom.com

### DIFF
--- a/bl31/aarch64/bl31_entrypoint.S
+++ b/bl31/aarch64/bl31_entrypoint.S
@@ -31,6 +31,7 @@
 #include <arch.h>
 #include <bl_common.h>
 #include <el3_common_macros.S>
+#include <platform_def.h>
 
 	.globl	bl31_entrypoint
 
@@ -42,6 +43,39 @@
 	 */
 
 func bl31_entrypoint
+#ifdef BL31_ENABLE_RELOCATION
+	/*
+	 * Below code snippet is used for relocating BL3-1.
+	 *
+	 *The BL3-1 relocation ensures that BL3-1 only runs from compiled
+	 *execution address. If previous boot-loader stage loads BL3-1 to
+	 *an address different from compiled execution address then BL3-1
+	 *relocation will copy BL3-1 to desired execution address and continue
+	 *from there.
+	 */
+
+	adr	x10, bl31_entrypoint
+	ldr	x11, __reloc_start
+	ldr	x12, __reloc_end
+	cmp	x10, x11
+	b.eq	_reloc_done
+1:	ldp	x13, x14, [x10], #16
+	stp	x13, x14, [x11], #16
+	cmp	x11, x12
+	b.lt	1b
+
+	ldr	x10, __reloc_done
+	br	x10
+	.align 3
+__reloc_start:
+	.dword BL31_BASE
+__reloc_end:
+	.dword __BL31_END__
+__reloc_done:
+	.dword _reloc_done
+_reloc_done:
+#endif
+
 #if !RESET_TO_BL31
 	/* ---------------------------------------------------------------
 	 * Preceding bootloader has populated x0 with a pointer to a


danh-arm:

danh-arm:

bl31/aarch64/bl31_entrypoint.S
```diff
@@ -42,6 +43,39 @@
 	 */
 
 func bl31_entrypoint
+#ifdef BL31_ENABLE_RELOCATION
```
danh-arm: As discussed in https://github.com/ARM-software/tf-issues/issues/382, BL31_DIFFERENT_LOAD_RUN_ADDRESS is a less confusing name

bl31/aarch64/bl31_entrypoint.S
```diff
@@ -42,6 +43,39 @@
...
+	 *from there.
+	 */
+
+	adr	x10, bl31_entrypoint
+	ldr	x11, __reloc_start
+	ldr	x12, __reloc_end
+	cmp	x10, x11
+	b.eq	_reloc_done
+1:	ldp	x13, x14, [x10], #16
+	stp	x13, x14, [x11], #16
```
danh-arm: As discussed in https://github.com/ARM-software/tf-issues/issues/382, implementation needs to check for a destructive overlapping copy "in the case that the end of the loaded image region overlaps the start of the run region". Suggest using ASM_ASSERT for this.

bl31/aarch64/bl31_entrypoint.S
danh-arm: Optional: Consider passing BL31_DIFFERENT_LOAD_RUN_ADDRESS through the build system by using the add_define make function in the top level Makefile. If you do this, BL31_DIFFERENT_LOAD_RUN_ADDRESS should be documented in the user guide.

danh-arm:
The commit message should more clearly state the use-case, e.g. re-use description here:
https://github.com/ARM-software/tf-issues/issues/382#issuecomment-204045436

sbranden:
if this patch serves no value to anyone else I think we can abandon it.  The use case will become less and less useful as we move fully to ATF bootloader.  Or, I can update commit message and you can merge if you like (but just means more code that may or may not work in the future)

vwadekar:
Hello Scott,

This patch makes sense for Tegra platforms as we use only the BL31
component from the entire ARM Trusted Firmware project. BL3-1 would be
running off of the secure SRAM on-chip for security/performance reasons and
has its own MMIO address which does not interfere with the DRAM addresses.

FWIW, we are currently using a similar trampoline code on our platforms.
Your change would allow our bootloaders to load the BL3-1 at any DRAM
location, program the reset vector, boot the CPU and then relocate to the
actual secure SRAM location.

Dan, if you think this change is OK to go in, then please review/merge this
PR. Since this feature is configurable, it would be easy to disable/enable
for platforms.

danh-arm:
Hi Varun/Scott

> Your change would allow our bootloaders to load the BL3-1 at any DRAM location...

There are potential security issues with this if that DRAM is accessible to non-secure code (e.g. firmware update code) prior to it being copied to secure SRAM, or if that code is left lying around afterwards. I would suggest this is generally only used as a development option. But that is an aside, and for you to decide on in your platform code.

> Dan, if you think this change is OK to go in...

As I said in https://github.com/ARM-software/tf-issues/issues/382#issuecomment-204336823, I'm willing to merge this in principle. I have already reviewed the code. If my comments are addressed (by either of you or someone else), I will merge. ARM has no particular interest in this so we're not going to rework this ourselves.

vwadekar:
Thanks Dan. The bl31 binary is encrypted and signed on our platforms. The
Memory controller is programmed with dram base/size to prepare a special
TZDRAM carveout during boot. The BL then decrypts bl31 to this TZDRAM
carveout. Since this carveout is movable we are not able to use the same
bl31 without recompiling for a new base. But with Scott's PR we don't have
to worry about this limitation.

FWIW, we already have a flavour of this change in our platform code
downstream.

danh-arm:
I'm closing this as the work seems to have stalled. Please raise a new PR with the rework if it's still important.

PR #571, opened 2016-03-31
drivers: add designware mmc driver

hzhuang1:
Changelog:
v2:
   * move dwmmc driver into another pull request.
   * reference return value to avoid build issue in release mode.

Support Designware eMMC driver. It's based on both IO block
and mmc driver.


hzhuang1:
Abandon. I will replace mmc by emmc instead.

PR #573, opened 2016-03-31
Mmc v2

hzhuang1:
Changelog:
v2:
   * Move dwmmc out of this pull request.
   * Rename mmc to emmc.
   * Fix the unused valure issue of release build.

https://github.com/ARM-software/tf-issues/issues/376


drivers/io/io_block.c
```diff
@@ -0,0 +1,298 @@
...
+	io_block_spec_t *region;
+
+	assert((dev_info->info != (uintptr_t)NULL) &&
+	       (spec != (uintptr_t)NULL) &&
+	       (entity->info == (uintptr_t)NULL));
+
+	block = (io_block_entity_t *)spec;
+	region = &block->region;
+	memcpy((void *)&block->buffer, (void *)dev_info->info,
+	       sizeof(io_block_spec_t));
```
danh-arm: From this I can infer that you expect platform code to pass an `io_block_spec_t` as the "block device spec" in its call to `io_dev_open()`. But:

*   It's not clear to me what this `io_block_spec_t` represents. Some comments or more info in the commit message would help.
*   It seems unnecessary to copy this into `block->buffer` since it's already stored in the global `block_dev_info.info`. If needed you could provide a macro that casts this to the appropriate type.

include/drivers/io/io_block.h
```diff
@@ -0,0 +1,55 @@
...
+
+typedef struct io_block_entity {
+	io_block_spec_t		region;
+	io_block_spec_t		buffer;
+	io_block_ops_t		ops;
+	size_t			file_pos;
+	int			block_size;
+	int			lba;
+	int			rpmb_enable;
+} io_block_entity_t;
```
danh-arm: It seems this structure is conflating 2 things; the "image spec" that is passed in the call to `block_open()`, and the internal structure that the driver uses to maintain state. As far as I can see, the "image spec" is entirely represented by the `region` field so you can just define this in platform code instead of an `io_block_entity`.

If you do need to define a new "image spec" type, then this should be defined in `io_storage.h` with the other related types. Hopefully this is not needed since it would be confusing for the "IO block driver" to require an image spec that is different to the existing `io_block_spec_t`.

`io_block_entity` should be private to the driver and declared in the c file or a private header.

include/drivers/io/io_block.h
```diff
@@ -0,0 +1,55 @@
...
+
+typedef struct io_block_ops {
+	size_t	(*read)(int lba, uintptr_t buf, size_t size);
+	size_t	(*write)(int lba, uintptr_t buf, size_t size);
+} io_block_ops_t;
+
+typedef struct io_block_entity {
+	io_block_spec_t		region;
+	io_block_spec_t		buffer;
+	io_block_ops_t		ops;
```
danh-arm: It looks like `buffer` and `ops` are fields that should be part of the "device spec" passed to `io_dev_open()`. Perhaps you need to define a new `io_block_dev_spec` structure to represent this?

include/drivers/io/io_block.h
```diff
@@ -0,0 +1,55 @@
...
+	size_t			file_pos;
+	int			block_size;
+	int			lba;
+	int			rpmb_enable;
```
danh-arm: rpmb is an eMMC specific thing. Should not be part of this interface.

drivers/io/io_block.c
```diff
@@ -0,0 +1,298 @@
...
+#include <assert.h>
+#include <debug.h>
+#include <errno.h>
+#include <io_block.h>
+#include <io_driver.h>
+#include <io_storage.h>
+#include <string.h>
+
+#define MIN_BLOCK_SIZE			64
+#define MIN_BLOCK_MASK			(MIN_BLOCK_SIZE - 1)
```
danh-arm: Why is this needed in the generic block driver? Should this check be in the eMMC driver instead?

drivers/io/io_block.c
```diff
@@ -0,0 +1,298 @@
...
+		      io_entity_t *entity)
+{
+	io_block_entity_t *block;
+	io_block_spec_t *region;
+
+	assert((dev_info->info != (uintptr_t)NULL) &&
+	       (spec != (uintptr_t)NULL) &&
+	       (entity->info == (uintptr_t)NULL));
+
+	block = (io_block_entity_t *)spec;
```
danh-arm: As stated elsewhere, it seems to me that `spec` should only contain `region`.

include/drivers/io/io_block.h
```diff
@@ -0,0 +1,55 @@
...
+	size_t	(*write)(int lba, uintptr_t buf, size_t size);
+} io_block_ops_t;
+
+typedef struct io_block_entity {
+	io_block_spec_t		region;
+	io_block_spec_t		buffer;
+	io_block_ops_t		ops;
+	size_t			file_pos;
+	int			block_size;
+	int			lba;
```
danh-arm: Why do you need to store the `lba` in this structure? It seems that this can be calculated at any time using the other fields in this structure.

drivers/io/io_block.c
```diff
@@ -0,0 +1,298 @@
...
+{
+	io_block_entity_t *block;
+	io_block_spec_t *region;
+
+	assert(entity->info != (uintptr_t)NULL);
+
+	block = (io_block_entity_t *)entity->info;
+	region = &block->region;
+
+	assert(offset < block->region.length);
```
danh-arm: Perhaps also check offset >=0

drivers/io/io_block.c
```diff
@@ -0,0 +1,298 @@
...
+
+	block->file_pos += offset;
+	block->lba = (block->file_pos + region->offset) / block->block_size;
+	assert(block->lba >= 0);
+	return 0;
+}
+
+static int block_read(io_entity_t *entity, uintptr_t buffer, size_t length,
+		      size_t *length_read)
```
danh-arm: Both the read and write functions could do with more comments or a big comment for each explaining what is going on.

include/drivers/io/io_block.h
danh-arm: Later: Actually, it looks like `block_size` might need to be part of the "image spec"

drivers/io/io_block.c
```diff
@@ -0,0 +1,298 @@
...
+				       count - skip);
+			}
+			left = left - (count - skip);
+		} else {
+			if (skip || padding)
+				count = block->ops.read(block->lba, buf->offset,
+							left);
+			else
+				count = block->ops.read(block->lba, buffer,
+							left);
```
danh-arm: So there are 4 different ways that the underlying `ops.read()` function is called. It's very difficult to understand what all the different cases are for without more comments.

include/drivers/io/io_block.h
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+#ifndef __IO_BLOCK_H__
+#define __IO_BLOCK_H__
+
+#include <io_storage.h>
+
+typedef struct io_block_ops {
+	size_t	(*read)(int lba, uintptr_t buf, size_t size);
+	size_t	(*write)(int lba, uintptr_t buf, size_t size);
```
danh-arm: `buf` arg in write function can be const.

include/drivers/emmc.h
```diff
@@ -0,0 +1,182 @@
...
+	unsigned int		reserved_3:		2;
+	unsigned int		spec_vers:		4;
+	unsigned int		csd_structure:		2;
+} emmc_csd_t;
+
+size_t emmc_read_blocks(int lba, uintptr_t buf, size_t size);
+size_t emmc_write_blocks(int lba, uintptr_t buf, size_t size);
+size_t emmc_erase_blocks(int lba, size_t size);
+size_t emmc_rpmb_read_blocks(int lba, uintptr_t buf, size_t size);
+size_t emmc_rpmb_write_blocks(int lba, uintptr_t buf, size_t size);
```
danh-arm: buf arg in write function can be const

include/drivers/emmc.h
```diff
@@ -0,0 +1,182 @@
...
+	unsigned int		tran_speed:		8;
+	unsigned int		nsac:			8;
+	unsigned int		taac:			8;
+	unsigned int		reserved_3:		2;
+	unsigned int		spec_vers:		4;
+	unsigned int		csd_structure:		2;
+} emmc_csd_t;
+
+size_t emmc_read_blocks(int lba, uintptr_t buf, size_t size);
+size_t emmc_write_blocks(int lba, uintptr_t buf, size_t size);
```
danh-arm: buf arg in write function can be const

drivers/emmc/emmc.c
```diff
@@ -0,0 +1,347 @@
...
+	       (ops_ptr->read != 0) &&
+	       (ops_ptr->write != 0) &&
+	       (clk != 0) &&
+	       ((width == EMMC_BUS_WIDTH_1) ||
+		(width == EMMC_BUS_WIDTH_4) ||
+		(width == EMMC_BUS_WIDTH_8)));
+	ops = ops_ptr;
+
+	emmc_enumerate(clk, width);
+}
```
danh-arm: This driver looks good as far as I can see.

Out of interest, do you have any need for eMMC "boot operation" mode? It seems to me that it would be good to have a "cut down" eMMC driver that only supports reading data out of the eMMC boot partition. Such a driver would be more suitable for inclusion in ROM code than a full eMMC driver.

include/drivers/io/io_block.h
hzhuang1: How about append a new helper in ops? like get_block_info()? Then I could get the block region and block size from this. Without region and block size, I can't calculate lba and file_pos.

drivers/emmc/emmc.c
hzhuang1: I must access eMMC boot operation (RPMB). In HiKey recovery mode, I need to update everything including RPMB. Otherwise, I could make HiKey working again when it becomes brick.

include/drivers/io/io_block.h
hzhuang1: Yes, I could remove it. Actually, the "rpmb_enable" field isn't used. I could declare different io_block_entity_t for boot area and user data area.

drivers/emmc/emmc.c
danh-arm: Yes, if you want your recovery mode IO drivers in BL1, then this will need to include write and RPMB support. On some platforms, it may be possible to put this additional driver in a separate non-secure image, keeping BL1 minimal.

By the way, when I say eMMC boot operation mode, I don't mean RPMB; I mean reading from the boot partition before CMD1 is issued.

hzhuang1:
Replaced by https://github.com/ARM-software/arm-trusted-firmware/pull/597

PR #578, opened 2016-04-01
Make improvements for host environment portability

EvanLloyd:
These changes enable building on host environments other than Linux/Unix.
This is for developer convenience, rather than a full solution, as the certification tool build is not tested.
This patch set adds changes for building on Windows CMD.EXE, Cygwin and MinGW.


danh-arm:
jenkins: add to whitelist

danh-arm:
jenkins: test this please

PR #590, opened 2016-04-12
Use unsigned long long instead of uintptr_t in TZC400/DMC500 drivers

yatharth-arm:
Currently the `tzc400_configure_region` and `tzc_dmc500_configure_region`
functions uses uintptr_t as the data type for `region_top` and `region_base`
variables, which will be converted to 32/64 bits for AArch32/AArch64
respectively. But the expectation is to keep these addresses at least 64 bit.

This patch modifies the data types to make it at least 64 bit by using
unsigned long long instead of uintptr_t for the `region_top` and
`region_base` variables. It also modifies the associated macros
`_tzc##fn_name##_write_region_xxx` accordingly.

Change-Id: I4e3c6a8a39ad04205cf0f3bda336c3970b15a28b

### DIFF
--- a/drivers/arm/tzc/tzc400.c
+++ b/drivers/arm/tzc/tzc400.c
@@ -191,8 +191,8 @@ void tzc400_configure_region0(tzc_region_attributes_t sec_attr,
  */
 void tzc400_configure_region(unsigned int filters,
 			  int region,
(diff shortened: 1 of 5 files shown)


danh-arm:
jenkins: retest this please

PR #592, opened 2016-04-13
Tf issues/391

sorenb-xlnx:


danh-arm:
This PR does not seem to be a correct workaround implementation for the errata in question. Could you describe the circumstances that lead you to hitting the errata? Or is this just a theorical issue?

In particular, the errata only occurs when HCR_EL2.VM is enabled and address translation is disabled for the current EL (and the current EL is EL3, EL2 or S-EL1). I presume that HCR_EL2.VM is enabled by EL2 after secure world cold boot in your environment? Generic TF code and the CPU libraries do not disable EL3 address translation after this point, nor does your upstream BL31 platform port as far as I can see, so there should be no issue there. If EL2 disables address translation, then it can clear HCR_EL2.VM itself without the help of TF.

The only problematic case I can see is if S-EL1 wants to disable address translation. It cannot modify HCR_EL2 so would need the help of EL3 (e.g. via a Fast SMC). But that seems unlikely.

Also, you seem to be unnecessarily saving and restoring HCR_EL2.

sorenb-xlnx:
Thus far it's just a theoretical issue and more about preventing hard to debug issues in the future.

I agree the ATF is unlikely to be problematic. But as you mention, a secure OS could cause this kind of problem, and those may be hard to control in what they are doing. Hence, a generic workaround in ATF seems to be desirable, as that would allow secure EL1 code to be written without making any additional assumptions regarding address translation. Also, if anybody decides to switch on/off address translation in EL3, this should be covered with this too.

The approach here is certainly rather brute force, but in theory should cover all affected scenarios in the simplest way (I think).

So, I guess we come down to:
 - Do you see ATF to be the correct place to put a workaround for this issue?
If so:
 - How would  you like it to be done?

I assume this patch would work. But yes, it is storing hcr_el2 under all circumstances. I guess that could be reduced to be only enabled when the errata item is selected, but that would make the #ifdeffery in the sys_regs_(re)store function a little more ugly.

athoelke:
I agree with Dan. The only points at which this may be a problem are after disabling address translation.

At CPU reset on A53 `HCR_EL2.VM=0`, so the cold and warm boot paths which may include some operation at EL3, EL2 and Secure EL1 when address translation is disabled are not impacted by this errata.

If EL3 code needs to disable address translation (for example, I'm thinking about Tegra's video memory support) then HCR_EL2 can be saved, modified and restored around that section of code if it might run on an A53.

There should be no need to disable adddress translation at Secure-EL1, but this is the case where some changes/EL3 support might be needed.

Unilaterally imposing an unnecessary extra register save/restore which will UNDEF on some implementations (those with no EL2) is inappropriate.

sorenb-xlnx:
On Thu, 2016-04-14 at 06:52:53 -0700, Andrew Thoelke wrote:
> I agree with Dan. The only points at which this may be a problem are after disabling address translation.
> 
> At CPU reset on A53 `HCR_EL2.VM=0`, so the cold and warm boot paths which may include some operation at EL3, EL2 and Secure EL1 when address translation is disabled are not impacted by this errata.

agree

> 
> If EL3 code needs to disable address translation (for example, I'm thinking about Tegra's video memory support) then HCR_EL2 can be saved, modified and restored around that section of code if it might run on an A53.

This is the question. Do we want platform code to
individually handle this case or can we provide a generic solution in
common code? I'd prefer a common solution over pushing this down to
platform maintainers to get right.

> There should be no need to disable adddress translation at Secure-EL1, but this is the case where some changes/EL3 support might be needed.

agree

> Unilaterally imposing an unnecessary extra register save/restore which will UNDEF on some implementations (those with no EL2) is inappropriate.

I just wanted to put something out here to talk about. Changing the way
this is done is expected. But it seems you don't want this at all.

danh-arm:
I think we could potentially add some support code into TF to help workaround this errata but it's never going to provide a blanket solution for all environments; it's going to require some co-operation with affected BL31 platform code or S-EL1 code. Clearing this bit on every entry into the secure world and restoring HCR_EL2 on every exit, for all CPU types, is over the top, even on systems affected by the errata. Even then this doesn't help if EL2 disables address translation without clearing the bit.

In practice disabling address translation at EL3/S-EL1 after cold boot is rare. Enabling it again later is even rarer. The Tegra platform only does this because dynamic translation table modification is not implemented yet (https://github.com/ARM-software/tf-issues/issues/310), so a point solution there (if required) is appropriate.

The errata text says this "is most likely to arise in powerdown code, if EL2 or EL3 software
disables address translation before the core is powered down". But the A53 powerdown code does not disable address translatation so does not need changing.

Potentially a cpu library hook could be added to `disable_mmu_el3`. The A53 implementation of this hook would clear the `HCR_EL2.VM` bit if the errata workaround was enabled. But this would be low value since it would still be up to the (platform) calling code to save/restore `HCR_EL2` if required.

Another potential modification would be to add an SMC (possibly as part of a new CPU Service) that allows secure world callers to ask for the `HCR_EL2.VM` bit to be cleared (or restored). It would be up to BL31 platform code to choose whether to include this service, and S-EL1 code to make correct use of it. This would be quite a lot of effort for low benefit unless there were concrete pull for it from Trusted OS vendors.

sorenb-xlnx:
OK. Thanks for the feedback.

PR #593, opened 2016-04-14
mt8173: Fix timing issue of mfg mtcmos power off

mtk09422:
In mt8173, there are totally 10 non-cpu mtcmos, so we cannot tell
if SPM finished the power control flow by 10 status bits of PASR_PDP_3.
So, extend PASR_PDP_3 status bits from 10 to 20 so that we can
make sure if the control action has been done precisely.

Change-Id: Ifd4faaa4173c6e0543aa8471149adb9fe7fadedc
Signed-off-by: Fan Chen <fan.chen@mediatek.com>


danh-arm:
jenkins: test this please

danh-arm:
jenkins: retest this please

PR #601, opened 2016-04-22
Cortex-A57 errata workarounds

sandrine-bailleux-arm:
This patch stack implements a number of Cortex-A57 errata workarounds in the CPU
reset handler. All of them are disabled by default and can be enabled by the platform
using the relevant build flags.


danh-arm:
off-topic comment: The CPU version that each errata applies to is documented in 4 separate places (the code, the makefile, the markdown document and the errata notice). This is probably over the top and we should remove this information from 2 of these places to improve maintainability.

This can be looked at separately so merging...

PR #606, opened 2016-04-26
Fix computation of L1 bitmask in the translation table lib

sandrine-bailleux-arm:
This patch fixes the computation of the bitmask used to isolate
the level 1 field of a virtual address. The whole computation needs
to work on 64-bit values to produce the correct bitmask value.
XLAT_TABLE_ENTRIES_MASK being a C constant, it is a 32-bit value
so it needs to be extended to a 64-bit value before it takes part
in any other computation.

This patch fixes this bug by casting XLAT_TABLE_ENTRIES_MASK as
an unsigned long long.

Note that this bug doesn't manifest itself in practice because
address spaces larger than 39 bits are not yet supported in the
Trusted Firmware.

### DIFF
--- a/lib/xlat_tables/xlat_tables_common.c
+++ b/lib/xlat_tables/xlat_tables_common.c
@@ -284,8 +284,9 @@ static mmap_region_t *init_xlation_table_inner(mmap_region_t *mm,
 	unsigned level_size_shift = L1_XLAT_ADDRESS_SHIFT - (level - 1) *
 						XLAT_TABLE_ENTRIES_SHIFT;
 	unsigned level_size = 1 << level_size_shift;
-	unsigned long long level_index_mask = XLAT_TABLE_ENTRIES_MASK <<
-							level_size_shift;
+	unsigned long long level_index_mask =
+		((unsigned long long) XLAT_TABLE_ENTRIES_MASK)
+		<< level_size_shift;
 
 	assert(level > 0 && level <= 3);
 
--- a/plat/arm/common/aarch64/arm_common.c
+++ b/plat/arm/common/aarch64/arm_common.c
@@ -163,6 +163,7 @@ const mmap_region_t *plat_arm_get_mmap(void)
 	return plat_arm_mmap;
 }
 
+#ifdef ARM_SYS_CNTCTL_BASE
 unsigned long long plat_get_syscnt_freq(void)
 {
 	unsigned long long counter_base_frequency;
@@ -176,3 +177,4 @@ unsigned long long plat_get_syscnt_freq(void)
 
 	return counter_base_frequency;
 }
+#endif /* ARM_SYS_CNTCTL_BASE */


plat/arm/common/aarch64/arm_common.c
```diff
@@ -163,6 +163,7 @@ const mmap_region_t *plat_arm_get_mmap(void)
 	return plat_arm_mmap;
 }
 
+#ifdef ARM_SYS_CNTCTL_BASE
```
sorenb-xlnx: I don't think this should be part of this PR.

lib/xlat_tables/xlat_tables_common.c
```diff
@@ -284,8 +284,9 @@ static mmap_region_t *init_xlation_table_inner(mmap_region_t *mm,
 	unsigned level_size_shift = L1_XLAT_ADDRESS_SHIFT - (level - 1) *
 						XLAT_TABLE_ENTRIES_SHIFT;
 	unsigned level_size = 1 << level_size_shift;
-	unsigned long long level_index_mask = XLAT_TABLE_ENTRIES_MASK <<
-							level_size_shift;
+	unsigned long long level_index_mask =
+		((unsigned long long) XLAT_TABLE_ENTRIES_MASK)
+		<< level_size_shift;
```
sorenb-xlnx: Probably just cosmetics, but would probably fix additional users too: I haven't tested it, but wouldn't it work to suffix the constant with ```ULL``` in the header and avoid the casting here?

plat/arm/common/aarch64/arm_common.c
sandrine-bailleux-arm: I agree, it was not intentional and is due to the integration branch and the master branch
not being in sync at the moment. Will be fixed soon.

lib/xlat_tables/xlat_tables_common.c
sandrine-bailleux-arm: (At the moment, init_xlation_table_inner() is the only place where 
XLAT_TABLE_ENTRIES_MASK is used so there is no additional user for now.)

I agree suffixing the constant with ULL would work as well but it is not necessarily what
one wants in all cases. XLAT_TABLE_ENTRIES_MASK equals 0x1ff, which easily
fits in 32 bits. Whether we need to cast it to a 64-bit value depends on the context in
which it is used so I think the cast should remain wherever the constant is used, not
in the definition itself.

sandrine-bailleux-arm: I don't understand what's happening, the integration and master branches have been synced
but this pull request is still showing 3 commits... I've created a new pull request, see #607,
as I couldn't figure out how to fix it. The code is exactly the same as this pull request. Sorry
for the confusion.

sorenb-xlnx: On Tue, 2016-04-26 at 08:18:51 -0700, Sandrine Bailleux wrote:
> > @@ -284,8 +284,9 @@ static mmap_region_t *init_xlation_table_inner(mmap_region_t *mm,
> >  	unsigned level_size_shift = L1_XLAT_ADDRESS_SHIFT - (level - 1) *
> >  						XLAT_TABLE_ENTRIES_SHIFT;
> >  	unsigned level_size = 1 << level_size_shift;
> > -	unsigned long long level_index_mask = XLAT_TABLE_ENTRIES_MASK <<
> > -							level_size_shift;
> > +	unsigned long long level_index_mask =
> > +		((unsigned long long) XLAT_TABLE_ENTRIES_MASK)
> > +		<< level_size_shift;
> 
> (At the moment, init_xlation_table_inner() is the only place where 
> XLAT_TABLE_ENTRIES_MASK is used so there is no additional user for now.)
> 
> I agree suffixing the constant with ULL would work as well but it is not necessarily what
> one wants in all cases. XLAT_TABLE_ENTRIES_MASK equals 0x1ff, which easily
> fits in 32 bits. Whether we need to cast it to a 64-bit value depends on the context in
> which it is used so I think the cast should remain wherever the constant is used, not
> in the definition itself.

makes sense

sandrine-bailleux-arm:
Closing this pull request in favour of the new (identical) one.

PR #608, opened 2016-04-27
Add support to change memory and TCR attributes

davwan01:


achingupta:
Hi David,

Had a very quick look at this PR. I am trying to understand the use case. If you are storing your code and data in non-cacheable memory then why bother enabling the MMU in the first place? This sounds like an exception rather than the norm which does not justify changes to the generic code. Please let me know what am I missing?

thanks,
Achin

danh-arm:
Hi Achin

Enabling the MMU not only enables caching; it restricts access to the required regions with the appropriate access permissions and (potentially) allows non-identity mappings. There may be other future platforms that require a particular BL to only use non-cacheable memory while taking advantage of the other MMU features, so it seems reasonable to me to support this use-case. 

Cheers

Dan.

plat/arm/common/aarch64/arm_common.c
```diff
@@ -52,40 +53,56 @@ extern const mmap_region_t plat_arm_mmap[];
...
 				coh_limit - coh_start,			\
 				MT_DEVICE | MT_RW | MT_SECURE);		\
 		mmap_add(plat_arm_get_mmap());				\
 		init_xlat_tables();					\
 									\
-		enable_mmu_el##_el(0);					\
+		if (type == MT_MEMORY) {				\
+			enable_mmu_el##_el(TCR_MT_DEFAULT);		\
+		} else {						\
+			assert(type == MT_NON_CACHEABLE);		\
```
achingupta: If code and data is not being marked with cache-able attributes then why restrict the alternative to being `NON_CACHEABLE`? This code is common to all ARM platforms. If a particular platform only supports NON_CACHEABLE then it should validate the `type` before calling `arm_configure_mmu_elX()`.

plat/arm/common/aarch64/arm_common.c
```diff
@@ -52,40 +53,56 @@ extern const mmap_region_t plat_arm_mmap[];
...
+			enable_mmu_el##_el(TCR_MT_NON_CACHEABLE);	\
```
achingupta: It is a bit strange that the `type` for the memory used by the image forces the memory attributes used for the page tables to be non-cacheable as well. Is it not conceivable to place page tables in cache-able and the image (code+data) in non-cacheable memory.

davwan01:
Hi Achin,
Yes, if the page tables are put in cacheable, the TCR should be cacheable.
This patch is for ARM common platforms and now they don't have this scenario yet.
So this is not a perfect solution to support all kinds of scenarios we could configure for images, page tables, and other spaces in the future.
We could expand the support for different configurations when we have on ARM common platforms based on idea of this patch.

achingupta:
Hi Dan,

Fair enough! I was fixated with caching since from what I vaguely recalled, it was the issue that motivated this change in the first place. If the intention is to cater for a hypothetical future use case, then let us not restrict ourselves to non-cacheable attributes. Also, the attributes for the page tables should be treated differently from the attributes of the translations it performs.

hth,
Achin

plat/arm/common/aarch64/arm_common.c
danh-arm: The TCR memory type should match the BL RW data memory type (given that the xlat_tables are stored in BL RW data on ARM platforms). The xlat_tables library currently only supports TCR memory types of `TCR_MT_DEFAULT` and `TCR_MT_NON_CACHEABLE`, so only the only supported values of `mem_type` are `MT_MEMORY` and `MT_NON_CACHEABLE`. So this assert is reasonable.

danh-arm:
We no longer think this PR is needed in the upstream.

PR #610, opened 2016-04-27
fip_create: Fix exit status for missing output filename (2)

bjackman:
Not sure whether I should submit against `integration` or `master` so I've done both, just reject whichever is wrong.


danh-arm:
The head branch should always be integration. I'll close the other one.

PR #621, opened 2016-05-12
mt8173: Reorganize plat SiP functions

mtk09422:
Due to the changes in Mediatek platform common code, we need to move
plat related SiP functions to plat folder.

Change-Id: I6b14b988235205a5858b4bf49043bc79d0512b06
Signed-off-by: Jimmy Huang <jimmy.huang@mediatek.com>


danh-arm:
This patch is included in #622 so I'll just merge it as one PR.

PR #624, opened 2016-05-20
Implement generic delay timer and use it on platforms

antonio-nino-diaz-arm:
Implemented generic delay timer, which is now used on FVP instead of the SP804 timer. The latter can still be used if the build option FVP_USE_SP804_TIMER is set to 1.

Substituted code in MediaTek and Rockchip platforms by calls to the generic delay timer.

Implemented a 32 bit version of plat_get_syscnt_freq().


sorenb-xlnx:
I wonder whether it's worth to go through this whole deprecation process for *plat_get_syscnt_freq()*, instead of just globally replacing it. But I guess it's the nicer way for non-upstream platforms.

The porting guide should be updated to the new API.

Acked-by: Sören Brinkmann <soren.brinkmann@xilinx.com>

danh-arm:
> I wonder whether it's worth to go through this whole deprecation process for `plat_get_syscnt_freq()` ...

I know it's a bit fiddly but in general we want to avoid breaks where we can. There are more non-upstream platforms than upstream ones. At some point, we'd like to release an incompatible version that cleans up all this deprecated stuff (no immediate plans for this yet).

Good point about the docs.

danh-arm:
Please raise a new PR targetting the integration branch instead of master.

antonio-nino-diaz-arm:
New PR, with updated porting guide: https://github.com/ARM-software/arm-trusted-firmware/pull/625

PR #628, opened 2016-05-24
Support for gpio reset v1

rkchrome:
Add the gpio configure to complete the system off/reset soc.

In order to properly reboot, the AP should not use the standar dreboot methods.  We need to use "AP_EC_WARM_RESET_REQ" .

The reason for this is that the reset method won't reset the TPM.

That means:
* I think the ARM Trusted Firmware is in charge of rebooting the machine, even when the kernel is running (PSCI is used to call ATF).  That means the firmware needs to do this.

* Also if the firmware sees that a Hardware Watchdog Reset happened it should log (using the event log) and then do a GPIO-based reset so that the TPM gets rebooted.

### DIFF
--- a/drivers/gpio/gpio.c
+++ b/drivers/gpio/gpio.c
@@ -80,6 +80,16 @@ void gpio_set_value(int gpio, int value)
 	ops->set_value(gpio, value);
 }
 
+void gpio_set_pull(int gpio, int pull)
+{
+	assert(ops);
+	assert(ops->set_pull != 0);
+	assert((pull == PULLNONE) || (pull == PULLUP) || (pull == PULLDOWN));
+	assert(gpio >= 0);
+
+	ops->set_pull(gpio, pull);
+}
+
 /*
  * Initialize the gpio. The fields in the provided gpio
  * ops pointer must be valid.
--- a/include/drivers/gpio.h
+++ b/include/drivers/gpio.h
@@ -37,17 +37,25 @@
 #define GPIO_LEVEL_LOW		0
 #define GPIO_LEVEL_HIGH		1
 
+enum {
+	PULLNONE = 0,
+	PULLUP,
+	PULLDOWN
+};
+
 typedef struct gpio_ops {
 	int (*get_direction)(int gpio);
 	void (*set_direction)(int gpio, int direction);
 	int (*get_value)(int gpio);
 	void (*set_value)(int gpio, int value);
+	void (*set_pull)(int gpio, int pull);
 } gpio_ops_t;
 
 int gpio_get_direction(int gpio);
 void gpio_set_direction(int gpio, int direction);
 int gpio_get_value(int gpio);
 void gpio_set_value(int gpio, int value);
+void gpio_set_pull(int gpio, int pull);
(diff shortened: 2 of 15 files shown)


include/drivers/gpio.h
```diff
@@ -37,17 +37,25 @@
 #define GPIO_LEVEL_LOW		0
 #define GPIO_LEVEL_HIGH		1
 
+enum {
+	PULLNONE = 0,
+	PULLUP,
+	PULLDOWN
+};
+
```
danh-arm: Use #defines for consistency with rest of API, e.g.

```
#define GPIO_PULL_NONE 0
#define GPIO_PULL_UP 1
#define GPIO_PULL_DOWN 2
```

drivers/gpio/gpio.c
```diff
@@ -80,6 +80,16 @@ void gpio_set_value(int gpio, int value)
...
 }
 
+void gpio_set_pull(int gpio, int pull)
+{
+	assert(ops);
+	assert(ops->set_pull != 0);
+	assert((pull == PULLNONE) || (pull == PULLUP) || (pull == PULLDOWN));
+	assert(gpio >= 0);
+
+	ops->set_pull(gpio, pull);
```
danh-arm: minor: Would be good to make it clear in the commit message that this is an optional function (since it is not asserted for in gpio_init()).

drivers/gpio/gpio.c
```diff
@@ -80,6 +80,16 @@ void gpio_set_value(int gpio, int value)
...
+}
+
```
danh-arm: Should provide `gpio_get_pull()` for consistency

plat/rockchip/rk3399/drivers/soc/soc.c
```diff
@@ -57,7 +57,18 @@ const mmap_region_t plat_rk_mmap[] = {
...
+	MAP_REGION_FLAT(GPIO1_BASE, GPIO1_SIZE,
+			MT_DEVICE | MT_RW | MT_SECURE),
+	MAP_REGION_FLAT(GPIO2_BASE, GPIO2_SIZE,
+			MT_DEVICE | MT_RW | MT_SECURE),
+	MAP_REGION_FLAT(GPIO3_BASE, GPIO3_SIZE,
+			MT_DEVICE | MT_RW | MT_SECURE),
+	MAP_REGION_FLAT(GPIO4_BASE, GPIO4_SIZE,
+			MT_DEVICE | MT_RW | MT_SECURE),
+	MAP_REGION_FLAT(GRF_BASE, GRF_SIZE,
+			MT_DEVICE | MT_RW | MT_SECURE),
```
danh-arm: minor: If these regions are not contiguous, you may end up using more memory in page tables than just mapping in the whole GPIO space as a single region.

danh-arm:
This PR should target the integration branch, not the master branch. Please create a new one with other comments addressed.

plat/rockchip/rk3399/drivers/soc/soc.c
rkchrome: Dan, we think about it and decide to use it for better debugging

PR #629, opened 2016-05-24
Bring IO storage dummy driver

ljerry:
Allow to handle cases where some images are pre-loaded (by debugger for
instance) without introducing many switches in files calling load_* functions.

Fixes: arm-software/tf-issues#398
Signed-off-by: Gerald Lejeune <gerald.lejeune@st.com>


drivers/io/io_dummy.c
```diff
@@ -0,0 +1,177 @@
...
+
+static const io_dev_connector_t dummy_dev_connector = {
+	.dev_open = dummy_dev_open
+};
+
+
+static const io_dev_funcs_t dummy_dev_funcs = {
+	.type = device_type_dummy,
+	.open = dummy_block_open,
+	.seek = NULL,
```
danh-arm: for completeness, would it not be better to provide dummy implementations for all functions?

drivers/io/io_dummy.c
```diff
@@ -0,0 +1,177 @@
...
+		assert(entity != NULL);
+
+		current_file.in_use = 1;
+		current_file.size = block_spec->length;
+		entity->info = (uintptr_t)&current_file;
+		result = 0;
+	} else {
+		WARN("A Dummy device is already active. Close first.\n");
+		result = -ENOMEM;
+	}
```
danh-arm: This is a dummy subset of the memmap device functionality, but it's not clear why some of the functionality is retained (e.g. current_file.in_use and current_file.size) while other parts are removed (e.g. current_file.base). It seems to be this should either be stripped down into a true dummy device or this should be called `IO_TYPE_MEMMAP_DUMMY` to make it clear that clients must use the same semantics as the `IO_TYPE_MEMMAP` device.

danh-arm: If this is tightly coupled with the MEMMAP device, then could they both share some code?

ljerry: Dummy IO implementation is widely inspired from Memmap IO one. Nevertheless, even if implementation is a kind of subset of Memmap IO, the underlying idea is clearly different as it is some sort of "Identity IO". Identity because already in place. This is why seek() is not implemented: only FIP IO makes use of seek() but there is no point to use FIP here since images are already in place. current_file.base was dropped because it is not needed in this scope.
This is why I am not sure that IO_TYPE_MEMMAP_DUMMY will be clearest naming. IO_TYPE_IDENTITY was foreseen for a while with no strong conviction that it is better from a clarity pov.

Some code may be shared with memmap implementation, it is a matter of adjusting the cursor between factorization and simplicity.

Where to put the factorization? Encapsulate memcpy() in a generic function (no idea here for its name) for IO memmap and define this same function to null for IO dummy?

danh-arm: I prefer IO_TYPE_DUMMY to IO_TYPE_IDENTITY since that makes it clearer that this device does not do much. I'll let this go as-is, although I think the whole IO framework could do with an overhaul at some point.

danh-arm:
jenkins:test this please

PR #631, opened 2016-05-25
Support for Mediatek MT6795 SoC

leon-chen-mtk:
This patch only support single core to boot to Linux kernel
through Trusted Firmware.
It also support 32 bit kernel and 64 bit kernel booting.
SMP and hotplug patch will be available for later SoC.

Signed-off-by: Leon Chen <leon.chen@mediatek.com>

### DIFF
--- a/plat/mediatek/mt6795/bl31_plat_setup.c
+++ b/plat/mediatek/mt6795/bl31_plat_setup.c
@@ -1,0 +1,496 @@
+/*
+ * Copyright (c) 2016, ARM Limited and Contributors. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * Neither the name of ARM nor the names of its contributors may be used
+ * to endorse or promote products derived from this software without specific
+ * prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+#include <arm_gic.h>
+#include <assert.h>
+#include <arch_helpers.h>
+#include <bl_common.h>
+#include <cci.h>
+#include <console.h>
+#include <context_mgmt.h>
+#include <debug.h>
+#include <mcucfg.h>
+#include <mmio.h>
+#include <mtk_sip_svc.h>
+#include <mtk_plat_common.h>
+#include <mt_cpuxgpt.h>
+#include <platform.h>
+#include <plat_private.h>
+#include <string.h>
+#include <xlat_tables.h>
+/*******************************************************************************
+ * Declarations of linker defined symbols which will help us find the layout
+ * of trusted SRAM
+ ******************************************************************************/
+unsigned long __RO_START__;
+unsigned long __RO_END__;
+
+unsigned long __COHERENT_RAM_START__;
+unsigned long __COHERENT_RAM_END__;
+
+/*
+ * The next 2 constants identify the extents of the code & RO data region.
+ * These addresses are used by the MMU setup code and therefore they must be
+ * page-aligned.  It is the responsibility of the linker script to ensure that
+ * __RO_START__ and __RO_END__ linker symbols refer to page-aligned addresses.
+ */
+#define BL31_RO_BASE (unsigned long)(&__RO_START__)
+#define BL31_RO_LIMIT (unsigned long)(&__RO_END__)
+
+/*
+ * The next 2 constants identify the extents of the coherent memory region.
+ * These addresses are used by the MMU setup code and therefore they must be
+ * page-aligned.  It is the responsibility of the linker script to ensure that
+ * __COHERENT_RAM_START__ and __COHERENT_RAM_END__ linker symbols
+ * refer to page-aligned addresses.
+ */
+#define BL31_COHERENT_RAM_BASE (unsigned long)(&__COHERENT_RAM_START__)
+#define BL31_COHERENT_RAM_LIMIT (unsigned long)(&__COHERENT_RAM_END__)
+
+/*
+ * Placeholder variables for copying the arguments that have been passed to
+ * BL3-1 from BL2.
+ */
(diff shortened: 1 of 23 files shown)


plat/mediatek/mt6795/bl31_plat_setup.c
```diff
@@ -0,0 +1,496 @@
...
+		mmap_add(plat_mmap);					\
+		init_xlat_tables();					\
+									\
+		enable_mmu_el ## _el(0);				\
+	}
+
+/* Define EL3 variants of the function initialising the MMU */
+DEFINE_CONFIGURE_MMU_EL(3)
+
+unsigned long long plat_get_syscnt_freq(void)
```
danh-arm: Now deprecated (following #625). Should change this to plat_get_syscnt_freq2, returning unsigned int.

plat/mediatek/mt6795/bl31_plat_setup.c
```diff
@@ -0,0 +1,496 @@
...
+			gteearg.atf_log_buf_start,
+			gteearg.atf_log_buf_size);
+
+		mmap_add_region(
+			gteearg.atf_log_buf_start &
+			~(PAGE_SIZE_2MB_MASK),
+			gteearg.atf_log_buf_start &
+			~(PAGE_SIZE_2MB_MASK),
+			PAGE_SIZE_2MB,
+			MT_DEVICE | MT_RW | MT_NS);
```
danh-arm: It seems this is mapping in some non-secure memory for logging purposes, based on information passed by the MTK preloader (BL2). This is a potential security attack vector so you should ensure this logging information is not available in production devices.

plat/mediatek/mt6795/bl31_plat_setup.c
```diff
@@ -0,0 +1,496 @@
...
+	} else{
+		gteearg.atf_aee_debug_buf_size = 0;
+		gteearg.atf_aee_debug_buf_start = 0;
+	}
+
+	/*
+	 * Without this, access to CPUECTRL from NS EL1
+	 * will cause trap into EL3
+	 */
+	enable_ns_access_to_cpuectlr();
```
danh-arm: Providing blanket NS access to CPUECTLR is a potential security hazard. It's preferable to provide a more specific interface to the required control bits.

plat/mediatek/mt6795/drivers/timer/mt_cpuxgpt.c
```diff
@@ -0,0 +1,111 @@
...
+{
+	normal_time_base = normal_base;
+	atf_time_base = atf_base;
+}
+
+uint64_t sched_clock(void)
+{
+	uint64_t cval;
+
+	cval = (((read_cntpct_el0() - atf_time_base)*1000)/13)
```
danh-arm: Better to not use frequency magic numbers, Can you not use SYS_COUNTER_FREQ_IN_TICKS?

plat/mediatek/mt6795/drivers/timer/mt_cpuxgpt.c
```diff
@@ -0,0 +1,111 @@
...
+		(uint32_t)(cval & 0xffffffff));
+}
+
+void sched_clock_init(uint64_t normal_base, uint64_t atf_base)
+{
+	normal_time_base = normal_base;
+	atf_time_base = atf_base;
+}
+
+uint64_t sched_clock(void)
```
danh-arm: This does not seem to be used anywhere. Also, TF contains a delay timer framework and a generic timer implementation (in drivers/delay_timer/generic_delay_timer.c), which might usable instead?

plat/mediatek/mt6795/drivers/uart/8250_console.S
```diff
@@ -0,0 +1,112 @@
+/*
```
danh-arm: This looks very similar to plat/mediatek/mt8173/drivers/uart/8250_console.S. Can this be re-used?

plat/mediatek/mt6795/drivers/uart/uart8250.h
```diff
@@ -0,0 +1,63 @@
+/*
```
danh-arm: This is identical to plat/mediatek/mt8173/drivers/uart/uart8250.h. Can it be re-used?

plat/mediatek/mt6795/include/mcucfg.h
```diff
@@ -0,0 +1,180 @@
+/*
```
danh-arm: This is very simlilar to plat/mediatek/mt8173/include/mcucfg.h. Is any code sharing possible?

plat/mediatek/mt6795/include/plat_macros.S
```diff
@@ -0,0 +1,110 @@
+/*
```
danh-arm: Looks very similar to plat/mediatek/mt8173/include/plat_macros.S

plat/mediatek/mt6795/plat_delay_timer.c
```diff
@@ -0,0 +1,52 @@
+/*
```
danh-arm: This can be removed and drivers/delay_timer/generic_delay_timer.c used instead. See https://github.com/ARM-software/arm-trusted-firmware/pull/625/commits/1d0b990e9d7e28fdf0de588815cd21d88c9fc2b7

plat/mediatek/mt6795/plat_pm.c
```diff
@@ -0,0 +1,512 @@
...
+	__asm__ volatile("mrs	%x0, cntp_ctl_el0"
+			 : "=r" (cntpctl) : : "memory");
+
+	return cntpctl;
+}
+
+static inline void write_cntpctl(uint64_t cntpctl)
+{
+	__asm__ volatile("msr	cntp_ctl_el0, %x0" : : "r"(cntpctl));
+}
```
danh-arm: These functions are not really necessary. You could add `DEFINE_SYSREG_RW_FUNCS(cntp_ctl_el0)` to arch_helpers.h instead.

danh-arm:
Regarding the PR description:

> This patch only support single core to boot to Linux kernel
through Trusted Firmware.
It also support 32 bit kernel and 64 bit kernel booting.
SMP and hotplug patch will be available for later SoC.

Are you saying this SoC will only ever support single core boot (to be fixed in future SoCs), or will SMP support on this SoC be addressed via a future patch? It's preferable to only upstream platforms when they are functionally mature.

plat/mediatek/mt6795/bl31_plat_setup.c
leon-chen-mtk: We will remove this function, since this is not comply to PSCI spec. SMPEN bit should be enabled in cortex_xx_reset_func

plat/mediatek/mt6795/bl31_plat_setup.c
leon-chen-mtk: We need to pass secure world log to normal world logging framework. This logging framework collect userspace/kernel/modem/secure world/other small MCU log. This log buffer write in secure world, read in normal world. Normal world attack should not access sensitive data or crash secure world, since secure world only write this buffer?  Or could you suggest us if we want log resides in memory buffer for logging system, other ways to transport secure world log to normal world except mapping the log memory for NS access? Need to design a handshake/protocol not just mapping to NS?

danh-arm: It's good that the log buffer is only written (not read) in the secure world, since you can't stop a malicious normal world entity from writing to this buffer.

My main concern here is that the log buffer could contain detailed information about how the secure world operates, which could be useful to an attacker. In general it's better if this information is suppressed in a production device. However, this is just a recommendation and will not gate upstreaming of your platform port.

plat/mediatek/mt6795/drivers/timer/mt_cpuxgpt.c
leon-chen-mtk: Hi Dan,
Do you mean I should just return 13000000 in plat_get_syscnt_freq() instead of define another macro for this number?

plat/mediatek/mt6795/drivers/timer/mt_cpuxgpt.c
leon-chen-mtk: We use sched_clock() to get system up time in our modified tf_printf function to print timestamp.

plat/mediatek/mt6795/include/mcucfg.h
leon-chen-mtk: MCU configuration register memory layout in 6795 is similar to 8173. However, in later SoC, the MCU configuration register differs, mcucfg.h content will change. mcucfg.h should still exist by platform SoC.

plat/mediatek/mt6795/include/plat_macros.S
leon-chen-mtk: So far mt6795 dump the same registers with  mt8173 while crash report generates. In later SoC, GiCv3 related registers will be dumped.

plat/mediatek/mt6795/drivers/timer/mt_cpuxgpt.c
danh-arm: I mean it's better to only define the frequency in one macro, and then use that macro where the frequency is needed. You already define SYS_COUNTER_FREQ_IN_TICKS, which is good, but here you should either use that macro or use the `cntfrq_el0` system register (if it has been initialized by this point), rather than using magic numbers that happen to have the same value.

PR #648, opened 2016-06-13
opteed: assume aarch64 for optee

ashutoshksingh:
OPTEE to execute in aarch64 bit mode, set it accordingly
when execution transitions from EL3 to EL1

Change-Id: I59f2f940bdc1aac10543045b006a137d107ec95f
Signed-off-by: Ashutosh Singh <ashutosh.singh@arm.com>

### DIFF
--- a/services/spd/opteed/opteed_main.c
+++ b/services/spd/opteed/opteed_main.c
@@ -148,7 +148,7 @@ int32_t opteed_setup(void)
 	 * state i.e whether AArch32 or AArch64. Assuming it's AArch32
 	 * for the time being.
 	 */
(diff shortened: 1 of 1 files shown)


sorenb-xlnx:
Why is one hardcoded handoff better than the other?
With this change existing flows break, still not providing a way to dynamically decide in what mode to hand off to.

ashutoshksingh:
Hi,

OPTEE-OS supports 64 bit builds so it makes logical sense to default to aarch64 for 64 bit platforms. Dynamic selection is possible based on optee header parsing, however that's a bigger change and will be taken up later.

Thanks,
Ashutosh

PR #655, opened 2016-06-27
bl31: Add error reporting registers

sorenb-xlnx:
This patch adds cpumerrsr_el1 and l2merrsr_el1 error reporting
registers to non_el3_sys_regs.

These registers hold the ECC errors on L1 and L2 Caches.

Signed-off-by: Naga Sureshkumar Relli <nagasure@xilinx.com>

### DIFF
--- a/include/lib/cpus/aarch64/cortex_a73.h
+++ b/include/lib/cpus/aarch64/cortex_a73.h
@@ -41,4 +41,9 @@
 
 #define CORTEX_A73_CPUECTLR_SMP_BIT	(1 << 6)
 
+/*******************************************************************************
+ * L2 Memory Error Syndrome register specific definitions.
+ ******************************************************************************/
+#define CORTEX_A73_L2MERRSR_EL1		S3_1_C15_C2_3   /* Instruction def. */
+
 #endif /* __CORTEX_A73_H__ */
--- a/include/lib/cpus/aarch64/cortex_a53.h
+++ b/include/lib/cpus/aarch64/cortex_a53.h
@@ -57,6 +57,11 @@
 #define CPUECTLR_FPU_RET_CTRL_MASK	(0x7 << CPUECTLR_FPU_RET_CTRL_SHIFT)
 
 /*******************************************************************************
+ * CPU Memory Error Syndrome register specific definitions.
+ ******************************************************************************/
+#define CPUMERRSR_EL1			S3_1_C15_C2_2	/* Instruction def. */
+
+/*******************************************************************************
  * CPU Auxiliary Control register specific definitions.
  ******************************************************************************/
 #define CPUACTLR_EL1			S3_1_C15_C2_0	/* Instruction def. */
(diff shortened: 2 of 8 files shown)


soby-mathew:
These are useful registers to report but my opinion is that these registers `cpumerrsr_el1` and `l2merrsr_el1` are CPU specific and therefore should be reported as part of CPU register dump function in cpu operations framework. For example, `cpuectlr_el1` is reported from Cortex A53 specific file [here](https://github.com/ARM-software/arm-trusted-firmware/blob/master/lib/cpus/aarch64/cortex_a53.S#L239).
All the applicable CPU specific register dumps would need to be modified to report these registers.

sorenb-xlnx:
Hi Soby,

On Tue, 2016-06-28 at 02:44:38 -0700, Soby Mathew wrote:
> These are useful registers to report but my opinion is that these registers `cpumerrsr_el1` and `l2merrsr_el1` are CPU specific and therefore should be reported as part of CPU register dump function in cpu operations framework. For example, `cpuectlr_el1` is reported from Cortex A53 specific file [here](https://github.com/ARM-software/arm-trusted-firmware/blob/master/lib/cpus/aarch64/cortex_a53.S#L239).
> All the applicable CPU specific register dumps would need to be modified to report these registers.

IIUC, the register contents are CPU-specific, but the registers themselves
are architecturally defined (correct me if I'm wrong). As we don't
interpret bitfields and just dump the full register content wouldn't it be
OK to just dump in from the common location?

	Thanks,
	Sören

achingupta:
These are not architecturally defined registers. They are CPU specific and might be common across a subset of CPUs. You will not find them in the ARMv8-A ARM. They should be reported as CPU specific registers as Soby suggested.

sorenb-xlnx:
On Thu, 2016-06-30 at 03:18:05 -0700, Achin Gupta wrote:
> These are not architecturally defined registers. They are CPU specific and might be common across a subset of CPUs. You will not find them in the ARMv8-A ARM. They should be reported as CPU specific registers as Soby suggested.

You're right. I mixed that up with some other register. We'll change
this.

	Thanks,
	Sören

sorenb-xlnx:
I updated the PR with an updated version of the change.

include/lib/cpus/aarch64/cortex_a73.h
```diff
@@ -41,4 +41,9 @@
 
 #define CORTEX_A73_CPUECTLR_SMP_BIT	(1 << 6)
 
+/*******************************************************************************
+ * L2 Memory Error Syndrome register specific definitions.
+ ******************************************************************************/
+#define CORTEX_A73_L2MERRSR_EL1		S3_1_C15_C2_3   /* Instruction def. */
+
```
danh-arm: So is CPUMERRSR_EL1 not defined for A73 (I haven't looked in the TRM)?

danh-arm:
Changes look OK to me. A couple of minor comments:

*   It would be good if the commit message listed which CPU libraries have been updated.
*   Does the A35 CPU library need updating in a similar way (I haven't checked the TRM)?

danh-arm:
jenkins: test this please

include/lib/cpus/aarch64/cortex_a73.h
sorenb-xlnx: No, it's not.

sorenb-xlnx:
I can update the commit message, no problem. I couldn't find any TRM for the A35, hence we didn't update that one.

sorenb-xlnx:
I pushed an updated commit message.

danh-arm:
jenkins: retest this please

danh-arm:
It seems the A35 TRM is not on the ARM infocenter yet so let's skip that one for now.

PR #662, opened 2016-07-08
Map read-only data as execute-never

sandrine-bailleux-arm:
This patch stack aims at providing a new build flag to isolate the code
and read-only data on separate memory pages so that the former can
be mapped as executable and the latter as execute-never.

This build flag is disabled by default but enabled on ARM platforms.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -108,6 +108,10 @@ PL011_GENERIC_UART		:= 0
 ENABLE_PMF			:= 0
 # Flag to enable PSCI STATs functionality
 ENABLE_PSCI_STAT	:= 0
+# Whether code and read-only data should be put on separate memory pages.
+# The platform Makefile is free to override this value.
+SEPARATE_CODE_AND_RODATA	:= 0
+
 
 ################################################################################
 # Checkpatch script options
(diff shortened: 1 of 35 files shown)


Makefile
```diff
@@ -108,6 +108,10 @@ PL011_GENERIC_UART		:= 0
 ENABLE_PMF			:= 0
 # Flag to enable PSCI STATs functionality
 ENABLE_PSCI_STAT	:= 0
+# Whether code and read-only data should be put on separate memory pages.
+# The platform Makefile is free to override this value.
+SEPARATE_CODE_AND_RODATA	:= 0
+
```
vwadekar: What's the argument against making this behavior as the default? (i.e. SEPARATE_CODE_AND_RODATA = 1 by default) I understand that this change fixes a security hole, so all platforms will benefit if we make this as the default option.

sorenb-xlnx:
As ZynqMP is affected, I gave this a spin and things seem to work fine.
Acked-by: Sören Brinkmann <soren.brinkmann@xilinx.com>

I also spent some time to migrate ZynqMP to use the code-rodata-split https://github.com/Xilinx/arm-trusted-firmware/tree/zynqmp/rodata-xn.

In case this setting becomes default, as @vwadekar suggested, feel free to integrate those changes into this PR, if it helps.

Makefile
sandrine-bailleux-arm: SEPARATE_CODE_AND_RODATA is a trade-off between security and memory consumption.
Setting this build flag to 1 by default would increase the memory footprint of all BL images by default
and on its own it doesn't bring any additional security if the platform port doesn't make use of
the finer-grained view of the image layout and map the code as executable and the read-only
data as execute-never.

sandrine-bailleux-arm:
@sorenb-xlnx Thanks for testing the patch on the ZynqMP platform, I'm glad to know
this PR didn't break anything for you. I will take a look at the patches migrating your
platform to to use the code-rodata split.

PR #664, opened 2016-07-13
include/common/asm_macros: update to build with GAS 2.26

martinlroth:
GNU Binutils v2.26 no longer builds arm-trusted-firmware.  It fails
with the error "Error: non-constant expression in ".if" statement"

The issue is described here:
https://www.mail-archive.com/linaro-toolchain%40lists.linaro.org/msg05689.html

Because the 2nd value in the align directive is the padding byte, by
changing '.align 11, 0' and '.align 7,0' to just 'align 11' and
'align 7', all that gets changed is the padding. Using GAS 2.26
and looking in the manuals back to GAS 2.15, the padding changes
from 0x00000000 to the value for nop (0xd403201f), which seems
appropriate anyway.

https://sourceware.org/binutils/docs-2.26/as/Align.html#Align

This issue is affecting the coreboot toolchain, which has been updated
to binutils 2.26, and cannot be reverted to 2.25 as other platforms
require the updated version.

Signed-off-by: Martin Roth <martinroth@chromium.org>

### DIFF
--- a/include/common/asm_macros.S
+++ b/include/common/asm_macros.S
@@ -75,7 +75,7 @@
 	 */
 	.macro vector_base  label
 	.section .vectors, "ax"
-	.align 11, 0
+	.align 11
 	\label:
 	.endm
 
@@ -88,7 +88,7 @@
 	 */
 	.macro vector_entry  label
 	.section .vectors, "ax"
-	.align 7, 0
+	.align 7
 	\label:
 	.endm
 


danh-arm:
This PR would effectively revert the change in https://github.com/ARM-software/arm-trusted-firmware/commit/79627dc37259781e578c47e1e63856dd0424b2a2. This was done deliberately as an extra level of protection against accidental execution of padding bytes in the exception vectors. I'd prefer not revert this simply to workaround a toolchain bug.

Is it not possible for Coreboot to carry this temporary workaround in a TF branch until a fixed toolchain version is available? I understand it's preferable for Coreboot to use the latest code but it seems to me these kind of problems will regularly occur if Coreboot directly uses the upstream master branches of dependent projects.

pgeorgi:
Is it acceptable to add a "nop" after the align 11,0 and align 7,0 (and before the symbol), respectively?

That would solve the issue (because the symbol is separated from the alignment, and thus makes the symbol clearly part of the "code" area, satisfying the assembler).

Downside: one more instruction in the exception handlers, but they look small enough for that to work out.

PR #665, opened 2016-07-14
Fix type usage and coding guideline issues

soby-mathew:
This patch stack reworks use of types within Trusted Firmware and fixes some coding
guideline warnings.


soby-mathew:
super-ceded by #667

PR #666, opened 2016-07-15
zynqmp: Map read-only data as execute-never

sorenb-xlnx:
Apply the added security of mapping read-only data as execute-never that was introduced in #662 to the ZynqMP port.


plat/xilinx/zynqmp/aarch64/zynqmp_common.c
```diff
@@ -40,10 +40,19 @@
...
+	},
+	{
+		DEVICE1_BASE, DEVICE1_BASE, DEVICE1_SIZE,
+		MT_DEVICE | MT_RW | MT_SECURE | MT_EXECUTE_NEVER,
+	},
+	{
+		CRF_APB_BASE, CRF_APB_BASE, CRF_APB_SIZE,
+		MT_DEVICE | MT_RW | MT_SECURE | MT_EXECUTE_NEVER,
+	},
+	{ 0 }
```
danh-arm: This change has no functional effect; device memory is always XN. Are you just trying to emphasise this? Since there is no commit message, it might make the reader think this improves security when it doesn't.

danh-arm:
jenkins:test this please

plat/xilinx/zynqmp/aarch64/zynqmp_common.c
sorenb-xlnx: I wasn't really aware of that. I just looked through the attributes that are implied in MT_DEVICE and didn't see anything regarding the XN. I'll drop this patch then.

sorenb-xlnx:
I updated the PR to just include the first patch of the original series. I dropped the second patch as it didn't add any value as @danh-arm pointed out.

danh-arm:
> I dropped the second patch as it didn't add any value ...

OK. Previous CI loop succeeded so merging...

PR #667, opened 2016-07-19
Introduce PSCI library

soby-mathew:
This patch set introduces PSCI library component. Alongside, the following changes are also done:
- A new library component `el3_runtime` is introduced.
- The type usage is reworked for more portability.
- The assembler and common platform helpers are rearranged so that the architecture specific
   functionality is separated from common functionality.


plat/xilinx/zynqmp/platform.mk
```diff
@@ -80,7 +80,7 @@ PLAT_BL_COMMON_SOURCES	:=	lib/xlat_tables/xlat_tables_common.c		\
 BL31_SOURCES		+=	drivers/arm/cci/cci.c				\
 				lib/cpus/aarch64/aem_generic.S			\
 				lib/cpus/aarch64/cortex_a53.S			\
-				plat/common/aarch64/plat_psci_common.c		\
+				plat/common/plat_psci_common.c			\
```
sorenb-xlnx: Nit: There seems to be some white space error.

soby-mathew: Thanks for pointing it out. I will rework it and issue new PR.

soby-mathew: It seems alright after all. Its just the extra tab that I had to add to keep the alignment :).

PR #668, opened 2016-07-19
Documentation for SEPARATE_CODE_AND_RODATA build flag

sandrine-bailleux-arm:
This patch documents the effect, cost and benefits of the
SEPARATE_CODE_AND_RODATA build flag.

### DIFF
--- a/docs/firmware-design.md
+++ b/docs/firmware-design.md
@@ -14,8 +14,9 @@ Contents :
 9.  [Memory layout of BL images](#9-memory-layout-of-bl-images)
 10. [Firmware Image Package (FIP)](#10--firmware-image-package-fip)
 11. [Use of coherent memory in Trusted Firmware](#11--use-of-coherent-memory-in-trusted-firmware)
-12. [Code Structure](#12--code-structure)
-13. [References](#13--references)
+12. [Isolating code and read-only data on separate memory pages](#12--isolating-code-and-read-only-data-on-separate-memory-pages)
+13. [Code Structure](#13--code-structure)
+14. [References](#14--references)
 
 
 1.  Introduction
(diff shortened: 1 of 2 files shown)


docs/firmware-design.md
```diff
@@ -1768,7 +1769,87 @@ whether coherent memory should be used. If a platform disables
...
+       | Exception vectors |
+       +-------------------+ 2 KB boundary
+       |     <Padding>     |
+       +-------------------+
+       |       Code        |
+       +-------------------+ BLx_BASE
+
+Even with this more condensed memory layout, the memory footprint is still
+bigger. When using a 4KB translation granule size, it adds up to (4KB - 1)
+padding bytes for each BL image and requires at least an extra page table.
```
athoelke: With this layout, the separation of read-only data will add zero or one page to the memory footprint compared to the original combined layout. But not more or less than that. There should also not be a need for an extra page table (or more).

sandrine-bailleux-arm: > With this layout, the separation of read-only data will add zero or one page to the memory footprint compared to the original combined layout. But not more or less than that.

I agree, that's what I meant by "up to (4KB - 1) padding bytes". That's zero or one page
(when using a 4KB granule size), isn't it? Sorry I might have misunderstood your point.

> There should also not be a need for an extra page table (or more).

You're right, now I realise I got confused. We just need an extra level3
translation table entry to map the extra memory page introduced to pad the
read-only data, if any. So that's zero or one extra translation table entry.

But still, potentially this could mean another page table, if all 512 entries
of the page table previously mapping the RO section were used. Do you agree?

The documentation definitely needs fixing in any case. Thanks for pointing it out.

PR #671, opened 2016-07-26
ARM platforms: Define common image sizes

antonio-nino-diaz-arm:
Compile option `ARM_BOARD_OPTIMISE_MMAP` has been renamed to
`ARM_BOARD_OPTIMISE_MEM` because it now applies not only to defines
related to the translation tables but to the image size as well.

The defines `PLAT_ARM_MAX_BL1_RW_SIZE`, `PLAT_ARM_MAX_BL2_SIZE` and
`PLAT_ARM_MAX_BL31_SIZE` have been moved to the file board_arm_def.h.
This way, ARM platforms no longer have to set their own values if
`ARM_BOARD_OPTIMISE_MEM=0` and they can specify optimized values
otherwise. The common sizes have been set to the highest values used
for any of the current build configurations.

This is needed because in some build configurations some images are
running out of space. This way there is a common set of values known
to work for all of them and it can be optimized for each particular
platform if needed.

The space reserved for BL2 when `TRUSTED_BOARD_BOOT=0` has been
increased. This is needed because when memory optimisations are
disabled the values for Juno of `PLAT_ARM_MMAP_ENTRIES` and
`MAX_XLAT_TABLES` are higher. If in this situation the code is
compiled in debug mode and with "-O0", the code won't fit.

Change-Id: I70a3d8d3a0b0cad1d6b602c01a7ea334776e718e

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -465,12 +465,13 @@ map is explained in the [Firmware Design].
     match the frame used by the Non-Secure image (normally the Linux kernel).
     Default is true (access to the frame is allowed).
 
-*   `ARM_BOARD_OPTIMISE_MMAP`: Boolean option to enable or disable optimisation
-    of page table and MMU related macros `PLAT_ARM_MMAP_ENTRIES` and
-    `MAX_XLAT_TABLES`. By default this flag is 0, which means it uses the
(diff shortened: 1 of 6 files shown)


danh-arm:
jenkins:retest this please

PR #674, opened 2016-07-28
rockchip: fixes typo and some bugs for suspend/resume tests

rkchrome:
1. Remove the AP_PWROFF in ATF, should configure it in kernel.
2. Save and restore the PWMs pin/regs for suspend/resume.
3. The pmusgrf reset-hold bits needs to be released. since the
   pmusgrf reset-hold bits needs to be held.
4. Configure the PMU power up/down cycles about delay 3ms.
5. With the MMIO register block as one big mapping.
6. Fix the build error with psci_entrypoint since PSCI lib updated.

Fixes the commit
9ec78bd ("rockchip: support the suspend/resume for rk3399").

Change-Id: I112806700bf433c87763aac23d22fa7e6a7f5264


danh-arm:
jenkins: test this please

PR #675, opened 2016-08-04
emmc: support CMD23

hzhuang1:
Support CMD23. When CMD23 is used, CMD12 could be avoided.

Two scenarios:
1. CMD17 for single block, CMD18 + CMD12 for multiple blocks.
2. CMD23 + CMD18 for both single block and multiple blocks.

The emmc_init() should initialize whether CMD23 is supported
or not.

### DIFF
--- a/include/drivers/emmc.h
+++ b/include/drivers/emmc.h
@@ -49,6 +49,7 @@
 #define EMMC_CMD13			13
 #define EMMC_CMD17			17
 #define EMMC_CMD18			18
+#define EMMC_CMD23			23
 #define EMMC_CMD24			24
 #define EMMC_CMD25			25
 #define EMMC_CMD35			35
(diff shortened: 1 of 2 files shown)


hzhuang1:
It's used to fix the issue in https://github.com/ARM-software/tf-issues/issues/412

include/drivers/emmc.h
```diff
@@ -177,6 +178,6 @@ size_t emmc_erase_blocks(int lba, size_t size);
 size_t emmc_rpmb_read_blocks(int lba, uintptr_t buf, size_t size);
 size_t emmc_rpmb_write_blocks(int lba, const uintptr_t buf, size_t size);
 size_t emmc_rpmb_erase_blocks(int lba, size_t size);
-void emmc_init(const emmc_ops_t *ops, int clk, int bus_width);
+void emmc_init(const emmc_ops_t *ops, int clk, int bus_width, int cmd23);
```
danh-arm: Your interface would be more future-proof if you made the extra argument a "flags" argument instead, with bit 0 indicating whether cmd23 was supported or not. You would also need a macro (or enum) to define the flag bit value. Similarly, the global var `emmc_cmd23` in the c file would be better as `emmc_flags`.

danh-arm:
minor: If your PR fixes an issue, then it is good to put a line like this in the commit message:
"Fixes ARM-software/tf-issues#412". Then the issue will be automatically closed when the PR is merged.

Also, it is only necessary to create an issue if there is going to be significant time between creating the issue and raising the PR. There is not much value in creating an issue if a fix for it appears at the same time.

include/drivers/emmc.h
hzhuang1: OK. I'll use emmc_flags instead.

PR #676, opened 2016-08-04
io: block: fix unaligned buffer

hzhuang1:
If buffer address parameter isn't aligned, it may cause
DMA issue in block device driver, as eMMC. Now check the buffer
address. If it's not aligned, use temporary buffer in io block
driver instead.


hzhuang1:
It's used to fix the issue in https://github.com/ARM-software/tf-issues/issues/413

danh-arm:
jenkins: test this please

PR #677, opened 2016-08-04
partition: check GPT partition table

hzhuang1:
Now only support GPT partition table. MBR partition table isn't
supported yet.

### DIFF
--- a/drivers/partition/gpt.c
+++ b/drivers/partition/gpt.c
@@ -1,0 +1,78 @@
+/*
+ * Copyright (c) 2016, ARM Limited and Contributors. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * Neither the name of ARM nor the names of its contributors may be used
+ * to endorse or promote products derived from this software without specific
+ * prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <assert.h>
+#include <debug.h>
+#include <errno.h>
+#include <gpt.h>
+#include <string.h>
+
+static int unicode_to_ascii(unsigned short *str_in, unsigned char *str_out)
+{
+	uint8_t *name = (uint8_t *)str_in;
+	int i;
+
+	assert((str_in != NULL) && (str_out != NULL) && (name[0] != '\0'));
+
+	/* check whether the unicode string is valid */
+	for (i = 1; i < (EFI_NAMELEN << 1); i += 2) {
+		if (name[i] != '\0')
+			return -EINVAL;
+	}
+	/* convert the unicode string to ascii string */
+	for (i = 0; i < (EFI_NAMELEN << 1); i += 2) {
+		str_out[i >> 1] = name[i];
+		if (name[i] == '\0')
+			break;
+	}
+	return 0;
+}
+
+int parse_gpt_entry(gpt_entry_t *gpt_entry, partition_entry_t *entry)
+{
+	int result;
+
+	assert((gpt_entry != 0) && (entry != 0));
+
+	if ((gpt_entry->first_lba == 0) && (gpt_entry->last_lba == 0)) {
+		return -EINVAL;
+	}
+
+	memset(entry, 0, sizeof(partition_entry_t));
+	result = unicode_to_ascii(gpt_entry->name, (uint8_t *)entry->name);
+	if (result != 0) {
+		return result;
+	}
+	entry->start = (uint64_t)gpt_entry->first_lba * PARTITION_BLOCK_SIZE;
+	entry->length = (uint64_t)(gpt_entry->last_lba -
+				   gpt_entry->first_lba + 1) *
+			PARTITION_BLOCK_SIZE;
+	return 0;
+}
--- a/drivers/partition/partition.c
+++ b/drivers/partition/partition.c
@@ -1,0 +1,229 @@
+/*
+ * Copyright (c) 2016, ARM Limited and Contributors. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * Neither the name of ARM nor the names of its contributors may be used
+ * to endorse or promote products derived from this software without specific
+ * prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <assert.h>
+#include <debug.h>
+#include <io_storage.h>
+#include <gpt.h>
+#include <mbr.h>
+#include <partition.h>
+#include <platform.h>
+#include <string.h>
+
+static uint8_t mbr_sector[PARTITION_BLOCK_SIZE];
+partition_entry_list_t list;
+
+#if LOG_LEVEL >= LOG_LEVEL_VERBOSE
+static void dump_entries(int num)
+{
+	char name[EFI_NAMELEN];
+	int i, j, len;
+
+	VERBOSE("Partition table with %d entries:\n", num);
+	for (i = 0; i < num; i++) {
+		len = snprintf(name, EFI_NAMELEN, "%s", list.list[i].name);
+		for (j = 0; j < EFI_NAMELEN - len - 1; j++) {
+			name[len + j] = ' ';
+		}
+		name[EFI_NAMELEN - 1] = '\0';
+		VERBOSE("%d: %s %lx-%lx\n", i + 1, name, list.list[i].start,
+			list.list[i].start + list.list[i].length - 4);
+	}
+}
+#else
+#define dump_entries(num)	((void)num)
+#endif
+
+/*
+ * Load the first sector that carries MBR header.
+ * The MBR boot signature should be always valid whether it's MBR or GPT.
+ */
+static int load_mbr_header(uintptr_t image_handle, mbr_entry_t *mbr_entry)
+{
+	size_t bytes_read;
+	uintptr_t offset;
+	int result;
+
+	assert(mbr_entry != NULL);
+	/* MBR partition table is in LBA0. */
+	result = io_seek(image_handle, IO_SEEK_SET, MBR_OFFSET);
+	if (result != 0) {
+		WARN("Failed to seek (%i)\n", result);
+		return result;
+	}
+	result = io_read(image_handle, (uintptr_t)&mbr_sector,
+			 PARTITION_BLOCK_SIZE, &bytes_read);
+	if (result != 0) {
+		WARN("Failed to read data (%i)\n", result);
+		return result;
+	}
+
+	/* Check MBR boot signature. */
+	if ((mbr_sector[PARTITION_BLOCK_SIZE - 2] != MBR_SIGNATURE_FIRST) ||
+	    (mbr_sector[PARTITION_BLOCK_SIZE - 1] != MBR_SIGNATURE_SECOND)) {
+		return -ENOENT;
+	}
+	offset = (uintptr_t)&mbr_sector + MBR_PRIMARY_ENTRY_OFFSET;
+	memcpy(mbr_entry, (void *)offset, sizeof(mbr_entry_t));
+	return 0;
+}
+
+/*
+ * Load GPT header and check the GPT signature.
+ * If partiton numbers could be found, check & update it.
+ */
+static int load_gpt_header(uintptr_t image_handle)
+{
+	gpt_header_t header;
+	size_t bytes_read;
+	int result;
+
+	result = io_seek(image_handle, IO_SEEK_SET, GPT_HEADER_OFFSET);
+	if (result != 0) {
+		return result;
+	}
+	result = io_read(image_handle, (uintptr_t)&header,
+			 sizeof(gpt_header_t), &bytes_read);
+	if ((result != 0) || (sizeof(gpt_header_t) != bytes_read)) {
+		return result;
+	}
+	if (memcmp(header.signature, GPT_SIGNATURE,
+		   sizeof(header.signature)) != 0) {
+		return -EINVAL;
+	}
+
+	/* partition numbers can't exceed PLAT_PARTITION_MAX_ENTRIES */
+	list.entry_count = header.list_num;
+	if (list.entry_count > PLAT_PARTITION_MAX_ENTRIES) {
+		list.entry_count = PLAT_PARTITION_MAX_ENTRIES;
+	}
+	return 0;
+}
+
+static int load_gpt_entry(uintptr_t image_handle, gpt_entry_t *entry)
+{
+	size_t bytes_read;
+	int result;
+
+	assert(entry != NULL);
+	result = io_read(image_handle, (uintptr_t)entry, sizeof(gpt_entry_t),
+			 &bytes_read);
+	if (sizeof(gpt_entry_t) != bytes_read)
+		return -EINVAL;
+	return result;
+}
+
+static int verify_partition_gpt(uintptr_t image_handle)
+{
+	gpt_entry_t entry;
+	int result, i;
+
+	for (i = 0; i < list.entry_count; i++) {
+		result = load_gpt_entry(image_handle, &entry);
+		assert(result == 0);
+		result = parse_gpt_entry(&entry, &list.list[i]);
+		if (result != 0) {
+			break;
+		}
+	}
+	if (i == 0) {
+		return -EINVAL;
+	}
+	/*
+	 * Only records the valid partition number that is loaded from
+	 * partition table.
+	 */
+	list.entry_count = i;
+	dump_entries(list.entry_count);
+
+	return 0;
+}
+
+int load_partition_table(unsigned int image_id)
+{
+	uintptr_t dev_handle, image_handle, image_spec = 0;
+	mbr_entry_t mbr_entry;
+	int result;
+
+	result = plat_get_image_source(image_id, &dev_handle, &image_spec);
+	if (result != 0) {
+		WARN("Failed to obtain reference to image id=%u (%i)\n",
+			image_id, result);
+		return result;
+	}
+
+	result = io_open(dev_handle, image_spec, &image_handle);
+	if (result != 0) {
+		WARN("Failed to access image id=%u (%i)\n", image_id, result);
+		return result;
+	}
+
+	result = load_mbr_header(image_handle, &mbr_entry);
+	if (result != 0) {
+		WARN("Failed to access image id=%u (%i)\n", image_id, result);
+		return result;
+	}
+	if (mbr_entry.type == PARTITION_TYPE_GPT) {
+		result = load_gpt_header(image_handle);
+		assert(result == 0);
+		result = io_seek(image_handle, IO_SEEK_SET, GPT_ENTRY_OFFSET);
+		assert(result == 0);
+		result = verify_partition_gpt(image_handle);
+	} else {
+		/* MBR type isn't supported yet. */
+		result = -EINVAL;
+		goto exit;
+	}
+exit:
+	io_close(image_handle);
+	return result;
+}
+
+const partition_entry_t *get_partition_entry(const char *name)
+{
+	int i;
+
+	for (i = 0; i < list.entry_count; i++) {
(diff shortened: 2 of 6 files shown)


hzhuang1:
It's used to fix https://github.com/ARM-software/tf-issues/issues/414

include/drivers/partition.h
```diff
@@ -0,0 +1,113 @@
...
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
```
danh-arm: I think this functionality needs splitting up and moving:
  * Pure GPT definitions/declararations should move to drivers/io/gpt.h
  * Pure MBR definitions/declarations should move to drivers/io/mbr.h
  * The remaining functionality appears to be closely tied to the io framework so should be in drivers/io/io_partition.h

include/drivers/partition.h
```diff
@@ -0,0 +1,113 @@
...
+#define GPT_ENTRY_OFFSET		(GPT_HEADER_OFFSET +		\
+					 PARTITION_BLOCK_SIZE)
+
+typedef struct partition_entry {
+	uint64_t		start;
+	uint64_t		length;
+	unsigned int		flags;
+	unsigned int		load_addr;
+	unsigned int		load_size;
+	int			id;
```
danh-arm: flags, load_addr, load_size and id don't appear to be used so why are they there?

include/drivers/partition.h
```diff
@@ -0,0 +1,113 @@
...
+	char			name[EFI_NAMELEN];
+} part_entry_t;
```
danh-arm: minor: prefer struct typedefs to match the struct name, i.e. partition_entry_t.

include/drivers/partition.h
```diff
@@ -0,0 +1,113 @@
...
+
+typedef struct partition_ops {
+	int (*handler)(const char *name, part_entry_t *entry);
+} partition_ops_t;
+
+typedef struct partition_mbr_entry {
```
danh-arm: minor: prefer struct typedefs to match the struct name, i.e. mbr_entry.

include/drivers/partition.h
```diff
@@ -0,0 +1,113 @@
...
+	uint8_t			status;
+	uint8_t			first_sector[MBR_CHS_ADDRESS_LEN];
+	uint8_t			type;
+	uint8_t			last_sector[MBR_CHS_ADDRESS_LEN];
+	uint32_t		first_lba;
+	uint32_t		sector_nums;
+} mbr_entry_t;
+
+typedef struct partition_gpt_entry {
```
danh-arm: minor: prefer struct typedefs to match the struct name, i.e. gpt_entry.

drivers/partition/partition.c
```diff
@@ -0,0 +1,255 @@
...
+	assert(entry != NULL);
+	for (i = 0; i < part_num_entries; i++) {
+		if (strcmp(name, entry_list[i].name) == 0) {
+			memcpy(entry, &entry_list[i], sizeof(part_entry_t));
+			return 0;
+		}
+	}
+	/* Check for pseudo entries out of partition table. */
+	if (part_ops.handler != NULL)
+		return part_ops.handler(name, entry);
```
danh-arm: What's a pseudo entry? Why does this function have such a vague name and why does it need to be supplied during init? If the same caller supplies this function during init then why can't they explicitly invoke this when get_partition_entry() returns error? Even if it is needed, wouldn't `get_pseudo_entry_from_table()` be a better name?

drivers/partition/partition.c
```diff
@@ -0,0 +1,255 @@
...
+	return -EINVAL;
+}
+
+void get_partition_entry_list(part_entry_t **list, int *num_entries)
```
danh-arm: Returning a const entry list would be better.

drivers/partition/partition.c
```diff
@@ -0,0 +1,255 @@
...
+}
+
+int get_partition_entry(const char *name, part_entry_t *entry)
+{
+	int i;
+
+	assert(entry != NULL);
+	for (i = 0; i < part_num_entries; i++) {
+		if (strcmp(name, entry_list[i].name) == 0) {
+			memcpy(entry, &entry_list[i], sizeof(part_entry_t));
```
danh-arm: Why bother copying instead of returning a const entry into the array?

include/drivers/partition.h
```diff
@@ -0,0 +1,113 @@
...
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef __PARTITION_H__
+#define __PARTITION_H__
+
+#include <types.h>
+
+#define MAX_PARTITION_NUM		128
```
danh-arm: Seems quite a large static define. Perhaps should it be an optional platform define with a default value?

drivers/partition/partition.c
```diff
@@ -0,0 +1,255 @@
...
+	result = io_read(image_handle, (uintptr_t)&mbr_sector,
+			 PARTITION_BLOCK_SIZE, &bytes_read);
+	if (result != 0) {
+		WARN("Failed to read data (%i)\n", result);
+		return result;
+	}
+
+	/* Check MBR boot signature. */
+	if ((mbr_sector[PARTITION_BLOCK_SIZE - 2] != 0x55) ||
+	    (mbr_sector[PARTITION_BLOCK_SIZE - 1] != 0xAA)) {
```
danh-arm: magic numbers

include/drivers/partition.h
```diff
@@ -0,0 +1,113 @@
...
+#define MBR_PRIMARY_ENTRY_NUMBER	4
+#define MBR_CHS_ADDRESS_LEN		3
+
+#define GPT_SIGNATURE			"EFI PART"
+#define GPT_ENTRY_OFFSET		(GPT_HEADER_OFFSET +		\
+					 PARTITION_BLOCK_SIZE)
+
+typedef struct partition_entry {
+	uint64_t		start;
+	uint64_t		length;
```
danh-arm: minor: exact width types not required here - `unsigned long long` preferred.

include/drivers/partition.h
```diff
@@ -0,0 +1,113 @@
...
+#define MBR_PRIMARY_ENTRY_OFFSET	0x1be
+#define MBR_PRIMARY_ENTRY_SIZE		0x10
+#define MBR_PRIMARY_ENTRY_NUMBER	4
+#define MBR_CHS_ADDRESS_LEN		3
+
+#define GPT_SIGNATURE			"EFI PART"
+#define GPT_ENTRY_OFFSET		(GPT_HEADER_OFFSET +		\
+					 PARTITION_BLOCK_SIZE)
+
+typedef struct partition_entry {
```
danh-arm: I'm strugglind to understand what this struct adds over and above the gpt_entry_t below.

drivers/partition/partition.c
```diff
@@ -0,0 +1,255 @@
...
+		return -EINVAL;
+	}
+	/* Check the first partition entry. */
+	entry = (mbr_entry_t *)((uintptr_t)&mbr_sector +
+				MBR_PRIMARY_ENTRY_OFFSET);
+	if (entry->type == PARTITION_TYPE_GPT)
```
danh-arm: Needs a comment to say that we only support the first entry being a GPT entry.

drivers/partition/partition.c
```diff
@@ -0,0 +1,255 @@
...
+		return -EAGAIN;
+	return -EINVAL;
+}
+
+static void dump_entries(int num)
```
danh-arm: This whole function body should be wrapped in a `#if LOG_LEVEL >= LOG_LEVEL_INFO` block.

drivers/partition/partition.c
```diff
@@ -0,0 +1,255 @@
...
+	result = verify_partition_mbr(image_handle);
+	if (result != 0) {
+		INFO("Can't find MBR partition table (%i)\n", result);
+		if (result != -EAGAIN)
+			goto exit;
+	} else {
+		/* Find MBR partition table. */
+		goto exit;
+	}
+
```
danh-arm: Need a comment to say that MBR is parsed OK and now need to verify that first entry is a GPT entry.

include/drivers/partition.h
hzhuang1: Do you mean that I also need to move partition.c from drivers/partition into drivers/io instead? Otherwise, I can't include the head files in drivers/io directory.

include/drivers/partition.h
hzhuang1: OK. I'll remove them.

drivers/partition/partition.c
hzhuang1: Those pseudo entries could confuse people. Let me move them into hikey implementation instead. I need them since I also need to update partition table in eMMC user area & BL1 in eMMC RPMB. There's no entry to describe partition table in the GPT/MBR. So I need a pseudo entry instead.

include/drivers/partition.h
danh-arm: Yes, I did mean that but I've changed my mind! I don't think it's necessary to put this in the IO directory but I still think the code needs splitting up, e.g. like this:

*   Pure GPT declararations should move to include/drivers/partition/gpt.h
*   Pure MBR declarations should move to include/drivers/partition/mbr.h
*   Remaining declarations should move to include/drivers/partition/partition.h
*   Pure GPT definitions (e.g. parse_gpt_entry()) should move to drivers/partition/gpt.c
*   Remaining code (that relies on the IO framework) should remain in drivers/paritition/partition.c

hzhuang1:
any comments?

drivers/partition/gpt.c
```diff
@@ -0,0 +1,82 @@
...
+#include <string.h>
+
+static int unicode_to_ascii(unsigned short *str_in, unsigned char *str_out)
+{
+	uint8_t *name = (uint8_t *)str_in;
+	int i;
+
+	assert((name[0] != '\0') && (str_in != NULL) && (str_out != NULL));
+
+	/* Calculate the unicode string length. */
```
davidcunado-arm: This comment does not reflect what the code is doing (which is validating that name is unicode).

drivers/partition/gpt.c
```diff
@@ -0,0 +1,82 @@
...
+#include <errno.h>
+#include <partition/partition.h>
+#include <string.h>
+
+static int unicode_to_ascii(unsigned short *str_in, unsigned char *str_out)
+{
+	uint8_t *name = (uint8_t *)str_in;
+	int i;
+
+	assert((name[0] != '\0') && (str_in != NULL) && (str_out != NULL));
```
davidcunado-arm: This assert will crash if str_in is NULL - in this case, name[0] will be dereferencing a NULL pointer.

drivers/partition/gpt.c
```diff
@@ -0,0 +1,82 @@
...
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <assert.h>
+#include <debug.h>
+#include <errno.h>
+#include <partition/partition.h>
+#include <string.h>
+
```
davidcunado-arm: It's a cleaner 'design' for .c files to include their own .h file - specifically here, partition/gpt.h.
To do this you'll need to move a few things around;
- Move definition of EFI_NAMELEN to partition.h
- Include partion.h in gpt.h

drivers/partition/gpt.c
```diff
@@ -0,0 +1,82 @@
...
+}
+
+int parse_gpt_entry(gpt_entry_t *gpt_entry, int entry_index)
+{
+	int result;
+
+	assert((gpt_entry != 0) && (entry_index >= 0) &&
+	       (entry_index < PLAT_PARTITION_MAX_ENTRIES));
+
+	/* exhaused partition entry */
```
davidcunado-arm: Comment doesn't make sense.

drivers/partition/partition.c
```diff
@@ -0,0 +1,214 @@
...
+ */
+
+#include <assert.h>
+#include <debug.h>
+#include <io_storage.h>
+#include <partition/partition.h>
+#include <platform.h>
+#include <string.h>
+
+#define MAX_PARTITION_ENTRIES		128
```
davidcunado-arm: This is already defined in partition.h - I'd remove from here.

drivers/partition/partition.c
```diff
@@ -0,0 +1,214 @@
...
+	int i, j, len;
+
+	INFO("Partition table with %d entries:\n", num);
+	for (i = 0; i < num; i++) {
+		len = snprintf(name, EFI_NAMELEN, "%s", entry_list[i].name);
+		for (j = 0; j < EFI_NAMELEN - len; j++) {
+			name[len + j] = ' ';
+		}
+		INFO("%d: %s %lx-%lx\n", i + 1, name, entry_list[i].start,
+		     entry_list[i].start + entry_list[i].length - 4);
```
davidcunado-arm: I think that this code is formatting the INFO string so that everything lines-up nicely. Is there a better way to do this? Perhaps using the '-' flag for left justification - e.g. %-10s to left-justify to a field of 10 characters.

drivers/partition/partition.c
```diff
@@ -0,0 +1,214 @@
...
+static void dump_entries(int num)
+{
+	char name[EFI_NAMELEN];
+	int i, j, len;
+
+	INFO("Partition table with %d entries:\n", num);
+	for (i = 0; i < num; i++) {
+		len = snprintf(name, EFI_NAMELEN, "%s", entry_list[i].name);
+		for (j = 0; j < EFI_NAMELEN - len; j++) {
+			name[len + j] = ' ';
```
davidcunado-arm: Do you need to add a NULL terminator?

drivers/partition/partition.c
```diff
@@ -0,0 +1,214 @@
...
+#include <string.h>
+
+#define MAX_PARTITION_ENTRIES		128
+
+static int part_num_entries;
+static uint8_t mbr_sector[PARTITION_BLOCK_SIZE];
+
+partition_entry_t entry_list[PLAT_PARTITION_MAX_ENTRIES];
+
+static int verify_partition_mbr(uintptr_t image_handle)
```
davidcunado-arm: load_mbr_partition could be a more descriptive name for this function.

davidcunado-arm: Perhaps if this function were to take an output param of mbr_entry_t, then the calling method could then check the type and decide whether to error out or then load a GPT partition info. This way you don't need to return an error code EAGAIN to know to look for a GPT partition.

drivers/partition/partition.c
```diff
@@ -0,0 +1,214 @@
...
+		}
+		INFO("%d: %s %lx-%lx\n", i + 1, name, entry_list[i].start,
+		     entry_list[i].start + entry_list[i].length - 4);
+	}
+}
+#else
+#define dump_entries(int num)	((void)0)
+#endif
+
+static int verify_partition_gpt(uintptr_t image_handle)
```
davidcunado-arm: Similar comment to that for the MBR function - split into load_gpt_header and load_gpt_entry and use output param to pass back the gtp_header_t and gtp_entry_t respectively.

drivers/partition/partition.c
```diff
@@ -0,0 +1,214 @@
...
+	if (i == 0) {
+		return -EINVAL;
+	}
+	part_num_entries = i;
+	dump_entries(part_num_entries);
+
+	return 0;
+}
+
+int load_partition_table(unsigned int image_id)
```
davidcunado-arm: Given comments to MBR and GPT functions, this function may flow better.

include/drivers/partition/gpt.h
```diff
@@ -0,0 +1,74 @@
...
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef __GPT_H__
+#define __GPT_H__
+
+#define EFI_NAMELEN			36
```
davidcunado-arm: As mentioned in another comment, move this definition to partition.h (or even a partion_common.h) and include appropriate header here.

include/drivers/partition/partition.h
```diff
@@ -0,0 +1,57 @@
...
+
+typedef struct partition_entry {
+	uint64_t		start;
+	uint64_t		length;
+	char			name[EFI_NAMELEN];
+} partition_entry_t;
+
+int load_partition_table(unsigned int image_id);
+const partition_entry_t *get_partition_entry(const char *name);
+const partition_entry_t *get_partition_entry_list(int *num_entries);
```
davidcunado-arm: Perhaps an alternative could be to have a struct that contains the number of entries and the array of entries.

drivers/partition/partition.c
hzhuang1: I think that this definition should be left. Since the maxium partition entries shouldn't exceed 128. This definition could avoid developer enter any wrong value exceeded 128 for PLAT_PARTITION_MAX_ENTRIES.

hzhuang1: Oh. I can use cassert() instead.
Sure. I'll remove the macro from here.

drivers/partition/partition.c
hzhuang1: I checked the tf_print(). It couldn't support "%10s" or "%-10s". It could only support "%s".

drivers/partition/gpt.c
```diff
@@ -0,0 +1,81 @@
...
+#include <errno.h>
+#include <partition/gpt.h>
+#include <string.h>
+
+static int unicode_to_ascii(unsigned short *str_in, unsigned char *str_out)
+{
+	uint8_t *name = (uint8_t *)str_in;
+	int i;
+
+	assert((name[0] != '\0') && (str_out != NULL));
```
davidcunado-arm: Consider the assert as follows - str_in != NULL && str_out != NULL && name[0] != \0

drivers/partition/partition.c
```diff
@@ -0,0 +1,229 @@
...
+}
+#else
+#define dump_entries(int num)	((void)0)
+#endif
+
+/*
+ * Load the first sector that carries MBR header.
+ * The MBR boot signature should be always valid whether it's MBR or GPT.
+ */
+static mbr_entry_t *load_mbr_header(uintptr_t image_handle)
```
davidcunado-arm: Apologies - to clarify, I was thinking of the following signature;

static int load_mbr_header(uintptr_t image_handle, mbr_entry_t * mbr_entry)

This was you can return both an error / success code and the mbr_entry_t object.

drivers/partition/partition.c
```diff
@@ -0,0 +1,229 @@
...
+	if ((result != 0) || (sizeof(gpt_header_t) != bytes_read)) {
+		return result;
+	}
+	if (memcmp(header.signature, GPT_SIGNATURE,
+		   sizeof(header.signature)) != 0) {
+		return -EINVAL;
+	}
+
+	/* partition numbers can't exceed PLAT_PARTITION_MAX_ENTRIES */
+	part_num_entries = header.part_num;
```
davidcunado-arm: Couldn't you use list.num here?

drivers/partition/partition.c
```diff
@@ -0,0 +1,229 @@
...
+		assert(result == 0);
+		result = parse_gpt_entry(&entry, i);
+		if (result != 0) {
+			break;
+		}
+	}
+	if (i == 0) {
+		return -EINVAL;
+	}
+	part_num_entries = i;
```
davidcunado-arm: Would be beneficial to explain why the entry count is being reset here.

drivers/partition/partition.c
```diff
@@ -0,0 +1,229 @@
...
+		}
+	}
+	return NULL;
+}
+
+const partition_entry_list_t *get_partition_entry_list(void)
+{
+	list.num = part_num_entries;
+	list.list = entry_list;
+	return &list;
```
davidcunado-arm: If list were to be populated directly, then this function could just return &list.

include/drivers/partition/partition.h
```diff
@@ -0,0 +1,66 @@
...
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef __PARTITION_H__
+#define __PARTITION_H__
+
+#include <cassert.h>
+#include <partition/mbr.h>
```
davidcunado-arm: Is this include needed here? If not, remove and include partition/mbr.h from partition.c

include/drivers/partition/partition.h
```diff
@@ -0,0 +1,66 @@
...
+#define EFI_NAMELEN			36
+
+typedef struct partition_entry {
+	uint64_t		start;
+	uint64_t		length;
+	char			name[EFI_NAMELEN];
+} partition_entry_t;
+
+typedef struct partition_entry_list {
+	partition_entry_t	*list;
```
davidcunado-arm: Any reason why this isn't partition_entry_t list[PLAT_PARTITION_MAX_ENTRIES]?
Also, perhaps entry_list would be better name?

include/drivers/partition/partition.h
```diff
@@ -0,0 +1,66 @@
...
+	int			num;
```
davidcunado-arm: Perhaps entry_count would be a better name?

include/drivers/partition/partition.h
```diff
@@ -0,0 +1,66 @@
...
+} partition_entry_list_t;
+
+int load_partition_table(unsigned int image_id);
+const partition_entry_t *get_partition_entry(const char *name);
+const partition_entry_list_t *get_partition_entry_list(void);
+void partition_init(unsigned int image_id);
+
+extern partition_entry_t entry_list[PLAT_PARTITION_MAX_ENTRIES];
```
davidcunado-arm: Given that you've provided an API to get the partition_entry_list_t, I'm assuming that all (read) access to this data should be via that API - in which case, this entry_list doesn't need to be declared here (or defined in the .c file).

include/drivers/partition/partition.h
```diff
@@ -0,0 +1,66 @@
...
+
+#endif	/* __PARTITION_GPT_H__ */
```
davidcunado-arm: Minor: /* __PARTITION_H__ */

drivers/partition/gpt.c
```diff
@@ -0,0 +1,81 @@
...
+	/* convert the unicode string to ascii string */
+	for (i = 0; i < (EFI_NAMELEN << 1); i += 2) {
+		str_out[i >> 1] = name[i];
+		if (name[i] == '\0')
+			break;
+	}
+	return 0;
+}
+
+int parse_gpt_entry(gpt_entry_t *gpt_entry, int entry_index)
```
davidcunado-arm: If the API also included partition_entry_t * parameter then the entry_list object would not need to be extern'ed.

drivers/partition/partition.c
davidcunado-arm: Ok - I assumed (incorrectly) that it did - apologies.

drivers/partition/gpt.c
```diff
@@ -0,0 +1,78 @@
...
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <assert.h>
+#include <debug.h>
+#include <errno.h>
+#include <partition/gpt.h>
```
danh-arm: Don't use relative paths to these partition headers. If the platform needs to use this functionality, it should add drivers/partition to the list of system includes in its .mk file.

drivers/partition/partition.c
```diff
@@ -0,0 +1,229 @@
...
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <assert.h>
+#include <debug.h>
+#include <io_storage.h>
+#include <partition/gpt.h>
+#include <partition/mbr.h>
+#include <partition/partition.h>
```
danh-arm: Don't use relative paths to these partition headers

include/drivers/partition/gpt.h
```diff
@@ -0,0 +1,75 @@
...
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef __GPT_H__
+#define __GPT_H__
+
+#include <partition/partition.h>
```
danh-arm: Don't use relative paths to these partition headers

drivers/partition/partition.c
```diff
@@ -0,0 +1,229 @@
...
+
+#if LOG_LEVEL >= LOG_LEVEL_INFO
+static void dump_entries(int num)
+{
+	char name[EFI_NAMELEN];
+	int i, j, len;
+
+	INFO("Partition table with %d entries:\n", num);
+	for (i = 0; i < num; i++) {
+		len = snprintf(name, EFI_NAMELEN, "%s", list.list[i].name);
```
danh-arm: `snprintf` is very expensive in terms of stack usage. I would suggest this function is only compiled in when `LOG_LEVEL` is set to `LOG_LEVEL_VERSBOSE`.

danh-arm:
jenkins:test this please

PR #679, opened 2016-08-11
Support pwm for rk3399

rkchrome:
This is all related to the PWM regulators and how to deal with them at suspend / resume time.


plat/rockchip/common/pmusram/pmu_sram_cpus_on.S
```diff
@@ -70,9 +70,7 @@ sys_resume:
 psram_data:
 	.quad	PSRAM_DT_BASE
 sys_wakeup_entry:
-#if !ERROR_DEPRECATED
 	.quad	psci_entrypoint
```
danh-arm: minor: Would be better to switch from using this deprecated function to using bl31_warm_entrypoint.

danh-arm:
jenkins:test this please

PR #688, opened 2016-08-22
support fastboot with usb

hzhuang1:
Fix to https://github.com/ARM-software/tf-issues/issues/419

### DIFF
--- a/drivers/fastboot/fastboot.c
+++ b/drivers/fastboot/fastboot.c
@@ -1,0 +1,558 @@
+/*
+ * Copyright (c) 2016, ARM Limited and Contributors. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * Neither the name of ARM nor the names of its contributors may be used
+ * to endorse or promote products derived from this software without specific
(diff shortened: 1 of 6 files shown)


drivers/synopsys/usb/dw_usb.c
```diff
@@ -0,0 +1,788 @@
+/*
```
danh-arm: Is the datasheet for this device publically avaiable, and does it allow distribution of an open source driver? I have the same IP concerns with this as the designware eMMC driver.

drivers/synopsys/usb/dw_usb.c
```diff
@@ -0,0 +1,788 @@
...
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <arch_helpers.h>
+#include <assert.h>
+#include <ctype.h>
+#include <debug.h>
+#include <dw_usb.h>
+#include <fastboot/fastboot.h>
+#include <fastboot/usb.h>
```
danh-arm: Don't use relative paths in system includes

drivers/fastboot/fastboot.c
```diff
@@ -0,0 +1,558 @@
+/*
```
danh-arm: As stated in the HiKey pull request, it should not be necessary to execute Fastboot in the secure world so I'm not convinced this driver needs to exist in the TF upstream. This could be part of a richer NS_BL2U that uses existing normal world firmware functionality that integrates Fastboot (e.g. UEFI or UBoot).

drivers/synopsys/usb/dw_usb.c
hzhuang1: Not yet. Actually it's same as the designware eMMC.

danh-arm:
Closing this until there's a solid use-case for this functionality residing in the secure world.

PR #697, opened 2016-09-08
rockchip: fix the scu idle for rk3399

rkchrome:
As rk3399 reported the d8/octane scores drop 10% with cpu idle.
The root cause is thc cpu cluster enter the slow mode.
We don't need switch the clock to 24MHz if cpu cluster enter the
retention mode. In order to improve performance, it just needs for
cluster enter powering off mode.

Also, we shouldn't do anything for hlvl if the system is off.

Change-Id: I2a02962a01343abd0cba47ed63192c1cdf88b119


davidcunado-arm:
jenkins: test this please

PR #704, opened 2016-09-13
GICv3: Allow either G1S or G0 interrupts to be configured

yatharth-arm:
Currently the GICv3 driver mandates that platform populate
both G1S and G0 interrupts. However, it is possible that a
given platform is not interested in both the groups and
just needs to specify either one of them.

This patch modifies the `gicv3_rdistif_init()` & `gicv3_distif_init()`
functions to allow either G1S or G0 interrupts to be configured.

Fixes ARM-software/tf-issues#400

Change-Id: I43572b0e08ae30bed5af9334f25d35bf439b0d2b

### DIFF
--- a/drivers/arm/gic/v3/gicv3_main.c
+++ b/drivers/arm/gic/v3/gicv3_main.c
@@ -120,10 +120,12 @@ void gicv3_driver_init(const gicv3_driver_data_t *plat_driver_data)
  ******************************************************************************/
 void gicv3_distif_init(void)
 {
+	unsigned int bitmap = 0;
+
 	assert(driver_data);
 	assert(driver_data->gicd_base);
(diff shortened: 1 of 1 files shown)


mpankaj:
Hi Yatharth,

Will it be possible to configure G1 interrupt in GICv2 in EL2. In EL3 I am able to configure G1 interrupts.
It seems there is configuration lockdown due to Lockable SPIs.
Please provide any pointers.

Thanks & Regards,
Pankaj

soby-mathew:
@mpankaj,
The ARM Trusted Firmware only configures the secure interrupts at EL3 as part of its initialization. The configuration of non secure interrupts (Group 1) interrupts are left to the Normal world (N-EL2/EL1).  I don't see any reason why a non secure interrupt configuration should be affected by the Exception level.

PR #716, opened 2016-09-23
AArch32: Fix detection of virtualization support

yatharth-arm:
The Virtualization field in the ID_PFR1 register has only 2
valid values (0 or 1) but it was incorrectly checked against
unrelated value tied to the SPSR register instead.

This patch fixes the detection of virtualization support by
using the valid values in BL1 context management code.

Change-Id: If12592e343770e1da90f0f5fecf0a3376047ac29


davidcunado-arm:
jenkins: test this please

PR #737, opened 2016-10-17
Patches for 8173 and OSI

freedomtan:
Patches for MT8173:
1. remove the use of ENABLE_PLAT_COMPAT
2. add PSCI_EXTENDED_STATE_ID supported
3. remove use of some old / deprecated APIs
4. Initial support OS Initiated idle (for genpd-based SoC idling, http://www.slideshare.net/linaroorg/bkk16410-soc-idling-cpu-cluster-pm)


danh-arm:
This PR needs several things fixing:

1.   Please target the integration branch, not the master branch.
2.   Please add a signed-off-by to all patches.
3.   For the mt1873 changes, this needs the approval of the MTK sub-maintainer: https://github.com/mtk09422.
4.   The OSI changes are very superficial and highly controversial. As far as I know there is no consensus in the kernel community that this is the correct direction to go. Even if it is, this requires major changes in TF that requires much more detailed design discussion and planning. For now I suggest you drop these changes here and start a dialogue with us about this support.

Regards

Dan.

mtk09422:
Hi @danh-arm , @freedomtan will drop OSI changes in next revision.
I am OK with this patch after the revision is done.

danh-arm:
This PR has stalled. Please raise a new PR with the requested changes when ready.

PR #722, opened 2016-09-28
Drop requirement for CLA in contribution.md

danh-arm:
It is no longer necessary for contributors to send a CLA to ARM
before making contributions. Contributors must instead add a
"Signed-off-by:" line to each commit, which certifies that the
contribution is made under the Developer Certificate of Origin
(DCO).

Update contributing.md to reflect this new policy and add a copy of
the DCO to the repository.

Change-Id: I7ca98bffc3bf57e8bdd51d763c24f13e415a328b
Signed-off-by: Dan Handley <dan.handley@arm.com>

PR #739, opened 2016-10-20
rockchip: fix A72 L2CTLR_DATA_RAM_LATENCY to 5

rkchrome:
The default value of L2CTLR_DATA_RAM_LATENCY is 2, depends to
the test result on rk3399, the A72 will need lower voltage for
high frequency if it's set to be 5, and almost no effect on performance.

Change-Id: I99a6a43edcc0c58f7775c10f4b85669dc3eff66d

### DIFF
--- a/plat/rockchip/common/aarch64/plat_helpers.S
+++ b/plat/rockchip/common/aarch64/plat_helpers.S
@@ -67,7 +67,7 @@ handler_a72:
 	 * Set the L2 Data RAM latency for Cortex-A72.
 	 * Set the L2 Tag RAM latency to for Cortex-A72.
 	 */
-	mov x0, #((2 << L2CTLR_DATA_RAM_LATENCY_SHIFT) |	\
+	mov x0, #((5 << L2CTLR_DATA_RAM_LATENCY_SHIFT) |	\
 			 (0x1 << 5))
 	msr	L2CTLR_EL1, x0
 	isb


Caesar-github:
David,  the signed-off thing is confusing me. Can we need add the  "signed-off-by" for commit? 
AFAIK, that's seem nobody add the signed-off for A-TF patches.

Due to  the checkpatch will report the warning, as following:
:6: WARNING: Do not use whitespace before Signed-off-by:
------------

That's weird for me.

davidcunado-arm:
Hi Caesar,
Re: signed-off - please take a look at [PR#722](https://github.com/ARM-software/arm-trusted-firmware/pull/722) as it details the changes.
This is a new change since v1.3 - patches since then have included a 'signed-off' field.
Not seen that warning that you mention - odd.

sandrine-bailleux-arm:
Hi Caesar, David,

> Not seen that warning that you mention - odd.

I have. At first sight, I think it's down to the way we use the checkpatch.pl script from the TF Makefile.
Basically we feed the output of the 'git log' command into the script's input stream. The problem here is,
'git log' indents the commit message with spaces (to make it nicer to read) and checkpatch.pl doesn't
like it, as flagged by the warning message.

So it looks like a problem in TF build system to me. I guess we're not using the checkpatch.pl script
in the intended way and we need to fix that.

It's only a warning message from checkpatch.pl's perspective anyway and our politics is to ignore 
those. Our automated tests won't gate a patch for that. In other words, you can safely ignore the
warning message, Caesar :)

Thanks for reporting the issue, though! I'll create a separate Github issue to fix our Makefile.

sandrine-bailleux-arm:
FYI, I've raised https://github.com/ARM-software/tf-issues/issues/432

rkchrome:
Got it, Thanks!

Updated the commit.

davidcunado-arm:
jenkins: test this please

PR #743, opened 2016-10-25
fixes the build warning from __deprecated

rkchrome:
It should not be used because there is (or there will be) a better
alternative in that should be used instead.
Then, warn about uses of the below functions marked as deprecated by
using the deprecated attribute, maybe we can return zero value to avoid
the warning.

bl31/bl31_context_mgmt.c: In function 'cm_get_context_by_mpidr':
bl31/bl31_context_mgmt.c:106:2: warning: 'platform_get_core_pos'
is deprecated (declared at include/plat/common/platform.h:284)
[-Wdeprecated-declarations]

plat/common/aarch64/plat_common.c: In function 'plat_get_syscnt_freq2':
plat/common/aarch64/plat_common.c:84:2: warning: 'plat_get_syscnt_freq'
is deprecated (declared at include/plat/common/platform.h:62)
[-Wdeprecated-declarations]

Change-Id: I7bc1ffca76d1ffdbf6858529dfe5fca2632f3f4a

### DIFF
--- a/include/plat/common/platform.h
+++ b/include/plat/common/platform.h
@@ -61,7 +61,10 @@ struct bl_params;
 /*******************************************************************************
  * Mandatory common functions
  ******************************************************************************/
-unsigned long long plat_get_syscnt_freq(void) __deprecated;
+static inline unsigned long long plat_get_syscnt_freq(void)
+{
+        return 0;
+}
 unsigned int plat_get_syscnt_freq2(void);
 
 int plat_get_image_source(unsigned int image_id,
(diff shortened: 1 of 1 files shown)


danh-arm:
What is the real concern here? The warnings you refer to can be safely ignored since the Makefile defines the compiler option `-Wno-error=deprecated-declarations`. All other warnings are interpreted as errors. If the warnings are bothering you, you can define `ERROR_DEPRECATED=1` in your platform port and they will go away (assuming you have removed all usage of deprecated instructions).

Your patch simply masks any usage of these deprecated functions, which may result in harder-to-diagnose issues.

rkchrome:
hmm, make sense. 
Abandoned, sorry for noise.

PR #744, opened 2016-10-26
fiptool: fix Segmentation fault when only --verbose option is given

masahir0y:
Fix the following bug:

  $ tools/fiptool/fiptool -v
  Segmentation fault (core dumped)


dp-arm:
Thanks for the patch.  Looks good to me.

davidcunado-arm:
jenkins: test this please

PR #656, opened 2016-06-28
Bakery locks fixes

vwadekar:
This branch fixes the issues discussed in https://github.com/ARM-software/tf-issues/issues/402

PR #761, opened 2016-11-21
Fix normal memory bakery lock implementation

soby-mathew:
This patch fixes an issue in the normal memory bakery lock
implementation. During assertion of lock status, there is a possibility
that the assertion could fail. This is because the previous update done
to the lock status by the owning CPU when not participating in cache
coherency could result in stale data in the cache due to cache maintenance
operations not propagating to all the caches. This patch fixes this issue
by doing an extra read cache maintenance operation prior to the assertion.

Fixes ARM-software/tf-issues#402

Change-Id: I0f38a7c52476a4f58e17ebe0141d256d198be88d

### DIFF
--- a/lib/locks/bakery/bakery_lock_normal.c
+++ b/lib/locks/bakery/bakery_lock_normal.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015, ARM Limited and Contributors. All rights reserved.
+ * Copyright (c) 2015-2016, ARM Limited and Contributors. All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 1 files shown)


danh-arm:
Hi @vwadekar

Soby went ahead and implemented his proposed fix to ARM-software/tf-issues#402. This will supersede your #656 PR. Let us know if you have any issues otherwise I will merge.

Regards

Dan.

davidcunado-arm:
Hi @vwadekar,

I'm planning on merging this tomorrow, WED 30-NOV - please let me know before then whether you have any issues?

Thanks,

David

vwadekar:
Hi David,

I won't be able to test this patch until Wed. But the implementation looks
similar to the one we have downstream.

One question - don't we want the code to run as part of release builds too
- as asserts won't be compiled there?

soby-mathew:
> One question - don't we want the code to run as part of release builds too - as asserts won't be compiled there?

The read in the assert() was accessing stale data on the cache line. Hence the need for cache maintenance ops prior to read. 
The idea is to keep the bakery lock info coherent with the memory just enough to ensure correctness of the algorithm, but not `always coherent` as it would be too expensive. Hence in release mode, the asserts are not present and the associated read cache operations need not be present too.

vwadekar:
Thanks Soby. That answers my question.

davidcunado-arm:
Hi @vwadekar 
I've closed your PR #656 as this PR resolves the issue and has now been merged.
Thanks,
David

PR #769, opened 2016-11-30
Check Trusty is present during the SPD's initialization

sandrine-bailleux-arm:
Add a debug assertion in the initialization function of Trusty's SPD
to check for the presence of Trusty. If Trusty is absent then the SPD's
setup function already detects it and returns an error code so the init
function will never been called. Therefore, a debug assertion is enough
to catch this improbable error case.


sandrine-bailleux-arm:
@vwadekar: Coverity has been complaining about a null pointer dereference in the trusty_init()
function so I've added a debug assertion like in the other SPDs. It is slightly pointless because the
init() function will never be called if the setup() function failed (which does check the address returned
by bl31_plat_get_next_image_ep_info()) but are you happy with this patch anyway?

vwadekar:
Thanks Sandrine. Looks good.

PR #783, opened 2016-12-15
Fixes for ARM Trusted Firmware Security Advisory TFV 1

danh-arm:


danh-arm:
I had to resolve some conflicts in this PR with https://github.com/ARM-software/arm-trusted-firmware/pull/781. I added my sign-off to the commits requiring resolution.

PR #789, opened 2016-12-19
plat_common: do not suppress runtime logs when DEBUG is defined

masahir0y:
When debugging, runtime logs will be useful.  For example,
psci_print_power_domain_map() will display details if compiled
with DEBUG option, but the console is needed to see the log.

### DIFF
--- a/plat/common/aarch64/plat_common.c
+++ b/plat/common/aarch64/plat_common.c
@@ -56,11 +56,13 @@ void bl32_plat_enable_mmu(uint32_t flags)
 
 void bl31_plat_runtime_setup(void)
 {
+#if !DEBUG
 	/*
 	 * Finish the use of console driver in BL31 so that any runtime logs
 	 * from BL31 will be suppressed.
 	 */
 	console_uninit();
+#endif
 }
 
 #if !ENABLE_PLAT_COMPAT


plat/common/aarch64/plat_common.c
```diff
@@ -56,11 +56,13 @@ void bl32_plat_enable_mmu(uint32_t flags)
 
 void bl31_plat_runtime_setup(void)
 {
+#ifndef DEBUG
```
jeenu-arm: I think DEBUG is always defined, to either 0 or 1. You might want to use `#if !DEBUG` to get the intended effect.

jeenu-arm: You just need to force-push your local branch again. I.e. just rerun your original push command with a -f. Github will update the PR automatically (as long as the remote branch is the same, I think).

masahir0y: I did the force-push. Thanks!

plat/common/aarch64/plat_common.c
```diff
@@ -56,11 +56,13 @@ void bl32_plat_enable_mmu(uint32_t flags)
 
 void bl31_plat_runtime_setup(void)
 {
+#if !DEBUG
```
danh-arm: With this change, BL31 runtime logs will be interleaved with the OS runtime logs, which may not be what all platforms want. In fact we introduced `bl31_plat_runtime_setup()` to avoid this issue. Therefore this change is unsuitable for a plat/common implementation. You are free to override this in your own platform how you like. If your platform has more than 1 UART, I would recommend using a different UART for TF and OS runtime logs (see `plat/arm/common/arm_bl31_setup.c` for an example of this).

masahir0y: OK, I will implement my own bl31_plat_runtime_setup().  Thanks!
Please discard this pull request.

PR #790, opened 2016-12-19
add utility macros to utils.h

masahir0y:


masahir0y:
Hi.

Yes, I need BIT(), min(), max() for my SoC port.  I will send my port, but I need some more time to brush up the code.  In the meanwhile, I am sending a pull request touching the common part.  Of course, I can duplicate the same definition of BIT(), min(), max() in my platform header, but I guess it is not the right thing to do.

You requested to make the macros more type-safe, but could you give me a clue?

At first, I wrote a macro like follows:
#define min(x, y) ({                           \
       __typeof__(x) _min1 = (x);                      \
       __typeof__(y) _min2 = (y);                      \
       (void) (&_min1 == &_min2);              \
       _min1 < _min2 ? _min1 : _min2; })

But, I found this would not compile for ARM Trusted Firmware.

ATF adopts '-std=c99', so my code was blocked by
'ISO C forbids braced-groups within expressions' error.

This prevent me from writing flexible code like Linux or other OSS projects.
Do you have an idea how to work around it and write a more safe macro
without using braces in the expression?

davidcunado-arm:
Hi @masahir0y 
Yes, when I tried to implement these macros, I had the same issues, but using the __extension__ seemed to resolve the build issue;

#define MAX(a, b) __extension__ ({      \
        __typeof__ (a) _a = (a);        \
        __typeof__ (b) _b = (b);        \
        _a > _b ? _a : _b;              \
        })

Anyway, I need to confirm with my colleagues whether we need to do this, but given the UK national holidays this will be delayed until early January - apologies.

The typesafe aspect apart, re-thinking... given you're planning on upstream your port then adding these utils.h is fine.

The min / max macro names should capitals - MIN / MAX. And as mentioned before we'll need to get Rockchip maintainer to approved changes to their platform port.

Thanks,

David

masahir0y:
> The min / max macro names should capitals - MIN / MAX

I am not sure about this.
We already have lower-case macros round_up() / round_down() in include/lib/utils.h
but I do not have a strong opinion about this.  So, I am fine with MIN/MAX.

In addition to them, I am using abs()  for my SoC port.

I can see the declaration for abs() in include/lib/stdlib/stdlib.h, but
no definition anywhere.

So, I wrote a stupid macro like this.
#define abs(x)                          ((x) < 0 ? -(x) : (x))

Of course, this is not safe for increment operator like   abs(x++),
but I could not figure out how to correct it without the extension.

If you are fine, I'd like to add  add  abs()  (or ABS() if you like capital letter) too.

Anyway, please let me know when there is an update for this discussion.

masahir0y:
As I have asked various questions in the "issues" forum, I have been hit by various problems when porting my SoCs.  This made me realize that it is difficult to upstream my code as I had expected first.  At this moment, I am not sure if I need this pull request or not.  I am still keeping this PR open for discussion, but you do not have to pull it.  You can close it anytime.  If it turned out necessary, I will re-open as another pull-req then.

danh-arm:
In general I prefer upper case for macro definitions, especially ones that take arguments, to avoid confusion between functions and macros. But as @masahir0y has observed, we are not consistent on this so I don't think this is a gating issue.

I like the proposed solution to add type safety to the macros, but since this is a GNU extension, it could perhaps be wrapped in a `#if __GNUC_' block.

masahir0y:
>  it could perhaps be wrapped in a #if __GNUC__ block.

After my porting re-design, I have less priority for min/max macros for now.

I dropped the 2/2 and rebased only BIT() commit on the latest branch.
Could you pull only BIT() move, please?
We have no more discussion for BIT().

If necessary, we can open the issues forum for further discussion for min/max.

davidcunado-arm:
jenkins: test this please

davidcunado-arm:
Changes look fine, but as you're making changes to rockship platform, need to get approval from the maintainer, @TonyXie06

TonyXie06:
It‘s good to provide a common macro for us。

davidcunado-arm:
Thanks, @TonyXie06!
I'll look to get this merged over this day or so.

PR #792, opened 2016-12-20
zynqmp: add "override" directive to mandatory options

masahir0y:
The platform.mk sets build options required for ZynqMP, but users
can still change them from the command line, like:

  make PLAT=zynqmp RESET_TO_BL31=0 CROSS_COMPILE=...

Then, the makefile shows an error message in that case:
  Using BL31 as the reset vector is only one option supported on ZynqMP.
  Please set RESET_TO_BL31 to 1.

If the option is not user-configurable, the makefile can specify
"override" to prevent users from changing it.  We do not need the
error message for the case that never happens.

Likewise, ENABLE_PLAT_COMPAT := 0 and PROGRAMMABLE_RESET_ADDRESS := 1
are mandatory to avoid build error.

Acked-by: Soren Brinkmann <soren.brinkmann@xilinx.com>

### DIFF
--- a/plat/xilinx/zynqmp/platform.mk
+++ b/plat/xilinx/zynqmp/platform.mk
@@ -27,12 +27,12 @@
 # ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
-ENABLE_PLAT_COMPAT := 0
-PROGRAMMABLE_RESET_ADDRESS := 1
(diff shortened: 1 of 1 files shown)


danh-arm:
jenkins: test this please

sorenb-xlnx:
It looks like something is wrong with this PR. There are a bunch of commits included that don't look like they should.

danh-arm:
Yes, we've been updating the integration branch due to some integration problems. In general, all pull requests should be made relative to the master branch to avoid problems like this. Please rebase on master.

masahir0y:
Rebase done!

PR #797, opened 2016-12-22
fiptool: Add support for operating on binary blobs using the UUID

dp-arm:
See individual commits for details.

### DIFF
--- a/include/lib/stdlib/sys/uuid.h
+++ b/include/lib/stdlib/sys/uuid.h
@@ -37,6 +37,9 @@
 /* Length of a node address (an IEEE 802 address). */
 #define	_UUID_NODE_LEN		6
 
+/* Length of UUID string including dashes. */
+#define _UUID_STR_LEN		36
+
 /*
  * See also:
  *      http://www.opengroup.org/dce/info/draft-leach-uuids-guids-01.txt
--- a/tools/fiptool/fiptool.c
+++ b/tools/fiptool/fiptool.c
@@ -172,6 +172,20 @@ static image_desc_t *new_image_desc(const uuid_t *uuid,
 	return desc;
 }
 
+static void set_image_desc_action(image_desc_t *desc, int action,
+    const char *arg)
+{
+	assert(desc != NULL);
+
+	if (desc->action_arg != DO_UNSPEC)
+		free(desc->action_arg);
+	desc->action = action;
+	desc->action_arg = NULL;
+	if (arg != NULL)
+		desc->action_arg = xstrdup(arg,
+		    "failed to allocate memory for argument");
+}
+
 static void free_image_desc(image_desc_t *desc)
 {
 	free(desc->name);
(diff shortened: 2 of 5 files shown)


masahir0y:
David reminded me of this pull-request, I just skimmed over the code.

The --blob option has a very long argument, but users will never type
such a long command directly, so I think we can live with it.

Just minor feedback from me.

[1] With this commit, toc_entries_len will be no longer used.  You can remove the last line in the tbbr_config.c and the declaration in the tbbr_config.h

[2] I personally prefer "if (!image)" to "if (image != NULL)"
     (avoid comparing with NULL or 0 explicitly), but I leave this coding style to maintainers.

dp-arm:
> The --blob option has a very long argument, but users will never type
> such a long command directly, so I think we can live with it.

Indeed.  I suspect people will script this part.  If you can afford recompiling the tool, you can always update the builtin table instead.

You might also find the functionality useful when operating on a FIP file that you did not generate yourself.  Such files may contain images that your version of fiptool might not understand via the builtin table.

masahir0y:
I have been thinking about this approach a bit more.

The advantage of the --blob option is that we can keep the fiptool platform-independent.

However, the tools/cert_create is already platform-dependent, right?

If the image added by --blob option needs authentication, I guess we will end up with expanding
   tools/cert_create/src/tbbr/rbb_key.c
   tools/cert_create/src/tbbr/rbb_cert.c

Likewise, we will need to touch 
  include/common/firmware_image_package.h
or something to make BL1/BL2 recognize the new blob added by --blob option.

So, I am not quite sure if the concept of --blob option is useful.

Another approach might be to allow the fiptool to compile
not only tools/fiptool/tbbr_config.c but also its platform-counterpart.

This will make the fiptool platform-dependent.
Instead, the advantage is that UUID is shared among BL1/BL2 and fiptool,
and also we need not type a long command.

At the moment, this feature is not urgent for me anyway.
Unless you have some reason to hurry up on your side, I hope more discussion about the approach.

danh-arm:
jenkins:retest this please

dp-arm:
> The advantage of the --blob option is that we can keep the fiptool platform-independent.

The --blob option is useful if you need to operate on a FIP file that contains images not known to fiptool via the builtin table, i.e. FIP files that you did not create yourself.  You can think of it as a kind of "raw" mode.  Depending on what you need to do with those FIP files, you might want to script this part.

> Another approach might be to allow the fiptool to compile
> not only tools/fiptool/tbbr_config.c but also its platform-counterpart.

I don't think the two approaches are mutually exclusive as they solve different problems.  If your boot flow includes some non-standard image, then this could be useful to have. Sure, you can abuse the --blob option to some effect but this is not the intended use-case.

BTW, do you currently use non-standard images in your boot flow?  If so, do you maintain patches to fiptool's tbbr_config.c and TF?  If you find this to be a pain point, you could provide a patch that allows platforms to include extra images in the TBBR configuration.

masahir0y:
> The --blob option is useful if you need to operate on a FIP file that contains images not known to fiptool via the builtin table, i.e. FIP files that you did not create yourself. You can think of it as a kind of "raw" mode. Depending on what you need to do with those FIP files, you might want to script this part.

OK, your intention is clear now.  Thanks for explaining this.

> BTW, do you currently use non-standard images in your boot flow? If so, do you maintain patches to fiptool's tbbr_config.c and TF? If you find this to be a pain point, you could provide a patch that allows platforms to include extra images in the TBBR configuration.

At first, I thought I needed a non-standard image in my boot flow, but after all, I am not using it for now.

masahir0y:
The fiptool in the integration branch is broken due to the pull request #797.
(I am fixing it in my pull request #810)

I will rebase this after the problem is fixed.

PR #799, opened 2016-12-25
fiptool: Alignment support + misc refactoring

masahir0y:


dp-arm:
> fiptool: support --align option to add desired alignment to image offset

The implementation looks good.  I would, however like to hear what other people have to say.  The patch might need to be rebased on top of my UUID pull request, but hold off on that until we have consensus.

afaerber:
On a brief look the --align option looks like what I was asking about for Amlogic S905 (Odroid-C2), but not tested yet.

tools/fiptool/fiptool.c
```diff
@@ -561,10 +570,29 @@ static void parse_plat_toc_flags(char *arg, unsigned long long *toc_flags)
 	*toc_flags |= flags << 32;
 }
 
+static int is_power_of_2(unsigned long x)
+{
+        return x && !(x & (x - 1));
+}
```
danh-arm: This function/macro is already available in utils.h

masahir0y: I did not know we can include include/lib/utils.h from host programs.
You need to be very careful about this
because mixture of target image code and host program code
often leads to maintenance headache.

Actually, we need to share include/common/firmware_image_package.h
between BL1/BL2 and FIPTOOL.  But, this is just one of some exceptions.

Currently, we do not have clear boundary between BL images and host-tools.
I believe this kind of work is really worthwhile.

At this moment, I do not want to include utils.h from FIPTOOL.
If we want to do it, we need to define a "shareable header directory" 
and move utils.h there.  But, this is a separate work.

For this pull request, I am opposed to including <utils.h> only for borrowing IS_POWER_OF_TWO().

dp-arm:
I just noticed that the usage() functions have not been updated to mention the --align option.

tools/fiptool/fiptool.c
dp-arm: I am inclined to go with masahir0y here.  I don't think we should try and generalize this proposition.  Even though utils.h is and will probably remain target agnostic, I don't think we should encourage sharing of code between TF and the host tools (as masahir0y already mentioned, we already share some code but this is mostly for convenience as the alternatives require more work).

masahir0y:
> I just noticed that the usage() functions have not been updated to mention the --align option.

Thanks.  I fixed it and pushed the tree forcibly.

Updates:
  - Add --align usage to create_usage(), update_usage(), and remove_usage().
  - Move FIP_ALIGN above FIP_NAME to keep make_helpers/defaults.mk sorted alphabetically.
  - Two more trivial commits on top of the alignment support commit

tools/fiptool/fiptool.c
danh-arm: Sorry, I didn't mean to suggest sharing this header with the tools - that was my mistake. It's fine as-is then.

danh-arm:
This PR needs rebasing. I'm not sure if this will conflict with the now merged #810. If so, you will either need to rebase on the integration branch or wait until master has caught up (in the next day or so).

dp-arm:
I started doing the rebase of this patch series, I will post the new one when ready.

tools/fiptool/fiptool.c
```diff
@@ -629,10 +662,11 @@ static void create_usage(void)
...
 static int update_cmd(int argc, char *argv[])
 {
-	struct option opts[toc_entries_len + 2];
+	struct option opts[toc_entries_len + 3];
 	char outfile[FILENAME_MAX] = { 0 };
 	fip_toc_header_t toc_header = { 0 };
 	unsigned long long toc_flags = 0;
+	unsigned long align = 1;
 	int pflag = 0;
 	int i;
```
dp-arm: add_opt(opts, i, "align", required_argument, OPT_ALIGN);

This is missing in the update and remove subcommands.  I have fixed this in the rebase that I did.

masahir0y:
> I started doing the rebase of this patch series, I will post the new one when ready.

I was working on this series, too.
(I have just been busy for other projects for a few days.)

I had also noticed the "info" command was not working.

Also, when I was testing my code, I found another regression added by your commit e0f083a.
So, I am doing more effort that I had posted this pull request first.

Anyway, I will update this pull request.

masahir0y:
I pushed a new version.

[1]  I noticed the current code base changes the image order by "update" command.
      I has been thinking if we should use the linked list or not.  Finally, I chose to
     revive replace_image() function to get back the original behavior.  
     ( If I remove the linked list, I saw a small problem with --blob option.  So,
       I am keeping Dimitris' design)

[2]  Some more bug-fixes, clean-ups.   This will allow us to work on new development
      on cleaner code base.

[3] Lastly, I rebased --align support patch on top of that.  The implementation was a little
     changed, but the basic idea is still the same.

tools/fiptool/fiptool.c
```diff
@@ -697,6 +703,24 @@ static void parse_plat_toc_flags(const char *arg, unsigned long long *toc_flags)
 	*toc_flags |= flags << 32;
 }
 
+static int is_power_of_2(unsigned long x)
+{
+        return x && !(x & (x - 1));
```
dp-arm: Minor thing, but this line is indented using spaces.

masahir0y:
> Minor thing, but this line is indented using spaces.

Good catch.  Thanks!

Fixed it and pushed forcibly.

danh-arm:
jenkins: test this please

PR #801, opened 2016-12-28
Macro cleanups

masahir0y:


danh-arm:
jenkins: test this please

masahir0y:
> This change looks good but you've removed all the copies of the comment block. Could you copy it here so it's not lost completely.

OK. I did so.

After some more consideration, I dropped
2/4 ( Move BL31_END and BL32_END)
4/4 (Add BL31_SIZE and BL32_SIZE)
because I'd like to take my time to find a better way for cleanups.

For now, please pull only two commits.

sandrine-bailleux-arm:
I'm fine with the macro renaming (BL_COHERENT_RAM_LIMIT -> BL_COHERENT_RAM_END) but I'd like it to be inconsistent across all macros of this kind. At the moment, we have a mix of *_END and *_LIMIT, which is not ideal from a clarity point of view. Could we either align all macros to use the new *_END naming scheme or drop the renaming?

davidcunado-arm:
jenkins: test this please

davidcunado-arm:
Ah - just seem comment from @sandrine-bailleux-arm.
@masahir0y - looks as if you have more changes / clean-ups to do; in which case would you prefer to switch back to using LIMIT, and then do the move to END across the board as part of your subsequent clean-up patches?

masahir0y:
> I'm fine with the macro renaming (BL_COHERENT_RAM_LIMIT -> BL_COHERENT_RAM_END) but I'd like it to be inconsistent across all macros of this kind. At the moment, we have a mix of *_END and *_LIMIT, 

We need to use both *_END an *_LIMIT correctly.
Basic rule is:

    *_BASE   +  *_MAX_SIZE  =  *_LIMIT
    *_BASE  +    *_SIZE   =  *_END

Generally, _LIMIT is defined platform_def.h because it shows SoC specific limitation
(for example, limitation comes from SRAM size.)
So, typical usage of _LIMIT is like:
      ASSERT(. <= BL31_LIMIT, "BL31 image has exceeded its limit.")

On the other hand, _END is derived from real image/section size,
which linker scripts tell us after compilation.

> in which case would you prefer to switch back to using LIMIT, and then do the move to END across the board as part of your subsequent clean-up patches?

My renaming  BL_COHERENT_RAM_LIMIT -> BL_COHRENT_RAM_END
is correct, so I do not want to revert.

The current usage of LIMIT in common_def.h is wrong.
  #define BL_CODE_LIMIT		(unsigned long)(&__TEXT_END__)
  #define BL_RO_DATA_LIMIT	(unsigned long)(&__RODATA_END__)
 
They should be renamed to BL_CODE_END, BL_RO_DATA_END, respectively.
If you want me to fix them, I can contribute.
(in this pull-request, if you request to update it)

sandrine-bailleux-arm:
Hi Masahiro,

Sorry I should have been clearer in the 1st place. I wasn't suggesting to use exclusively *_END or only
*_LIMIT. I completely agree with your analysis above, *_END and *_LIMIT both need to exist, as they
have different meanings. What I meant was, your patch fixes an incorrect usage of *_LIMIT but some
others remain and could benefit from the same fix. We would welcome such a contribution :)

Regards,
Sandrine

danh-arm:
I agree it would be good to fix the other instances in common_def.h in this PR, although that should not gate merging of this. @masahir0y - Please indicate if you intend to make further changes, otherwise we will merge.

masahir0y:
> Please indicate if you intend to make further changes, otherwise we will merge.

I will update the PR soon.
  - Rename the other macros
 - Rebase on the latest integration branch

masahir0y:
Update done!

danh-arm:
jenkins: retest this please

danh-arm:
Latest version looks good to me. For future reference, in general it's better to rebase on master than integration branch, since occasionally we have to rewind the integration branch. It should be OK in this case, though.

masahir0y:
> For future reference, in general it's better to rebase on master than integration branch, since occasionally we have to rewind the integration branch. It should be OK in this case, though.

The contribution guide (https://github.com/ARM-software/arm-trusted-firmware/blob/master/contributing.md) says:
   Submit a pull request to the arm-trusted-firmware integration branch

Is this wrong information?

danh-arm:
> Is this wrong information?

No. You should always submit to the integration branch, but it's better in general if topics are rebased against the master branch. This allows us to integrate patches in the order we choose, and avoids the need for us to ask for rebases when there are problems on the integration branch. The exception to this is when there are multiple open PRs that touch the same piece of code. In that case we either have to wait for the first PR to get onto master or get subsequent PRs to rebase on integration.

PR #802, opened 2017-01-04
rockchip: Build m0 firmware without standard libraries

pgeorgi:
Depending on the compiler used, it might try to link in libc even though
it's not required. Stop it from doing that.


danh-arm:
jenkins: test this please

PR #803, opened 2017-01-06
TBB: fix comment about MBEDTLS_KEY_ALG default

masahir0y:
This comment block says the default algorithm is ESDSA, while the
code obviously sets the default to RSA:

  ifeq (${MBEDTLS_KEY_ALG},)
      MBEDTLS_KEY_ALG            :=      rsa
  endif

The git log of commit 7d37aa171158 ("TBB: add mbedTLS authentication
related libraries") states available options are:

  * 'rsa' (for RSA-2048) (default option)
  * 'ecdsa' (for ECDSA-SECP256R1)

So, my best guess is the comment block is wrong.

The mismatch between the code and the comment is confusing. Fix it.


danh-arm:
jenkins: test this please

PR #807, opened 2017-01-10
uart: 16550: Fix getc

nmenon:
tbz check for RDR status is to check for a bit being zero.
Unfortunately, we are using a mask rather than the bit position.

Further as per http://www.ti.com/lit/ds/symlink/pc16550d.pdf (page 17),
LSR register bit 0 is Data ready status (RDR), not bit position 2.

Update the same to match the specification.

Reported-by: Sekhar Nori <nsekhar@ti.com>


danh-arm:
This looks good but please target the integration branch, not the master branch.

nmenon:
Done.

danh-arm:
jenkins: test this please

PR #808, opened 2017-01-12
Fix parallel building

masahir0y:
Soren reports build fails if -j option is given:

  $ make -j16 CROSS_COMPILE=aarch64-linux-gnu-
  Building fvp
  make: *** No rule to make target 'build/fvp/release/bl1/',
                    needed by 'build/fvp/release/bl1/bl1.ld'.  Stop.
  make: *** Waiting for unfinished jobs....

The cause of the failure is that $(dir ) leaves a trailing / on the
directory names.   It must be ripped off to let Make create the
directory.

Reported-by: Soren Brinkmann <soren.brinkmann@xilinx.com>

### DIFF
--- a/make_helpers/build_macros.mk
+++ b/make_helpers/build_macros.mk
@@ -227,11 +227,12 @@ endef
 # MAKE_LD generate the linker script using the C preprocessor
 #   $(1) = output linker script
 #   $(2) = input template
+#   $(3) = BL stage (2, 2u, 30, 31, 32, 33)
 define MAKE_LD
 
 $(eval DEP := $(1).d)
 
-$(1): $(2) | $(dir ${1})
+$(1): $(2) | bl$(3)_dirs
 	@echo "  PP      $$<"
 	$$(Q)$$(CPP) $$(CPPFLAGS) -P -D__ASSEMBLY__ -D__LINKER__ $(MAKE_DEP) -o $$@ $$<
 
@@ -297,7 +298,7 @@ define MAKE_BL
         $(eval BL_LINKERFILE := $(BL$(call uppercase,$(1))_LINKERFILE))
         # We use sort only to get a list of unique object directory names.
         # ordering is not relevant but sort removes duplicates.
-        $(eval TEMP_OBJ_DIRS := $(sort $(BUILD_DIR)/ $(dir ${OBJS})))
+        $(eval TEMP_OBJ_DIRS := $(sort $(BUILD_DIR)/ $(dir ${OBJS} ${LINKERFILE})))
         # The $(dir ) function leaves a trailing / on the directory names
         # We append a . then strip /. from each, to remove the trailing / characters
         # This gives names suitable for use as make rule targets.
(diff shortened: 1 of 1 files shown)


davidcunado-arm:
jenkins: test this please.

sandrine-bailleux-arm:
Hello,

I couldn't reproduce the build failure on my machine using:
`make -j16 CROSS_COMPILE=aarch64-linux-gnu-`

Could you please tell me what system you are building on? Is it Windows or Linux?

Regards,
Sandrine

masahir0y:
> Could you please tell me what system you are building on? Is it Windows or Linux?

Linux (Ubuntu 16.04LTS).

masahiro@pug:~/ref/arm-trusted-firmware$ cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=16.04
DISTRIB_CODENAME=xenial
DISTRIB_DESCRIPTION="Ubuntu 16.04.1 LTS"
masahiro@pug:~/ref/arm-trusted-firmware$ getconf _NPROCESSORS_ONLN
8
masahiro@pug:~/ref/arm-trusted-firmware$ uname -a
Linux pug 4.4.0-59-generic #80-Ubuntu SMP Fri Jan 6 17:47:47 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
masahiro@pug:~/ref/arm-trusted-firmware$ git describe
v0.2-1644-g9eb4d4d
masahiro@pug:~/ref/arm-trusted-firmware$ make realclean 
  REALCLEAN
masahiro@pug:~/ref/arm-trusted-firmware$ make -j8 CROSS_COMPILE=aarch64-linux-gnu- 
Building fvp
make: *** No rule to make target 'build/fvp/release/bl1/', needed by 'build/fvp/release/bl1/bl1.ld'.  Stop.
make: *** Waiting for unfinished jobs....

jeenu-arm:
I'm able to reproduce the issue on 16.04 (that comes with make 4.1, as opposed to 3.81 on my system, and hence presumably the change in behaviour).

I agree that your change fixes the problem, but I'm not sure that's where the fix needs to go.

The rule `MAKE_PREREQ_DIR` is in charge of creating pre-requisite directories, and that's where this should go. I.e., that should look like:

```
$(patsubst %/,%,$(1)) $(patsubst %/,%,$(1))/: $(2)
        ...
```

I.e., the rule handles both `foo` and `foo/`. Both `unix.mk` and `windows.mk`, and comments therein needs changing to the same effect.

The above change would handle pre-requisite directory creation from everywhere, not just for LD pre-processing. With this change, the special treatment in `build_macros.mk`, lines 301 to 304 aren't required either.

Hope that makes sense. Are you happy to re-submit your PR? I'm equally happy to do that at our end too.

davidcunado-arm:
@masahir0y - see comments from jeenu - could you make the suggested changes to the PR?

masahir0y:
At first, my interest in this pull request was just to fix the problem,
not to refactor the code.
This topic is less significant because the big picture of the build system is already a clap.
However, if we really want to talk about "where the fix need to go"
I am opposed to Jeenu's opinion.

MAKE_PREREQ_DIR is intended to compensate the difference among operating systems.
It is not intended to append or strip trailing slashes; this task is common across OSes.
So, it should be cared outside of {unix.mk, windows.mk}.
Otherwise, we would end up with duplicating as many "$(patsubst %/,%,$(1)) $(patsubst %/,%,$(1))/"
as the number of OSes.   This is ugly.

Just looking at the current design,

bl${1}_dirs: | ${OBJ_DIRS}

seems the central place of creating directories.  So, my new solution is to make MAKE_LD look like MAKE_C and MAKE_S.  Comments are welcome.

danh-arm:
jenkins: retest this please

jeenu-arm:
From a utilitarian stand point, yes, what's in this pull request isn't wrong; besides it fixes a usability bug. So I'm happy with it as it's.

PR #810, opened 2017-01-14
Fix fiptool bug introduced by recent rework

masahir0y:


danh-arm:
jenkins: test this please

PR #815, opened 2017-01-20
drivers: add designware emmc driver

hzhuang1:
Support Designware eMMC driver. It's based on both IO block
and eMMC driver.

### DIFF
--- a/drivers/synopsys/emmc/dw_mmc.c
+++ b/drivers/synopsys/emmc/dw_mmc.c
@@ -1,0 +1,424 @@
+/*
+ * Copyright (c) 2016-2017, ARM Limited and Contributors. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * Neither the name of ARM nor the names of its contributors may be used
+ * to endorse or promote products derived from this software without specific
+ * prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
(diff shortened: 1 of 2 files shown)


hzhuang1:
It's used to fix issue https://github.com/ARM-software/tf-issues/issues/376.

drivers/synopsys/emmc/dw_mmc.c
```diff
@@ -0,0 +1,423 @@
...
+		udelay(500);
+		data = mmio_read_32(base + DWMMC_RINTSTS);
+
+		if (data & err_mask)
+			return -EIO;
+		if (data & INT_DTO)
+			break;
+		if (--timeout == 0) {
+			ERROR("%s, RINTSTS:0x%x\n", __func__, data);
+			assert(timeout > 0);
```
danh-arm: Perhaps worth panicking here instead of asserting to avoid tight loop in release builds.

drivers/synopsys/emmc/dw_mmc.c
```diff
@@ -0,0 +1,423 @@
...
+		op |= CMD_RESP_EXPECT;
+		break;
+	default:
+		op |= CMD_RESP_EXPECT | CMD_CHECK_RESP_CRC;
+		break;
+	}
+	timeout = TIMEOUT;
+	do {
+		data = mmio_read_32(base + DWMMC_STATUS);
+		assert(--timeout > 0);
```
danh-arm: Perhaps worth panicking here instead of asserting to avoid tight loop in release builds.

danh-arm:
jenkins: test this please

hzhuang1:
Fixed the assert issue.

masahir0y:
I asked @hzhuang1 about io_block.c in the following question: 
https://github.com/ARM-software/tf-issues/issues/447
I have not got any response, though.

danh-arm:
jenkins: retest this please

hzhuang1:
@masahir0y I'm sorry that I missed your email. I'll check your patch and give response soon.

danh-arm:
While I agree that question should be answered, as far as I can see that is independent to this PR so should not block merging. Am I right?

masahir0y:
> While I agree that question should be answered, as far as I can see that is independent to this PR so should not block merging. Am I right?

Ah, right.

I misunderstood this patch by the commit message "It's based on both IO block ...".
This patch seems actually independent from the IO block implementation.
So, nothing is blocked.  Please go ahead!

PR #816, opened 2017-01-23
partition: fix the path of head file

hzhuang1:
Add partition path of head file in both head and source file.
Otherwise, partition path has to be filled in PLAT_INCLUDES of
platform.mk.

Fixes arm-software/tf-issues#450.

### DIFF
--- a/drivers/partition/gpt.c
+++ b/drivers/partition/gpt.c
@@ -31,7 +31,7 @@
 #include <assert.h>
 #include <debug.h>
 #include <errno.h>
-#include <gpt.h>
+#include <partition/gpt.h>
 #include <string.h>
 
 static int unicode_to_ascii(unsigned short *str_in, unsigned char *str_out)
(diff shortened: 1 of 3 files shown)


danh-arm:
jenkins: test this please

danh-arm:
In reviewing this, I notice we already add a long list of directories to the system include path in the top-level Makefile, some of which are not used by all platforms/images. The simplest fix to your problem would be to just add the partition directory to the list there. However, I'm not sure that is the best thing to do generally. One alternative (general) solution is to use relative paths everywhere (like this PR does) but that would make the source code more ugly and difficult to maintain. Another solution is to devolve building the include list to the individual image/platform/SPD makefiles.

I'm currently thinking we should just keep adding these to the top-level Makefile until such time this becomes unscaleable. Any other thoughts?

hzhuang1:
OK. I'll refer to head files in platform makefiles.

danh-arm:
I'm going to close this then as we expect the platform makefiles to add this directory to PLAT_INCLUDES.

We have a more general task (internal ref https://jira.arm.com/browse/GENFW-1219) to devolve these includes from the top-level Makefile to the individual BL/SPD/platform makefiles (this task is not currently scheduled).

PR #819, opened 2017-01-26
Resolve build errors flagged by GCC 6.2

davidcunado-arm:
With GCC 6.2 compiler, more C undefined behaviour is being flagged as
warnings, which result in build errors in ARM TF build.

The specific issue that this patch resolves is the use of (1 << 31),
which is predominantly used in case statements, where 1 is represented
as a signed int. When shifted to msb the behaviour is undefined.

The resolution is to specify 1 as an unsigned int using a convenience
macro ULL(). A duplicate macro MAKE_ULL() is replaced.

Fixes ARM-software/tf-issues#438

Change-Id: I08e3053bbcf4c022ee2be33a75bd0056da4073e1

### DIFF
--- a/include/bl32/payloads/tlk.h
+++ b/include/bl32/payloads/tlk.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015, ARM Limited and Contributors. All rights reserved.
+ * Copyright (c) 2015-2017, ARM Limited and Contributors. All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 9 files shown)


davidcunado-arm:
@ronan22 this PR includes fix from your PR #804.

davidcunado-arm:
@jenswi-linaro I've made a change to the opteed code to address gcc 6.2 build errors, as reported in ARM-software/tf-isses/438
Can you approve this change?

danh-arm:
The build failure is due to a checkpatch false positive, which we do not intend to fix:
```
services/spd/trusty/smcall.h:44: ERROR: space prohibited after that '&' (ctx:WxW)
+		(((((unsigned int) (fastcall)) & 0x1) << 31) |	\
 		                               ^
```
We will merge this in the next day or so if there are no comments.

PR #820, opened 2017-01-26
Add --align <value> to fiptool to specify the image alignment (rebased patch series)

dp-arm:


masahir0y:
I have not checked this in detailed yet.

I am seeing another problem, so I will update my #799.  (hopefully soon)
If you do not rush this, please compare the two,
then judge which one you like.

dp-arm:
We are not in a rush, but if you think this pull request is OK I encourage you to base your work on top of it.  I can then squash as necessary before merging.

If something in this PR is not okay, then let me know and I can fix it. Then proceed by putting more patches on top of it.

masahir0y:
Your fix-up seems correct in the sense that it makes "info" command work with --align option.

Since your rework (specifically commit e0f083a), the "update" command changes the order of images.
For example, if I update BL2, the FIP will contain BL31, BL33, BL2, in this order.
remove_image() drops the old image from the linked list,
then add_image() adds the new image at the tail of the linked list.
The images are packed in the new order.
It works, but I am not comfortable.  I am not asking the tool to change the order of the images.

I believe the linked list was a bad choice in the first place.

This motivated me to re-work again.
Also, copying "uuid", "size" between "struct image" and "struct fip_toc_entry"
and you are now adding "offset", right?
So, the two structures are the same except the "buffer" pointer.

If you want to fix the design, I will leave this task to you.
(but the change of the infrastructure will affect my implementation.)

My preferred development style is, "fix underlying issues" then "expand features".

We can swap the order when we are in a rush for new features, then fix the design later.  However, this often requires more efforts in a long run, and the last thing I want to do in general.

dp-arm:
> Since your rework (specifically commit e0f083a), the "update" command changes the order of images.
> For example, if I update BL2, the FIP will contain BL31, BL33, BL2, in this order.
> remove_image() drops the old image from the linked list,
> then add_image() adds the new image at the tail of the linked list.
> The images are packed in the new order.
> It works, but I am not comfortable. I am not asking the tool to change the order of the images.

I was aware of this particularity and I am glad to see it fixed in your PR.

> I believe the linked list was a bad choice in the first place.

May have been indeed.  The alternative is to use an array as before and either impose a maximum limit on number of images or use realloc().  It may be simpler but it is up for discussion.

> This motivated me to re-work again.
> Also, copying "uuid", "size" between "struct image" and "struct fip_toc_entry"
> and you are now adding "offset", right?
> So, the two structures are the same except the "buffer" pointer.
> 
> If you want to fix the design, I will leave this task to you.
> (but the change of the infrastructure will affect my implementation.)

That bit bugged me as well.  I prefer your approach of embedding the toc_entry for now.
I was also thinking whether we really do need two separate lists (one for descriptors and one
for images).

Maybe the code could be changed to only ever traverse the descriptor list and access
the images through the descriptor node.  Let me think about it.

> My preferred development style is, "fix underlying issues" then "expand features".
> 
> We can swap the order when we are in a rush for new features, then fix the design later. However, this often requires more efforts in a long run, and the last thing I want to do in general.

Yup, fully agree.

masahir0y:
> That bit bugged me as well. I prefer your approach of embedding the toc_entry for now.
I was also thinking whether we really do need two separate lists (one for descriptors and one
for images).

> Maybe the code could be changed to only ever traverse the descriptor list and access
the images through the descriptor node. Let me think about it.


Right.  I was expecting this comment!

When I was tackling on the re-order problem of the "update" command,
I came up with two approaches:

[1] Revive replace_image() function, keeping two separate lists as-is.

[2] Remove the linked list of images, then add a pointer to image into struct image_desc.
    Traverse the descriptor list, then images are accessed via `desc->image`


Actually, I tried both.  I found pros and cons for both approaches.


[1] revives the original behavior as we had before your re-work.
The "update" command seems to work nicely also with your new --blob option.

I see one minor issue in the following use case:

1. Assume we have FIP with BL2, BL31, BL32, BL33
2. Remove BL32 with `fiptool remove --tos-fw  fip.bin`
3. Re-add BL32 with `fiptool update --tos-fw bl32.bin  fip.bin`

In this case, we get FIP with BL2, BL31, BL33, BL32 in this order.
(a new image is appended at the tail)
I think this is the only case to create an FIP that includes images
in a different order than the descriptor order


[2] always handles images in the order of descriptor.
So, "remove and re-add" gives the same result as a single "update" command.  This is good.
The con is, the "update" command changes the order of user-defined blobs given by --blob.


After all, I ended up with [1] in my pull request,
but [2] should be OK as well if you like.
Whether we take either way, we have pros and cons.
(or, if somebody suggests a better idea, that would be really appreciated...)

dp-arm:
> After all, I ended up with [1] in my pull request,
> but [2] should be OK as well if you like.
> Whether we take either way, we have pros and cons.
> (or, if somebody suggests a better idea, that would be really appreciated...)

Thanks for all the work.  We'll go with approach [1] as implemented in your PR.  I will look into approach [2] at a later date.

PR #823, opened 2017-01-30
Add -fno-builtin to CFLAGS

douglas-raillard-arm:
Disable the automatic substitution of functions with builtins. The
existing -ffreestanding option should already do this but explicitly
adding -fno-builtin reduces the risk of compiler variation. With this
option, GCC is not supposed to be able to make assumptions on what the
function does, which could otherwise lead to security-sensitive code
removal.

This can lead to potentially less efficient code but improves
predictability of what code is actually compiled into the binary.

Change-Id: I06ad151c61318bd1b00d84976f051d2d94314acc


danh-arm:
jenkins: retest this please

danh-arm:
This PR seems to have inherited the checkpatch failure from #819. The problem should go away once master is caught up.

PR #826, opened 2017-01-31
Decouple PSCI stat residency calculation from PMF

dp-arm:
The original PSCI stat implementation was coupled with PMF.  This patch introduces 3 optional PSCI interfaces that a platform can implement in order to calculate the PSCI stat residency without necessarily using PMF.


davidcunado-arm:
jenkins: test this please

PR #828, opened 2017-02-03
Fiptool cleanup

masahir0y:


davidcunado-arm:
jenkins: test this please

tools/fiptool/fiptool.c
```diff
@@ -826,7 +826,7 @@ static void create_usage(void)
 	    "pointed to by file.\n");
 	printf("  --plat-toc-flags <value>\t16-bit platform specific flag field "
 	    "occupying bits 32-47 in 64-bit ToC header.\n");
-	fputc('\n', stderr);
+	printf("\n");
```
dp-arm: Good catch, generally usage output should all go to stderr but in this case it does not really matter because fiptool cannot be used in a pipeline so I am okay with this.

tools/fiptool/fiptool.c
```diff
@@ -52,34 +52,6 @@
 #define OPT_PLAT_TOC_FLAGS 1
 #define OPT_ALIGN 2
 
-static image_desc_t *lookup_image_desc_from_uuid(const uuid_t *uuid);
```
dp-arm: I don't really like this to be honest.  I prefer the table to stay at the top because it is data.

masahir0y: Why do you believe data should come before functions?

tools/fiptool/fiptool.c
masahir0y: I do not have a strong opinion about whether help messages should go to stdout or stderr, as long as I see consistency.   Whether stdout or stderr is a matter of taste.  For example,  "git help" outputs usage messages to stdout.

antonio-nino-diaz-arm:
This patch targets the master branch instead of integration.

tools/fiptool/fiptool.c
dp-arm: I prefer global data to be at the top, so one can easily find them/see them when opening the file before reading any of the code.

masahir0y: The "cmds" is prefixed with "static", so it is no more global than any other static functions.

The "cmds" is an array of function callbacks.
The callback functions are usually defined above the ops array to save declarations.
We see the same pattern everywhere in drivers.

    static int my_foo_read(...)
    {
           ...
    }

    static int my_foo_write(...)
    {
            ...
    }

    static struct foo_driver_ops my_foo_ops = {
            .read = my_foo_read,
            .write = my_foo_write,
            ....
    };
 
If you look at Linux Kernel, drivers are like that.  Callback functions first, then operation structure that contains them.

If you want to see some examples in ATF,
"git grep -C10 DECLARE_RT_SVC".

Of course, it is possible to put the DECLARE_RT_SVC() at the top of file if you like
(you need to add some declarations at the very top), what does it buy you?

masahir0y:
> As per @antonio-nino-diaz-arm comment, please target integration branch.

It is easier to do that, but I want to be sure about this.

Currently, both master and integration point to the same commit bcc2bf09.
So, does it make any difference?

On more thing, in the pull request #801, I made it target to the integration branch at first,
but Dan said "Latest version looks good to me. For future reference, in general it's better to rebase on master than integration branch, since occasionally we have to rewind the integration branch. It should be OK in this case, though."

This time, I targeted the master branch (by chance), then you guys suggested me the integration branch.
From what I saw in the forum, the statement varies case by case.  Do we really need to be picky about this?

antonio-nino-diaz-arm:
The process is to base your changes on the latest master branch because that one is known to be working correctly, and it is clear that any problem during the pull request checks is introduced by the changes of the pull request. The target has to be the integration branch because that way multiple branches can be merged to integration, tested together and then merged back to master when we do more in-depth checks and are sure that everything works as expected. All of that is done without actually affecting the master branch in case there is a problem introduced by the merges that wasn't detected by the tests done for every pull request. That way, even if someone happened to clone the repository when integration was having problems, master would still be working correctly so that person would still have a working copy of the repository.

tools/fiptool/fiptool.c
dp-arm: This boils down to preferences.  TF is not consistent in this regard either, just check drivers/io/io_block.c.
I will let others voice their opinion and we can decide later.

masahir0y: Right.  I had noticed that.  drivers/io/* are not nice.  They need much reworking.  I like to avoid declarations for static functions.   There is no point to describe the same thing twice.

masahir0y:
> The process is to base your changes on the latest master branch because that one is known to be working correctly, and it is clear that any problem during the pull request checks is introduced by the changes of the pull request. The target has to be the integration branch because that way multiple branches can be merged to integration, tested together and then merged back to master when we do more in-depth checks and we are sure that everything works as expected. All of that is done without actually affecting the master branch in case there is a problem introduced by the merges that wasn't detected by the tests done for every pull request. That way, even if someone happened to clone the repository when integration was having problems, master would still be working correctly so that person would still have a working copy of the repository.

I know this.
You are explaining how the maintainers manage the two branches, right?
And you are talking about a different thing from my question.

At this point of time (Fri Feb  3 17:19:06 UTC 2017), both branches (master and integration)
point to the same commit bcc2bf0 ("Merge pull request #821 from jeenu-arm/errata-printing").
The three commits in my pull request are based on it.

In this situation,
-    `git rebase master fiptool`
-    `git rebase integration fiptool`
-    `git rebase bcc2bf0 fiptool`
are all equivalent.  So, nothing happens.

    $ git remote get-url upstream
    https://github.com/ARM-software/arm-trusted-firmware.git
    $ git fetch  upstream 
    $ git log --oneline -1 upstream/master 
    bcc2bf0 Merge pull request #821 from jeenu-arm/errata-printing
    $ git log --oneline -1 upstream/integration 
    bcc2bf0 Merge pull request #821 from jeenu-arm/errata-printing
    $ git log --oneline -1  master
    bcc2bf0 Merge pull request #821 from jeenu-arm/errata-printing
    $ git log --oneline -1  integration 
    bcc2bf0 Merge pull request #821 from jeenu-arm/errata-printing
    $ git log --oneline -4  fiptool
    eaecaa4 fiptool: re-order function definitions to delete declarations
    a15ce4d fiptool: stretch out printf strings for readability and grep'ability
    6c10070 fiptool: print new lines of usage to stdout instead of stderr
    bcc2bf0 Merge pull request #821 from jeenu-arm/errata-printing
    $ git rebase integration fiptool
    Current branch fiptool is up to date.
    $ git rebase master fiptool
    Current branch fiptool is up to date.
    $ git rebase bcc2bf0 fiptool
    Current branch fiptool is up to date.

Accepting this pull request means pulling commit eaecaa4.
git is not even aware of whether I based my work on master or integration.

Is my point clearer?

masahir0y:
Sorry, my misunderstanding.
I thought the maintainer could choose integration or master when pulling this.

When we use git locally, it is a puller's choice which branch he pull things into.
But GitHUB is probably different.
The requester specifies the target branch, and the puller (maintainer) cannot change this.

Will fix the request soon.

tools/fiptool/fiptool.c
davidcunado-arm: As @dp-arm states, TF code-base is not consistent in this regard and so this change doesn't bring this file into alignment with the rest of the codebase - if would trade inconsistency from one direction to another!
The rest of the PR looks good - could you update the PR without this change?

masahir0y: OK.  The last one has been replaced with a new one.
(I will plan to fix this issue and make it consistent, though.)

masahir0y:
Rebased.

davidcunado-arm:
jenkins: test this please

PR #832, opened 2017-02-06
RK3399 ARM TF clean up 20170206

rkchrome:
Hi,
  So far, it looks like that we have not updated the ARM-TF related codes on upstream for RK3399. In order to prevent the project codes and upstream's of the growing differences, we need to update the latest patches from the Chromium project to synchronize to the upstream.

Which mainly updates following:

1/ updates DDR sleep/resume, and DDR freq features
2/ updates RK3399 M0 features
3/ clean up ARM-TF codes style for RK3399

Thanks.


acgzx:
Could any one (ARM) help to review these?

Thanks.

PR #837, opened 2017-02-13
build: Use separate CFLAGS for tools

douglas-raillard-arm:
Sharing the same CFLAGS for both the firmware and host tools is not
useful and can lead to compilation failure when CFLAGS is set on the
command line.

This patch make MAKEOVERRIDES special variable empty so make's command
line variable definitions are not implicitly forwarded to sub make
invocations. However, make options are still forwarded like "-s" option
to make the build silent.

Change-Id: Ia921c02beb1fd2041696a26160b9e7974ae000fa

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,11 @@ VERSION_MINOR			:= 3
 # Default goal is build all images
 .DEFAULT_GOAL			:= all
 
+# Avoid any implicit propagation of command line variable definitions to
+# sub-Makefiles, like CFLAGS that we reserved for the firmware images'
+# usage. Other command line options like "-s" are still propagated as usual.
+MAKEOVERRIDES =
+
 MAKE_HELPERS_DIRECTORY := make_helpers/
 include ${MAKE_HELPERS_DIRECTORY}build_macros.mk
 include ${MAKE_HELPERS_DIRECTORY}build_env.mk


masahir0y:
Curious.  As far as I tested, MAKEOVERRIDES did not work as your comment stated.
Could you provide me some test patterns?

douglas-raillard-arm:
Hi @masahir0y, the build fails without this patch when invoking the top-level Makefile with CFLAGS='-mgeneral-regs-only' for example (or any other aarch64/32 specific compiler flag).
I've tested the forwarding of other options (interactions between MAKOVERRIDES and MAKEFLAGS) in a separate scratch Makefile.

masahir0y:
Maybe are you misunderstanding the usage of CFLAGS?

The doc/user-guide.md says as follows:

    Extra debug options can be passed to the build system by setting `CFLAGS`:
    CFLAGS='-O0 -gdwarf-2'                                     \
    make PLAT=<platform> DEBUG=1 V=1 all


The user guide is saying to do

    $ CFLAGS=-mgeneral-regs-only make CROSS_COMPILE=aarch64-linux-gnu- fip BL33=readme.md

instead of

    $ make CFLAGS=-mgeneral-regs-only make CROSS_COMPILE=aarch64-linux-gnu- fip BL33=readme.md

I tried this:

    $ make CFLAGS=-mgeneral-regs-only CROSS_COMPILE=aarch64-linux-gnu-  fip BL33=readme.md 
      CC      fiptool.c
    gcc: error: unrecognized command line option ‘-mgeneral-regs-only’
    make[1]: *** [fiptool.o] Error 1
    make: *** [tools/fiptool/fiptool] Error 2
    $ CFLAGS=-mgeneral-regs-only make CROSS_COMPILE=aarch64-linux-gnu-  fip BL33=readme.md 
      CC      fiptool.c
      CC      tbbr_config.c
      LD      fiptool

    Built fiptool successfully

    Trusted Boot Firmware BL2: offset=0xB0, size=0x4188, cmdline="--tb-fw"
    EL3 Runtime Firmware BL31: offset=0x4238, size=0x6090, cmdline="--soc-fw"
    Non-Trusted Firmware BL33: offset=0xA2C8, size=0x21C1, cmdline="--nt-fw"

    Built build/fvp/release/fip.bin successfully

masahir0y:
Having said that, the usage of CFLAGS is unclear.  In my opinion, the solution with MAKEOVERRIDES is  difficult to understand because MAKEOVERRIDES is only effective after entering sub makes.

As a comparison, I've pushed my pull request #839.  I hope this will make the intention clearer.

douglas-raillard-arm:
This is indeed the current documented way of achieving this result. However, there are shortcomings with the use of environment variables:

* It is inconsistent with the way of changing CFLAGS for tools' Makefile, because they use a direct assignment to CFLAGS which overrides the value from the environment. OTH, using a parameter forces the value (and then ignores the assignments).

* The tools' CFLAGS probably need the same kind of fix as what the main TF Makefile got a long time ago, which is having a TF_CFLAGS for internal Makefile use and a user-facing CFLAGS variable to allow appending options without overriding the mandatory one. Doing that would bring us to square one as user-defined CFLAGS would now be shared again.

In short, the only reason why setting CFLAGS as an environment variable currently works is because the tools' Makefiles are written in a straightforward style that does not let the user do much changes.

The only ways to get out of this situation once and for all is either to remove any implicit parameters passing from the sub-make invocations as I did, or to introduce a new TOOLS_CFLAGS variable so there is no name clash. The first solution has been chosen as it allows using the conventional CFLAGS name.
Also, the tools' Makefiles belong to a different logical build system as they have different needs, compiler and purposes. Breaking any implicit link between these Makefiles is therefore more logical. The TF Makefiles' hierarchy is not impacted by this change as Makefiles are included instead of being called recursively.

I hope this shed some light on the rational behind this patch.

danh-arm:
I agree in hindsight that it would be better if TF used unique build options for each Makefile rather than the broadly names CFLAGS across several Makefiles. However, changing this now would introduce a compatibility break for calling scripts, with minimal benefit. Unless there is a use-case for passing tool specific compile flags through the top-level Makefile and into the tools Makefile, then I'd rather leave the calling semantics as-is.

Therefore I think the solution in this PR is sufficient and #839 should be closed.

masahir0y:
This pull-request is not straight-forward to fix the problem.  Adding "override" to tools' Makefile would be easier to understand.  Or another way is to rename CFLAGS to HOSTCFLAGS in tools' Makefiles.

danh-arm:
> This pull-request is not straight-forward to fix the problem.

It seems straight-forward to me; it is a one-line fix with clear documentation.

> Adding "override" to tools' Makefile would be easier to understand. 

If you mean use `override variable = value`, that would prevent passing a command line CFLAGS to the tools Makefile when invoked directly (although admittedly this isn't properly supported anyway).

> Or another way is to rename CFLAGS to HOSTCFLAGS in tools' Makefiles.

That could work, although it would be a bigger change and I think proper override support we need to be added at the same time. As Douglas said, he chose this option in order to keep the conventional CFLAGS variable.

douglas-raillard-arm:
Also, using override on specific variables leaves the door opened for more unintended "variable leakage". Only CFLAGS seems to suffer from that for now but this can affect any variable that happens to have the same name.

PR #839, opened 2017-02-13
Build: pass extra options via TF_EXTRA_CFLAGS instead of CFLAGS

masahir0y:
The doc/user-guide.md says as follows:

  Extra debug options can be passed to the build system by
  setting `CFLAGS`:

  CFLAGS='-O0 -gdwarf-2'                                     \
  make PLAT=<platform> DEBUG=1 V=1 all

The usage of CFLAGS is confusing.

If CFLAGS is passed as an environment like above, it works.  On the
other hand, if CFLAGS is passed from the command line, it causes
an error.  For example,

  make CFLAGS=-mgeneral-regs-only CROSS_COMPILE=aarch64-linux-gnu- \
  fip BL33=...

will fail to build the fiptool because CFLAGS will also override the
instance in tools/fiptool/Makefile.  Sharing the same CFLAGS for
both the firmware and host tools is problematic.  Actually, CFLAGS
will be a potential name conflict because I also see it in
plat/rockchip/rk3399/drivers/m0/Makefile.  Let's use a different
flag, TF_EXTRA_CFLAGS, for the extra opitons for firmware.  This can
be passed via either an environment or a command line option.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -150,7 +150,8 @@ ASFLAGS			+=	$(CPPFLAGS) $(ASFLAGS_$(ARCH))			\
 				-Wa,--fatal-warnings
 TF_CFLAGS		+=	$(CPPFLAGS) $(TF_CFLAGS_$(ARCH))		\
 				-ffreestanding -fno-builtin -Wall -std=c99 -Os	\
-				-ffunction-sections -fdata-sections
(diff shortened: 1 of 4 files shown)


danh-arm:
As discussed in #837, we prefer the solution there.

PR #850, opened 2017-02-24
cleanups of IO layer

masahir0y:


danh-arm:
We discussed this PR internally. As far as we can see, there is no compelling reason to choose this "avoid forward declarations" style over the existing "keep static data at the top" style. Other projects may lean towards one style or other, but this is more of de-facto situation than a strong policy. Both are equally valid and the author/maintainer of each module can choose whichever style they prefer.

Therefore this PR comes under the category of "gratuitous change", which simply adds code churn and risk of regression. We don't intend to merge it.

masahir0y:
> We discussed this PR internally.

I wanted you to do discussion in the public forum,
like you suggested me to do so before.
(Although I already got smells that this project pretends to be open.)

> Hi Yamada-san.
>
>This sounds OK but could you use the public forum where possible?
>https://github.com/arm-software/tf-issues/issues
>
>Thanks
>
>Dan.

>Therefore this PR comes under the category of "gratuitous change", which simply adds code churn and risk of regression. We don't intend to merge it.

I believed this was a cleanup because we need not double the same information for nothing.
Anyway, I apologize with lots of "gratuitous changes".   I will try my best to not bother you in the future.

PR #855, opened 2017-03-01
Enable OS Initiated mode

freedomtan:
Make OS Initiated mode run. Genpd-based SoC idle [1]
relies on OS Initiated mode to work. This patch adds
a function for setting PSCI suspend mode to either
platform coordinated mode or OS Initiated mode.

[1] https://www.spinics.net/lists/arm-kernel/msg526814.html

Signed-off-by: Koan-Sin Tan <koansin.tan@gmail.com>

### DIFF
--- a/include/lib/psci/psci.h
+++ b/include/lib/psci/psci.h
@@ -83,6 +83,7 @@
 #define PSCI_NODE_HW_STATE_AARCH64	0xc400000d
 #define PSCI_SYSTEM_SUSPEND_AARCH32	0x8400000E
 #define PSCI_SYSTEM_SUSPEND_AARCH64	0xc400000E
+#define PSCI_SET_SUSPEND_MODE		0x8400000F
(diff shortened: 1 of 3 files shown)


davidcunado-arm:
jenkins: test this please

danh-arm:
Hi @freedomtan 

This patch doesn't really enable OSI mode; it provides a fake implementation of PSCI SET_SUSPEND_MODE to make the PSCI client think the firmware is running in OSI mode, when in fact it is still running in Platform Initiated mode. This is not acceptable for upstream.

Also, we're not convinced that implementing OSI mode provides any significant benefits, there we do not intend to implement it unless there is evidence to the contrary.

Regards

Dan.

danh-arm:
I'm closing this as it does not really implement OSI mode. The requirement to implement OSI mode in TF should be discussed separately.

PR #856, opened 2017-03-01
Introduce version 2 of the translation tables library

antonio-nino-diaz-arm:
The main differences between this version and the previous one are:

- Support for different contexts of translation tables.
- Support for dynamic mapping and unmapping of memory regions. It can be enabled in a per-image basis.

For more details, check the individual commit messages.

### DIFF
--- a/lib/xlat_tables_v2/xlat_tables_internal.c
+++ b/lib/xlat_tables_v2/xlat_tables_internal.c
@@ -563,11 +563,14 @@ static void xlat_desc_print(uint64_t desc)
 }
 
 static const char * const level_spacers[] = {
-	"",
-	"  ",
-	"    ",
-	"      "
+	"[LV0] ",
+	"  [LV1] ",
+	"    [LV2] ",
+	"      [LV3] "
 };
+
+static const char *invalid_descriptors_ommited =
+		"%s(%d invalid descriptors omitted)\n";
 
 /*
  * Recursive function that reads the translation tables passed as an argument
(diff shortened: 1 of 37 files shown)


vwadekar:
@antonio-nino-diaz-arm 

Thanks for the PR. I would like to test this on Tegra platforms before we go ahead with the merge. ETA 3/2.

danh-arm:
Hi @vwadekar 

> I would like to test this on Tegra platforms before we go ahead with the merge

The new code is "off" by default so should not impact you unless you pull in the new sources. However, we'd be really interested to hear your feedback so we'll hold off merging until we're hear back from you.

Cheers

Dan.

vwadekar:
@danh-arm 

We have a few use cases where the dynamic memory map/unmap code would help. Hence, I wanted to try that out before we merge.

vwadekar:
@danh-arm @antonio-nino-diaz-arm 

Verified (on Tegra210) that we are able to dynamically map NS memory, access it using the VA and then unmap it.

As a negative test, I also tried to access the unmapped NS memory using its VA and saw an abort. So this PR does what it is supposed to.

Please go ahead with the merge.

antonio-nino-diaz-arm:
Thanks for testing this, @vwadekar

lib/xlat_tables_v2/xlat_tables_internal.c
```diff
@@ -0,0 +1,1101 @@
...
+
+	desc = addr_pa;
+	/*
+	 * There are different translation table descriptors for level 3 and the
+	 * rest.
+	 */
+	desc |= (level == XLAT_TABLE_LEVEL_MAX) ? PAGE_DESC : BLOCK_DESC;
+	/*
+	 * Deduce other fields of the descriptor based on MT_NS, MT_RW and
+	 * ACCESS_FLAG attributes.
```
sandrine-bailleux-arm: Actually, we always set the access flag unconditionally, it is not deduced from any MT_* attribute so this comment is misleading. I'd suggest rewording, something like:

```
Always set the access flag, as TF doesn't manage access flag faults.
Deduce other fields of the descriptor based on the MT_NS and MT_RW memory region attributes.
```

antonio-nino-diaz-arm: Done.

vwadekar:
@antonio-nino-diaz-arm 

Hi, when can we expect this code to be merged?

davidcunado-arm:
@vwadekar 
It's in!

PR #857, opened 2017-03-01
ARM Cortex-A53 erratum 855873 workaround

Andre-ARM:
Enable a firmware level workaround for the ARM Cortex-A53 erratum 855873 by setting a bit in the ACTLR register. Applies to revisions r3 and *higher* of the core. The workaround gets enabled for all platforms known to have an affected core.


davidcunado-arm:
jenkins: test this please

danh-arm:
jenkins: add to whitelist

docs/cpu-specific-build-macros.md
```diff
@@ -60,6 +60,12 @@ For Cortex-A53, following errata build flags are defined :
      CPU. This needs to be enabled only for revision <= r0p3 of the CPU. From
      r0p4 and onwards, this errata is enabled by default.
 
+*   `ERRATA_A53_855873`: This applies errata 855873 workaround to Cortex-A53
+     CPUs. Though the erratum is present in every revision of the CPU,
+     this workaround is only applied to CPUs from r0p3 onwards, which feature
+     a chicken bit in CPUACTLR_EL1 to enable a hardware workaround.
+     Earlier revisions of the CPU are meant to be patched by the rich OS.
```
nmenon: Noob here: what is rich OS? only other usage seems to be in plat/xilinx/zynqmp/pm_service/pm_svc_main.c -> did we mean Secure OS like OPTEE?

Andre-ARM: "Rich OS" is the generic term for a general purpose OS like Linux, as opposed to a secure OS which is meant to be lean to have a smaller attack surface.
Originally I had "Linux" in there, but was asked to change this to "rich OS" to be more generic (as it could be *BSD as well, for instance).

nmenon: but this could be anything in "non secure", right? -> now that you explain and noticing EL1 register change, it makes sense. but I think might be good to have some explanation in the documentation - something like "rich OS (e.g. Linux)" or the like for Noobs like me.

Andre-ARM: What I meant to say is that earlier revision of the core suffer from other errata, which need an equivalent software workaround in _every_ code that is running, possibly by patching code. Linux handles this like this: https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/commit/?id=301bcfac42897dbd1b0b3c1be49f24654a1bc49e
Cortex-A53 r0p3 fixed those errata, also introduced a chicken bit to avoid code patching in the future, just in case. As it turns out we can now use this feature to fix the (new) erratum 855873.

nmenon: @Andre-ARM got it. thanks. makes sense.

davidcunado-arm:
@Andre-ARM 
Can you rebase you PR and re-submit, please?

PR #860, opened 2017-03-02
Patches for platforms with hardware-assisted coherency

jeenu-arm:


docs/user-guide.md
```diff
@@ -328,6 +328,15 @@ performed.
 *   `HANDLE_EA_EL3_FIRST`: When defined External Aborts and SError Interrupts
     will be always trapped in EL3 i.e. in BL31 at runtime.
 
+*   `HW_ASSISTED_COHERENCY`: On most ARM systems to-date, platform-specific
+    software operations are required for CPUs to enter and exit coherency.
+    However, there exists newer systems where CPUs' entry to and exit from
+    coherency is managed in hardware. Such systems require software to only
```
nmenon: is this implying global monitory past the cluster providing coherence across clusters?

jeenu-arm: (I had left a comment earlier, but it seemed that I started a review in Github, rather than publishing the comment itself. So I'm not sure if my earlier reply was publicly visible; I re-compose my earlier reply).

Without being specific, this only implies that, as soon a CPU is out of reset and enter PSCI state coordination, it's coherent with all other CPUs in the system, and is "ready to go".

Typically in multi-cluster ARM systems, the clusters are connected via. CCI or similar IP to keep all clusters coherent. This means that, before CPUs in those clusters can be assumed to be coherent, the CCI needs programming. Because of this, TF presently delays assuming (a sub-optimal but necessary step to ensure correctness) them to be coherent until the IP is programmed.

However, with this build option, TF is told that all CPUs in the system are coherent with each other as soon as they're out of reset, without needing programming of any IP. I.e. regardless of which cluster they're in, they all _effectively_ belong to one and same coherency domain.

nmenon: @jeenu-arm  thanks. very interesting. I might have expected coherency to be active only after MMUs are active. I am guessing the patch is due to sizeable performance advantage by skipping the syncs off. if PTE driven coherency, then initial stages of cache invalidations might still be valid.

but thanks for the explanation. will keep a watch.

lib/psci/psci_off.c
```diff
@@ -119,10 +119,9 @@ int psci_do_cpu_off(unsigned int end_pwrlvl)
 #endif
 
 	/*
-	 * Arch. management. Perform the necessary steps to flush all
-	 * cpu caches.
+	 * Arch. management. Initiate power down sequence.
 	 */
-	psci_do_pwrdown_cache_maintenance(psci_find_max_off_lvl(&state_info));
```
nmenon: what if there is a level 3 cache at this point? do we just skip the flush down to memory prior to powering off?

lib/psci/psci_suspend.c
```diff
@@ -121,13 +121,11 @@ static void psci_suspend_to_pwrdown_start(unsigned int end_pwrlvl,
...
 	/*
-	 * Arch. management. Perform the necessary steps to flush all
-	 * cpu caches. Currently we assume that the power level correspond
-	 * the cache level.
+	 * Arch. management. Initiate power down sequence.
 	 * TODO : Introduce a mechanism to query the cache level to flush
 	 * and the cpu-ops power down to perform from the platform.
 	 */
-	psci_do_pwrdown_cache_maintenance(max_off_lvl);
+	psci_do_pwrdown_sequence(max_off_lvl);
```
nmenon: if we do have a level 3 cache and we would reach low power state where the sub system is completely powered off, the stale data in level3 will be lost, no? explicit flush out might be needed?

lib/psci/psci_private.h
```diff
@@ -38,6 +38,29 @@
 #include <psci.h>
 #include <spinlock.h>
 
+#if HW_ASSISTED_COHERENCY
+/*
+ * On systems with hardware-assisted coherency, make PSCI cache operations NOP,
+ * as PSCI participants are cache-coherent, and there's no need for explicit
+ * cache maintenance operations or barriers to coordinate their state.
+ */
+#define psci_flush_dcache_range(addr, size)
```
nmenon: #define psci_flush_dcache_range(...)

just in case the cache apis change?

OR we should just make it static inline function to ensure that the types are forced

static inline ... psci_flush_dcache_range (.. a, .. b)
{
#ifndef HW_ASSISTED_COHERENCY
flush_dcache_range(a, b);
#endif 
}

I am no good at github reviews.. so just trying to illustrate.

Makefile
```diff
@@ -305,6 +305,11 @@ ifeq (${ARCH},aarch32)
     endif
 endif
 
+# When building for systems with hardware-assisted coherency, there's no need to
+# use USE_COHERENT_MEM. Require that USE_COHERENT_MEM must be set to 0 too.
+ifeq ($(HW_ASSISTED_COHERENCY)-$(USE_COHERENT_MEM),1-1)
+$(error USE_COHERENT_MEM cannot be enabled with HW_ASSISTED_COHERENCY)
```
nmenon: I think the message here should probably be: USE_COHERENT_MEM is redundant given HW_ASSISTED_COHERENCY is defined, use one.

lib/psci/psci_off.c
soby-mathew: @nmenon, This is a problem that needs to be solved between the CPU- Operations framework and the platform. If you take a look at the CPU- operations that are currently implemented, all of the CPUs only implement till Level 2 cache and hence they flush only till Level 2 in their cluster power down sequence. So , in you case, if the Level 3 cache is part of the CPU Architecture, then you would do the cache flush for the same in the corresponding CPU-operations for that CPU.

Otherwise, if the additional level of Cache is part of the platform (eg: system caches in interconnect), then you would do the L3 cache flush in the platform hooks for the particular platform.

Hope that helps.

lib/psci/psci_suspend.c
soby-mathew: Same answer as my comments above.

vwadekar:
@jeenu-arm 

Hi, what is the plan for this PR?

davidcunado-arm:
@vwadekar 
We're planning on merging, pending whether any documentation / comment updates would help clarify.

lib/psci/psci_private.h
soby-mathew: Sorry, this comment got masked in the PR. IMO, this macro is a simple wrapper to the underlying function with both the parameters being passed directly to the underlying function. Hence, in case HW_ASSISTED_COHERENCY is defined, the function prototype of flush_dache_range ensures the type checking of the parameters to this macro. When HW_ASSISTED_COHERENCY is not defined, it is a NOP and it doesnt matter anyway. Hence I think the macro achieves the types enforcement when it matters.

Makefile
soby-mathew: > I think the message here should probably be: USE_COHERENT_MEM is redundant given HW_ASSISTED_COHERENCY is defined, use one.

Although in effect what you say is true, it sounds a bit vague to me.  Since this is a build error, IMO it is better to give a clear message that the 2 options are incompatible.  Hence the message.

PR #867, opened 2017-03-16
Add io_retry mechanism for boot redundancy

sbranden:
Add io_retry mechanism to io_storage.
With io_retry in place BL1 and BL2 can both search for the FIP image
in multiple locations.  And, the io_memmap device knows where to look
for the image based on the retry count.
There doesn't seem to be a way to cleanly pass a retry count between the
various components in the current ATF software architecture otherwise.

### DIFF
--- a/bl1/bl1_main.c
+++ b/bl1/bl1_main.c
@@ -42,6 +42,7 @@
 #include <utils.h>
 #include "bl1_private.h"
 #include <uuid.h>
+#include <io_storage.h>
 
 /* BL1 Service UUID */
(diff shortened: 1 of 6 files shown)


sbranden:
Here is the boot redundancy we have in place based on the limitations of the io_fip and io_storage architectures.   Perhaps someone else has some ideas on how to better abstract the mechanism or this is sufficient.  We have added additional io_ drivers to load from other boot sources (eMMC, UART, NAND, SPI) utilizing this code as is.

PR #868, opened 2017-03-16
Add io_retry mechanism for boot redundancy

sbranden:
Add io_retry mechanism to io_storage.
With io_retry in place BL1 and BL2 can both search for the FIP image
in multiple locations.  And, the io_memmap device knows where to look
for the image based on the retry count.
There doesn't seem to be a way to cleanly pass a retry count between the
various components in the current ATF software architecture otherwise.

### DIFF
--- a/bl1/bl1_main.c
+++ b/bl1/bl1_main.c
@@ -42,6 +42,7 @@
 #include <utils.h>
 #include "bl1_private.h"
 #include <uuid.h>
+#include <io_storage.h>
 
 /* BL1 Service UUID */
 DEFINE_SVC_UUID(bl1_svc_uid,
@@ -198,17 +199,18 @@ void bl1_load_bl2(void)
 
 	INFO("BL1: Loading BL2\n");
 
+	do {
 #if LOAD_IMAGE_V2
-	err = load_auth_image(BL2_IMAGE_ID, image_info);
+		err = load_auth_image(BL2_IMAGE_ID, image_info);
 #else
-	/* Load the BL2 image */
-	err = load_auth_image(bl1_tzram_layout,
-			 BL2_IMAGE_ID,
-			 image_info->image_base,
-			 image_info,
-			 ep_info);
-
-#endif /* LOAD_IMAGE_V2 */
+		/* Load the BL2 image */
+		err = load_auth_image(bl1_tzram_layout,
+				      BL2_IMAGE_ID,
+				      image_info->image_base,
+				      image_info,
+				      ep_info);
+#endif
+	} while (err && io_retry() > 0);
 
 	if (err) {
 		ERROR("Failed to load BL2 firmware.\n");
--- a/bl2/bl2_image_load.c
+++ b/bl2/bl2_image_load.c
@@ -35,6 +35,7 @@
 #include <bl_common.h>
 #include <debug.h>
 #include <errno.h>
+#include <io_storage.h>
 #include <platform.h>
 #include <platform_def.h>
 #include <stdint.h>
@@ -212,7 +213,10 @@ entry_point_info_t *bl2_load_images(void)
 	entry_point_info_t *bl31_ep_info;
 	int e;
 
-	e = load_scp_bl2();
+	do {
+		e = load_scp_bl2();
+	} while (e && io_retry() > 0);
+
 	if (e) {
 		ERROR("Failed to load SCP_BL2 (%i)\n", e);
 		plat_error_handler(e);
@@ -244,7 +248,10 @@ entry_point_info_t *bl2_load_images(void)
 	bl31_ep_info->args.arg0 = (unsigned long) bl2_to_bl31_params;
 	bl2_plat_set_bl31_ep_info(NULL, bl31_ep_info);
 #else
-	e = load_bl31(bl2_to_bl31_params, bl31_ep_info);
+	do {
+		e = load_bl31(bl2_to_bl31_params, bl31_ep_info);
+	} while (e && io_retry() > 0);
+
 	if (e) {
 		ERROR("Failed to load BL31 (%i)\n", e);
 		plat_error_handler(e);
--- a/drivers/io/io_fip.c
+++ b/drivers/io/io_fip.c
@@ -188,11 +188,6 @@ static int fip_dev_init(io_dev_info_t *dev_info, const uintptr_t init_params)
 static int fip_dev_close(io_dev_info_t *dev_info)
 {
 	/* TODO: Consider tracking open files and cleaning them up here */
-
-	/* Clear the backend. */
-	backend_dev_handle = (uintptr_t)NULL;
-	backend_image_spec = (uintptr_t)NULL;
-
 	return 0;
 }
 
--- a/drivers/io/io_memmap.c
+++ b/drivers/io/io_memmap.c
@@ -32,9 +32,14 @@
 #include <debug.h>
 #include <io_driver.h>
 #include <io_storage.h>
+#include <platform_def.h>
 #include <string.h>
 #include <utils.h>
 
+#ifndef PLAT_FIP_ATTEMPT_OFFSET
+/* Attempt to find the FIP every 0x20000 bytes */
+#define PLAT_FIP_ATTEMPT_OFFSET		0x20000
+#endif
 /* As we need to be able to keep state for seek, only one file can be open
  * at a time. Make this a structure and point to the entity->info. When we
  * can malloc memory we can change this to support more open files.
@@ -50,6 +55,7 @@ typedef struct {
 } file_state_t;
 
 static file_state_t current_file = {0};
+static int last_num_retries = -1;
 
 /* Identify the device type as memmap */
 io_type_t device_type_memmap(void)
@@ -123,7 +129,13 @@ static int memmap_block_open(io_dev_info_t *dev_info, const uintptr_t spec,
 			     io_entity_t *entity)
 {
 	int result = -ENOMEM;
+	int num_retries;
 	const io_block_spec_t *block_spec = (io_block_spec_t *)spec;
+
+	num_retries = io_get_num_retries();
+
+	if (num_retries < 0)
+		return result;
 
 	/* Since we need to track open state for seek() we only allow one open
 	 * spec at a time. When we have dynamic memory we can malloc and set
@@ -134,7 +146,8 @@ static int memmap_block_open(io_dev_info_t *dev_info, const uintptr_t spec,
 		assert(entity != NULL);
 
 		current_file.in_use = 1;
-		current_file.base = block_spec->offset;
+		current_file.base = (block_spec->offset & 0xFFFFFFFFFF000000) +
+				     num_retries * PLAT_FIP_ATTEMPT_OFFSET;
 		/* File cursor offset for seek and incremental reads etc. */
 		current_file.file_pos = 0;
 		current_file.size = block_spec->length;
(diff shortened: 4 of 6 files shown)


davidcunado-arm:
@sbranden 
Could you elaborate on what issue this patch is looking to address? Why would a retry mechanism be needed?
Is this something that can be addressed at the lower-levels, transparently to  users of io_storage?

sbranden:
Details are in tf_issue_383.  We need a method to support boot redundancy and alternate boot sources.  In tf_issue_383 it was indicated by ARM that this is not something they were planning on working on.  But, it is a necessary feature to have fail safe boot on real devices and support attempting to boot from various boot sources.  Unfortunately ATF code doesn't have a simply method to support this.  And, the io_storage and authentication method is tightly coded in a complex way.  I think I would have to rewrite the ATF code quite a bit to do it the way I would be happy with.  So, the solution I came up with is this patchset.  A simple loop is added to the upper layer and the low level io_storage driver(s) perform a different operation based on which retry count they are doing.  By doing so I did not need to touch the rest of the code and kept remaining functionality intact.  We've tested and it provides all the functionality we currently need to boot from various sources (SRAM, NAND, SPI, eMMC, UART, USB).

danh-arm:
Hi @sbranden

I think we need to understand your requirements a bit better here. On the face of it, this functionality appears to be at wrong level of abstraction.

If there is an image load failure in the middle of cold boot (for whatever reason), then it does not seem safe to me to continue, since there may be a mismatch between the previously loaded (and perhaps executed) images and the images from the retry location (for example if they were different firmware versions). This could potentially be leveraged by an attacker, for example to bypass the rollback mechanism, which may only be enforced by root certificates.

To achieve retry behaviour, I would expect your implementation of `plat_error_handler()` to increment the persistent "retry register" and then reboot. Your implementation of `plat_get_image_source()` would then use the "retry register" value to select a different image source. This wouldn't require any changes in generic code.

Alternatively (or as well as), you could do some simple probing in `plat_get_image_source()` to detect whether the boot source is valid before trying the next source. We do this in the ARM FVP platform port, where we first look for the image in flash and then fall back to loading via semihosting (see `plat_arm_get_alt_image_source()` in `plat/arm/board/fvp/fvp_io_storage.c`.

Let us know if this doesn't work for you.

Regards

Dan.

sbranden:
Hi @danh-arm,

The requirement is to support boot redundancy.  The current ATF assumption there is only one valid location for an image and is not how a reliable boot ROM should operate.  If the first location in flash does not contain a valid image then additional locations need to be searched.  This needs to be done without resetting the device.

Various boot sources might be selected at run time (not compile time) via platform specific means: straps, OTP, registers, etc).
For SRAM Memory directly this may involve only searching one location.
For NAND, this would involve scan multiple blocks of NAND.
For SPI, a similar, yet different, search pattern may be performed.
For eMMC, this means checking BP1 and BP2, possibly user partition(s), at various locations. 
For SD Card FAT partitions this may involve search for different FAT filenames or partitions
For USB mass storage devices this may involve similar operations as SD Cards
For UART transfers this may involve a retry.
Also, some vendors may wish to use a GPT, some may not.

As you can see - the number of locations tried depends on the boot source selected and is platform specific.  Different vendors may wish to have their io_retry mechanism implemented differently.   Also, order of checking may be different depending on vendor implementation or other information passed to boot rom on boot - it is vendor dependant.

I don't see any security hole you are mentioning - if one image is not valid we search for another.  Pretty simple.

Yes, I don't like the level of abstraction that was needed but due to the overly-complex ATF io_storage implementation this was the simplest mechanism I could add.  But, we managed to implement and it works quite flawlessly right now.  I fear bringing all of this up here will just create something more complicated than needs to be and will now break what we already have working.

danh-arm:
Hi @sbranden 

I'm not questioning that platforms need retry support and that these mechanisms are very platform specific. I'm just questioning what support and platform hooks are needed in generic code.

The platform function `plat_get_image_source()` seems to cover the main requirement to me. During the first invocation of this function, you can iterate through the various boot sources, attempting to open the first image. If the "open" fails, you can try the next source until a valid one is found. Subsequent image loads can then select the same source.

In the rarer case that an image is corrupted/invalid, the cost of a reboot seems small since we're so early in the boot sequence. Am I underestimating the likely performance cost of this on your platforms?

I'm still uneasy about images being loaded from multiple sources during the same boot as it seems to open up additional attack vectors, although after looking again at the rollback implementation, I think my specific concern there is unfounded. The alternative is to start the BL's whole image load process again when one image fails to load, although that creates other issues around cleanup of previously loaded images.

Your implementation retains the retry value between each image load. However, it doesn't seem to cater for the case where earlier boot sources are valid for later image loads (i.e. the retry value only goes forward). Wouldn't a better implementation try the current retry value, then start from the first source again on failure (skipping the one it's already tried)?

Once we've agreed what's needed, I suspect it's going to be easier for us to implement this our side as we've got more understanding of the IO framework and image loading code. I think there should be fewer changes in BL main code and more in the BL common loading code.

Regards

Dan.

sbranden:
"The platform function plat_get_image_source() seems to cover the main requirement to me. During the first invocation of this function, you can iterate through the various boot sources, attempting to open the first image. If the "open" fails, you can try the next source until a valid one is found. Subsequent image loads can then select the same source."

Above might work, if we have a mechanism to load from the same source at a different location.  What call do we make to do so?

"In the rarer case that an image is corrupted/invalid, the cost of a reboot seems small since we're so early in the boot sequence. Am I underestimating the likely performance cost of this on your platforms?"

Unfortunately boot times are critical in one of our use cases.  We must fully initialize some PHYs on board power up under tight timing constraints.  Simply issuing a reset one more time will take too much time.

"I'm still uneasy about images being loaded from multiple sources during the same boot as it seems to open up additional attack vectors, although after looking again at the rollback implementation, I think my specific concern there is unfounded. The alternative is to start the BL's whole image load process again when one image fails to load, although that creates other issues around cleanup of previously loaded images."

Yes, I would like the whole image load process to start again when one image load fails but current codebase does not allow it.  It would be great if it did.

"Your implementation retains the retry value between each image load. However, it doesn't seem to cater for the case where earlier boot sources are valid for later image loads (i.e. the retry value only goes forward). Wouldn't a better implementation try the current retry value, then start from the first source again on failure (skipping the one it's already tried)?"

No, starting from first source again on failure is no good.  Starting the entire image load process when one fails (at the next retry location) is what really should be done.

"Once we've agreed what's needed, I suspect it's going to be easier for us to implement this our side as we've got more understanding of the IO framework and image loading code. I think there should be fewer changes in BL main code and more in the BL common loading code."

I agree it will likely be easier for you to implement.  But you didn't want to previously based on the issue I filed.

danh-arm:
Hi @sbranden 

> Above might work, if we have a mechanism to load from the same source at a different location. What call do we make to do so?

You would open the image using a different image spec, for example `io_open()` on a memmap device using an alternative `io_block_spec_t`. The `arm_io_storage.c` file has an example of this.

> Unfortunately boot times are critical in one of our use cases.

OK, understood.

> Yes, I would like the whole image load process to start again when one image load fails but current codebase does not allow it. It would be great if it did.

OK, I think we should pursue this idea.

> No, starting from first source again on failure is no good.

I agree this is probably over-flexibility and doesn't really solve the requirement properly.

> I agree it will likely be easier for you to implement. But you didn't want to previously based on the issue I filed.

I didn't really understand then what you wanted from generic code. I've got a better idea now. We will discuss more internally and let you know what we plan to do.

Regards

Dan.

danh-arm:
We had a discussion about this and some new things came up...

If an image in the middle of the list fails to load and we start loading the images again, there needs to be some cleanup of the images already loaded. As a minimum, this means resetting internal state but there may need to be some image specific cleanup (e.g. undoing of bl2_plat_handle_post_image_load()). This could be managed via another platform function. although if this cleanup failed there would be no choice but to reset.

Some cold boot operations can only be done once. For example, on platforms with an SCP_BL2, the running SCP image cannot be loaded onto the SCP again without resetting. Another example is the TZC init that occurs during `bl2_platform_setup()` on ARM platforms. So any failure after this point would have to restart from just after `bl2_platform_setup()`, resulting in image loads from different sources, which is what we wanted to avoid.

With TBBR enabled, loading an image may implicitly involve loading several parent images (e.g. certificates). The Chain of Trust design involve a tree structure up to the Root of Trust. Partially rolling back the Chain of Trust state on image load failure looks difficult and error prone.

All in all, this looks like a lot of complexity so I'm keen to exhaust simpler options first. I'm going to make some assertions in the following so please correct me if I'm wrong...

It seems to me that once the first image has successfully loaded from the first valid boot source, any failure after this point is highly unlikely, and if there is one, calling `plat_error_handler()` is reasonable. I notice in your current patch that BL2 only retries loading SCP_BL2 and BL31, and not the later images.

The first image that BL1 loads from flash is BL2 (or its certificates) so if my previous assertion is correct, this redundancy support is only really needed in BL1 (or the first image in BL2 for platforms that don't use TF BL1). 

As stated before, overriding `plat_get_image_source()` allows simple probing to find a valid boot source. Adding a retry wrapper around the first image load would ensure full authentication of the first image from a valid boot source, but in practice would this really be significantly more robust than simple image probing to find the first valid boot source?

If load failures after the first successful image load do need to be mitigated, then I think I'd prefer to try to keep going using a new boot source source (as this PR currently does), due to the complexity of trying to cleanup and start again. However, if we go that route I think it would be better (as previously stated) to allow platforms to go through all boot sources rather than just the sources after the current one. It could be left to the platform to choose the next boot source to try (if any).

sbranden:
There is a solution to the failures in the middle of loading and my current implementation:  using redundancy.  I think it might be the simplest way to continue to use thinking about it more now. 

You simply program the same fip image more than once.  Then, if first portion of image is found to be valid it is loaded.  Then if next portion of image is invalid(corrupted) io_retry is called and next copy of image is loaded.

For paranoid users we have they actually program 4 copies of the image.  Then 4 copies of a recovery image.  It's pretty hard to brick the device due to flash errors due to decay or power losses in the middle of flash upgrades.  As much or as little boot redundancy as desired can be used.

sbranden:
For a complete different boot source we added code to select the next boot sources (not in this commit).
First source and second boot sources are selected via platform specific mechanisms.
When switching boot sources you need to open a new device.  io_retry is used on the current device only and you don't open the device every time - and it is device specific how many retries you wish to do before going to (possible) next boot source.

sbranden:
"I notice in your current patch that BL2 only retries loading SCP_BL2 and BL31, and not the later images."

This is simply because those are the only images we currently load.  If we need to load the any other images in the fip we will need to add the io_retry to those as well.

danh-arm:
OK, I see the main thing you're trying to add is redundancy within a single boot source due to for example, flash corruption. However, I don't think this should be done at the IO device level; it should be done at the `load_auth_image()` level. Your implementation would not detect small corruptions, where the file is read correctly but then fails authentication.

I think a platform specific retry mechanism can be implemented around `load_auth_image()`. The generic code would then not care whether the retry comes from the same or a different boot source. In the former case, your plat_get_image_source() would need to select a different image spec depending on the value of the platform specific retry value.

I'll get something along these lines scheduled and then you can see if this meets your needs.

sbranden:
Current patchset actually meets all my needs.  But I do agree it is not
clean and layered properly.  I worked with what was present in current
design.

danh-arm:
Hi @sbranden. We finally got around to implementing what we think is a cleaner way to support this boot redundancy use-case:
https://github.com/ARM-software/arm-trusted-firmware/pull/1141

Any feedback on that would be appreciated.

davidcunado-arm:
Closing as #1141 has been merged.

PR #869, opened 2017-03-17
Allow platform specific ERROR, NOTICE, WARNING, INFO, VERBOSE messages

sbranden:
Allow ERROR, NOTICE, WARNING, INFO, and VERBOSE
messages to be implemented on a platform specific basis.

Default remains current implementation.

Fixes ARM-software/tf-issues#462

### DIFF
--- a/docs/porting-guide.md
+++ b/docs/porting-guide.md
@@ -203,6 +203,14 @@ platform port to define additional platform porting constants in
     Currently, this macro is used by the Generic PSCI implementation to size
     the array used for PSCI_STAT_COUNT/RESIDENCY accounting.
 
+*   **#define : PLAT_LOG_MACROS **
+
+    Override the default log macros with platform specific logging mechanism.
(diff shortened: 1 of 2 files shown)


sbranden:
Updated commit to add info in porting-guide for new platform specific log macro overrides

sorenb-xlnx:
In https://github.com/ARM-software/tf-issues/issues/462, you mentioned that you also want to make the prints to be switchable at run-time. Since the sole reason for this seems to be to reduce printed characters, wouldn't the run-time switch make this PR obsolete?

davidcunado-arm:
jenkins: test this please

sbranden:
@sorenb-xlnx, please have a look inside this commit and you will see that the log functions can be replaced at the platform level.  So, whatever a platform wishes (run-time, shorten prints, etc) can be done at the platform level.

soby-mathew:
👎  My comments are mentioned here: https://github.com/ARM-software/tf-issues/issues/462

PR #873, opened 2017-03-20
Move plat/common source file definitions to generic Makefiles

dp-arm:
These source file definitions should be defined in generic
Makefiles so that all platforms can benefit. Ensure that the
symbols are properly marked as weak so they can be overridden
by platforms.

NOTE: This change is a potential compatibility break for
non-upstream platforms.

Change-Id: I7b892efa9f2d6d216931360dc6c436e1d10cffed


vwadekar:
@dp-arm 

Looks good for Tegra

davidcunado-arm:
@TonyXie06 
Ok with the changes to rockchip platforms?

TonyXie06:
It's OK for us.

PR #874, opened 2017-03-22
mbedtls: Namespace TF specific macros

dp-arm:
These macros are not part of mbed TLS so they should not be prefixed
with `MBEDTLS_` to avoid potential collision in the future. Use the
`TBBR_` suffix to highlight that they only used in TF.

`MBEDTLS_KEY_ALG` was not modified because that is documented and used
by platforms to select the key algorithm.

Change-Id: Ief224681715c481691c80810501830ce16e210b0


danh-arm:
Regarding:

`MBEDTLS_KEY_ALG was not modified because that is documented and used by platforms to select the key algorithm.`

One way around this would be to supersede this with TBBR_KEY_ALG but provide some compatibility by setting this to MBEDTLS_KEY_ALG (and printing a warning) if MBEDTLS_KEY_ALG is defined and !ERROR_DEPRECATED. Then we could remove this code in a future non-compatible release.

PR #875, opened 2017-03-23
Tegra186 platform support v2

vwadekar:
Tegra186 platform changes from downstream


davidcunado-arm:
jenkins: test this please

davidcunado-arm:
jenkins: test this please

plat/nvidia/tegra/soc/t186/drivers/smmu/smmu.c
```diff
@@ -438,18 +444,24 @@ void tegra_smmu_save_context(void)
 	assert(!((pgshift != PGSHIFT) || (cb_size != CB_SIZE)));
 #endif
 
+	assert((smmu_ctx_addr >= tzdram_base) && (smmu_ctx_addr <= tzdram_end));
```
davidcunado-arm: @vwadekar 
Ahead of merging, can you check that the conditions in this assert are what you intended? If reading correctly, I think that the second check should consider the size of the smmu_ctx_regs array.
Assuming this needs to be corrected, let me know if you want to update this PR or submit a separate PR to address just this?

vwadekar: @davidcunado-arm 

We just want to make sure that the address being used to store the context is a valid TZDRAM address. But I see what you are saying. I will try to pass the destination size to the smmu function as well. Need some time to figure out, so will push another PR to fix this.

davidcunado-arm: @vwadekar 
Ok, I'll merge the PR as is.

PR #879, opened 2017-03-27
ARM platforms: Add support for MT bit in MPIDR

Summer-ARM:
This patch modifies some of the functions in ARM platform layer to cater
for the case when multi-threading `MT` is set in MPIDR. A new build flag
`ARM_PLAT_MT` is added, and when enabled, the functions accessing MPIDR
now assume that the `MT` bit is set for the platform and access the bit
fields accordingly.

Also, a new API plat_arm_get_cpu_pe_count is added when `ARM_PLAT_MT` is
enabled, returning the PE count within the physical cpu corresponding to
`mpidr`.

Change-Id: I04ccf212ac3054a60882761f4087bae299af13cb
Signed-off-by: Summer Qin <summer.qin@arm.com>

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -197,6 +197,12 @@ performed.
     by the interrupt management framework. Default is 2 (that is, version 2.0).
     This build option is deprecated.
 
+*   `ARM_PLAT_MT`: This flag determines whether the ARM platform layer has to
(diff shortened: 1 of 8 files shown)


davidcunado-arm:
jenkins: test this please

davidcunado-arm:
jenkins: add to whitelist

PR #883, opened 2017-03-30
Allow platform specific ERROR, NOTICE, WARNING, INFO, VERBOSE messages

sbranden:
Allow ERROR, NOTICE, WARNING, INFO, and VERBOSE
messages to be implemented on a platform specific basis.

Default remains current implementation.

Fixes ARM-software/tf-issues#462

### DIFF
--- a/docs/porting-guide.md
+++ b/docs/porting-guide.md
@@ -203,6 +203,14 @@ platform port to define additional platform porting constants in
     Currently, this macro is used by the Generic PSCI implementation to size
     the array used for PSCI_STAT_COUNT/RESIDENCY accounting.
 
+*   **#define : PLAT_LOG_MACROS **
+
+    Override the default log macros with platform specific logging mechanism.
+    See default_log_macros.h for details on macro functionality to replace.
+    To override, define PLAT_LOG_MACROS and create plat_log_macros.h.
+    For template, copy default_log_macros.h to your platform specific include
+    path and rename to plat_log_macros.h, then edit log macros.
+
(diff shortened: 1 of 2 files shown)


danh-arm:
The solution here isn't exactly what I had in mind in [my comment here](https://github.com/ARM-software/tf-issues/issues/462#issuecomment-289762576). I was thinking the log definitions in debug.h would change to something like:

```
#if LOG_LEVEL >= LOG_LEVEL_NOTICE
# ifdef PLAT_LOG
#  define NOTICE(...)	PLAT_LOG(LOG_LEVEL_NOTICE, __VA_ARGS__)
# else
#  define NOTICE(...)	tf_printf("NOTICE:  " __VA_ARGS__)
# endif
#else
# define NOTICE(...)
#endif
```

Then the platform definition might look something like:

```
#define LOG_PREFIX_LOG_LEVEL_NOTICE "N:     "
#define PLAT_LOG(_log_level, ...) my_printf(LOG_PREFIX_##_log_level __VA_ARGS__)
```

However, I'm now thinking it would be more future proof to change all `tf_printf()` calls across the codebase into `plat_printf()`, and add a new"#define PLAT_LOG_PREFIX(_log_level)". We should also fix the code size issue with the repeated prefix strings at the same time. This is a more invasive change, which I suggest we tackle our side. I'll get this scheduled.

sbranden:
Yes, plat_printf will be needed with your implementation.

davidcunado-arm:
Closing as #1093 has now been merged which supersedes this patch.

PR #890, opened 2017-04-05
Build: add generic way to include SCP_BL2 into FIP image

masahir0y:
If SCP_BL2 is passed in from the command line, it is recognized by
make_helpers/tbbr/tbbr_tools.mk, and the cert_create tool generates
the corresponding key and content certificates.

On the other hand, the top-level Makefile does not care SCP_BL2, so
the --scp-fw option is not passed to the fiptool.  As far as I see
plat/arm/css/common/css_common.mk, it looks like a platform's job to
add $(eval $(call FIP_ADD_IMG,SCP_BL2,--scp-fw)).

We can make the top-level Makefile kind enough to take care of it.
This is useful when we want to have optional SCP_BL2 firmware.

Adjust css_common.mk so that Juno still requires SCP_BL2 by default.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -341,6 +341,11 @@ ifdef BL2_SOURCES
         endif
 endif
 
+# If SCP_BL2 is given, we always want FIP to include it.
+ifdef SCP_BL2
+        NEED_SCP_BL2		:=	yes
+endif
+
 # Process TBB related flags
 ifneq (${GENERATE_COT},0)
         # Common cert_create options
@@ -534,6 +539,10 @@ endif
 ifeq (${NEED_BL2},yes)
 $(if ${BL2}, $(eval $(call MAKE_TOOL_ARGS,2,${BL2},tb-fw)),\
 	$(eval $(call MAKE_BL,2,tb-fw)))
+endif
+
+ifeq (${NEED_SCP_BL2},yes)
+$(eval $(call FIP_ADD_IMG,SCP_BL2,--scp-fw))
 endif
 
 ifeq (${NEED_BL31},yes)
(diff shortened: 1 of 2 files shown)


danh-arm:
I can see the benefit here if your (non-ARM) platform has an SCP_BL2. Is that case?

I'm a bit concerned about increasing the number of images that the top level Makefile is aware of, and if we do, where do we stop. For example, if we're adding support for SCP_BL2, I guess we should also add support for NS_BL2U and SCP_BL2U in the FWU FIP?

In general it would be better if the top level Makefile didn't have all this BL specific functionality, but we are where are so I'm OK with this change in principle.

masahir0y:
> I can see the benefit here if your (non-ARM) platform has an SCP_BL2. Is that case?

Yes.  My platform has two modes in terms of system power control.
[1] The main SoC runs by itself  (no SCP)
[2] The main SoC runs under control of SCP

For [1], PSCI support is limited.  (For example, SYSTEM_OFF is not supported.)
But, keeping SCP optional is still useful because SCP firmware is developed by other people.

>For example, if we're adding support for SCP_BL2, I guess we should also add support for NS_BL2U and SCP_BL2U in the FWU FIP?

If somebody needs those.

>In general it would be better if the top level Makefile didn't have all this BL specific functionality, but we are where are so I'm OK with this change in principle.

I notice the Makefile is too cluttered.  At least, NEED_* is the same pattern, so I think it should be OK.

danh-arm:
OK, I guess support for the other images can be added when others need it.

davidcunado-arm:
jenkins: test this please

davidcunado-arm:
jenkins: test this please

PR #892, opened 2017-04-06
rockchip/rk3399: the printf changed to tf_printf for console output

rkchrome:
The printf() isn't used by the firmware itself, just by the tools under
the ./tools/ folder.

Change-Id: Ie699ccb54a5be9a2cbbd7b8d4193b57075a2f57a
Signed-off-by: Caesar Wang <wxt@rock-chips.com>

PR #893, opened 2017-04-06
Replace tf_printf occurrences with ERROR

antonio-nino-diaz-arm:
The amount of console output is controlled by the LOG_LEVEL build
option. Using tf_printf without any #ifdef depending on the LOG_LEVEL
doesn't give the user that flexibility.

This patch replaces all occurrences of tf_printf that prints error, but
aren't dependent on LOG_LEVEL, with the ERROR macro.

Change-Id: Ib5147f14fc1579398a11f19ddd0e840ff6692831
Signed-off-by: Antonio Nino Diaz <antonio.ninodiaz@arm.com>

### DIFF
--- a/drivers/arm/cci/cci.c
+++ b/drivers/arm/cci/cci.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015, ARM Limited and Contributors. All rights reserved.
+ * Copyright (c) 2015-2017, ARM Limited and Contributors. All rights reserved.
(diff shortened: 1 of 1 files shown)


antonio-nino-diaz-arm:
Addresses the problems in https://github.com/ARM-software/tf-issues/issues/467 that https://github.com/ARM-software/arm-trusted-firmware/pull/892 doesn't fix.

PR #896, opened 2017-04-10
Move defines in utils.h to utils_def.h to fix shared header compile i…

sbranden:
…ssues

utils.h is included in various header files for the defines in it.
Some of the other header files only contain defines.  This allows the
header files to be shared between host and target builds for shared defines.

Recently types.h has been included in utils.h as well as some function
prototypes.

Because of the inclusion of types.h conflicts exist building host tools
abd these header files now.  To solve this problem,
move the defines to utils_def.h and have this included by utils.h and
change header files to only include utils_def.h and not pick up the new
types.h being introduced.

Fixes ARM-software/tf-issues#461


davidcunado-arm:
jenkins: test this please

davidcunado-arm:
@sbranden 
The build failed when building with EL3_PAYLOAD_BASE, e.g.:
make V=1 DEBUG=1 PLAT=juno SPIN_ON_BL1_EXIT=1 EL3_PAYLOAD_BASE=0x80000000

plat/arm/css/common/css_bl2_setup.c: In function 'bl2_platform_setup':
plat/arm/css/common/css_bl2_setup.c:101:2: error: implicit declaration of function 'zero_normalmem' [-Werror=implicit-function-declaration]
  zero_normalmem((void *)ARM_TRUSTED_SRAM_BASE, 128);
  ^
cc1: all warnings being treated as errors
make: *** [build/juno/debug/bl2/css_bl2_setup.o] Error 1

But thinking about, perhaps an alternate approach would be to 'revert' utils.h to the version before 308d359b260d888f024a2d26c76cd4a50789e432 and instead put the content added to utils.h in that patch into its own file - e.g. utils_mem.h (yeah not a great name - I'll see if I can come up with a new one over the w/end!).

I think that this approach will need less changes all round - thoughts?

sbranden:
@davidcunado-arm, I see a number of other header files are named xxx_def.h  which is why I moved the defines from utils.h to utils_def.h.  If you'd like to move the function prototypes and #include out of utils.h elsewhere please do so.  Chanding whatever C files to include the necessary utils.h will solve the compile issue.

Or, reverting the change you mentioned in utils.h and moving it to another header file will work as well.

Either way will work.

danh-arm:
I think splitting the file into utils.h and utils_def.h as Scott has done is the right thing to do. It's just that css_bl2_setup.c now needs to include utils.h. That looks like minor rework to me...

masahir0y:
Scott's motivation is to share header(s) between TF and hosttools.
This is generally a bad idea.  As others pointed out, some pieces of macros should be copied to a host header.
For example, Linux has separate copies of headers under "tools/include/" directory.

davidcunado-arm:
@sbranden 
As per yours and Dan's comment, splitting out the defines from utils.h into utils_def.h aligns with how this has been done for other defines.
So with the current:

- PR, utils.h doesn't strictly need to include utils_def.h, it's there for compatibility - alternatively, this include can be guarded by #if !ERROR_DEPRECATED

- To fix the check failure, include utils.h in css_bl2_setup.c, as Dan suggested.

With these minor changes, we can progress this PR.

sbranden:
Included utils.h in css_bl2_setup.c.

I think further work (in a different commit to cleanup the inclusion of utils.h everywhere) would need to be done before adding ERROR_DEPRECATED into the utils.h right now so I left it out of the commit.

antonio-nino-diaz-arm:
If you have done a force push to the same branch used to create the PR everything in the PR should have been updated to point at new commit automatically (as far as I see, that's what has happened). Are you talking about something else?

sbranden:
I did a force push - just wanted to see if that the correct procedure to follow.  Because after doing so I can not access the "view changes" button in the comment above that I show below.  So I don't know if you added comments to the review:

davidcunado-arm requested changes 7 days ago View changes
See other comment

davidcunado-arm:
@sbranden 
Could you resolve the conflict in plat_arm.h?
Also, to add ERROR_DEPRECATED around utils_def.h include in utils.h only requires an additional change to plat/arm/board/juno/juno_trng.c - change include utils.h to utils_def.h.
I've got the patch file, which I can email to you (as I cannot seem to attach it here!!!)

sbranden:
@davidcunado-arm, please send me the patch to incorporate to my username (same as github) @broadcom.com

Thanks

sbranden:
Merged @davidcunado-arm change to include utils.h in juno_trng.c, change to add ERROR_DEPRECATED to utils.h, and resolved conflict in plat_arm.h

davidcunado-arm:
jenkins: test this please

PR #897, opened 2017-04-10
Tegra: memctrl_v1: enable 'xlat_table_v2' library

vwadekar:
This patch enables the 'xlat_table_v2' library for the Tegra Memory
Controller driver. This library allows us to dynamically map/unmap
memory regions, with MMU enabled.

The Memory Controller driver maps/unmaps non-overlapping Video Memory
region, to clean it of any secure contents, before it resizes the
region.


davidcunado-arm:
jenkins: test this please

PR #898, opened 2017-04-13
PSCI: Build option to enable D-Caches early in warmboot

soby-mathew:
This patch introduces a build option to enable D-cache early on the CPU
after warm boot. This is applicable for platforms which do not require
interconnect programming to enable cache coherency (eg: single cluster
platforms). If this option is enabled, then warm boot path enables
D-caches immediately after enabling MMU.

Fixes ARM-Software/tf-issues#456

Change-Id: I44c8787d116d7217837ced3bcf0b1d3441c8d80e

### DIFF
--- a/bl32/sp_min/aarch32/entrypoint.S
+++ b/bl32/sp_min/aarch32/entrypoint.S
@@ -236,23 +236,26 @@ func sp_min_warm_entrypoint
 	 *
 	 * The PSCI implementation invokes platform routines that enable CPUs to
 	 * participate in coherency. On a system where CPUs are not
-	 * cache-coherent out of reset, having caches enabled until such time
-	 * might lead to coherency issues (resulting from stale data getting
-	 * speculatively fetched, among others). Therefore we keep data caches
-	 * disabled while enabling the MMU, thereby forcing data accesses to
-	 * have non-cacheable, nGnRnE attributes (these will always be coherent
-	 * with main memory).
+	 * cache-coherent without appropriate platform specific programming,
+	 * having caches enabled until such time might lead to coherency issues
+	 * (resulting from stale data getting speculatively fetched, among
+	 * others). Therefore we keep data caches disabled even after enabling
+	 * the MMU for such platforms.
 	 *
-	 * On systems where CPUs are cache-coherent out of reset, however, PSCI
-	 * need not invoke platform routines to enter coherency (as CPUs already
-	 * are), and there's no reason to have caches disabled either.
+	 * On systems with hardware-assisted coherency, or on single cluster
+	 * platforms, such platform specific programming is not required to
+	 * enter coherency (as CPUs already are); and there's no reason to have
+	 * caches disabled either.
 	 */
-#if HW_ASSISTED_COHERENCY
-	mov	r0, #0
-#else
 	mov	r0, #DISABLE_DCACHE
-#endif
 	bl	bl32_plat_enable_mmu
+
+#if HW_ASSISTED_COHERENCY || WARMBOOT_ENABLE_DCACHE_EARLY
+	ldcopr	r0, SCTLR
+	orr	r0, r0, #SCTLR_C_BIT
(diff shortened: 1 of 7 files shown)


vwadekar:
@soby-mathew 

Thanks for this PR. I will test and report. Hopefully, this week or early next week.

bl32/sp_min/aarch32/entrypoint.S
```diff
@@ -236,24 +236,27 @@ func sp_min_warm_entrypoint
...
+	 * platforms, such platform specific programming is not required to
+	 * enter coherency (as CPUs already are); and there's no reason to have
+	 * caches disabled either.
 	 */
-#if HW_ASSISTED_COHERENCY
-	mov	r0, #0
-#else
 	mov	r0, #DISABLE_DCACHE
-#endif
 	bl	bl32_plat_enable_mmu
```
vwadekar: Are the platforms responsible for stack memory mgmt with this approach?

lib/psci/psci_suspend.c
```diff
@@ -302,7 +302,7 @@ void psci_cpu_suspend_finish(unsigned int cpu_idx,
 	 */
 	psci_plat_pm_ops->pwr_domain_suspend_finish(state_info);
 
-#if !HW_ASSISTED_COHERENCY
+#if !(HW_ASSISTED_COHERENCY || WARMBOOT_ENABLE_DCACHE_EARLY)
 	/* Arch. management: Enable the data cache, stack memory maintenance. */
```
vwadekar: stack memory mgmt would be skipped with this approach. Is this intended?

vwadekar:
@soby-mathew 

Hi Soby,

This PR works well with Tegra210 and Tegra186 SoCs. You can go ahead with the merge.

Thanks.

Acked-by: Varun Wadekar <vwadekar@nvidia.com>

bl32/sp_min/aarch32/entrypoint.S
soby-mathew: @vwadekar ,
Usually if the D-cache is enabled while the stack is in use, we have to perform invalidation of the `used` stack to prevent hitting stale cache data. Not needed in this case, because the D-cache is only enabled after stack is fully unwound (i.e used stack size = 0) in line 254.

So platform need not perform stack mgmt when WARMBOOT_ENABLE_DCACHE_EARLY is enabled.

lib/psci/psci_suspend.c
soby-mathew: The D-cache is enabled early when WARMBOOT_ENABLE_DCACHE_EARLY is enabled. As mentioned in my previous comment, stack management is not required for this case.

soby-mathew:
Thanks @vwadekar for the confirmation. So will you be raising a PR to enable WARMBOOT_ENABLE_DCACHE_EARLY for Tegra210 and Tegra186 ?

vwadekar:
Yes, Soby. I will start a PR for the Tegra ports.

docs/user-guide.md
```diff
@@ -508,6 +508,12 @@ performed.
     to a string formed by concatenating the version number, build type and build
     string.
 
+*   `WARMBOOT_ENABLE_DCACHE_EARLY` : Boolean option to enable D-cache early on
+    the CPU after warm boot. This is applicable for platforms which do not
+    require interconnect programming to enable cache coherency (eg: single
+    cluster platforms). If this option is enabled, then warm boot path
+    enables D-caches immediately after enabling MMU. This option defaults to 0.
+
```
danh-arm: minor: I would prefer if the description for HW_ASSISTED_COHERENCY were updated to say if HW_ASSISTED_COHERENCY is enabled then this implies WARMBOOT_ENABLE_DCACHE_EARLY is also enabled.

soby-mathew: OK

davidcunado-arm:
jenkins: test this please

PR #900, opened 2017-04-14
drivers: ti: uart: remove UART_FCR read-modify-write

vwadekar:
UARTFCR is a write-only register and so a read-modify-write is not
applicable. This patch updates the programming sequence to remove this
logic and write to it directly from console_core_putc().

Signed-off-by: Vincent Zhang <xunz@nvidia.com>

### DIFF
--- a/drivers/ti/uart/aarch64/16550_console.S
+++ b/drivers/ti/uart/aarch64/16550_console.S
@@ -115,9 +115,6 @@ func console_core_putc
 	b.ne	1b
 	mov	w2, #0xD		/* '\r' */
 	str	w2, [x1, #UARTTX]
-	ldr	w2, [x1, #UARTFCR]
-	orr	w2, w2, #UARTFCR_TXCLR
-	str	w2, [x1, #UARTFCR]
 
 	/* Check if the transmit FIFO is full */
 2:	ldr	w2, [x1, #UARTLSR]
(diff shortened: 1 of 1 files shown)


vwadekar:
@nmenon 

Hi Nishanth,

Can you please review?

r-vignesh:
There is no need to touch UART_FCR register in console_core_putc. The
check for whether transmit FIFO is full is sufficient before writing to
UART TX FIFO. In fact setting UARTFCR_TXCLR immediately after a byte is
written to FIFO might even result in loss of that byte, if UART hasn't
sent that byte out.

Soft resetting TX and RX FIFO once during console init should be sufficient.

UARTFCR_DMAEN should not be set, as ATF is not using DMA with UART.

davidcunado-arm:
jenkins: test this please

vwadekar:
David, I think Vignesh is not happy with the current implementation. Please
don't merge this PR until we get an Acknowledgement from    Nishanth or
Vignesh.

nmenon:
@vwadekar thanks. that is correct - the patch currently is wrong and can have sideeffects depending on SoC.. enabling DMA will trigger DMA events outof the hardware block and depending on the SoC and DMA infrastructure all kinds of weird stuff can happen.

So, NAK.

vwadekar:
@r-vignesh @nmenon 

I have tried to fix your concerns. Please review and let me know if this is OK.

nmenon:
LGTM, @r-vignesh -> ok with you?

r-vignesh:
@vwadekar  thanks! The change itself looks good. 
@nmenon IMO, we need a patch to remove setting of UARTFCR_DMAEN bit in console_core_init.

davidcunado-arm:
jenkins: test this please

davidcunado-arm:
@vwadekar 
PR looks fine - I'll give it a couple of days to see if there are comments from anyone else, and then I'll merge it.

vwadekar:
Agreed. Thanks.

PR #903, opened 2017-04-19
ARM platforms: Add option to use xlat tables lib v1

antonio-nino-diaz-arm:
ARM platforms have migrated to the translation tables library v2.
However, for testing purposes, it can be useful to temporarily switch
back to the old version.

This patch introduces the option `ARM_XLAT_TABLES_LIB_V1`, that switches
to v1 of the library when is set to 1. By default, it is 0, so that ARM
platforms use the new version unless specifically stated.

Updated User Guide.

Change-Id: I53d3c8dd97706f6af9c6fca0364a88ef341efd31
Signed-off-by: Antonio Nino Diaz <antonio.ninodiaz@arm.com>

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -568,6 +568,10 @@ performed.
     -   `tdram` : Trusted DRAM (if available)
     -   `dram`  : Secure region in DRAM (configured by the TrustZone controller)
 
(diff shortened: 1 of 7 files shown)


antonio-nino-diaz-arm:
@sbranden, you could use this option instead of reverting the changes like you said here: https://github.com/ARM-software/tf-issues/issues/469

PR #905, opened 2017-04-20
tspd:FWU:Fix usage of SMC_RET0

antonio-nino-diaz-arm:
SMC_RET0 should only be used when the SMC code works as a function that
returns void. If the code of the SMC uses SMC_RET1 to return a value to
signify success and doesn't return anything in case of an error (or the
other way around) SMC_RET1 should always be used to return clearly
identifiable values.

This patch fixes two cases in which the code used SMC_RET0 instead of
SMC_RET1.

It also introduces the define SMC_OK to use when an SMC must return a
value to tell that it succeeded, the same way as SMC_UNK is used in case
of failure.

### DIFF
--- a/bl1/bl1_fwu.c
+++ b/bl1/bl1_fwu.c
@@ -109,7 +109,7 @@ register_t bl1_fwu_smc_handler(unsigned int smc_fid,
 		break;
 	}
 
-	SMC_RET0(handle);
+	SMC_RET1(handle, SMC_UNK);
 }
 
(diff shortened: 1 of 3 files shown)


antonio-nino-diaz-arm:
Closing as this PR targets the master branch.

danh-arm:
For future reference, GitHub now allows people with commit rights to change the target branch, so there's no need to close PRs for this reason.

PR #908, opened 2017-04-20
Tegra: Control inclusion of helper code used for asserts

antonio-nino-diaz-arm:
One assert depends on code that is conditionally compiled based on the
DEBUG define. This patch modifies the conditional inclusion of such code
so that it is based on the ENABLE_ASSERTIONS build option.

Platform Makefile changed to set `ENABLE_ASSERTIONS` to 1 instead of
the deprecated option `ASM_ASSERTION`. This also pulls in C assertions
in release mode.

This patch has to be applied on top of the changes of https://github.com/ARM-software/arm-trusted-firmware/pull/906

### DIFF
--- a/plat/nvidia/tegra/soc/t132/plat_psci_handlers.c
+++ b/plat/nvidia/tegra/soc/t132/plat_psci_handlers.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015-2016, ARM Limited and Contributors. All rights reserved.
+ * Copyright (c) 2015-2017, ARM Limited and Contributors. All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are met:
@@ -132,7 +132,7 @@ int tegra_soc_pwr_domain_off(const psci_power_state_t *target_state)
 
 int tegra_soc_pwr_domain_suspend(const psci_power_state_t *target_state)
 {
-#if DEBUG
+#if ENABLE_ASSERTIONS
 	int cpu = read_mpidr() & MPIDR_CPU_MASK;
 
 	/* SYSTEM_SUSPEND only on CPU0 */
--- a/plat/nvidia/tegra/common/tegra_common.mk
+++ b/plat/nvidia/tegra/common/tegra_common.mk
@@ -31,8 +31,8 @@
 CRASH_REPORTING		:=	1
 $(eval $(call add_define,CRASH_REPORTING))
 
-ASM_ASSERTION		:=	1
-$(eval $(call add_define,ASM_ASSERTION))
+ENABLE_ASSERTIONS	:=	1
+$(eval $(call add_define,ENABLE_ASSERTIONS))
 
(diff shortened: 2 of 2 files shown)


antonio-nino-diaz-arm:
Could you take a look at this, @vwadekar ?

antonio-nino-diaz-arm:
This build is expected to fail as the patches in https://github.com/ARM-software/arm-trusted-firmware/pull/906 are needed for this to build without errors.

plat/nvidia/tegra/soc/t132/plat_psci_handlers.c
```diff
@@ -132,7 +132,7 @@ int tegra_soc_pwr_domain_off(const psci_power_state_t *target_state)
 
 int tegra_soc_pwr_domain_suspend(const psci_power_state_t *target_state)
 {
-#if DEBUG
+#if ENABLE_ASSERTIONS
```
vwadekar: @antonio-nino-diaz-arm 

I wonder why this file was changed. Are we supposed to replace all such #if DEBUG checks with this new macro?

davidcunado-arm: The change of DEBUG to ENABLE_ASSERTIONS is needed here because PR #906 added support to enable assertions in release builds as well as debug builds. 
And so to include auxiliary code that is only used for assertions (the local cpu in this case) we can no longer use DEBUG - we need to use another define that is in sync with when assertions are enabled, rather than aligning with the build-type; ENABLE_ASSERTIONS is the that new define.

vwadekar:
Thanks for the explanation, David. I am OK with this PR.

davidcunado-arm:
jenkins: test this please

davidcunado-arm:
@vwadekar 
The tegra platforms have moved since Antonio created this patch - would you be able to rebase and update the patch to cover the auxiliary assert code added to plat/nvidia/tegra/common/drivers/smmu/smmu.c?
https://github.com/ARM-software/arm-trusted-firmware/blob/master/plat/nvidia/tegra/common/drivers/smmu/smmu.c#L92

vwadekar:
@davidcunado-arm 

Sure, I can create a PR. Do you want me to push a new PR fir the smmu change? Or include this change in the new PR as well?

davidcunado-arm:
@vwadekar 
Update this PR as without the update to smmu.c the checks fail. Thanks!

vwadekar:
@davidcunado-arm 

Thanks. I will push the changes. Would I be able to push to the branch antonio-nino-diaz-arm:an/asserts-release-nvidia ?

vwadekar:
@davidcunado-arm 

I am not able to push to Antonio's repo. I think we need to start a new PR. I have made the necessary changes and cherry-picked Antonio's patch from this PR. Let me know.

vwadekar:
Started a new PR - #920

antonio-nino-diaz-arm:
It seems that your PR takes care of this so I'll close this one.

PR #910, opened 2017-04-20
Add AArch32 support for Juno

dp-arm:
This patch series adds AArch32 support  for Juno.  The boot flow is slightly convoluted because Juno cannot boot in AArch32 directly.  As a result, BL1/BL2 run in AArch64 and at the end of BL1, a trampoline is written at the high vector address (0xFFFF0000).  A warm reset in AArch32 is then requested.  The CPU starts execution at the high vector address and jumps into the SP_MIN runtime.


omnium21:
Excellent! This works for me:
Tested-by: Ryan Harkin <ryan.harkin@linaro.org>

davidcunado-arm:
jenkins: test this please

PR #912, opened 2017-04-21
Tegra: smmu: make the context save sequence robust

vwadekar:
This patch sanity checks the SMMU context created by the platform
code. The first entry contains the size of the array; which the
driver now verifies before moving on with the save.

This patch also fixes an error in the calculation of the size of
the context that gets copied to TZDRAM.


vwadekar:
@davidcunado-arm 

Hi David,

Can you prioritize this PR? The error is the size calculation actually breaks System Suspend. This PR fixes that problem.

davidcunado-arm:
jenkins: test this please

vwadekar:
Thanks.

davidcunado-arm:
No problem!

PR #914, opened 2017-04-21
fiptool: Support non-decimal --align arguments

afaerber:
An alignment value of 0x4000 is much easier to type than 16384,
so enhance get_image_align() to recognize a 0x prefix.

### DIFF
--- a/tools/fiptool/fiptool.c
+++ b/tools/fiptool/fiptool.c
@@ -646,7 +646,7 @@ static unsigned long get_image_align(char *arg)
 	unsigned long align;
 
 	errno = 0;
-	align = strtoul(arg, &endptr, 10);
+	align = strtoul(arg, &endptr, 0);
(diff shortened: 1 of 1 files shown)


tools/fiptool/fiptool.c
```diff
@@ -646,7 +646,10 @@ static unsigned long get_image_align(char *arg)
 	unsigned long align;
 
 	errno = 0;
-	align = strtoul(arg, &endptr, 10);
+	if (strlen(arg) > 2 && strncmp(arg, "0x", 2) == 0)
+		align = strtoul(arg + 2, &endptr, 16);
```
dp-arm: I would rather see strtoul() use base 0 (it will detect if it is hex, dec or octal automatically).  Avoids having to detect the base by scanning for 0x etc.

afaerber:
Thanks, changed, appears to work for `0x`.

davidcunado-arm:
jenkins: test this please

PR #916, opened 2017-04-24
fiptool: Add --align-reserve command line option

afaerber:
Allow each image to reserve some space before alignment.

Use case is --align 0x4000 --align-reserve 64 to make sure sufficient
room is available to prepend (or append) vendor-specific headers.

Helps resolve afaerber/meson-tools#3.

### DIFF
--- a/tools/fiptool/fiptool.c
+++ b/tools/fiptool/fiptool.c
@@ -51,6 +51,7 @@
 #define OPT_TOC_ENTRY 0
 #define OPT_PLAT_TOC_FLAGS 1
 #define OPT_ALIGN 2
+#define OPT_ALIGN_RESERVE 3
 
 static int info_cmd(int argc, char *argv[]);
 static void info_usage(void);
@@ -511,7 +512,8 @@ static void info_usage(void)
 	exit(1);
 }
 
-static int pack_images(const char *filename, uint64_t toc_flags, unsigned long align)
+static int pack_images(const char *filename, uint64_t toc_flags,
+	unsigned long align, unsigned long align_reserve)
 {
 	FILE *fp;
 	image_desc_t *desc;
@@ -546,7 +548,7 @@ static int pack_images(const char *filename, uint64_t toc_flags, unsigned long a
 		if (image == NULL)
 			continue;
 		payload_size += image->toc_e.size;
-		entry_offset = (entry_offset + align - 1) & ~(align - 1);
+		entry_offset = (entry_offset + align_reserve + align - 1) & ~(align - 1);
 		image->toc_e.offset_address = entry_offset;
 		*toc_entry++ = image->toc_e;
 		entry_offset += image->toc_e.size;
@@ -653,6 +655,19 @@ static unsigned long get_image_align(char *arg)
 	return align;
 }
 
+static unsigned long get_image_align_reservation(char *arg)
+{
+	char *endptr;
+	unsigned long align_reserve;
+
+	errno = 0;
+	align_reserve = strtoul(arg, &endptr, 0);
+	if (*endptr != '\0' || errno != 0)
+		log_errx("Invalid alignment: %s", arg);
+
+	return align_reserve;
+}
+
 static void parse_blob_opt(char *arg, uuid_t *uuid, char *filename, size_t len)
 {
 	char *p;
@@ -673,7 +688,7 @@ static int create_cmd(int argc, char *argv[])
 	struct option *opts = NULL;
 	size_t nr_opts = 0;
 	unsigned long long toc_flags = 0;
-	unsigned long align = 1;
+	unsigned long align = 1, align_reserve = 0;
 
 	if (argc < 2)
 		create_usage();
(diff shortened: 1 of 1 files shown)


tools/fiptool/fiptool.c
```diff
@@ -753,6 +773,7 @@ static void create_usage(void)
 	printf("\n");
 	printf("Options:\n");
 	printf("  --align <value>\t\tEach image is aligned to <value> (default: 1).\n");
+	printf("  --align-reserve <value>\t\tEach image is padded by <value> (default: 0).\n");
```
dp-arm: Use single \t here and in update_usage().

afaerber: As you wish.
Can you think of any ways to improve the wording while at it? I clearly copied the --align example, but it might be clearer to append "bytes" to both? I.e., to distinguish it from which byte value is used to pad up to the alignment size, which I assume is always 0x00.

dp-arm: We should probably mention the units indeed.

dp-arm:
BTW, what is the mechanism by which you prepend those headers? Do you post-process the FIP using another tool?

afaerber:
Yes, see the commit message. The tools are proprietary `aml_encrypt_gxb` (there's also `aml_encrypt_gxl`, `aml_encrypt_gxtvbb` and `aml_encrypt_txl` versions not yet investigated in detail) and my Open Source replacement `amlbootsig` in the mentioned meson-tools.

Affected are the development boards Hardkernel Odroid-C2 and presumably FriendlyElec NanoPi K2 (both Amlogic S905 aka Meson GXB), Khadas VIM (S905X aka GXL), as well as several TV boxes based on these and related Amlogic SoCs.

I still have one ugly UUID addition that I am holding back. It seems that at least one image's offset and size are hardcoded in Odroid-C2's BL2, so I first want to investigate whether we can work around it by choosing other official UUIDs in the expected sort order.

One other change in the postprocessing was an alignment of the image **size** to 16 bytes (see `padlen` in my code).

Since I am barely just finding out by black-box-testing what the header fields may mean, I do not see integrating those headers into `fiptool` as a realistic option - in case that's what you were getting at.

Independently, it seems there are some licensing inconsistencies that will make upstreaming BLx driver code as BSD-3-clause difficult. Amlogic's U-Boot tarballs and Hardkernel's u-boot_firmware repository contain some code I didn't investigate further yet in "bl21" and "scp_task" dirs - Hardkernel has licensed the repository as GPL-3.0.

dp-arm:
I see.  I was just wondering whether the headers you are referring to could be additional images.

dp-arm:
So thinking about it I have two concerns.  First of all, those headers cannot be authenticated because they are essentially "phantom" blobs.  To authenticate them they would need to be presented as an image with their own UUID and corresponding hash in the content certificate.

My second concern is a consequence of the fact that these headers cannot be authenticated.  The platform code will need to parse these headers and do appropriate checks.  Failing to do these checks properly can result in security issues depending on the content of those headers.  For example, assuming there is a load address field in those headers, the attacker could modify the field and point it anywhere.

masahir0y:
[1] The option name "--pad" would be clearer?  For example, dtc seems to do like this.
[2] If you want to make the units clearer in the help message, you can mention '--align <bytes>' instead of '--align <value>'.  dtc seems to do like this.

The following is a snippet of 'dtc --help'.

  -S, --space <arg>          
	Make the blob at least <bytes> long (extra space)
  -p, --pad <arg>            
	Add padding to the blob of <bytes> long (extra space)
  -a, --align <arg>          
	Make the blob align to the <bytes> (extra space)

masahir0y:
Bah, all the angle brackets disappeared.

I wanted to mean like follows.  I hope it is OK this time...

[2] If you want to make the units clearer in the help message, you can mention `--align <bytes>` instead of `--align <value>`. dtc seems to do like this.

The following is a snippet of 'dtc --help'.

    -S, --space <arg>
        Make the blob at least <bytes> long (extra space)
    -p, --pad <arg>
        Add padding to the blob of <bytes> long (extra space)
    -a, --align <arg>
        Make the blob align to the <bytes> (extra space)

dp-arm:
Indeed, that would be clearer.

danh-arm:
Hi @afaerber. Do you have any response to the comment [here](https://github.com/ARM-software/arm-trusted-firmware/pull/916#issuecomment-296968003) to this PR?

afaerber:
Hi @danh-arm, as you know, I am not associated with Amlogic and do not understand why I should defend their boot ROM or BLx code. Thus, discussions about whether or not some hypothetical authentication feature we're not using works or works not in that theoretical scenario is utterly useless.
Given time, I will look into making minor tweaks to the help texts and update this pull.

danh-arm:
Hi @afaerber. We're not asking you to defend Amlogic's boot code; we just wondered if you had more information on how it works, mainly just to understand the bigger picture. However, you're right that any security concerns in that firmware don't really affect this PR or the wider ARM TF Trusted Board Boot feature. So we'll merge this once you've done those minor tweaks.

dp-arm:
This PR has been superseded by PR "fiptool: Add --pad command line option".

danh-arm:
Superseded by #952

PR #919, opened 2017-04-26
Update terminology: standard SMC to yielding SMC

davidcunado-arm:
Since Issue B (November 2016) of the SMC Calling Convention document
standard SMC calls are renamed to yielding SMC calls to help avoid
confusion with the standard service SMC range, which remains unchanged.

http://infocenter.arm.com/help/topic/com.arm.doc.den0028b/ARM_DEN0028B_SMC_Calling_Convention.pdf

This patch adds a new define for yielding SMC call type and deprecates
the current standard SMC call type. The tsp is migrated to use this new
terminology and, additionally, the documentation and code comments are
updated to use this new terminology.

Change-Id: I0d7cc0224667ee6c050af976745f18c55906a793


davidcunado-arm:
jenkins: test this please

PR #922, opened 2017-04-28
Migrate secure payload dispatchers to new SMC terminology

davidcunado-arm:
Since Issue B (November 2016) of the SMC Calling Convention document
standard SMC calls are renamed to yielding SMC calls to help avoid
confusion with the standard service SMC range, which remains unchanged.

http://infocenter.arm.com/help/topic/com.arm.doc.den0028b/ARM_DEN0028B_SMC_Calling_Convention.pd

A previous patch introduced a new define for yielding SMC call type.
This patch updates the secure payload dispatchers (except the TSPD) to
use this new define and also migrates the code to use the new
terminology.

Change-Id: I3d2437c04e3b21fdbd32019f55c066c87679a5bf


davidcunado-arm:
@jenswi-linaro @vwadekar

Could you review the changes to optee, tlk and trusty SPDs?

NOTE - these commits are dependent on #919

include/bl32/payloads/tlk.h
```diff
@@ -36,18 +36,22 @@
 /*
  * Generate function IDs for the Trusted OS/Apps
  */
+#if !ERROR_DEPRECATED
 #define TLK_TOS_STD_FID(fid)	((fid) | 0x72000000 | (0 << 31))
 #define TLK_TA_STD_FID(fid)	((fid) | 0x70000000 | (0 << 31))
+#endif
```
vwadekar: Can we delete the STD_FID macros? Since it is just a rename of the macros, I dont see a problem.

services/spd/trusty/smcall.h
```diff
@@ -48,9 +48,13 @@
 		)
 
 #define SMC_FASTCALL_NR(entity, fn)	SMC_NR((entity), (fn), 1, 0)
-#define SMC_STDCALL_NR(entity, fn)	SMC_NR((entity), (fn), 0, 0)
 #define SMC_FASTCALL64_NR(entity, fn)	SMC_NR((entity), (fn), 1, 1)
+#if !ERROR_DEPRECATED
```
vwadekar: Can we delete the STDCALL macros?

include/bl32/payloads/tlk.h
danh-arm: This file (and the Trusty) equivalent are public APIs. They (or a version of them) may be used by the OS, so this code aimed to maintain compatibility and provide a migration path, with a view to removing the old defines at a later date.

If it's considered unlikely that the OS will use these files directly (e.g. if there's no link between this copy and the OS's copy), then there may be no need to provide this migration path. Are you sure this is what you want?

davidcunado-arm: @vwadekar 
As per @danh-arm comment, is it ok to *not* have this migration path and simply remove the current STD-based APIs?

vwadekar:
Yes. Since the SMC ID remain the same we can remove the previous macros. We
are already modifying the SPDs, so I don't see any problems.

davidcunado-arm:
@vwadekar 
Ok - I'll remove the migration path (and the current STD-based APIs) and update the PR.
Thanks!

davidcunado-arm:
@jenswi-linaro 
Ok with these changes to opteed? Please let me know by Monday 8th May - assuming no f/b I'll look to merge then during UK work day.

PR #923, opened 2017-05-01
xlat lib: Don't set mmap_attr_t enum to be -1

nmenon:
-1 is not a defined mmap_attr_t type. Instead of using invalid enum
types, we can either choose to define a INVALID type OR handle the
condition specifically.

Since the usage of mmap_region_attr is limited, it is easier to just
handle the error condition specifically and return 0 or -1 depending
on success or fail.

Fixes: ARM-Software/tf-issues#473
Fixes: 28fa2e9ee8f4 ("xlat lib: Use mmap_attr_t type consistently")

### DIFF
--- a/lib/xlat_tables/xlat_tables_common.c
+++ b/lib/xlat_tables/xlat_tables_common.c
@@ -273,15 +273,19 @@ static uint64_t mmap_desc(mmap_attr_t attr, unsigned long long addr_pa,
 }
 
 /*
- * Returns attributes of area at `base_va` with size `size`. It returns the
- * attributes of the innermost region that contains it. If there are partial
- * overlaps, it returns -1, as a smaller size is needed.
+ * Look for the innermost region that contains the area at `base_va` with size
+ * `size`. Populate *attr with the attributes of this region.
+ *
+ * On success, this function returns 0.
+ * If there are partial overlaps (meaning that a smaller size is needed) or if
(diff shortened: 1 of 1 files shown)


davidcunado-arm:
jenkins: test this please

lib/xlat_tables/xlat_tables_common.c
```diff
@@ -277,11 +277,11 @@ static uint64_t mmap_desc(mmap_attr_t attr, unsigned long long addr_pa,
  * attributes of the innermost region that contains it. If there are partial
  * overlaps, it returns -1, as a smaller size is needed.
```
sandrine-bailleux-arm: Could you please update the function comment to better reflect its new behaviour?

I had an internal version of this patch sitting in our internal review system, so here's what I had written, but feel free to diverge from that:
```
/*
 * Look for the innermost region that contains the area at `base_va` with size
 * `size`. Populate *attr with the attributes of this region.
 *
 * On success, this function returns 0.
 * If there are partial overlaps (meaning that a smaller size is needed) or if
 * the region can't be found in the given area, it returns -1. In this case the
 * value pointed by attr should be ignored by the caller.
 */
```

nmenon:
@sandrine-bailleux-arm Thanks. updated.

davidcunado-arm:
jenkins: test this please

PR #924, opened 2017-05-02
Fix execute-never permissions in xlat tables libs

antonio-nino-diaz-arm:
Translation regimes that only support one virtual address space (such as the ones for EL2 and EL3) can flag memory regions as execute-never by setting to 1 the XN bit in the Upper Attributes field in the translation tables descriptors. Translation regimes that support two different virtual address spaces (such as the one shared by EL1 and EL0) use bits PXN and UXN instead.

The Trusted Firmware runs at EL3 and EL1, it has to handle translation tables of both translation regimes, but the previous code handled both regimes the same way, as if both had only 1 VA range.

When trying to set a descriptor as execute-never it would set the XN bit correctly in EL3, but it would set the XN bit in EL1 as well. XN is at the same bit position as UXN, which means that EL0 was being prevented from executing code at this region, not EL1 as the code intended. Therefore, the PXN bit was unset to 0 all the time. The result is that, in AArch64 mode, read-only data sections of BL2 weren't protected from being executed.

This patch adds support of translation regimes with two virtual address spaces to both versions of the translation tables library, fixing the execute-never permissions for translation tables in EL1.

The library currently does not support initializing translation tables for EL0 software, therefore it does not set/unset the UXN bit. If EL1 software needs to initialize translation tables for EL0 software, it should use a different library instead.

### DIFF
--- a/lib/xlat_tables_v2/xlat_tables_private.h
+++ b/lib/xlat_tables_v2/xlat_tables_private.h
@@ -108,6 +108,13 @@ typedef struct {
 	/* Set to 1 when the translation tables are initialized. */
 	int initialized;
 
+	/*
+	 * Bit mask that has to be ORed to the rest of a translation table
+	 * descriptor in order to prohibit execution of code at the exception
+	 * level of this translation context.
+	 */
+	uint64_t execute_never_mask;
+
 } xlat_ctx_t;
 
 #if PLAT_XLAT_TABLES_DYNAMIC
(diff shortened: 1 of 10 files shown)


lib/xlat_tables_v2/xlat_tables_private.h
```diff
@@ -108,6 +108,13 @@ typedef struct {
 	/* Set to 1 when the translation tables are initialized. */
 	int initialized;
 
+	/*
+	 * Bit mask that has to be ORed to the rest of a translation table
+	 * descriptor in order to prohibit execution of code at the exception
+	 * level of this translation context.
+	 */
+	uint64_t execute_never_mask;
```
achingupta: AFAIU, a `xlat_ctx_t` is associated with a unique translation regime. If so, then would it not be more scalable to add a field in this structure that indicates the EL or the translation regime it corresponds to. If the EL is then passed to `xlat_desc()` then the choice between PXN and XN could be made there.
This way if EL3 is creating translation tables on behalf of S-EL1, then there will be no reliance on the `CurrentEL`. Any future EL specific differences between translation regimes could be catered for as well.
Please let me know if I am misunderstanding anything

danh-arm: An earlier version did as you suggest but this means that a calculation of the mask to use from the EL needs to be done for every xlat table created. This way is more efficient. If the EL is needed in the context later, this can be easily added without affecting compatibility.

PR #925, opened 2017-05-03
Use SPDX license identifiers

dp-arm:
To make software license auditing simpler, use SPDX[0] license
identifiers instead of duplicating the license text in every file.

NOTE: Files that have been imported by FreeBSD have not been modified.

[0]: https://spdx.org/

Change-Id: I80a00e1f641b8cc075ca5a95b10607ed9ed8761a

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,31 +1,7 @@
 #
 # Copyright (c) 2013-2017, ARM Limited and Contributors. All rights reserved.
 #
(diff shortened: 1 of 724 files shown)


davidcunado-arm:
Due to the number of files this PR touches, we intend to merge very soon - please raise any concerns / feedback by THU 4th MAY-17. I'll look to merge towards the end of the UK workday.

Thanks!

PR #932, opened 2017-05-10
Use SPDX license identifiers for remaining files

dp-arm:
Change-Id: I7f54f45db65f32481cc05e1bd2c9c683b756e19a


davidcunado-arm:
@TonyXie06 
FYI - we've migrated to using SPDX license identifiers; this patch updates the new files added for rk3328 platform.

PR #934, opened 2017-05-12
Break circular dependency in FVP and Nvidia header files

sandrine-bailleux-arm:


sandrine-bailleux-arm:
Hi @vwadekar,

Could you please have a look at the 2nd patch in this pull request and make sure you're happy with it?

Regards,
Sandrine

danh-arm:
This is so minor that I don't think we need Varun's approval. I'll just merge it.

PR #936, opened 2017-05-12
Simplify assert() to reduce memory usage

antonio-nino-diaz-arm:
The behaviour of assert() now depends on the value of the new optional platform define `PLAT_LOG_LEVEL_ASSERT`. This defaults to `LOG_LEVEL` if not defined by the platform.

- If `PLAT_LOG_LEVEL_ASSERT` >= `LOG_LEVEL_VERBOSE`, it prints the file name, line and asserted expression.
- If `PLAT_LOG_LEVEL_ASSERT` >= `LOG_LEVEL_INFO`, it prints the file name and line.
- If not, it doesn't print anything.

Note the old behaviour was to print the function name whereas now it prints the file name. This reduces memory usage because the file name is shared between all assert calls in a given file. Also, the default
behaviour in debug builds is to no longer print the asserted expression, greatly reducing the string usage.

For FVP debug builds this change saves approximately:
```
              No TBBR    TBBR
        BL1    1.6 KB   2.2 KB
        BL2    1.7 KB   2.1 KB
        BL31   2.6 KB   3.3 KB
```

### DIFF
--- a/docs/porting-guide.md
+++ b/docs/porting-guide.md
@@ -535,6 +535,17 @@ optionally be defined:
     PLAT_PARTITION_MAX_ENTRIES	:=	12
     $(eval $(call add_define,PLAT_PARTITION_MAX_ENTRIES))
 
+The following constant is optional. It should be defined to override the default
+behaviour of the `assert()` function (for example, to save memory).
+
+*   **PLAT_LOG_LEVEL_ASSERT**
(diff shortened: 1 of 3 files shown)


antonio-nino-diaz-arm:
I've just pushed a new version of this PR, could you check if it works for you?

sbranden:
@antonio-nino-diaz-arm 
This looks like it will work.

PR #938, opened 2017-05-15
Collect headers shared between TF and host-tools into include/tools_share

masahir0y:


davidcunado-arm:
jenkins: test this please

tools/cert_create/Makefile
```diff
@@ -61,8 +54,7 @@ all: clean ${BINARY}
 
 ${BINARY}: ${OBJECTS} Makefile
 	@echo "  LD      $@"
-	@echo 'const char build_msg[] = "Built : "__TIME__", "__DATE__; \
-                const char platform_msg[] = "${PLAT}";' | \
```
danh-arm: Why have you lost the platform_msg[] initialization here? It is used by the tool for debug output.

antonio-nino-diaz-arm:
The checks have failed because of a false positive in the copyright check script:  the file `include/tools_share/uuid.h` doesn't have the regular ARM copyright header and it's not in the list of exceptions. All other tests are fine.

include/tools_share/tbbr_oid.h
```diff
@@ -1,21 +1,17 @@
...
 
 /*
  * The following is a list of OID values defined and reserved by ARM, which
  * are used to define the extension fields of the certificate structure, as
  * defined in the Trusted Board Boot Requirements (TBBR) specification,
  * ARM DEN0006C-1.
- *
- * Non-ARM platform owners that wish to align with the TBBR should define
- * constants with the same name in their own platform port(s), using their
- * own OIDs obtained from the ITU-T.
```
danh-arm: While this patch makes sense from an implementation point of view, I'm not sure if it's acceptable to allow all TBBR-compliant platforms to use OIDs that are assigned to ARM Limited. I'll clarify this with the TBBR authors...

tools/cert_create/Makefile
masahir0y: With platform_oid.h deprecated, the cert_create tool is no longer platform-dependent.
So, the message "Target platform ..." makes no sense.  What's why.

We should keep host tools platform-independent where possible because:

[1] We can avoid recompiling under tools/ directory for build-test across all platforms

[2] fiptool and cert_create may be useful for some people to manipulate FIP image outside the ATF project.
For example, developers working on U-Boot or EDK2 (=BL33) probably no interest in building ARM Trusted Firmware.   What they want to do is to generate a certificate for BL33 with cert_create and update the the fip.bin with fiptool, then flush the new fip.bin to the board.
So, they may finally decide to provide a package of the host tools for some distributions (such as deb, rpm package).  This will be pretty easy if the host tools have no dependency on a particular platform.

         $ sudo apt-get install atf-tools

Then, pre-built fiptool and cert_create will be installed in /usr/bin directory on their system.
Similar things often happen.  For example, Debian provides a package "u-boot-tools".

If the cert_create depends on a particular platform, this would be difficult.  If cert_create built for ARM Ltd. contains ARM's OID, it would not work for Socionext Inc.

danh-arm: Thanks for the extra info. I understand it's desirable to keep the host tools platform independent where possible. That should be achievable in the environments you care about but I don't think this is always going to be possible (e.g. for non-TBBR compliant platforms or platforms that really do want to use their own OIDs).

The device-side code currently has platform interfaces that enable selection of TBBR related code. Ideally the tools should have the same thing but this hasn't been implemented yet. Although the cert_create tool is currently hardcoded to use TBBR definitions, this may not be the case in future.

Also, I'm concerned about compatibility with platforms that do provide their own platform_oid.h, which would now be ignored by your patch.

I think a compromise solution is to allow the TBBR definitions to be used by default, while still allowing platforms to override if they want to. So something like the following:

- Moving the board_arm_oid.h definitions to include/tools_share/tbbr_oid.h (as you have done) looks OK.

- There needs to be a mechanism for a platforms's platform_oid.h to be used if required. There could be a build flag (e.g. USE_TBBR_DEFS), which controls whether platform_oid.h or tbbr_oid.h is used. This should default to false but could default to true if ERROR_DEPRECATED is also true.

- The upstream platforms can then enable USE_TBBR_DEFS and remove their platform_oid.h, although ideally this should come in a separate patch to prove this is backward compatible.

- cert_create should have the same mechanism. The build flag could also be used to control the platform_msg[] initalisation; when true the message could be hardcoded to "TBBR Generic" or similar.

- Potentially other parts of the codebase could use USE_TBBR_DEFS to control whether TBBR source files are pulled in, instead of leaving this to platforms to define. However, that is a bit off-topic and can come later.

Does that sound OK?

masahir0y: Given the current situation of TBB, is it worthwhile?

As I pointed out in the issue "Question: How much memory is required to enable Trusted Board Boot? #448",  TBBR is too over-implemented to fit in typical SRAM size in competetive mass products.  So, TBBR is designed to work on a $1000 board like Juno or emulators.

If you really believe this is the right thing to do, yes, we could do that by sprinkling #ifdef.
But, what does it buy us? and how long should we carry USE_TBBR_DEFS?

One more thing, even if somebody uses their own platform_oid.h in downstream code, this commit is not a breakage.

If the downstream users follow TBBR, it will work seamlessly.
Only the difference is the OIDs will be replaced with ones assigned for ARM Ltd.
This is not a problem because both TF and cert_create reference tbbr_oid.h
instead of platform_oid.h   The users may not even notice that.

If they do not follow TBBR, they are not using drivers/auth/tbbr/tbbr_cot.c.
Instead, they are using their internal authentication table where
platform_oid.h is not mandatory.   They are probably not using cert_create either
because this tool is so hard-coded.

In either way, I could not get the point of you suggestion.

danh-arm: I refute the suggestion that TBBR is _designed_ to only work on expensive devices. As stated in https://github.com/ARM-software/tf-issues/issues/448, there are are number of things that can be done to improve the implementation but these are not fundamental design issues.

Regarding OID compatibility, it's very difficult to know how the codebase is being used, since most vendor's platform code is not upstream. It's not a simple case of vendors either using the upstream TBBR implementation as-is or not at all. Many vendors pick and choose the code they need, which may or may not result in a non-TBBR compliant implementation. But we still need to try to provide compatibility where possible.

For example, it's possible that some vendors are using the upstream device side code (or a forked version of it) but their own certificate creation tools. Or their device side code version may not be in lock-step with their cert_create version. Silently changing the OIDs under their feet could easily result in authentication failures. Doing this with no migration path is not acceptable (in my opinion). Also, some vendors may simply prefer to use their own OIDs, especially in the case where their various firmware images are owned by different entities.

I'm happy for non-ARM vendors to use the ARM Limited OIDs if they want, and I'm also happy for this to eventually be the default behaviour. But I don't want to force this on all vendors.

Regarding "how long should we carry USE_TBBR_DEFS?", I think some vendors will always need this option, although the default option can change from false to true at some point. That "some point" is likely to be when we make a v2.0 release, at which time we can remove all conditional code under the ERROR_DEPRECATED flag.

masahir0y:
I pushed a new version.

tools/cert_create/Makefile
```diff
@@ -27,6 +28,12 @@ MAKE_HELPERS_DIRECTORY := ../../make_helpers/
 include ${MAKE_HELPERS_DIRECTORY}build_macros.mk
 include ${MAKE_HELPERS_DIRECTORY}build_env.mk
 
+ifeq (${USE_TBBR_DEFS},1)
```
danh-arm: minor: It would be good to add a comment here saying that cert_tool is platform independent in this case

make_helpers/defaults.mk
```diff
@@ -128,6 +128,9 @@ TRUSTED_BOARD_BOOT		:= 0
 # Build option to choose whether Trusted firmware uses Coherent memory or not.
 USE_COHERENT_MEM		:= 1
 
+# Use tbbr_oid.h instead of platform_oid.h
+USE_TBBR_DEFS			:= 0
+
```
masahir0y: Apparently, the conditional does not work as expected in defaults.mk.

ERROR_DEPRECATED can be overridden much later in platform.mk
but it is always 0 in defaults.mk.  So the code you suggested is equivalent to my code.
So, I assume you suggested me to move this code to a different place.

Or, is it better to do like this?

    USE_TBBR_DEFS              = $(ERROR_DEPRECATED)

plat/arm/board/common/board_arm_trusted_boot.c
```diff
@@ -1,13 +1,17 @@
...
  */
 
 #include <arm_def.h>
 #include <assert.h>
 #include <platform.h>
+#if USE_TBBR_DEFS
+#include <tbbr_oid.h>
+#else
 #include <platform_oid.h>
+#endif
```
masahir0y: In the previous feedback, you said
>The upstream platforms can then enable USE_TBBR_DEFS and remove their platform_oid.h, although ideally this should come in a separate patch to prove this is backward compatible.

So, I followed it.

plat/arm/board/fvp/fvp_trusted_boot.c
```diff
@@ -1,13 +1,18 @@
...
  *
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
 #include <assert.h>
+#if USE_TBBR_DEFS
+#include <tbbr_oid.h>
+#else
 #include <platform_oid.h>
+#endif
```
masahir0y: Ditto.  Anybody who has interest can work on it as a separate work, right?

tools/cert_create/Makefile
```diff
@@ -10,6 +10,7 @@ V		:= 0
 DEBUG		:= 0
 BINARY		:= ${PROJECT}${BIN_EXT}
 OPENSSL_DIR	:= /usr
+USE_TBBR_DEFS	?= 1
```
masahir0y: When Make is invoked from the top Makefile, `USE_TBBR_DEFS=${USE_TBBR_DEFS}` is passed correctly down to tools/cert_create/Makefile, so `USE_TBBR_DEFS ?= 1` is not effective.

`USE_TBBR_DEFS ?= 1` was added here in order to avoid build failure when cert_create is built directly in tools/cert_create directory.

You said as follows in pull request #837:
> (that would prevent passing a command line CFLAGS to the tools Makefile when invoked directly (although admittedly this isn't properly supported anyway)

Seemed you liked to cater to such a use-case, so I fixed the build error just in case.

In upstream, building in tools/cert_create never succeeds:

    $ cd tools/cert_create; make
    ../../make_helpers/plat_helpers.mk:56: *** "Error: Invalid platform. The following platforms are available: fvp|juno|mt6795|mt8173|qemu|rk3368|rk3399|tegra|zynqmp".  Stop.

If you prefer to keeping backward compatibility including the build failure, I can change it to `USE_TBBR_DEFS ?= 0` to get the build fail again.

tools/cert_create/Makefile
```diff
@@ -47,6 +55,9 @@ else
 	Q :=
 endif
 
+$(eval $(call add_define,USE_TBBR_DEFS))
+CFLAGS += ${DEFINES}
+
```
masahir0y: I could not understand what you mean.

I think my code passes `USE_TBBR_DEFS` to C files unconditionally.
What is your intention of moving it to line 24?

make_helpers/defaults.mk
danh-arm: Good point. Your alternative code looks better.

plat/arm/board/common/board_arm_trusted_boot.c
danh-arm: I guess I wasn't clear enough. I was suggesting the platform make file (arm_common.mk for the ARM platforms) should set USE_TBBR_DEFS to 1 and use tbbr_oid.h unconditionally in the source files.

plat/arm/board/fvp/fvp_trusted_boot.c
danh-arm: You are right that the ARM platform code changes are not essential. They can be done separately by someone else if you prefer not to.

tools/cert_create/Makefile
danh-arm: OK, I misunderstood what this code is doing. It can stay as-is but It would be good to add a comment explaining that the value passes from the top-level makefile is used by default.

Also, the user guide says "Specifying the platform is mandatory since the tool is platform specific." This is no longer true so could do with updating.

tools/cert_create/Makefile
danh-arm: Sorry, my mistake; I misread the code. Leave this as-is.

make_helpers/defaults.mk
masahir0y: Unfortunately, I found the alternative code did not work with `$(call assert_boolean, ...)`

So, I moved the ERROR_DEPRECATED check to the top Makefile  (after including platform.mk).

tools/cert_create/Makefile
masahir0y: I thought about it a little more.   Currently, the tool can be built directly in tools/cert_create if PLAT is passed from the command line.  As far as I understood, tools/cert_create/Makefile was implemented in that way.

After all I removed `USE_TBBR_DEFS ?= 1` entirely.   The direct build will be possible by passing correct combination PLAT and USE_TBBR_DEFS from the command line.  This does not look very nice, but we can excuse here.

I updated the user guide.

tools/cert_create/Makefile
masahir0y: I did so.

masahir0y:
I updated and pushed.

Makefile
```diff
@@ -354,6 +354,11 @@ ifneq (${FIP_ALIGN},0)
 FIP_ARGS += --align ${FIP_ALIGN}
 endif
 
+# If ERROR_DEPRECATED is defined, use cert_create built with generic TBBR
+ifeq (${ERROR_DEPRECATED},1)
+USE_TBBR_DEFS := 1
+endif
+
```
danh-arm: The problem with this is that it overrides any setting of USE_TBBR_DEFS in platform.mk.

Perhaps you could remove the initialization of USE_TBBR_DEFS in defaults.mk (as this is a special case) and then initialize here only in the case that it has not been previously defined.

So:

```
USE_TBBR_DEFS              ?= $(ERROR_DEPRECATED)
```

masahir0y:
Updated again.  I think I reflected all request.

plat/arm/board/common/board_arm_trusted_boot.c
danh-arm: minor: changing this file both here and in the next patch is unnecessary but the end result is good!

danh-arm:
jenkins: test this please

danh-arm:
Sorry, this failed to build all non-ARM platforms:

```
Makefile:450: *** USE_TBBR_DEFS must be boolean.  Stop.
Build failed with code 2
```

From reading the make docs, it seems the `?=` syntax results in a recursively expanded variable (like '='), whereas we want a simply expanded variable (like `:=`). Perhaps the following will work instead?

```
ifeq ($(origin FOO), undefined)
USE_TBBR_DEFS := ${ERROR_DEPRECATED}
endif
```

Could you try this, and check 1-2 of the other upstream platforms?

Perhaps at the same time you drop the changes to `board_arm_trusted_boot.c` and `fvp_trusted_boot.c` in the 2nd patch (to address my "minor" comment).

masahir0y:
> Could you try this, and check 1-2 of the other upstream platforms?

I think the root cause is the wrong implementation of assert_boolean.
So, I fixed it.  Then, I moved USE_TBBR_DEFS back to defaults.mk


> Perhaps at the same time you drop the changes to board_arm_trusted_boot.c and fvp_trusted_boot.c in the 2nd patch (to address my "minor" comment).

OK, done.

danh-arm:
jenkins: retest this please

danh-arm:
Your change to assert_boolean looks good. Let's see if the CI works now...

PR #940, opened 2017-05-15
Allow TF to be built using clang or ARM Compiler 6

dp-arm:


dp-arm:
Accidentally targeted the master branch, closing.

PR #941, opened 2017-05-15
Allow TF to be built using clang or ARM Compiler 6

dp-arm:


vwadekar:
Can we use this patch and compile TF with clang provided by the Android
build system? Can you provide links to the AC6 compiler?

vwadekar:
@dp-arm 

Hi,

I had started a discussion some time back to move to AC6, so this is a welcome change. Can you please help me understand the end goal? Are we planning to move away from the GCC toolchain? 

-Varun

danh-arm:
Hi @vwadekar 

There are no plans to move away from GCC; we're just diversifying the TF toolchain support. At the moment that just means having the option of using the AC6/LLVM compiler. We may look into the assembler/linker in future but that would be significantly more work for possibly little additional benefit. At the moment, no further work is planned in this area.

Regarding AC6 links, I refer you to the ARM website. This toolchain comes bundled with the DS-5 development environment:
https://developer.arm.com/products/software-development-tools/compilers/arm-compiler

Regards

Dan.

dp-arm:
> Can we use this patch and compile TF with clang provided by the Android
> build system? Can you provide links to the AC6 compiler?

I have tested this using clang 3.9 and 4.0.  You can give it a try and see if it works.

vwadekar:
@dp-arm 

Thanks. I will test locally with the Android build system. You dont have to wait for my ack though.

docs/user-guide.md
```diff
@@ -104,6 +107,18 @@ Download the Trusted Firmware source code from Github:
 
         export CROSS_COMPILE=<path-to-aarch32-gcc>/bin/arm-linux-gnueabihf-
 
+    It is possible to build Trusted Firmware using clang or ARM Compiler 6.
+    To do so `CC` needs to point to the clang or armclang binary. Only the
+    compiler is switched; the assembler and linker need to be provided by
+    the GNU toolchain, thus `CROSS_COMPILE` should be set as described above.
+
+    ARM Compiler 6 will be selected when the base name of the path assigned
```
vwadekar: I'm not sure I understand. Can you please add an example to the docs?

docs/user-guide.md
```diff
@@ -104,6 +107,18 @@ Download the Trusted Firmware source code from Github:
...
+    to `CC` matches the string 'armclang'.
+
+    Clang will be selected when the base name of the path assigned to `CC`
```
vwadekar: Same

Makefile
```diff
@@ -125,11 +125,18 @@ OD			:=	${CROSS_COMPILE}objdump
 NM			:=	${CROSS_COMPILE}nm
 PP			:=	${CROSS_COMPILE}gcc -E
 
-ASFLAGS_aarch64		=	-mgeneral-regs-only
-TF_CFLAGS_aarch64	=	-mgeneral-regs-only -mstrict-align
+ifneq ($(findstring clang,$(notdir $(CC))),)
```
vwadekar: Is the name "clanggcc" right? The value for CC above would turn out to be "clanggcc' AFAICT.

docs/user-guide.md
dp-arm: I will add an example of this in the documentation.

dp-arm:
> HOSTCC should point to clang, correct? The following worked for me:
> 
> make PLAT=tegra TARGET_SOC=t186 DEBUG=1 SPD=trusty CROSS_COMPILE=~/aarch64-linux-gnu_4.9-2015.05/bin/aarch64-linux-gnu- HOSTCC=~/clang/bin/clang bl31

HOSTCC is only used when building host tools (i.e. fiptool, cert_create) and it is optional.  You should be setting CC to point to clang instead.  This will switch the default compiler when building TF itself.  CROSS_COMPILE still needs to be set as that is used to locate other tools like the linker, assembler etc.

The clang linker is not currently used because it does not seem to be able to parse our linker scripts.  We will be looking into this at a later date.

Regarding the assembler, the argument is similar in that some of the assembler macros do not seem to be properly parsed.  There is less of an argument to move to the clang assembler though.

Makefile
dp-arm: The only reason we use findstring here is to support both CC=/foo/bar/clang and CC=/foo/bar/clang-X.Y

So base name of the path should normally be clang or clang-X.Y.

PR #946, opened 2017-05-18
Tegra misra fixes v1

vwadekar:
Patches to fix some defects found by the MISRA scanner.


include/lib/utils_def.h
```diff
@@ -57,16 +57,19 @@
...
- * scripts and assembler files.
+ * The GNU assembler and linker do not support the 'ul' and 'ull' suffix (it
+ * causes the build process to fail) therefore the suffix is omitted when used
+ * in linker scripts and assembler files.
 */
 #if defined(__LINKER__) || defined(__ASSEMBLY__)
+# define  UL(_x)	(_x)
 # define ULL(_x)	(_x)
 #else
+# define  UL(_x)	(_x##ul)
```
antonio-nino-diaz-arm: We avoid using `long` instead of `long long` because it causes problems when compiling for AArch32. In your platform it doesn't matter because it's only built for AArch64, but I'm not sure about adding this to the generic code.

https://github.com/ARM-software/arm-trusted-firmware/wiki/ARM-Trusted-Firmware-Coding-Guidelines#use-of-built-in-c-and-stdlib-data-types

> Avoid use of `long`. This is guaranteed to be at least 32-bit but given that `int` is 32-bit on ARM platforms, there is no use for it. For integers at least 64-bit, use `long long`.

vwadekar: @antonio-nino-diaz-arm 

Is there any #if check we can add to this code so that it works for 32/64 bit architectures?

antonio-nino-diaz-arm: As far as I know, using `ULL()` should be enough because it ensures that the constant is treated as a 64-bit value when compiling for both architectures.

antonio-nino-diaz-arm: I don't think I understand what you are trying to say. For 32-bit variables you need to use 'U', not 'UL', and for 64-bit you need 'ULL'. Can you give me an example of a case in which this would be a problem?

davidcunado-arm:
@vwadekar 
Could you split out the changes to the generic / non-tegra code out into a separate PR?

include/lib/utils_def.h
davidcunado-arm: Yes, agree with @antonio-nino-diaz-arm  - we should avoid using long / unsigned long as this could be ambiguous re: size.
If you need a 4-byte type, then add a U() macro.
Also, looking at the rest of the patch, it seems the changes are inconsistent - some defines are wrapped by the new macro, others have the suffix added.
Perhaps best to migrate to the macro rather than adding a suffix?
(NOTE the generic code is also similarly inconsistent!)

vwadekar:
@davidcunado-arm 

I dont agree that the generic code (1 patch) needs its own PR. We should keep all relevant code together. The generic code is required by Tegra code, so makes sense to keep them as part of one PR.

davidcunado-arm:
@vwadekar 
Ok - but let's at least move them into their own commits.

vwadekar:
@davidcunado-arm 

Will do

vwadekar:
Closing this PR as it was opened against the master branch.

vwadekar:
Started PR #953

PR #949, opened 2017-05-23
Reduce code size when building with Trusted Board Boot enabled

antonio-nino-diaz-arm:
This pull request introduces two changes that reduce the code size of BL1 and BL2 by approximately 4.6 KB when the Trusted Firmware is built with Trusted Board Boot enabled. Other images aren't affected by this change.


antonio-nino-diaz-arm:
Done. Does the new version work for you, @sbranden?

davidcunado-arm:
@sbranden 
Will merge early next week, unless you have further feedback - thanks!

PR #952, opened 2017-05-25
fiptool: Add --pad command line option

dp-arm:
Allow each image to reserve some padding space before alignment.

Use case is --align 0x4000 --pad 64 to make sure sufficient
room is available to prepend (or append) vendor-specific headers.

Helps resolve afaerber/meson-tools#3.

Change-Id: I31e59456600b72cd30cde0f653ec6523480a5823

### DIFF
--- a/tools/fiptool/fiptool.c
+++ b/tools/fiptool/fiptool.c
@@ -28,6 +28,7 @@
 #define OPT_TOC_ENTRY 0
 #define OPT_PLAT_TOC_FLAGS 1
 #define OPT_ALIGN 2
+#define OPT_PAD 3
 
 static int info_cmd(int argc, char *argv[]);
 static void info_usage(void);
@@ -488,7 +489,8 @@ static void info_usage(void)
 	exit(1);
 }
 
-static int pack_images(const char *filename, uint64_t toc_flags, unsigned long align)
+static int pack_images(const char *filename, uint64_t toc_flags,
+	unsigned long align, unsigned long padding)
 {
 	FILE *fp;
 	image_desc_t *desc;
@@ -523,7 +525,7 @@ static int pack_images(const char *filename, uint64_t toc_flags, unsigned long a
 		if (image == NULL)
 			continue;
 		payload_size += image->toc_e.size;
-		entry_offset = (entry_offset + align - 1) & ~(align - 1);
+		entry_offset = (entry_offset + padding + align - 1) & ~(align - 1);
 		image->toc_e.offset_address = entry_offset;
 		*toc_entry++ = image->toc_e;
 		entry_offset += image->toc_e.size;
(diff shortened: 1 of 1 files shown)


dp-arm:
> Thanks, but "dp-arm" is not an acceptable Signed-off-by.

All of my patches use that form.

afaerber:
It violates contributing.md and the Linux DCO.

masahir0y:
> Use case is --align 0x4000 --pad 64 to make sure sufficient
room is available to prepend (or append) vendor-specific headers.

From this statement, it was not clear whether `--pad` means `--pre-pad` or `--post-pad`.

If I understood the source code correctly, it looks like `--post-pad`.
It it OK if you are sure `--pre-pad` is unneeded.

Here, I mean `--pre-pad` aligns the start of extra space,
and `--post-pad` aligns the start of image bodies.

dp-arm:
> Here, I mean --pre-pad aligns the start of extra space,
> and --post-pad aligns the start of image bodies.

Yes, I think it is --post-pad according to this definition.  I guess the reason --pre-pad is not going to be particularly useful is because the image will not run correctly if the start of the image contains extra headers (offsets will be incorrect).  There's no magic done in the TF loader to cope with this situation.

afaerber:
Actually the Amlogic firmware does need a pre-pad, but the proprietary tool (and currently my tool) assumes a post-pad. For a --pre-pad much more code would need to be changed than just the alignment lines. Not sure if it's worth implementing that, given that a simple cat command could achieve the same effect for a fixed size header. On the other hand, by that argument, I could just cat a dd if=/dev/zero of=pad bs=1 count=N file and use unmodified --align and drop the PR. Thoughts?

masahir0y:
> On the other hand, by that argument, I could just cat a dd if=/dev/zero of=pad bs=1 count=N file to the end of my blobs and use unmodified --align and drop the PR. Thoughts

Yeah, I was also thinking this.

Append (or prepend) headers, then put the images into FIP
 will be easier than
Make extra spaces in FIP, then embed headers

tools/fiptool/fiptool.c
```diff
@@ -729,7 +749,8 @@ static void create_usage(void)
 	printf("fiptool create [opts] FIP_FILENAME\n");
 	printf("\n");
 	printf("Options:\n");
-	printf("  --align <value>\t\tEach image is aligned to <value> (default: 1).\n");
+	printf("  --align <bytes>\t\tEach image is aligned to <bytes> (default: 1).\n");
+	printf("  --pad <bytes>\t\t\tEach image is padded by <bytes> (default: 0).\n");
```
danh-arm: minor: Could say "... padded at the end by...". Same for other cases below.

danh-arm:
Hi @afaerber. We don't have a strong view on whether it's better to use `dd` before putting images into the FIP or to use the feature in this PR. If you say this PR satisfies your needs and you will use it, we will merge, otherwise we'll drop it.

danh-arm:
jenkins: retest this please

dp-arm:
@afaerber ping

davidcunado-arm:
@afaerber  
We're still pending your feedback on this PR. Please could you confirm that this PR satisfies your needs and you will use it? 
I'll look to either close it or merge it by Wednesday 28-JUN depending on what your feedback is.

davidcunado-arm:
Closing PR - it has not been confirmed that this change will be used. Please re-open if this PR is still required.

PR #954, opened 2017-05-26
Partition 4kb

hzhuang1:
Fix https://github.com/ARM-software/tf-issues/issues/480


masahir0y:
> Partition block size must match with storage device size. Since
there're different block size in different storage size. For
example, eMMC block size is 512, and UFS block size is 4096.
Replace PARTITION_BLOCK_SIZE by PLAT_PARTITION_BLOCK_SIZE. User
sets the right block size in his platform.

I have not checked the code, but I just wondered this from your log.
If two or more devices are connected, should PLAT_PARTITION_BLOCK_SIZE be set to 512 or 4096?

I think ARM Trusted Firmware mostly decides parameters at compile time by macros and disallows run-time flexibility; for example, if a valid image is not found in storage A, then storage B is checked.

One more thing, I work for a SoC vendor, and we need to support 100 boards (including different boot devices), but I do not want to build  100 images with different set of compile options.  I try my best to support multiple boards/modes with a single image, but it is sometimes difficult due to the framework limitation.  We should consider that when a new feature is added. Thought?

danh-arm:
> If two or more devices are connected, should PLAT_PARTITION_BLOCK_SIZE be set to 512 or 4096?

Good point. It looks like the current API doesn't support multiple devices with differing block sizes. One of the patches here changes `partition_init()` to take a `size` argument, but then only uses that to assert it is less than `PLAT_PARTITION_BLOCK_SIZE`. Could this `size` argument be used for subsequent reads instead of `PLAT_PARTITION_BLOCK_SIZE`? Given that client (platform) code now also sizes the arrays, there should be no need for a platform option at all.

hzhuang1:
On 2017/5/30 22:22, danh-arm wrote:
>     If two or more devices are connected, should
>     PLAT_PARTITION_BLOCK_SIZE be set to 512 or 4096?
> 
> Good point. It looks like the current API doesn't support multiple 
> devices with differing block sizes. One of the patches here changes 
> |partition_init()| to take a |size| argument, but then only uses that to 
> assert it is less than |PLAT_PARTITION_BLOCK_SIZE|. Could this |size| 
> argument be used for subsequent reads instead of 
> |PLAT_PARTITION_BLOCK_SIZE|? Given that client (platform) code now also 
> sizes the arrays, there should be no need for a platform option at all.
> 
I need to check whether partition driver could get block size from io 
block driver. But do we need to support loading partition table from two 
or more block devices at the same time? I think that supporting one 
partition table with multiple block devices is enough. We could 
determine partition table is in which block device at runtime.

danh-arm:
Hi @hzhuang1 . Any progress here? Otherwise I will close.

hzhuang1:
@danh-arm, No resource on it. I think you could close it first. When it's ready, I'll open it again.

davidcunado-arm:
As requested - closing PR, will be re-opened when needed.

PR #963, opened 2017-06-01
Add SCMI power domain and system power protocol support

soby-mathew:


danh-arm:
jenkins: retest this please

PR #970, opened 2017-06-07
Enable pmf rt instr hikey

vingu-linaro:
This patch enable PMF and runtime instrumentation for hikey

### DIFF
--- a/plat/hisilicon/hikey/hisi_sip_svc.c
+++ b/plat/hisilicon/hikey/hisi_sip_svc.c
@@ -1,0 +1,84 @@
+/*
+ * Copyright (c) 2016-2017, ARM Limited and Contributors. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include <hisi_sip_svc.h>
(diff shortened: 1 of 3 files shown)


danh-arm:
jenkins: test this please

danh-arm:
Hi @hzhuang1. Can you ack this? It looks OK to me.

hzhuang1:
Acked.

Are these register definitions only for hikey? Do we have the same register definitions on hikey960? If they're same, maybe we can write it as common code.

danh-arm:
I agree this code could be shared with hikey960 but that will be easier to do after hikey960 is integrated. There's no need to delay this PR.

vingu-linaro:
I haven't look in details hikey960 code but this code doesn't use anything specific to hikey so it should be easy to add hikey960 support in a 2nd step after it has been integrated

danh-arm:
jenkins: retest this please

danh-arm:
We're having infrastructure issues with this...

danh-arm:
jenkins: retest this please

PR #977, opened 2017-06-09
bl32: add secure interrupt handling in AArch32 sp_min

etienne-lms:
Propose a minimal secure interrupt service in sp_min for the AArch32
implementation.

Change-Id: I1f7d1dc7b21f6f90011b7f3fcd921e455592f5e7
Signed-off-by: Etienne Carriere <etienne.carriere@st.com>

### DIFF
--- a/bl32/sp_min/aarch32/entrypoint.S
+++ b/bl32/sp_min/aarch32/entrypoint.S
@@ -18,6 +18,17 @@
 	.globl	sp_min_entrypoint
 	.globl	sp_min_warm_entrypoint
 
+	.macro route_fiq_to_sp_min reg
+		/* -----------------------------------------------------
+		 * FIQs are secure interrupts trapped by Monitor and non
+		 * secure is not allowed to mask the FIQs.
+		 * -----------------------------------------------------
+		 */
+		ldcopr	\reg, SCR
+		orr	\reg, \reg, #SCR_FIQ_BIT
+		bic	\reg, \reg, #SCR_FW_BIT
+		stcopr	\reg, SCR
+	.endm
 
 vector_base sp_min_vector_table
 	b	sp_min_entrypoint
@@ -27,7 +38,7 @@ vector_base sp_min_vector_table
 	b	plat_panic_handler	/* Data abort */
 	b	plat_panic_handler	/* Reserved */
 	b	plat_panic_handler	/* IRQ */
-	b	plat_panic_handler	/* FIQ */
+	b	handle_fiq		/* FIQ */
 
 
 /*
@@ -91,6 +102,10 @@ func sp_min_entrypoint
 	mov	r0, #0
 	mov	r1, #0
 #endif /* RESET_TO_SP_MIN */
+
+#if SP_MIN_WITH_SECURE_FIQ
+	route_fiq_to_sp_min r4
+#endif
 
 	bl	sp_min_early_platform_setup
 	bl	sp_min_plat_arch_setup
@@ -166,6 +181,44 @@ func handle_smc
 endfunc handle_smc
 
 /*
+ * Secure Interrupts handling function for SP_MIN.
+ */
+func handle_fiq
+#if !SP_MIN_WITH_SECURE_FIQ
+	b plat_panic_handler
+#else
+	/* FIQ has a +4 offset for lr compared to preferred return address */
+	sub	lr, lr, #4
+	/* On SMC entry, `sp` points to `smc_ctx_t`. Save `lr`. */
+	str	lr, [sp, #SMC_CTX_LR_MON]
+
+	smcc_save_gp_mode_regs
+
+	/*
+	 * AArch32 architectures need to clear the exclusive access when
+	 * entering Monitor mode.
+	 */
+	clrex
+
+	/* load run-time stack */
+	mov	r2, sp
+	ldr	sp, [r2, #SMC_CTX_SP_MON]
+
+	/* Switch to Secure Mode */
+	ldr	r0, [r2, #SMC_CTX_SCR]
+	bic	r0, #SCR_NS_BIT
+	stcopr	r0, SCR
+	isb
+
+	push	{r2, r3}
+	bl	sp_min_fiq
+	pop	{r0, r3}
+
+	b	sp_min_exit
+#endif
+endfunc handle_fiq
+
+/*
  * The Warm boot entrypoint for SP_MIN.
  */
 func sp_min_warm_entrypoint
@@ -212,6 +265,10 @@ func sp_min_warm_entrypoint
 	 */
 	mov	r0, #DISABLE_DCACHE
 	bl	bl32_plat_enable_mmu
+
+#if SP_MIN_WITH_SECURE_FIQ
+	route_fiq_to_sp_min r0
+#endif
 
 #if HW_ASSISTED_COHERENCY || WARMBOOT_ENABLE_DCACHE_EARLY
 	ldcopr	r0, SCTLR
--- a/bl32/sp_min/sp_min_main.c
+++ b/bl32/sp_min/sp_min_main.c
@@ -207,3 +207,19 @@ void sp_min_warm_boot(void)
 	copy_cpu_ctx_to_smc_stx(get_regs_ctx(cm_get_context(NON_SECURE)),
 			next_smc_ctx);
 }
+
+#if SP_MIN_WITH_SECURE_FIQ
+/******************************************************************************
+ * This function is invoked on secure interrupts. By construction of the
+ * SP_MIN, secure interrupts can only be handled when core executes in non
+ * secure state.
+ *****************************************************************************/
+void sp_min_fiq(void)
+{
+	uint32_t id;
+
+	id = plat_ic_acknowledge_interrupt();
+	sp_min_plat_fiq_handler(id);
+	plat_ic_end_of_interrupt(id);
+}
+#endif /* SP_MIN_WITH_SECURE_FIQ */
--- a/include/bl32/sp_min/platform_sp_min.h
+++ b/include/bl32/sp_min/platform_sp_min.h
@@ -17,4 +17,7 @@ void sp_min_plat_runtime_setup(void);
 void sp_min_plat_arch_setup(void);
 entry_point_info_t *sp_min_plat_get_bl33_ep_info(void);
 
+/* Platforms that enable SP_MIN_WITH_SECURE_FIQ shall implement this api */
+void sp_min_plat_fiq_handler(uint32_t id);
+
 #endif /* __PLATFORM_SP_MIN_H__ */
--- a/bl32/sp_min/sp_min.mk
+++ b/bl32/sp_min/sp_min.mk
@@ -37,3 +37,9 @@ endif
 RESET_TO_SP_MIN	:= 0
 $(eval $(call add_define,RESET_TO_SP_MIN))
 $(eval $(call assert_boolean,RESET_TO_SP_MIN))
+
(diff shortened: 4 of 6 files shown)


etienne-lms:
my apologies, this change is buggy, I did not rebase it.
fix is coming...

bl32/sp_min/aarch32/entrypoint.S
```diff
@@ -176,24 +176,28 @@ func handle_fiq
...
-	push	{r4}			/* push SCR */
-	bic	r0, r4, #SCR_NS_BIT
-	stcopr	r0, SCR
+	ldr	r1, [r2, #SMC_CTX_SCR]
+	bic	r1, #SCR_NS_BIT
+	stcopr	r1, SCR
 	isb
 
-	and 	r0, r4, #SECURE
+	push	{r2}
```
dp-arm: Don't we need 8 byte alignment here? (push some other dummy reg as well)

etienne-lms: true, i'll update. thanks.

etienne-lms:
Squashed + rebased + fixed from review comment (alignment issue on stack usage).

soby-mathew:
@etienne-lms , we would like to understand more on the use case for FIQ handler in SP_MIN. The original intention of SP_MIN was to serve as a minimal image for just PSCI. Adding the FIQ handler now increases the scope and we want to understand the direction we are heading with this.

Also, the function `sp_min_fiq` takes `security_state` as an argument. Is that added to follow exisiting function prototypes or does it mean you need to allow interrupts to be triggered when executing in Secure mode ?

Also we need to understand where the responsibility of programming the SCR.FIQ bit lies (platform layer or generic layer).

etienne-lms:
@soby-mathew, the idea is to allow one to use the sp_min as a very basic runtime secure service stack, but not only the PSCI stack. Such service would indeed be quite simple since operating mode of the sp_min. Among such basic services could be handling some specific event we don't want to rely on "less safe" nonsecure world: forcing a platform reset or wiping some data from in secure ram on specific HW event detection (i.e some hw tampering).
Maybe such would require a specific configuration directive ? (`SP_MIN_WITH_SECURE_ISR=y`)

`security_state` handling in proposed `sp_min_fiq()` only aims at complying with generic prototype of `plat_sp_min_interrupt_handler()`. It is not expected to be meaningful and could be arbitrary force when calling `plat_sp_min_interrupt_handler()`, instead of being relayed from the FIQ root handler.

Regarding the setting of SCR.FIQ, we planned to configure it from the platform, but it could be set from the sp_min boot entrypoints.

soby-mathew:
Thanks @etienne-lms. I think we can clarify the following regarding the interrupt handler :
* SP_MIN can only handle FIQs. 
* It cannot pre-empt the secure world and the FIQ will only ever interrupt when executing in Normal world. So the FIQ handler need not save the secure register context.
* The build flag (SP_MIN_WITH_SECURE_ISR) will wrap all the code related to this.

The userguide would need updation with details of the build flag.

> security_state handling in proposed sp_min_fiq() only aims at complying with generic prototype of plat_sp_min_interrupt_handler(). It is not expected to be meaningful and could be arbitrary force when calling plat_sp_min_interrupt_handler(), instead of being relayed from the FIQ root handler.

OK, we can force this flag to be `non-secure` from generic code.

> Regarding  the setting of SCR.FIQ, we planned to configure it from the platform, but it could be set from the sp_min boot entrypoints

Yes, preferably the SCR.FIQ bit is set at both cold boot and warm boot entrypoints depending on the build flag.

etienne-lms:
Ok @soby-mathew. Thanks for the feedback. I will update the P-R accordingly.

Since this is restricted to FIQ, I propose to rename the build directive to `SP_MIN_WITH_SECURE_FIQ`.

By the way, do you prefer `SP_MIN_WITH_SECURE_ISR=y|n` or `SP_MIN_WITH_SECURE_ISR=0|1` ?

dp-arm:
> By the way, do you prefer SP_MIN_WITH_SECURE_ISR=y|n or SP_MIN_WITH_SECURE_ISR=0|1
I would go with the latter form as it is more consistent with what we use currently for other options.

etienne-lms:
P-R updates squashed and rebased.

bl32/sp_min/sp_min_main.c
```diff
@@ -207,3 +207,25 @@ void sp_min_warm_boot(void)
...
+	id = plat_ic_acknowledge_interrupt();
+	plat_sp_min_interrupt_handler(id);
+	plat_ic_end_of_interrupt(id);
+}
+
+#pragma weak plat_sp_min_interrupt_handler
+void plat_sp_min_interrupt_handler(uint32_t id)
+{
+	VERBOSE("[sp_min] secure interrupt %d\n", (int)id);
+}
```
danh-arm: I don't see the value in this weak definition. The feature has no value unless the platform provides a useful implementation. I suggest removing to force the platform to provide an implementation.

include/bl32/sp_min/platform_sp_min.h
```diff
@@ -17,4 +17,6 @@ void sp_min_plat_runtime_setup(void);
 void sp_min_plat_arch_setup(void);
 entry_point_info_t *sp_min_plat_get_bl33_ep_info(void);
 
+void plat_sp_min_interrupt_handler(uint32_t id);
```
danh-arm: The other functions in this file are called `sp_min_plat_xxx()`. You should align with this. Also wouldn't `sp_min_plat_fiq_handler()` be more accurate?

include/common/aarch32/el3_common_macros.S
```diff
@@ -195,6 +195,18 @@
...
+	 * FIQs are secure interrupts trapped by Monitor and non
+	 * secure is not allowed to mask the FIQs.
+	 * -----------------------------------------------------
+	 */
+	ldcopr	r0, SCR
+	orr	r0, r0, #SCR_FIQ_BIT
+	bic	r0, r0, #SCR_FW_BIT
+	stcopr	r0, SCR
+#endif
+
```
danh-arm: The `el3_entrypoint_common` macro is also used by BL1. This code would be better placed in `sp_min_entrypoint`.

bl32/sp_min/aarch32/entrypoint.S
```diff
@@ -166,6 +166,44 @@ func handle_smc
...
+#else
+	/* FIQ has a +4 offset for lr compared to preferred return address */
+	sub	lr, lr, #4
+	/* On SMC entry, `sp` points to `smc_ctx_t`. Save `lr`. */
+	str	lr, [sp, #SMC_CTX_LR_MON]
+
+	smcc_save_gp_mode_regs
+
+	/*
+	 * ARMv7 architectures need to clear the exclusive access when
```
etienne-lms: ok.

bl32/sp_min/sp_min_main.c
etienne-lms: Ok

include/bl32/sp_min/platform_sp_min.h
etienne-lms: right and true. i will update. thanks.

include/common/aarch32/el3_common_macros.S
etienne-lms: right, my fault, sorry. I see 2 ways of implementing this, which do you prefer ?
- either add the sequence at 2 places in bl32/sp_min/aarch32/entrypoint.S,
in `sp_min_entrypoint` and `sp_min_warm_entrypoint`.
- or condition this to `#if defined(IMAGE_BL32) && SP_MIN_WITH_SECURE_FIQ`

danh-arm: I think the latter is fine. It's better to avoid duplicated code.

etienne-lms:
Rebased, comments addressed.
Updated the doc change:
```patch
+   Platforms that enable FIQ handling in SP_MIN shall implement
+   the api `sp_min_plat_fiq_handler()`.
```

include/common/aarch32/el3_common_macros.S
```diff
@@ -195,6 +195,18 @@
 	cps	#MODE32_mon
 	isb
 
+#if defined(IMAGE_BL32) && SP_MIN_WITH_SECURE_FIQ
```
soby-mathew: nit: Would prefer to avoid SP_MIN build flags in el3_common macros. Perhaps an additional argument to this macro (_route_fiq?) and SP_MIN entry point can encode this argument based on SP_MIN_WITH_SECURE_FIQ flag ?

Please dont do the rework yet. Let me confirm that the suggestion is alright.

etienne-lms: ok. i see your idea. 
If you feel this implementation better suites the atf generic way, I am fine.

A drawback a see is to clearly define what such `_route_fiq` argument would mean on AArch64 version of the macro `el3_entrypoint_common`. And digressions on fiq/irq/where-are-the-secure-interrupts.

What about a new directive `BL32_ROUTE_SECURE_FIQ=undefined|0|1` that would be set from `SP_MIN_WITH_SECURE_FIQ` in the **make_helpers/default.mk**. **el3_common_macros.S** would rely on `BL32_ROUTE_SECURE_FIQ` not `SP_MIN_WITH_SECURE_FIQ`.

etienne-lms: Following discussion from https://github.com/ARM-software/arm-trusted-firmware/pull/977#pullrequestreview-46814003, i will move back this SCR init so that it remains in the sp_min in **bl32/sp_min/aarch32/entrypoint.S**.

etienne-lms:
Thanks for the feedback. Yes, i will squash and update the P-R.

Makefile
```diff
@@ -455,6 +455,7 @@ $(eval $(call assert_boolean,USE_COHERENT_MEM))
 $(eval $(call assert_boolean,USE_TBBR_DEFS))
 $(eval $(call assert_boolean,WARMBOOT_ENABLE_DCACHE_EARLY))
 $(eval $(call assert_boolean,ENABLE_SPE_FOR_LOWER_ELS))
+$(eval $(call assert_boolean,SP_MIN_WITH_SECURE_FIQ))
```
soby-mathew: Sorry, I didn't notice this before. This check has to go into sp_min.mk.

Makefile
```diff
@@ -495,6 +496,7 @@ $(eval $(call add_define,USE_COHERENT_MEM))
 $(eval $(call add_define,USE_TBBR_DEFS))
 $(eval $(call add_define,WARMBOOT_ENABLE_DCACHE_EARLY))
 $(eval $(call add_define,ENABLE_SPE_FOR_LOWER_ELS))
+$(eval $(call add_define,SP_MIN_WITH_SECURE_FIQ))
```
soby-mathew: Same comment as above.

make_helpers/defaults.mk
```diff
@@ -119,6 +119,10 @@ SPD				:= none
 # image. This is meant to help debugging the post-BL2 phase.
 SPIN_ON_BL1_EXIT		:= 0
 
+# Flag to allow SP_MIN to handle FIQ interrupts in monitor mode. The platform
+# port is free to override this value. It is default disabled.
+SP_MIN_WITH_SECURE_FIQ 		:= 0
+
```
soby-mathew: This has to go to `sp_min.mk`.

davidcunado-arm:
@etienne-lms 
Need to re-base following doc migration to .rst format. Also, I think there are some comments from Soby that still need addressing.
Thanks!

danh-arm:
jenkins: test this please

PR #986, opened 2017-06-15
CSS:Fix scpi "get_power_state" when ARM_PLAT_MT is set

jagadeeshujja:
The ARM_PLAT_MT bit enables the support for MT bit in
MPIDR format. This means that the level 0 affinity
represents the thread and CPU / Cluster levels are
at affinity level 1 and 2 respectively.
This was not catered for in the scpi 'css_scp_get_power_state, API.
Since the SCPI driver can only cater for single threaded CPUs,
this patch fixes the problem by catering for this shift by
effectively ignoring the Thread (level 0) affinity level.

Change-Id: If44f55c9fb2773c8d3f8a9bbcf5420a6f7409dfe
Signed-off-by: jagadeesh ujja <jagadeesh.ujja@arm.com>


davidcunado-arm:
jenkins: test this please

davidcunado-arm:
jenkins: add to whitelist

PR #988, opened 2017-06-19
plat: Hikey960: fix the CPU hotplug

Leo-Yan:
In CPU off callback function, the old code uses the function
hisi_test_pwrdn_allcores() to check if all CPUs in cluster have been
powered off and if it's valid then power off the whole cluster. But the
function hisi_test_pwrdn_allcores() only maintains the different power
states only for CPU suspend/resume flow, so it cannot return correct
states for CPU on/off flow.

This patch is to change use hisi_test_cpu_down() to check if all CPUs
have been powered off, so that can power off the whole cluster properly
when all CPUs in cluster have been hotplugged off.

Signed-off-by: Tao Wang <kevin.wangtao@hisilicon.com>

### DIFF
--- a/plat/hisilicon/hikey960/hikey960_pm.c
+++ b/plat/hisilicon/hikey960/hikey960_pm.c
@@ -102,7 +102,7 @@ void hikey960_pwr_domain_off(const psci_power_state_t *target_state)
 	hisi_powerdn_core(cluster, core);
 
 	/* check if any core is powered up */
-	if (hisi_test_pwrdn_allcores(cluster, core)) {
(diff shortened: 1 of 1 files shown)


danh-arm:
jenkins: test this please

davidcunado-arm:
jenkins: test this please

Leo-Yan:
@davidcunado-arm Just ensure I don't miss anything,  is this a command to trigger Jenkins for automatic testing? :)

davidcunado-arm:
@Leo-Yan 
Yes, this is to trigger our internal testing sysrtem, but can only be done by arm team members.

PR #989, opened 2017-06-20
Adding coreboot support

jwerner-chromium:
This patch series adds code that allows Trusted Firmware to more deeply integrate with coreboot on platforms where coreboot is acting as BL2. The most important feature for now is access to coreboot's in-memory firmware console to log boot messages and crash dumps, which helps triaging issues on systems that don't have a serial console hooked up. It is only used by the Rockchip platform for now, but most of the code is generic and can be used by all future coreboot-based platforms.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -276,6 +276,11 @@ endif
 # This can be overridden by the platform.
 include lib/cpus/cpu-ops.mk
 
+ifeq (${COREBOOT},1)
+# Include coreboot support library when building with coreboot as BL2
+include lib/coreboot/coreboot.mk
+endif
+
 ifeq (${ARCH},aarch32)
 NEED_BL32 := yes
 
@@ -330,6 +335,12 @@ endif
 # use USE_COHERENT_MEM. Require that USE_COHERENT_MEM must be set to 0 too.
 ifeq ($(HW_ASSISTED_COHERENCY)-$(USE_COHERENT_MEM),1-1)
 $(error USE_COHERENT_MEM cannot be enabled with HW_ASSISTED_COHERENCY)
+endif
+
+ifeq (${COREBOOT},1)
+	ifneq (${ARCH},aarch64)
+	$(error "coreboot only supports Trusted Firmware on AArch64.")
+	endif
 endif
 
 ################################################################################
@@ -428,6 +439,7 @@ endif
 ################################################################################
 
 $(eval $(call assert_boolean,COLD_BOOT_SINGLE_CPU))
+$(eval $(call assert_boolean,COREBOOT))
 $(eval $(call assert_boolean,CREATE_KEYS))
 $(eval $(call assert_boolean,CTX_INCLUDE_AARCH32_REGS))
 $(eval $(call assert_boolean,CTX_INCLUDE_FPREGS))
@@ -470,6 +482,7 @@ $(eval $(call add_define,ARM_ARCH_MAJOR))
 $(eval $(call add_define,ARM_ARCH_MINOR))
 $(eval $(call add_define,ARM_GIC_ARCH))
 $(eval $(call add_define,COLD_BOOT_SINGLE_CPU))
+$(eval $(call add_define,COREBOOT))
 $(eval $(call add_define,CTX_INCLUDE_AARCH32_REGS))
 $(eval $(call add_define,CTX_INCLUDE_FPREGS))
 $(eval $(call add_define,ENABLE_ASSERTIONS))
--- a/docs/porting-guide.rst
+++ b/docs/porting-guide.rst
@@ -2445,9 +2445,12 @@ of the CPU to enable quick crash analysis and debugging. It requires that a
 console is designated as the crash console by the platform which will be used to
 print the register dump.
 
-The following functions must be implemented by the platform if it wants crash
-reporting mechanism in BL31. The functions are implemented in assembly so that
-they can be invoked without a C Runtime stack.
+By default, the definitions in `plat/common/aarch64/platform\_helpers.S` will
+cause the crash output to be routed over the same console that was registered
+for boot messages. If necessary, platforms may supply their own definitions of
+the following functions to define special crash output behavior. The functions
+are implemented in assembly so that they can be invoked without a C Runtime
+stack.
 
 Function : plat\_crash\_console\_init
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -2480,12 +2483,11 @@ Function : plat\_crash\_console\_flush
 ::
 
     Argument : void
-    Return   : int
+    Return   : void
 
 This API is used by the crash reporting mechanism to force write of all buffered
 data on the designated crash console. It should only use general purpose
-registers x0 and x1 to do its work. The return value is 0 on successful
-completion; otherwise the return value is -1.
+registers x0 and x1 to do its work.
 
 Build flags
 -----------
--- a/drivers/console/aarch64/console.S
+++ b/drivers/console/aarch64/console.S
@@ -3,103 +3,240 @@
  *
  * SPDX-License-Identifier: BSD-3-Clause
  */
-#include <asm_macros.S>
 
-	.globl	console_init
+#include <console_macros.S>
+
+	.globl	console_register
+	.globl	console_unregister
 	.globl	console_uninit
+	.globl	console_start_crash
 	.globl	console_putc
 	.globl	console_getc
 	.globl	console_flush
 
 	/*
-	 *  The console base is in the data section and not in .bss
-	 *  even though it is zero-init. In particular, this allows
+	 *  The console list pointer is in the data section and not in
+	 *  .bss even though it is zero-init. In particular, this allows
 	 *  the console functions to start using this variable before
 	 *  the runtime memory is initialized for images which do not
 	 *  need to copy the .data section from ROM to RAM.
 	 */
-.section .data.console_base ; .align 3
-	console_base: .quad 0x0
+.section .data.console_list ; .align 3
+	console_list: .quad 0x0
+.section .data.console_state ; .align 0
+	console_state: .byte (1 << CONSOLE_FLAG_BOOT_SHIFT)
 
+	/* -----------------------------------------------
+	 * void console_register(console_info_t *console)
+	 * Function to insert a new console structure into
+	 * the console list. Should usually only be called
+	 * by declare_console expansions.
+	 * In : x0 - address of console structure
+	 * Clobber list: x0, x1, x14
+	 * -----------------------------------------------
+	 */
+func console_register
+	adrp	x14, console_list
+	ldr	x1, [x14, :lo12:console_list]	/* X1 = first struct in list */
+	str	x0, [x14, :lo12:console_list]	/* list head = new console */
+	str	x1, [x0]			/* new console next ptr = X1 */
+	ret
+endfunc console_register
+
+#if !ERROR_DEPRECATED
+	/* Set up compatiblity symbols for platforms using old console API. */
+	declare_console core
+
+	.globl	console_init
 	/* -----------------------------------------------
 	 * int console_init(uintptr_t base_addr,
 	 * unsigned int uart_clk, unsigned int baud_rate)
-	 * Function to initialize the console without a
-	 * C Runtime to print debug information. It saves
-	 * the console base to the data section.
+	 * Deprecated legacy function for initializing a
+	 * single console. Current platforms should call
+	 * console_<driver>_register for the relevant
+	 * driver instead. This needs to clear the console
+	 * list first because some legacy platforms called
+	 * console_init more than once and expect it to
+	 * overwrite previous console state.
 	 * In: x0 - console base address
 	 *     w1 - Uart clock in Hz
 	 *     w2 - Baud rate
-	 * out: return 1 on success else 0 on error
-	 * Clobber list : x1 - x4
+	 * Clobber list : x0 - x4, x14, x15
 	 * -----------------------------------------------
 	 */
 func console_init
-	/* Check the input base address */
-	cbz	x0, init_fail
-	adrp	x3, console_base
-	str	x0, [x3, :lo12:console_base]
-	b	console_core_init
-init_fail:
-	ret
+	adrp	x14, console_list
+	str	xzr, [x14, :lo12:console_list]
+	b	console_core_register
 endfunc console_init
+#endif
+
+	/* -----------------------------------------------
+	 * console_info_t *console_unregister(
+	 *                        console_info_t *console)
+	 * Function to find a specific console in the list
+	 * of currently active consoles and remove it.
+	 * Should usually only be called by
+	 * declare_console expansions.
+	 * In: x0 - address of console structure to remove
+	 * Out: x0 - removed address, or NULL if not found
+	 * Clobber list: x0, x1, x14
+	 * -----------------------------------------------
+	 */
+func console_unregister
+	adrp	x14, console_list
+	add	x14, x14, :lo12:console_list	/* X14 = ptr to first struct */
+	ldr	x1, [x14]			/* X3 = first struct */
+
+unregister_loop:
+	cbz	x1, unregister_not_found
+	cmp	x0, x1
+	b.eq	unregister_found
+	ldr	x14, [x14]			/* X14 = next ptr of struct */
+	ldr	x1, [x14]			/* X1 = next struct */
+	b	unregister_loop
+
+unregister_found:
+	ldr	x1, [x1]			/* X1 = next struct */
+	str	x1, [x14]			/* prev->next = cur->next */
+	ret
+
+unregister_not_found:
+	mov	x0, #0				/* return NULL if not found */
+	ret
+endfunc console_unregister
 
 	/* -----------------------------------------------
 	 * void console_uninit(void)
-	 * Function to finish the use of console driver.
-	 * It sets the console_base as NULL so that any
-	 * further invocation of `console_putc` or
-	 * `console_getc` APIs would return error.
+	 * Function to switch from boot to runtime state.
+	 * The name is misleading for compatibility with
+	 * the legacy API -- this used to uninitialize all
+	 * consoles, but now it just quiesces consoles
+	 * that are not marked as "runtime".
+	 * Clobber list: x0, x1
 	 * -----------------------------------------------
 	 */
 func console_uninit
-	mov	x0, #0
-	adrp	x3, console_base
-	str	x0, [x3, :lo12:console_base]
+	mov	w0, #(1 << CONSOLE_FLAG_RUNTIME_SHIFT)
+	adrp	x1, console_state
+	strb	w0, [x1, :lo12:console_state]
 	ret
 endfunc console_uninit
 
+	/* -----------------------------------------------
+	 * int console_start_crash(void)
+	 * Function to switch to crash state. Returns 1 in
+	 * x0 for convenience with plat_crash_console_init
+	 * Out: x0 - always 1
+	 * Clobber list: x0, x1
+	 * -----------------------------------------------
+	 */
+func console_start_crash
+	mov	w0, #(1 << CONSOLE_FLAG_CRASH_SHIFT)
+	adrp	x1, console_state
+	strb	w0, [x1, :lo12:console_state]
+	mov	x0, #1
+	ret
+endfunc console_start_crash
+
 	/* ---------------------------------------------
-	 * int console_putc(int c)
-	 * Function to output a character over the
-	 * console. It returns the character printed on
-	 * success or -1 on error.
+	 * void console_putc(int c)
+	 * Function to output a character. Calls all
+	 * console's putc() handlers in succession.
 	 * In : x0 - character to be printed
-	 * Out : return -1 on error else return character.
-	 * Clobber list : x1, x2
+	 * Clobber list : x1, x2, x14, x15
 	 * ---------------------------------------------
 	 */
 func console_putc
-	adrp	x2, console_base
-	ldr	x1, [x2, :lo12:console_base]
-	b	console_core_putc
+	mov	x15, x30
+	adrp	x14, console_list
+	ldr	x14, [x14, :lo12:console_list]	/* X14 = first console struct */
+
+putc_loop:
+	cbz	x14, putc_done
+	adrp	x1, console_state
+	ldrb	w1, [x1, :lo12:console_state]
+	ldr	x2, [x14, #CONSOLE_INFO_FLAGS]
+	tst	w1, w2
+	b.eq	putc_continue
+	ldr	x1, [x14, #CONSOLE_INFO_BASE]
+	ldr	x2, [x14, #CONSOLE_INFO_PUTC]
+	cbz	x2, putc_continue
+	blr	x2
+putc_continue:
+	ldr	x14, [x14]			/* X14 = next struct */
+	b	putc_loop
+
+putc_done:
+	ret	x15
 endfunc console_putc
 
 	/* ---------------------------------------------
 	 * int console_getc(void)
-	 * Function to get a character from the console.
-	 * It returns the character grabbed on success
-	 * or -1 on error.
-	 * Clobber list : x0, x1
+	 * Function to get a character from any console.
+	 * Keeps looping through all consoles' getc()
+	 * handlers until one of them returns a
+	 * character, then stops iterating and returns
+	 * that character to the caller.
+	 * Out : x0 - read character, or -1 if none
+	 * Clobber list : x0, x1, x14, x15
 	 * ---------------------------------------------
 	 */
 func console_getc
-	adrp	x1, console_base
-	ldr	x0, [x1, :lo12:console_base]
-	b	console_core_getc
+	mov	x15, x30
+getc_try_again:
+	adrp	x14, console_list
+	ldr	x14, [x14, :lo12:console_list]	/* X14 = first console struct */
+
+getc_loop:
+	cbz	x14, getc_try_again		/* if list ended, start again */
+	adrp	x0, console_state
+	ldrb	w0, [x0, :lo12:console_state]
+	ldr	x1, [x14, #CONSOLE_INFO_FLAGS]
+	tst	w0, w1
+	b.eq	getc_continue
+	ldr	x0, [x14, #CONSOLE_INFO_BASE]
+	ldr	x1, [x14, #CONSOLE_INFO_GETC]
+	cbz	x1, getc_continue
+	blr	x1
+	cmp	x0, #0				/* if X0 >= 0: return */
+	b.ge	getc_done
+getc_continue:
+	ldr	x14, [x14]			/* X14 = next struct */
+	b	getc_loop
+
+getc_found:
+	ret	x15
 endfunc console_getc
 
 	/* ---------------------------------------------
-	 * int console_flush(void)
+	 * void console_flush(void)
 	 * Function to force a write of all buffered
-	 * data that hasn't been output. It returns 0
-	 * upon successful completion, otherwise it
-	 * returns -1.
-	 * Clobber list : x0, x1
+	 * data that hasn't been output. Calls all
+	 * console's flush() handlers in succession.
+	 * Clobber list : x0, x1, x2, x3, x14, x15
 	 * ---------------------------------------------
 	 */
 func console_flush
-	adrp	x1, console_base
-	ldr	x0, [x1, :lo12:console_base]
-	b	console_core_flush
+	mov	x15, x30
+	adrp	x14, console_list
+	ldr	x14, [x14, :lo12:console_list]	/* X14 = first console struct */
+
+flush_loop:
+	cbz	x14, flush_done
+	adrp	x1, console_state
+	ldrb	w1, [x1, :lo12:console_state]
+	ldr	x2, [x14, #CONSOLE_INFO_FLAGS]
+	tst	w1, w2
+	b.eq	flush_continue
+	ldr	x1, [x14, #CONSOLE_INFO_BASE]
+	ldr	x2, [x14, #CONSOLE_INFO_FLUSH]
+	cbz	x2, flush_continue
+	blr	x2
+flush_continue:
+	ldr	x14, [x14]			/* X14 = next struct */
+	b	flush_loop
+
+flush_done:
+	ret	x15
 endfunc console_flush
--- a/include/common/aarch64/console_macros.S
+++ b/include/common/aarch64/console_macros.S
@@ -1,0 +1,157 @@
+/*
+ * Copyright (c) 2017, ARM Limited and Contributors. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+#ifndef __CONSOLE_MACROS_S__
+#define __CONSOLE_MACROS_S__
+
+#include <asm_macros.S>
+#include <console.h>
+
+/*
+ * This file defines macros for use by console drivers that want to hook into
+ * the generic console API. Each console driver needs to include this file, add
+ * a declare_console directive and implement at least func_console_init. All
+ * other functions are optional. Functions need to be declared through these
+ * macros to provide backwards-compatiblity with the previous console API for
+ * older platform code. All functions need to take care to obey the register
+ * restrictions outlined below and not change any register not explicitly listed
+ * in the clobber list. In addition to the listed registers, implementations are
+ * always allowed to use the intra-procedure-call registers (X16 and X17) as
+ * temporary scratch registers within a single function.
+ */
+
+	/* ------------------------------------------------
+	 * int console_<driver>_init(uintptr_t baseaddr,
+	 *                           ...args...)
+	 * Initializes the console. Will be called directly
+	 * from console_<driver>_register and receive any
+	 * arguments in x0-x7 unchanged from it. x0 is
+	 * intended to be a base address pointer, but can
+	 * be used for any purpose. It will be stored in
+	 * the console structure and passed in to every
+	 * other callback for this console.
+	 * In: x0 - driver-specific base address
+	 *     x1...x7 - optional extra arguments
+	 * Out: x0 - 1 for success, 0 for error
+	 * Clobber list: x0 - x7
+	 * ------------------------------------------------
+	 */
+.macro func_console_init _driver
+	func console_\_driver\()_init
+	.globl console_core_init
+	.weak console_core_init
+	console_core_init:
+.endm
+
+	/* ------------------------------------------------
+	 * void console_<driver>_putc(int c, uintptr_t base)
+	 * Outputs the character c on the console. Has no
+	 * return value and must preserve x0 even on error.
+	 * In: x0 - character to output
+	 *     x1 - baseaddr passed in on initialization
+	 * Clobber list: x1, x2 (NOT x0!!!)
+	 * ------------------------------------------------
+	 */
+.macro func_console_putc _driver
+	func console_\_driver\()_putc
+	.globl console_core_putc
+	.weak console_core_putc
+	console_core_putc:
+.endm
+
+	/* ------------------------------------------------
+	 * int console_<driver>_getc(uintptr_t baseaddr)
+	 * Reads a character from the console.
+	 * In: x0 - baseaddr passed in on initialization
+	 * Out: x0 - read character, or -1 if not available
+	 * Clobber list: x0, x1
+	 * ------------------------------------------------
+	 */
+.macro func_console_getc _driver
+	func console_\_driver\()_getc
+	.globl console_core_getc
+	.weak console_core_getc
+	console_core_getc:
+.endm
+
+	/* ------------------------------------------------
+	 * void console_<driver>_flush(uintptr_t baseaddr)
+	 * Write out all buffered data. Should only return
+	 * after the buffers have been completely flushed
+	 * (or on an unrecoverable error condition).
+	 * In: x0 - baseaddr passed in on initialization
+	 * Clobber list: x0, x1, x2, x3
+	 * ------------------------------------------------
+	 */
+.macro func_console_flush _driver
+	func console_\_driver\()_flush
+	.globl console_core_flush
+	.weak console_core_flush
+	console_core_flush:
+.endm
+
+.macro declare_console _driver _boot=1 _runtime=0 _crash=1
+	.section .data.console_\_driver\()_info
+	.align 3
+	console_\_driver\()_info:
+	.quad 0x0
+	.quad (\_boot << CONSOLE_FLAG_BOOT_SHIFT) | \
+	      (\_runtime << CONSOLE_FLAG_RUNTIME_SHIFT) | \
+	      (\_crash << CONSOLE_FLAG_CRASH_SHIFT)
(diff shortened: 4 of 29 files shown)


jwerner-chromium:
@Caesar-github FYI for the Rockchip parts. (Note that there are also patches on the coreboot side needed for this to actually do something... I'll upload those to coreboot.org once this is accepted here.)

davidcunado-arm:
@jwerner-chromium 
Apologies for lack of progress - other internal are distracting attention! I have started review, and will continue next week. Will look to get f/b to you next week. Again apologies.

include/lib/coreboot/cbmem_console.h
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright (c) 2017, ARM Limited and Contributors. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef __CBMEM_CONSOLE_H
```
davidcunado-arm: ARM TF include guard format -> \_\_CBMEM_CONSOLE_H__

lib/coreboot/coreboot_table.c
```diff
@@ -44,6 +46,7 @@ struct cb_entry {
 	uint32_t size;
 	union {
 		struct cb_serial serial;
+		uint64_t uint64;
```
davidcunado-arm: Is there a more meaningful name that could be used?

lib/stdlib/putchar.c
```diff
@@ -15,6 +16,8 @@
 int putchar(int c)
 {
 	int res;
+	if (COREBOOT)
```
davidcunado-arm: If coreboot support is known at compile time then this can be #if.

include/common/aarch64/debug_macros.S
```diff
@@ -0,0 +1,32 @@
+/*
+ * Copyright (c) 2014-2016, ARM Limited and Contributors. All rights reserved.
```
davidcunado-arm: Copyright date - 2014-2017

bl31/aarch64/crash_reporting.S
```diff
@@ -5,6 +5,7 @@
  */
 #include <arch.h>
 #include <asm_macros.S>
+#include <debug_macros.S>
```
davidcunado-arm: includes should be in alphabetical order

docs/porting-guide.md
```diff
@@ -2267,7 +2267,7 @@ register x0.
 
 This API is used by the crash reporting mechanism to force write of all buffered
 data on the designated crash console. It should only use general purpose
-registers x0 and x1 to do its work. The return value is 0 on successful
+registers x0, x1, x2 and x3 to do its work. The return value is 0 on successful
```
davidcunado-arm: Need to rebase the patch as documentation now migrated to .rst format files.

lib/coreboot/coreboot_table.c
```diff
@@ -0,0 +1,105 @@
...
+#include <mmio.h>
+#include <string.h>
+
+/*
+ * Structures describing coreboot's in-memory descriptor tables. See
+ * <coreboot>/src/commonlib/include/commonlib/coreboot_tables.h for
+ * canonical implementation.
+ */
+
+struct cb_header {
```
jwerner-chromium: Sorry, I don't really understand what you mean by "anonymous struct" here. To me, an anonymous struct is a struct without a name, like this:

struct {
  uint32_t a;
  uint32_t b;
} myvariable;

But all the structs you pointed out here are normal structs with names? I just have a nested annonymous union in cb_entry below, I can change that if you want...

bl31/plat_params.c
```diff
@@ -17,6 +18,13 @@ void bl31_plat_params_setup(void *from_bl2, bl31_plat_param_handler_t handler) {
 			continue;
 
 		switch (param->tag) {
+		case BL31_PLAT_PARAM_COREBOOT_TABLE:
+			if (COREBOOT)
```
jwerner-chromium: Yeah, I can change that if you like. Is that required ARM TF style?

I generally prefer this style over preprocessor #if for constants, because it reads nicer and it allows the compiler to syntax check even the code branch that is not enabled for the current build configuration (making it easier to prevent bit rot for rarely-used paths). But I can comply with your style if you want.

Makefile
```diff
@@ -256,8 +256,18 @@ include lib/stack_protector/stack_protector.mk
 # makefile may use all previous definitions in this file)
 ################################################################################
 
+# Exclude coreboot support by default, override by command line or plat file
```
jwerner-chromium: Yeah, the way I want to use it, it does.

The thing is that coreboot isn't really a per-platform constant. Some platforms support booting from coreboot, but they may still be booted from a different BL2 implementation as well. For example, Rockchip's RK3399 supports booting from coreboot or U-Boot, I think.

I want to add this constant here and then add COREBOOT=1 to the make invocation command line that the coreboot build system uses to build Trusted Firmware. This way, if you build for coreboot it will always automatically be 1, and if you build standalone it will always automatically be 0. You don't have to set anything manually anymore.

include/lib/coreboot/coreboot_table.h
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright (c) 2017, ARM Limited and Contributors. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef __COREBOOT_TABLE_H
```
jwerner-chromium: Okay, I'll update all of these.

bl31/plat_params.c
```diff
@@ -0,0 +1,26 @@
+/*
```
jwerner-chromium: Sure, will do.

bl31/bl31.mk
```diff
@@ -12,6 +12,7 @@ BL31_SOURCES		+=	bl31/bl31_main.c				\
 				bl31/aarch64/runtime_exceptions.S		\
 				bl31/aarch64/crash_reporting.S			\
 				bl31/bl31_context_mgmt.c			\
+				bl31/plat_params.c				\
```
jwerner-chromium: Yeah, it could, although I feel that would just add extra busy-work for platforms trying to use it. Do you think building it by default would be an issue? Note that this does add no extra bytes to the binary if the platform doesn't use it, since Trusted Firmware is linked with --gc-sections and will garbage-collect all unused code and data.

I can put it in common/, but I felt that bl31/ was the more appropriate place since it only makes sense for use in BL31 (whereas common/ seems to contain code that is useful in more than one BL).

bl31/aarch64/crash_reporting.S
jwerner-chromium: Will fix.

docs/porting-guide.md
jwerner-chromium: Will fix.

lib/coreboot/coreboot_table.c
davidcunado-arm: In general structs / enums defined in ARM TF are typedef'ed as follows:
typedef struct param_header {
	uint8_t type;		/* type of the structure */
	uint8_t version;    /* version of this structure */
	uint16_t size;      /* size of this structure in bytes */
	uint32_t attr;      /* attributes: unused bits SBZ */
} param_header_t;

typedef enum {
	AFF_STATE_ON = U(0),
	AFF_STATE_OFF = U(1),
	AFF_STATE_ON_PENDING = U(2)
} aff_info_state_t;

I may have mixed my terms re: "anonymous structs", put if you looks through the ARM TF code you'll see that structs and enums are typedef'ed. Sorry for the confusion!

lib/coreboot/cbmem_console.c
```diff
@@ -0,0 +1,60 @@
...
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include <coreboot/cbmem_console.h>
+#include <xlat_tables_v2.h>
+
+/*
+ * The console is initially validated by coreboot, so we can trust it. But at
+ * runtime the non-secure OS could manipulate the size field to trick us into
+ * making bad memory accesses. Therefore we need to store the known-good size
```
danh-arm: If this memory is accessible to the non-secure OS, then having this mapped into BL31 at runtime is a security attack vector. I'll have to trust that you've thought through all the implications of this.

include/lib/coreboot/cbmem_console.h
```diff
@@ -0,0 +1,16 @@
+/*
```
danh-arm: This console functionality is really a driver so would be better in (include)/drivers/coreboot. The other stuff is OK in lib/coreboot though.

lib/stdlib/putchar.c
```diff
@@ -15,6 +16,8 @@
...
+		coreboot_cbmc_putc(c);
```
danh-arm: Hooking into the stdlib putchar is not appropriate. This should be implemented within the existing console framework.

bl31/aarch64/crash_reporting.S
```diff
@@ -60,8 +61,11 @@ intr_excpt_msg:
 	 * Helper function to print newline to console.
 	 */
 func print_newline
+	mov	x3, x30
 	mov	x0, '\n'
-	b	plat_crash_console_putc
+	crash_console_putc
```
danh-arm: I don't think this patch will be necessary if you hook into the console framework properly in the earlier patches.

include/lib/coreboot/params.h
```diff
@@ -0,0 +1,22 @@
...
+#define __COREBOOT_PARAMS_H
+
+/*
+ * CAREFUL: This header is included directly from coreboot sources. Do not add
+ * any further includes unless you make sure that they're available in coreboot
+ * and provide all required declarations.
+ */
+#include <bl31_plat_params.h>
+
+struct bl31_coreboot_table_param {
```
davidcunado-arm: Yes - apologies, got terms mixed-up; the usual format / style for structs / enums in ARM TF is that they are typedef'ed.
Sorry for the confusion!

lib/coreboot/coreboot_table.c
davidcunado-arm: Yes - I got my terms mixed-up; my comments should have been that structs / enums are typdef'ed.

include/lib/coreboot/params.h
danh-arm: OK, but there is no requirement to use typedefs!

davidcunado-arm: Ok - my mistake on that!

include/lib/coreboot/cbmem_console.h
jwerner-chromium: Okay, sure.

lib/coreboot/cbmem_console.c
jwerner-chromium: Yeah, I know. I'm pretty confident that this is bullet-proof but happy to get more eyes on the code, of course. (Only coreboot_cbmc_putc examines the contents of the buffer at runtime, and it's checking the address it writes to ensure that it cannot lie outside the buffer.)

lib/stdlib/putchar.c
jwerner-chromium: [whoops, I thought I commented this last time already]

I always prefer a C if() over a preprocessor #if even for macro constants where possible, because I feel it makes the code flow easier to read and it allows the compiler to syntax check both branches (so it's harder for bit rot to creep into code paths of rarely used configurations when stuff gets refactored over the years). But if official Trusted Firmware style says to always use #if for those cases I can change this, of course.

include/lib/coreboot/coreboot_table.h
jwerner-chromium: I'll write them as David suggested for now, and then you can bulk edit all headers later if you need.

bl31/bl31.mk
danh-arm: Regarding the "official dynamic configuration mechanism", don't hold your breath on that. I intend to discuss the design at the next Linaro Connect conference in September, with a view to starting the implementation shortly after. So we're looking at towards the end of this year. In terms of tracking, we'll either use https://github.com/ARM-software/tf-issues/issues/454 or create a new (more specific) issue.

In terms of mechanism, we're still looking at using the device tree format, but we'll try not to lock all platforms into using this if it doesn't meet your needs. I can see that console config while the MMU is disabled is particularly problematic.

drivers/console/aarch64/console.S
```diff
@@ -95,10 +95,15 @@ endfunc console_getc
...
-	 * Clobber list : x0, x1
+	 * Clobber list : x0, x1, x2, x3, x4
 	 * ---------------------------------------------
 	 */
 func console_flush
+#if COREBOOT
+	mov	x4, x30
+	bl	coreboot_cbmc_flush
+	mov	x30, x4
+#endif
```
danh-arm: For the crash console, wouldn't it be possible for the coreboot console to allocate some scratch space for its `plat_crash_console_putc()` implementation to use instead of registers? Any initialization needed can be done within `plat_crash_console_init()`.

I see the general idea of providing branch points at the top level for different console implementations, however I don't think this is very scalable or maintainable since it distributes console specific functionality across several places in generic code. You've already had to plug into 3 different places in generic code to make this work and more may be needed in future.

I don't really understand this comment:

> I feel that the decision which consoles drivers to use shouldn't really be hidden in the platform...

The general philosophy in TF is that platform code is precisely where variations like this should be implemented. We want to work towards reducing build flags in generic code, not proliferating them.

bl31/bl31.mk
jwerner-chromium: Do you mean that your mechanism likely won't be parsed before the MMU is enabled? But that would mean that it also cannot be used to pass serial console configuration information in general, right? Is that intended? Or do you expect to enable the MMU before the serial console in the future?

If what you're building is just fundamentally not suited for the coreboot table pointer, we should maybe just special-case it and find a different solution (it's just one pointer, after all). How about adding a new, optional field to bl31_params_t, for example? I would really like to find a solution where not every platform has to come up with its own custom way of fitting this into the bl31_plat_params somehow.

drivers/console/aarch64/console.S
jwerner-chromium: > For the crash console, wouldn't it be possible for the coreboot console to allocate some scratch space for its plat_crash_console_putc() implementation to use instead of registers? Any initialization needed can be done within plat_crash_console_init().

Well... I could have each function store its return pointer in static scratch space, yeah. Since the crash console is not reentrant that should work. It would just look kinda weird.

But going back to your idea with the "splitting" driver, I also don't see how I could do that without running into naming clashes. The generic console code just calls function names like console_core_putc, and the individual drivers implement those. I can't really shim a layer in-between that and leave both of the existing sides untouched (at least not without heavy linker magic). I think this only really works if the splitting happens in the generic code already (or if the whole console API was made more flexible and would work by registering function pointers at runtime or something like that).

> The general philosophy in TF is that platform code is precisely where variations like this should be implemented. We want to work towards reducing build flags in generic code, not proliferating them.

Does that also count for flags which represent user decisions rather than modeling hardware? As I tried to say, I don't think console types (the ones you'd like to mix and match) are really a property of the platform. Sure, some types may only be supported on some subset of platforms, but the decision on which of them to enable for a particular build is made by the person building it, not predetermined by the hardware. So I feel like this is similar to something like DEBUG=1 or LOG_LEVEL=30, not a hardware-describing option like ARM_CCI_PRODUCT_ID=500. For those, doesn't it make much more sense to keep them in generic code so they don't need to be duplicated individually by all platforms?

bl31/bl31.mk
danh-arm: I think the mechanism needs to be flexible as to whether the config is parsed before or after its passed to each BL. MMU enablement is separate issue; in future, it might be possible to keep the MMU enabled when passing from one BL to the next, but that requires changes to the handover interface and is not something we're currently looking at.

I want to find a solution that works for you, as well as for partners who want a more traditional device tree approach.

drivers/console/aarch64/console.S
danh-arm: > (or if the whole console API was made more flexible and would work by registering function pointers at runtime or something like that)

Yes, that was what I was envisaging. The "splitting" console API would be a struct of function pointers corresponding to the existing console API. This would be much like the PSCI pm-ops API. Each console driver would register a struct containing function pointers unique to that driver. The platform would also provide a macro to determine the number of supported consoles (i.e. how much memory to allocate for the array of structs).

I'm sorry but I disagree that choosing console types is a generic user config option. Each console type implies a particular kind of hardware availability, which the platform should be responsible for defining. The set of potential console types does not look bounded to me and would make the generic code more cluttered and less maintainable.

jwerner-chromium: > Yes, that was what I was envisaging. The "splitting" console API would be a struct of function pointers corresponding to the existing console API. This would be much like the PSCI pm-ops API. Each console driver would register a struct containing function pointers unique to that driver. The platform would also provide a macro to determine the number of supported consoles (i.e. how much memory to allocate for the array of structs).

Sorry, I'm still not quite sure I understand you right. Do you want me to change the whole API in the generic code (meaning that console_putc would no longer just call console_core_putc, but would instead look up a function pointer to call from a struct that was previously registered)? I don't suppose you do, because it sounds like you want to keep all of these changes out of the generic code?

My problem right now is that the generic code is just calling console_core_putc, and the ti/16550_console.S driver that I want to continue to use unchanged is implementing console_core_putc. If I want to write a splitting driver that can call several implementations, it also needs to implement console_core_putc... but then that would clash with the existing implementation from the 16550 driver. As long as the generic API just consists of calling well-known function names, how could I implement a splitting driver that can wrap the unchanged existing console drivers without naming clashes?

danh-arm: I would expect all console drivers that need to support splitting would move the body of their `console_core_putc()` (and other functions) into exported functions with unique names in a new file (e.g. `ti16550_console_core_putc()` in `16550_multi_console.S`). The splitting driver would provide a strong definition of `console_core_putc()`, which would iterate through all the registered splitting drivers. The platform ports would include the new file for each console driver instead of the existing one.

Thinking about it some more, perhaps this could be an extension to the existing API using the name "multi console API" instead of "splitting". In hindsight it would be good if the existing API to supported multiple consoles like this but changing the existing APIs would not be backwards compatible for platforms.

jwerner-chromium: Okay, how about this: I add a new series of assembler macros like

 .macro func_console_putc _name
 func \_name
 .weak console_single_putc
 console_single_putc:
 .endm

Then I rewrite all the console drivers to use this (with a unique name) instead of the existing 'func console_core_putc' (and equivalent for the other functions). I would change drivers/console/.../console.S to use structs with function pointers, with a new function to register them (e.g. console_register). It would still provide console_init for backwards compatibility, which would call console_register with a struct containing console_core_putc and friends as function pointers.

This way, existing platforms that only link a single console driver and call console_init should continue to work unchanged. Newer platforms could start linking more than one driver and use console_register instead (so that console_init is garbage-collected by the linker and the duplicate .weak symbols aren't referenced so that they shouldn't cause a problem).

Does that sound like the kind of system you want?

danh-arm: I'm struggling to visualise what you mean here. I like the idea of introducing a `console_register()` function, which supersedes use of `console_init()`, while re-implementing the latter in terms of the former to keep compatibility. However, I don't understand what is achieved by your series of assembler macros and duplicate weak functions, or why that is better than what I suggested. Perhaps we just need to see some prototype code?

jwerner-chromium: I'm just trying to come up with a system that doesn't require every driver to essentially be written twice. If drivers implement a new API in a separate file as you suggested, I'm afraid it would create a mess of drivers that only support one API but not the other depending on what the implementer needed. I think it would be nicer if we can rewrite the existing drivers to a framework that can support both APIs at once, so that future driver additions will also support both automatically.

I'll try to put this into code and see how it works out in practice.

danh-arm: I agree it would be good if we can make this work without adding new source files for each driver, while keeping compatibility for unmodified drivers, including non-upstream ones. I'll wait to see some code for this before commenting further.

jwerner-chromium:
Hi Dan,

Sorry, this took a while, had to figure out some tricky details (and find a few generic bugs in the process). But here's a new version with a function-pointer based console API as we discussed, that's still fully backwards-compatible to old platform code or old console drivers. Let me know what you think!

sandrine-bailleux-arm:
Hi @jwerner-chromium,

I'm afraid Dan is not in the office this week so he won't be able to provide any feedback until next week.

Makefile
```diff
@@ -276,6 +276,11 @@ endif
 # This can be overridden by the platform.
 include lib/cpus/cpu-ops.mk
 
+ifeq (${COREBOOT},1)
+# Include coreboot support library when building with coreboot as BL2
+include lib/coreboot/coreboot.mk
+endif
+
```
danh-arm: It shouldn't be necessary to change the generic Makefile. Platforms can just pull coreboot.mk as required.

make_helpers/defaults.mk
```diff
@@ -38,6 +38,10 @@ BASE_COMMIT			:= origin/master
 # The platform Makefile is free to override this value.
 COLD_BOOT_SINGLE_CPU		:= 0
 
+# Flag to compile in coreboot support code. Exclude by default. The coreboot
+# Makefile system will set this when compiling TF as part of a coreboot image.
+COREBOOT			:= 0
+
```
danh-arm: This can be moved to coreboot.mk

plat/common/aarch64/platform_helpers.S
```diff
@@ -55,31 +55,31 @@ func plat_report_exception
...
 	/* -----------------------------------------------------
-	 * Placeholder function which should be redefined by
-	 * each platform.
+	 * void plat_crash_console_putc(int character)
+	 * Output through the normal console by default.
 	 * -----------------------------------------------------
 	 */
 func plat_crash_console_putc
-	ret
+	b	console_putc
```
danh-arm: Actually, it's not clear that there are any C consoles out there, so this could be a sensible change. However, it should probably be in a separate patch is it doesn't seem part of the general API change.

docs/porting-guide.rst
```diff
@@ -2445,9 +2445,12 @@ of the CPU to enable quick crash analysis and debugging. It requires that a
 console is designated as the crash console by the platform which will be used to
 print the register dump.
 
-The following functions must be implemented by the platform if it wants crash
-reporting mechanism in BL31. The functions are implemented in assembly so that
-they can be invoked without a C Runtime stack.
+By default, the definitions in `plat/common/aarch64/platform\_helpers.S` will
+cause the crash output to be routed over the same console that was registered
+for boot messages. If necessary, platforms may supply their own definitions of
```
danh-arm: If you can guarantee that early crashes on platforms that don't override these functions, results in nothing being output and the plat_panic_handler() being correctly called, then this should be OK. However, this seems to be separate to the main part of this patch and could come in a later patch. The additional comments you suggest would help too.

docs/porting-guide.rst
```diff
@@ -2480,12 +2483,11 @@ Function : plat\_crash\_console\_flush
 ::
 
     Argument : void
-    Return   : int
+    Return   : void
```
danh-arm: I see, although only flush() is updated not putc().

The return codes are there generally to follow the stdio.h equivalent functions. I can see there's not much value in returning error codes from the crash functions but I'm not very comfortable with discarding this information from all the underlying drivers, in case they are used by other top-level APIs (e.g. stdlib clients).

In the case of the multi-console API, you could make a policy decision on what to do, for example try outputting on all consoles even on error, and report the first error you come across.

In any case, this change in error code handling is separate to the main part of this patch and it would be easier to review if it were in a separate patch (you might be detecting a theme to my comments here!).

drivers/console/aarch64/console.S
```diff
@@ -3,103 +3,240 @@
...
-	 * Function to initialize the console without a
-	 * C Runtime to print debug information. It saves
-	 * the console base to the data section.
+	 * Deprecated legacy function for initializing a
+	 * single console. Current platforms should call
+	 * console_<driver>_register for the relevant
+	 * driver instead. This needs to clear the console
+	 * list first because some legacy platforms called
+	 * console_init more than once and expect it to
+	 * overwrite previous console state.
```
danh-arm: OK, I agree console_init() should be deprecated. Ideally the declaration in the header should add the __deprecated keyword but we can only do that if AArch32 is updated too.

Regarding multiple instances of the same driver, I agree its unlikely this will be needed (for now). However, if this were needed in future, it would be possible to create a simple memory pool rather than full dynamic memory allocation. Anyway, let's ignore this for now.

drivers/console/aarch64/console.S
```diff
@@ -3,103 +3,240 @@
...
 endfunc console_init
+#endif
+
+	/* -----------------------------------------------
+	 * console_info_t *console_unregister(
+	 *                        console_info_t *console)
+	 * Function to find a specific console in the list
+	 * of currently active consoles and remove it.
+	 * Should usually only be called by
+	 * declare_console expansions.
```
danh-arm: > This is just the helper, the function that should be called is console_(driver)_unregister() (which will be created as part of declare_console).

Hmm, that's not very obvious, especially when console_[un]register() is declared in console.h but not console_(driver)_[un]register(). It would be easier to understand (and maintain) if the functions/macros to be invoked by C code were declared in console.h, and any other functions were hidden in a private header.

danh-arm: > Also, note that the boot console should normally not be unregistered.
But that's how some existing platforms expect things to work. The boot and runtime console might want to use the same driver but on different UARTs. The only way to do that is to unregister the boot console and register the runtime console.

I don't see why the framework needs to maintain this global console state, as long as the correct consoles are registered/unregistered by the platform during the runtime transistion. The crash console can just blindly output on any registered consoles, whether they're boot or runtime consoles.

drivers/console/aarch64/console.S
```diff
@@ -3,103 +3,240 @@
...
-	 * Function to finish the use of console driver.
-	 * It sets the console_base as NULL so that any
-	 * further invocation of `console_putc` or
-	 * `console_getc` APIs would return error.
+	 * Function to switch from boot to runtime state.
+	 * The name is misleading for compatibility with
+	 * the legacy API -- this used to uninitialize all
+	 * consoles, but now it just quiesces consoles
+	 * that are not marked as "runtime".
+	 * Clobber list: x0, x1
```
danh-arm: I agree that (some) consoles should be kept around for crash reporting, but that should be done by the platform calling register/unregister appropriately. This console_uninit() function should continue to unregister all consoles, even if the platforms no longer use it. The decision as to whether a console is a runtime console should be made by the platform, not the console driver.

drivers/console/aarch64/console.S
```diff
@@ -3,103 +3,240 @@
...
+	 * Out: x0 - always 1
+	 * Clobber list: x0, x1
+	 * -----------------------------------------------
+	 */
+func console_start_crash
+	mov	w0, #(1 << CONSOLE_FLAG_CRASH_SHIFT)
+	adrp	x1, console_state
+	strb	w0, [x1, :lo12:console_state]
+	mov	x0, #1
+	ret
```
danh-arm: OK, I understand how this works now, but I don't think it's necessary (see earlier comment).

drivers/console/aarch64/console.S
```diff
@@ -3,103 +3,240 @@
...
+endfunc console_start_crash
+
 	/* ---------------------------------------------
-	 * int console_putc(int c)
-	 * Function to output a character over the
-	 * console. It returns the character printed on
-	 * success or -1 on error.
+	 * void console_putc(int c)
+	 * Function to output a character. Calls all
+	 * console's putc() handlers in succession.
```
danh-arm: As I said in another comment, I'm not comfortable with removing the error codes, at least not as a side-effect of this patch. I'll try to canvas other opinions though. I agree this is source compatible but it's going to be important to keep function declarations and definitions the same, as this is a MISRA rule we want to comply with.

include/common/aarch64/console_macros.S
```diff
@@ -0,0 +1,157 @@
...
+	 * be used for any purpose. It will be stored in
+	 * the console structure and passed in to every
+	 * other callback for this console.
+	 * In: x0 - driver-specific base address
+	 *     x1...x7 - optional extra arguments
+	 * Out: x0 - 1 for success, 0 for error
+	 * Clobber list: x0 - x7
+	 * ------------------------------------------------
+	 */
+.macro func_console_init _driver
```
danh-arm: I agree the console_core_xxx definitions could be part of declare_console, although a better name for that might be register_console, as it's both declaring and defining things. You can keep the func_console_xxx macros if you like, but they should only contain the func console_\_driver\()_xxx prologue, and yes there should be endfunc equivalents. Given that, I'm not sure they have much value, though.

include/common/aarch64/console_macros.S
```diff
@@ -0,0 +1,157 @@
...
+	func console_\_driver\()_init
+	.globl console_core_init
+	.weak console_core_init
```
danh-arm: You're right that .globl doesn't mean strong (sorry about that), although I think declaring a symbol as weak effectively makes it global. I don't think allowing multiple weak definitions is very robust, due to the ambiguity as to which one the linker chooses. We intend to ban this pattern in future to comply as best we can with certain MISRA rules. We should only allow at most one weak and one strong definition.

If these console_core_xxx declarations were hidden behind a build flag (e.g. ALLOW_MULTIPLE_CONSOLES) then there should only be a single (strong) definition of console_core_xxx for unmigrated platforms, so declaring them as weak would not be necessary. Migrated platforms can set ALLOW_MULTIPLE_CONSOLES to 1 so that these functions are not declared. ALLOW_MULTIPLE_CONSOLES could default to the value of ERROR_DEPRECATED.

include/common/aarch64/console_macros.S
```diff
@@ -0,0 +1,157 @@
...
+		adrp	x0, console_\_driver\()_info
+		add	x0, x0, :lo12:console_\_driver\()_info
+		b	console_unregister
+	endfunc console_\_driver\()_unregister
+
+	/* Mark all callbacks as weak so their implementation is optional. */
+	.weak console_\_driver\()_putc
+	.weak console_\_driver\()_getc
+	.weak console_\_driver\()_flush
+	.weak console_\_driver\()_uninit
```
danh-arm: OK, but there needs to a comment to explain what's going on here.

include/drivers/console.h
```diff
@@ -7,14 +7,64 @@
...
+#include <cassert.h>
+#include <types.h>
+
+typedef struct console_info {
+	struct console_info *next;
+	u_register_t flags;
+	uintptr_t baseaddr;
+	void (*putc)(int character, uintptr_t baseaddr);
+	int (*getc)(uintptr_t baseaddr);
+	int (*flush)(uintptr_t baseaddr);
```
danh-arm: I see, thanks.

plat/common/aarch64/platform_helpers.S
```diff
@@ -55,31 +55,31 @@ func plat_report_exception
...
-	ret
+	b	console_start_crash
 endfunc plat_crash_console_init
 
 	/* -----------------------------------------------------
-	 * Placeholder function which should be redefined by
-	 * each platform.
+	 * void plat_crash_console_putc(int character)
+	 * Output through the normal console by default.
 	 * -----------------------------------------------------
```
soby-mathew: nit: Good to update the clobber list to the functions here.

jwerner-chromium: Will do.

docs/porting-guide.rst
danh-arm: > I think the code should definitely always run through all consoles even if some of them fail, so the question is just what we want to return in the failure case. We could either return -1 if at least one console fails or return -1 only if all consoles fail.

I already said what I think - the former.

> I really can't come up with a scenario where either of these results would be terribly useful for a caller and allow it to make an important decision. Consider that almost all characters output by Trusted Firmware go through tf_printf() anyway, which already throws the return code of putchar() away and returns void.

Not all TF code is upstream and some of it may be used by other projects. Although the error code is not useful in current upstream client code, making a unilateral decision to remove it is unnecessary and off-topic.

> I think we just have a very different situation from what the stdio.h API was designed for (where output can be piped into files or devices and errors like "out of space" are common and meaningful) and keeping this functionality isn't worth all the extra effort and complexity it requires.

Who's to say the code may not get used in future in the way you describe (for piping into devices/files). Keeping this functionality doesn't seem much effort to me (an extra register save), and dealing with the compatibility issues is more effort.

drivers/console/aarch64/console.S
danh-arm: > I would like to get away from the forced console_init(baseaddr, clock, baud) function signature, because for some consoles those parameters may not make sense and they'll want different ones instead. That's why every driver can declare their own console_xxx_register() function (e.g. the one in include/drivers/ti/uart/uart_16550.h) with the specific signature they need.

Yes, I have no objection to that. I was just saying it's confusing to have console_register() in console.h when the expectation is that clients call console_xxx_register().

> In fact, now that I think of it it's not really useful to have C declarations for console_register() and console_unregister() anyway. Let me remove those so it's an implementation detail of the assembly code.

Agreed.

drivers/console/aarch64/console.S
danh-arm: It's just that the signature has to match the implementation. A #ifdef in the header would solve this as you suggest but I don't like the unnecessary break anyway.

drivers/console/aarch64/console.S
danh-arm: I think we're talking at cross purposes here. As I said, I don't think you need to store the console state anywhere. At runtime, any crashes should be reported on the registered runtime console(s); at boot time, any crashes should be reported on the register boot console(s). The default bl31_plat_runtime_setup() should continue to uninit/unregister all consoles but platforms can do something different if they want.

include/common/aarch64/console_macros.S
danh-arm: Sorry, register_console indeed would be confusing with the existing console_register functions. However, register_console would be more in keeping with existing REGISTER_XXX macros in the codebase. Perhaps the console_register functions could be renamed, to say console_add or console_bind?

include/common/aarch64/console_macros.S
danh-arm: The point of the MISRA rules generally (which I agree with) is to not rely on people not writing bad code where possible, but enforce it through coding standards. In this case, we can enforce the correct usage (and not break a MISRA rule) by not declaring the symbol as weak. We can remove the build flag in future when we decide we only support ALLOW_MULTIPLE_CONSOLES = 1. I suspect other parts of the implementation could be hidden behind this build flag too.

danh-arm:
Due to the limitations of GitHub review, it's becoming almost impossible to follow what's happening in this PR. Given that we don't seem to be converging on a solution, I suggest any open issues are resolved in a teleconference, and new PR raised with the result.

include/common/aarch64/console_macros.S
jwerner-chromium: Okay, I'll add an explicit build flag. At least that will give some more freedom in designing the other parts of the API as well.

drivers/console/aarch64/console.S
danh-arm: OK, now I understand the behaviour you're expecting. This is different to the behaviour expected by the platforms I know about. For the latter, we expect there to be 2 serial consoles, one for boot and one for runtime. The boot console is handed over to other boot components after TF boot (normal world firmware and eventually Linux). On those platforms, the crash console is currently statically configured to be the same as the runtime console.

As far as I can see, both our use-cases can be implemented by both our suggestions, although yours requires additional state. To implement your expected behaviour with my suggestion, the platform would override bl31_plat_runtime_setup() to disable the serial boot console and leave the runtime memory console active, and also override plat_crash_console_init to re-enable the serial console for crashes (or enable it the first time for early crashes).

The advantages to your suggestion that I can see, are a) there's generally no need to override plat_crash_console_init(), unless you want console output for early crashes, and b) some platforms can do all console registration/enablement in one place (platform setup). a) seems quite weak as overriding this function is easy and early crash debugging is generally important. Also, platforms that do override plat_crash_console_init will need to remember to call console_start_crash to keep things working. b) doesn't help the platforms I mentioned before, as it's only possible to have one serial console registered at a time, so the platforms would need to unregister/register the console anyway during the runtime switch.

Finally, even if we do retain this console state, this shouldn't change the implementation of console_uninit() in this unintuitive way; it should change the plat/common implementation of bl31_plat_runtime_setup().

include/common/aarch64/console_macros.S
danh-arm: console_xxx_enable() and console_xxx_disable() work for me, although that could be confusing if we retain the console state, since console enablement would also depend on the console state.

danh-arm:
> But I also think we are through with most issues now. 

I hope so. I just find the GitHub UI difficult to use for complicated PRs like this.

drivers/console/aarch64/console.S
jwerner-chromium: > Also, platforms that do override plat_crash_console_init will need to remember to call console_start_crash to keep things working.

My intention is that platforms which do override plat_crash_console_init (I assume this would mostly be done temporarily to debug early boot failures) would also override plat_crash_console_putc and plat_crash_console_flush, so they wouldn't rely on the common infrastructure at all and don't need to call console_crash_start.

Also, getting plat_crash_console_init right for this case becomes much harder when your console is no longer just a simple UART with hardcoded initialization parameters. For example, the base address for the memory console I want to add is passed into BL31 via a parameter from BL2 (coreboot), which points to a memory structure that needs to be parsed at boot time because it is no longer accessible after MMU initialization. It would be hard to find that value again during a crash (especially because plat_crash_console_init does not permit a C environment, so you would force all console discovery to be written in assembly with a restricted register set). It's possible to store this in global variables in platform code, sure, but isn't it much easier if the console infrastructure does this for you?

> doesn't help the platforms I mentioned before, as it's only possible to have one serial console registered at a time

If this is the main point of contention I'd rather expand the system to support that as well, instead of implementing something that I think makes the use case I'm trying to solve more painful. Would you be okay with that?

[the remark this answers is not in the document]
danh-arm: > My intention is that platforms which do override plat_crash_console_init (I assume this would mostly be done temporarily to debug early boot failures) would also override plat_crash_console_putc and plat_crash_console_flush, so they wouldn't rely on the common infrastructure at all and don't need to call console_crash_start.

OK, that's a fair assumption, although I wouldn't want to preclude platforms from using the new common infrastructure in their plat_crash_console_xxx implementations (I don't think that's the case).

> Also, getting plat_crash_console_init right for this case becomes much harder when your console is no longer just a simple UART ...

OK, I can see there's value for your use-case in initializing all consoles in one place (early platform setup) and changing console state as required. For this to work in the use-case I mentioned, I think you'd need to at least allow the console flags to be set during `console_\_driver\()_register`, so that platforms can change the crash console later (e.g. during the runtime switch).

To allow centralised console registration in my use-case, we'd need to allow multiple instances of the same console to be registered, and the ability to change the console flags after registration. But that could be implemented later if needed.

jwerner-chromium:
Superseded by PR #1103

PR #991, opened 2017-06-20
hikey960: migrate to use A53 specific defines

davidcunado-arm:
The patch fb7d32e5881ef2445e8fe2305005f5590d4a7cfa migrated the CPU
libraries to have unique defines, prefixing them with the CPU name.

This patch migrates the hikey960 platform port to use the A53 specific
defines.

Change-Id: Id76f544b0b236bbd4974ab5ffa1203f073c20021


davidcunado-arm:
@hzhuang1 
As per commit message, the hikey960 port was not migrated to the CPU specific defines, and with the addition of the hikey960 build into the CI we're getting build breaks - therefore we've expedited merging this PR. 

Apologies if this is not the fix you would have done - but please submit new PR if that's the case; I'll review it as a matter of urgency.

PR #993, opened 2017-06-21
rockchip: support to use hdcp for rk3399

rkchrome:
If kernel and loader have not been supported the paramater, the ATF's hdcp never used for rk3399.


davidcunado-arm:
@rkchrome 
The PR is targeting master - please can you update so that it points to integration?

davidcunado-arm:
jenkins: test this please

plat/rockchip/rk3399/drivers/dp/cdn_dp.c
```diff
@@ -41,3 +44,46 @@ __asm__(
...
+
+uint64_t dp_hdcp_ctrl(uint64_t type)
+{
+	switch (type) {
+	case HDCP_KEY_DATA_START_TRANSFER:
+		memset(&key, 0x00, sizeof(key));
+		hdcp_key_pdata = (uint64_t *)&key;
+		return 0;
+	case HDCP_KEY_DATA_START_DECRYPT:
+		if (hdcp_key_pdata == (uint64_t *)(&key + 1))
```
danh-arm: If I understand correctly, the idea is for the normal world to make multiple calls to `dp_hdcp_store_key()` until `key` is completely populated, then HDCP_KEY_DATA_START_DECRYPT is called. For this condition to be true, it seems that `key` must be aligned on 6 x 64-bit word boundary, but I don't see any checks that this is the case.

plat/rockchip/rk3399/drivers/dp/cdn_dp.h
```diff
@@ -0,0 +1,65 @@
...
+	CDN_DP_HDCP_KEY_LEN = 326,
+};
+
+struct cdn_dp_hdcp_key_1x {
+	uint8_t ksv[CDN_DP_HDCP_KSV_LEN];
+	uint8_t device_key[CDN_DP_HDCP_DPK_LEN];
+	uint8_t sha1[CDN_DP_HDCP_SHA_LEN];
+	uint8_t uid[CDN_DP_HDCP_UID_LEN];
+	uint16_t seed;
+	uint8_t reserved[10];
```
danh-arm: Given that you are writing directly into the memory allocated by one of these structs, it seems you are either relying on this structure being packed or there being a specific amount of padding. You should probably either declare this as packed or explicitly insert the assumed padding.

plat/rockchip/rk3399/drivers/dp/cdn_dp.c
```diff
@@ -0,0 +1,89 @@
+/*
+ * Copyright (c) 2017, ARM Limited and Contributors. All rights reserved.
+ *
```
Caesar-github: Done

plat/rockchip/rk3399/drivers/dp/cdn_dp.c
Caesar-github: Hmm.... you are right here, I think.

plat/rockchip/rk3399/drivers/dp/cdn_dp.h
Caesar-github: Yes, sound resonable.

plat/rockchip/rk3399/drivers/dp/cdn_dp.c
Caesar-github: +struct cdn_dp_hdcp_key_1x {
 +	uint8_t ksv[CDN_DP_HDCP_KSV_LEN];
 +	uint8_t device_key[CDN_DP_HDCP_DPK_LEN];
 +	uint8_t sha1[CDN_DP_HDCP_SHA_LEN];
 +	uint8_t uid[CDN_DP_HDCP_UID_LEN];
 +	uint16_t seed;
 +	uint8_t reserved[CDN_DP_HDCP_RESERVED_LEN];
 +} __aligned(8);

Maybe, can we should align the 8byte?

davidcunado-arm: @Caesar-github 
Apologies for the slow response.
I don't think that the __align(8) will work - perhaps a CASSERT on the size of cdn_dp_hdcp_key_1x to be a multiple of 6x64 bits; this would warn you if that struct changed in size to not be aligned with 6x64 bits.

davidcunado-arm:
@Caesar-github 
I've replied to your question - from merge point of view, the alignment question is not a blocker, but probably better to sort in one go.
Let me know what you want to do - but given our release activities at the moment, the merge may be delayed until later this week anyway.

plat/rockchip/rk3399/drivers/dp/cdn_dp.c
rkchrome: Yup, that should be better for me. 

For example:

+/* Checks the cdn_dp_hdcp_key_1x must be aligned on 6 x 64-bit word boundary */
 +CASSERT(sizeof(struct cdn_dp_hdcp_key_1x) % HDCP_KEY_1X_STORE_DATA_ALIGN_SIZE, \
 +	assert_hdcp_key_1x_store_data_align_size_mismatch);

davidcunado-arm:
jenkins: test this please

davidcunado-arm:
jenkins: test this please

PR #997, opened 2017-06-22
aarch64: Enable Statistical Profiling Extensions for lower ELs

dp-arm:
This PR depends on "Fully initialise essential control registers #995".


dp-arm:
Looking into the failure, will update as needed.

PR #998, opened 2017-06-22
Power down the CPU after platforms complete their power down sequence

vwadekar:
This patch moves the actual CPU power down code after the platform's power down
handler during the CPU_OFF and CPU_SUSPEND sequence. This allows platforms to
work with Dcache and MMU enabled for the power down sequence. Platforms usually
power down other hardware components, save context to memory from theri power
down handlers. Once they are done with their sequence, the core can be prepared
for power down - exit SMP, flush dcache, disable dcache, etc.

Platforms like Tegra will definitely benefit from this change in policy.

### DIFF
--- a/lib/psci/psci_off.c
+++ b/lib/psci/psci_off.c
@@ -83,6 +83,12 @@ int psci_do_cpu_off(unsigned int end_pwrlvl)
 	psci_stats_update_pwr_down(end_pwrlvl, &state_info);
 #endif
 
+	/*
+	 * Plat. management: Perform platform specific actions to turn this
+	 * cpu off e.g. exit cpu coherency, program the power controller etc.
+	 */
+	psci_plat_pm_ops->pwr_domain_off(&state_info);
+
 #if ENABLE_RUNTIME_INSTRUMENTATION
 
 	/*
@@ -104,12 +110,6 @@ int psci_do_cpu_off(unsigned int end_pwrlvl)
 		RT_INSTR_EXIT_CFLUSH,
 		PMF_NO_CACHE_MAINT);
 #endif
-
-	/*
-	 * Plat. management: Perform platform specific actions to turn this
-	 * cpu off e.g. exit cpu coherency, program the power controller etc.
-	 */
-	psci_plat_pm_ops->pwr_domain_off(&state_info);
 
 #if ENABLE_PSCI_STAT
 	plat_psci_stat_accounting_start(&state_info);
--- a/lib/psci/psci_suspend.c
+++ b/lib/psci/psci_suspend.c
@@ -173,9 +173,6 @@ void psci_cpu_suspend_start(entry_point_info_t *ep,
 	psci_stats_update_pwr_down(end_pwrlvl, state_info);
 #endif
 
-	if (is_power_down_state)
-		psci_suspend_to_pwrdown_start(end_pwrlvl, ep, state_info);
-
 	/*
 	 * Plat. management: Allow the platform to perform the
 	 * necessary actions to turn off this cpu e.g. set the
@@ -183,6 +180,9 @@ void psci_cpu_suspend_start(entry_point_info_t *ep,
 	 * program the power controller etc.
 	 */
 	psci_plat_pm_ops->pwr_domain_suspend(state_info);
+
+	if (is_power_down_state)
+		psci_suspend_to_pwrdown_start(end_pwrlvl, ep, state_info);
 
 #if ENABLE_PSCI_STAT
 	plat_psci_stat_accounting_start(state_info);


vwadekar:
@danh-arm @davidcunado-arm 

Can you please provide feedback when you get a chance? Thanks.

lib/psci/psci_off.c
```diff
@@ -83,6 +83,12 @@ int psci_do_cpu_off(unsigned int end_pwrlvl)
 	psci_stats_update_pwr_down(end_pwrlvl, &state_info);
 #endif
 
+	/*
+	 * Plat. management: Perform platform specific actions to turn this
+	 * cpu off e.g. exit cpu coherency, program the power controller etc.
+	 */
+	psci_plat_pm_ops->pwr_domain_off(&state_info);
```
soby-mathew: Moving this handler above the `cpu power down` ops will break for platforms which have dependancy on this (eg: CCI programming). So we cannot do this change generically.

Introducing an `early` pwr_domain_off/pwr_domain_suspend hook is an option but that requires further discussion. On a similar note, we could have `late` pwr_domain_in_finisher/pwr_domain_suspend_finisher hook for operations that needs to be done after `psci_do_pwrup_cache_maintenance()`.

vwadekar:
Thanks Soby. We already have a late suspend finisher. Do you think that
would be a good candidate for the CCI operations?

soby-mathew:
The CCI operations need to be done prior to `psci_do_pwrup_cache_maintenance()` so it wouldnt be a good candidate for the later suspend finisher. Other drivers like GIC are possible candidates.

danh-arm:
As Soby says, we can't move the existing hooks but I'm OK in principle with adding new early `pwr_domain_off/pwr_domain_suspend` hooks and late `pwr_domain_on_finisher/pwr_domain_suspend_finisher hooks`. Note these hooks would only have any relevance for !HW_ASSISTED_COHERENCY systems so probably should only be called in that case.

vwadekar:
I'm a bit confused. Why do we need new hooks. We already have the
suspend/off handlers and then the _wfi handler. Why can't we use the _wfi
as the late suspend handler?

vwadekar:
@soby-mathew 

>> But as I remember, the power up path is unique for Tegra since it enables D-cache early in warm boot. So it wouldn't matter to you anyway?

The power up path looks OK for Tegra as of now, since we merged the change to enable D-cache early for platforms like Tegra.

vwadekar:
>> The CCI operations need to be done prior to psci_do_pwrup_cache_maintenance()

I am assuming this is a typo (?) The platform's suspend handler gets called after the CPU power down routine now.

vwadekar:
FWIW, I think psci_suspend_to_pwrdown_start() will have to be broken down to call parts of it before 'psci_plat_pm_ops->pwr_domain_suspend'. At least, 'psci_spd_pm->svc_suspend(max_off_lvl)' needs to be called earlier for sure.

But if we can agree on path forward, I can try to come up with a change.

soby-mathew:
@vwadekar, 
> I am assuming this is a typo (?) The platform's suspend handler gets called after the CPU power down routine now.

I was referring to the power up path in my comment. To clarify things , let me explain again.
In the power down path during suspend, there are 2 hooks invoked currently : `psci_plat_pm_ops->pwr_domain_suspend()` and `psci_plat_pm_ops->pwr_domain_pwr_down_wfi()`.

Both the hooks are invoked with Caches OFF. psci_plat_pm_ops->pwr_domain_pwr_down_wfi() is invoked outside the psci_locks and as mentioned in the porting guide  "actions performed in this hook must be local to the CPU or the platform must ensure that races between multiple CPUs cannot occur". Hence this hook cannot be used for CCI programming and is not considered at all.

I was proposing to add a 3rd hook in the power down path, `psci_plat_pm_ops->pwr_domain_suspend_early()`, which will be invoked before `psci_do_pwrdown_cache_maintenance()`. This hook will run before caches are disabled and hence all platform operations which do not have dependancy on cache state can be done in this hook.

Similarly, for the power up path, we can possibly add `psci_plat_pm_ops->pwr_domain_suspend_finish_late()` which will get invoked after `psci_do_pwrup_cache_maintenance()`. This will run after the caches are enabled during power-up and hence platform operations which do no have dependancy on cache state can be done here.

These new hooks are not required when HW_ASSISTED_COHERENCY=1 because the caches are not disabled when powering down and are enabled early on power up.

Hope it is clear now.

vwadekar:
@soby-mathew 

I think adding one more handler for suspend is an overkill. We should try to keep dcache enabled for _suspend and dcache disabled for the _wfi suspend handlers. We dont have a late suspend finisher as of now, so adding one would be a good idea.

Having said that, please post the outcome of the internal meetings here. Thanks.

vwadekar:
@soby-mathew 

`WARMBOOT_ENABLE_DCACHE_EARLY` is enabled for Tegra platforms, so dcache is already enabled when CPUs power up.

soby-mathew:
> I think adding one more handler for suspend is an overkill. We should try to keep dcache enabled for _suspend and dcache disabled for the _wfi suspend handlers. We dont have a late suspend finisher as of now, so adding one would be a good idea.

Perhaps it is an overkill. As I mentioned in my previous comment, we cannot use _wfi for CCI programming and change cache state for _suspend handler because :

1. They are outside psci locks and that means the cluster state is not guaranteed to stay when executing the _wfi handler. This means that the `final` CPU on the cluster at the point of releaing the psci_lock maynot be `final` when executing  _wfi handler, because another CPU in the cluster could have woken up and captured the psci lock and requires CCI coherency to be ON.

Hence, although using _wfi handler for CCI programming might work for Tegra (as cluster power down transitions are more controlled), this is not an optimization that can be done for other platforms (even single cluster ones if it requires CCI programming).

2.  The effects of programming the CCI is not isolated to the local CPU and needs to be done in _suspend handler.  _wfi handler must only used for operations which are local to the CPU.

This mandates that _suspend handler is used for such programming and is invoked with D-Cache OFF.

> WARMBOOT_ENABLE_DCACHE_EARLY is enabled for Tegra platforms, so dcache is already enabled when CPUs power up.

Yes, hence late suspend finisher is not useful in your case.

> Having said that, please post the outcome of the internal meetings here. Thanks.

Currently I dont see a strong case for introducing additional pm hooks. Although they might bring some slight performance benefit, the platform interface complexity and the chances of platforms getting the some operation (like CCI programming) in the wrong hook increases. Given the direction the hardware platforms are evolving (HW_ASSISTED_COHERENCY), these additional hooks are less relevant.

Hence there is no strong appetite from our end for this.

vwadekar:
@soby-mathew 

Thanks for the info. Even though Juno/FVP wont benefit from this, Tegra would need this for sure. I believe other platforms would also benefit if dcache is enabled during suspend. What do you suggest? How should we move forward?

danh-arm:
@vwadekar 

As Soby explained, moving the existing hooks is not acceptable since it changes the semantics for all platforms, which will cause regressions. The only way forward is to add a new hook (the suggestion is `psci_plat_pm_ops->pwr_domain_suspend_early()`), which platforms may choose to use or not. Since your platform is the only known platform that would benefit from this, the onus is on you to implement it, plus the corresponding changes needed to Tegra code.

I would expect platforms to have to implement 1 or both of `pwr_domain_suspend()/pwr_domain_suspend_early()`, and this can be asserted in the code. The semantics of the new hook and the changed expectations on the platform will need documenting in the porting guide.

For symmetry, it would make sense to also provide generic support for `psci_plat_pm_ops->pwr_domain_suspend_finish_late()`, but since this is not useful to you, you can skip this.

vwadekar:
@danh-arm 

Thanks. I will push a PR with the relevant changes.

vwadekar:
Created PR #1021 with the early suspend handler implementation.

PR #999, opened 2017-06-23
Fix Tegra CFLAGS usage

douglas-raillard-arm:
Use TF_CFLAGS instead of CFLAGS, to allow CFLAGS to be overriden from
the make command line.

Change-Id: I3e5726c04bcd0176f232581b8be2c94413374ac7


douglas-raillard-arm:
Hi @vwadekar,

Can you confirm this is not causing any issue for you platform ? 

Thanks,
Douglas

vwadekar:
Verified (compilation only) that this change does not break Tegra builds.

PR #1002, opened 2017-06-27
Apply workarounds for A53 Cat A Errata 835769 and 843419

douglas-raillard-arm:
These errata are only applicable to AArch64 state. See the errata notice
for more details:
http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.epm048406/index.html

Introduce the build options ERRATA_A53_835769 and ERRATA_A53_843419.
Enable both of them for Juno.

Apply the 835769 workaround as following:
* Compile with -mfix-cortex-a53-835769
* Link with --fix-cortex-a53-835769

Apply the 843419 workaround as following:
* Link with --fix-cortex-a53-843419

The erratum 843419 workaround can lead the linker to create new sections
suffixed with "*.stub*" and 4KB aligned. The erratum 835769 can lead the
linker to create new "*.stub" sections with no particular alignment.

Also add support for LDFLAGS_aarch32 and LDFLAGS_aarch64 in Makefile for
architecture-specific linker options.

Change-Id: Iab3337e338b7a0a16b0d102404d9db98c154f8f8

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -151,6 +151,7 @@ TF_CFLAGS		+=	$(CPPFLAGS) $(TF_CFLAGS_$(ARCH))		\
 				-ffreestanding -fno-builtin -Wall -std=gnu99	\
 				-Os -ffunction-sections -fdata-sections
 
+LDFLAGS			+=	$(LDFLAGS_$(ARCH))
 LDFLAGS			+=	--fatal-warnings -O1
 LDFLAGS			+=	--gc-sections
 
--- a/lib/cpus/aarch64/cortex_a53.S
+++ b/lib/cpus/aarch64/cortex_a53.S
@@ -138,6 +138,26 @@ func check_errata_855873
 	b	cpu_rev_var_hs
 endfunc check_errata_855873
 
+/*
+ * Errata workaround for Cortex A53 Errata #835769.
+ * This applies to revisions <= r0p4 of Cortex A53.
+ * This workaround is statically enabled at build time.
+ */
+func check_errata_835769
+	mov	x1, #0x04
+	b	cpu_rev_var_ls
+endfunc check_errata_835769
+
+/*
+ * Errata workaround for Cortex A53 Errata #843419.
+ * This applies to revisions <= r0p4 of Cortex A53.
+ * This workaround is statically enabled at build time.
+ */
+func check_errata_843419
+	mov	x1, #0x04
+	b	cpu_rev_var_ls
+endfunc check_errata_843419
+
 	/* -------------------------------------------------
 	 * The CPU Ops reset function for Cortex-A53.
 	 * Shall clobber: x0-x19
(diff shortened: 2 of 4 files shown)


douglas-raillard-arm:
Hi @vwadekar, @hzhuang1, @mtk09422, @rkchrome, @sorenb-xlnx, 

This patch enables a compile/link time workaround for an A53 errata (revision <= r0p4). It may apply to some other platforms than Juno.

Thanks,
Douglas

sorenb-xlnx:
@douglas-raillard-arm : Thanks for the heads up. I think zynqmp is affected too.

rkchrome:
Hi Douglas,

The rk3399 don't need enable it, maybe the older rk3368 SoCs have to enable it.

RK3399-A53:
*843419(A): A load or store might access an incorrect address  (read REVIDR_EL1, fixed by hardware)
*835769(A): AArch64 multiply-accumulate instruction might produce incorrect result (read REVIDR_EL1, fixed by hardware)

The A53 on rk3399 is not susceptible to erratum 843419 as evidenced by the read of the REVIDR_EL1:
  REVIDR_EL1 0x00000180


Thanks.
-Caesar

hzhuang1:
Hi Douglas,

I checked both HiKey and HiKey960 platform.
  * A53 r0p3 is used on HiKey platform. The value of REVIDR_EL1 is 0xc0.
  * A53 r0p4 is used on HiKey960 platform. The value of REVIDR_EL1 is 0x180.

So I think this issue has been already fixed on both HiKey and HiKey960. We don't need to enable it.

antonio-nino-diaz-arm:
Actually, @hzhuang1, "A53 r0p3 is used on HiKey platform. The value of REVIDR_EL1 is 0xc0.". Bit 8 isn't set, so it seems it is affected by 843419.

hzhuang1:
@antonio-nino-diaz-arm, Thanks for the correction. I'm always using uefi-tools to build ATF & UEFI. Is the link parameters for Juno merged into uefi-tools git repository?

douglas-raillard-arm:
@hzhuang1, are you talking about this repository ? https://git.linaro.org/uefi/uefi-tools.git
If that is the case, it looks like it is merely calling into the ATF build system. With this patch, the linker options are enabled by the ATF Makefiles when PLAT=juno so I assume it should be fine (see plat/arm/board/juno/platform.mk).

The same can be done for other platforms than Juno in the respective platform Makefiles if that is relevant to avoid the need of specifying on the command line.

sorenb-xlnx:
Zynqmp should be ok:
```
 revidr_el1:         00000180
```

sorenb-xlnx:
This is now causing warnings on unaffected HW:
```
WARNING: BL31: cortex_a53: errata workaround for 835769 was missing!
WARNING: BL31: cortex_a53: errata workaround for 843419 was missing!
```

danh-arm:
Sorry @sorenb-xlnx - it looks like we need to enhance our errata checking code for these errata (and possibly others) to use REVIDR_EL1 as well as MIDR_EL1. We'll create a task for this but until then I think you'll have to live with those warnings.

hzhuang1:
@douglas-raillard-arm I found that --fix-cortex-a53-835769 is available from GCC4.9, and --fix-cortex-a53-843419 is available from GCC5.0. But nobody checks the GCC version in Makefile. and there's no README on this.

douglas-raillard-arm:
Hi @hzhuang1, --fix-cortex-a53-843419 is actually a linker option that has been introduced in binutils 2.26 (in addition to be available through a -mfix-cortex-a53-843419 GCC flag) . We do not currently check the compiler version as we only support the current Linaro release 17.04 with a GCC version newer than 5.0. Some documentation patches are in flight to make it bit more explicit.

PR #1006, opened 2017-06-28
Improve format of exception vectors in BL1 description

robertovargas-arm:
Without the additional newlines all the text becomes a single paragraph
and next newlines are ignored.

Change-Id: I783198477f654e3923fcabb21248f2bc62c33e9d


davidcunado-arm:
jenkins: add to whitelist

PR #1012, opened 2017-06-30
rockchip/rk3399: fixes the typo and the WARNINGS during suspend/resume

rkchrome:
This patch fixes the two things as follows:

1) rk3399_flash_l2_b" seems to be a typo. That's "flush", not "flash".

2) fixes the warnings log.
We always hit the warnings thing during the suspend, as below log:
..
[   51.022334] CPU5: shutdown
[   51.025069] psci: CPU5 killed.
INFO:    sdram_params->ddr_freq = 928000000
WARNING: rk3399_flash_l2_b:reg 28830380,wait

When the L2 completes the clean and invalidate sequence, it asserts the
L2FLUSHDONE signal. The SoC can now deassert L2FLUSHREQ signal and then
the L2 deasserts L2FLUSHDONE.

Then, a loop without a delay isn't really great to measure time. We should
probably add a udelay(10) or so in there and then maybe replace the WARN()
after the loop. In the actual tests, the L2 cache will take ~4ms by
default for big cluster.

In the real world that give 10ms for the enough margin, like the
ddr/cpu/cci frequency and other factors that will affect it.

Change-Id: I55788c897be232bf72e8c7b0e10cf9b06f7aa50d

### DIFF
--- a/plat/rockchip/rk3399/drivers/pmu/pmu.c
+++ b/plat/rockchip/rk3399/drivers/pmu/pmu.c
@@ -362,19 +362,23 @@ static void pmu_power_domains_resume(void)
 	clk_gate_con_restore();
 }
 
-void rk3399_flash_l2_b(void)
+void rk3399_flush_l2_b(void)
 {
 	uint32_t wait_cnt = 0;
 
 	mmio_setbits_32(PMU_BASE + PMU_SFT_CON, BIT(L2_FLUSH_REQ_CLUSTER_B));
 	dsb();
 
+	/*
+	 * The Big cluster flush L2 cache took ~4ms by default, give 10ms for
+	 * the enough margin.
+	 */
 	while (!(mmio_read_32(PMU_BASE + PMU_CORE_PWR_ST) &
 		 BIT(L2_FLUSHDONE_CLUSTER_B))) {
 		wait_cnt++;
-		if (wait_cnt >= MAX_WAIT_COUNT)
-			WARN("%s:reg %x,wait\n", __func__,
-			     mmio_read_32(PMU_BASE + PMU_CORE_PWR_ST));
+		udelay(10);
+		if (wait_cnt == 10000 / 10)
+			WARN("L2 cache flush on suspend took longer than 10ms\n");
 	}
 
 	mmio_clrbits_32(PMU_BASE + PMU_SFT_CON, BIT(L2_FLUSH_REQ_CLUSTER_B));
@@ -391,7 +395,7 @@ static void pmu_scu_b_pwrdn(void)
 		return;
 	}
 
-	rk3399_flash_l2_b();
+	rk3399_flush_l2_b();
 
 	mmio_setbits_32(PMU_BASE + PMU_SFT_CON, BIT(ACINACTM_CLUSTER_B_CFG));
 


plat/rockchip/rk3399/drivers/pmu/pmu.c
```diff
@@ -362,19 +362,23 @@ static void pmu_power_domains_resume(void)
...
+	 * the enough margin.
+	 */
 	while (!(mmio_read_32(PMU_BASE + PMU_CORE_PWR_ST) &
 		 BIT(L2_FLUSHDONE_CLUSTER_B))) {
 		wait_cnt++;
-		if (wait_cnt >= MAX_WAIT_COUNT)
-			WARN("%s:reg %x,wait\n", __func__,
-			     mmio_read_32(PMU_BASE + PMU_CORE_PWR_ST));
+		udelay(10);
+		if (wait_cnt == 10000 / 10)
```
danh-arm: This loop used to use MAX_WAIT_COUNT, now it uses a magic number. Would be better as a #define.

plat/rockchip/rk3399/drivers/pmu/pmu.c
```diff
@@ -362,19 +362,23 @@ static void pmu_power_domains_resume(void)
...
+			WARN("L2 cache flush on suspend took longer than 10ms\n");
```
danh-arm: This loop is unbounded (although that was also the case before). Would be better to eventually exit (with a panic) rather than hang.

plat/rockchip/rk3399/drivers/pmu/pmu.c
```diff
@@ -362,19 +362,23 @@ static void pmu_power_domains_resume(void)
...
 	}
```
danh-arm: This pattern of looping on a register until MAX_WAIT_COUNT has passed is used in several other places in this function. It seems that if you're changing this instance, then those other instances should be updated as well?

plat/rockchip/rk3399/drivers/pmu/pmu.c
rkchrome: feel like there's no use in having this constant here, it just means we have to look up the value when we read it. I think it would be better to put a literal 1000 in here (or, even better, (10000 / 10)).

plat/rockchip/rk3399/drivers/pmu/pmu.c
rkchrome: We should probably rewrite all of these to use a udelay or some other form of reliable timekeeping.
but I don't want that to hold up this patch since this is the one that's actually annoying.

Maybe the another patch will pull request to change them later.

plat/rockchip/rk3399/drivers/pmu/pmu.c
rkchrome: Indeed, maybe have the panic() is better for loop. 
But I don't decide the delay time. (since many factors maybe affect it). So we use the WARN() for here.

rkchrome:
Dan, thanks your comments! But I don't ready to fix it, unless that's necessary. :)

danh-arm:
jenkins; test this please

davidcunado-arm:
jenkins: test this please

PR #1015, opened 2017-07-01
readme: add Socionext UniPhier to the supported platform list

masahir0y:


danh-arm:
We only update the readme file just before a release. We will be pushing an update that includes a line for UniPhier in the next few days.

PR #1019, opened 2017-07-03
CPU_DATA_LOG2SIZE depends on cache line size

etienne-lms:
Platform may use specific cache sizes. Since CPU_DATA_LOG2SIZE
depends on the cache line size, this change sets CPU_DATA_LOG2SIZE
according to CACHE_WRITEBACK_SHIFT which is already set by several
platforms. In case the platform does not specify the cache line size,
ATF assumes the legacy value of 64 byte large cache lines.

Change-Id: Ic2d49ffe0c3e51649425fd9c8c99559c582ac5a1
Signed-off-by: Etienne Carriere <etienne.carriere@st.com>

### DIFF
--- a/include/lib/el3_runtime/cpu_data.h
+++ b/include/lib/el3_runtime/cpu_data.h
@@ -7,12 +7,15 @@
 #ifndef __CPU_DATA_H__
 #define __CPU_DATA_H__
 
+#include <platform_def.h>	/* CACHE_WRITEBACK_GRANULE required */
+
 #ifdef AARCH32
 
 #if CRASH_REPORTING
 #error "Crash reporting is not supported in AArch32"
 #endif
 #define CPU_DATA_CPU_OPS_PTR		0x0
+#define CPU_DATA_CRASH_BUF_OFFSET	0x4
 
 #else /* AARCH32 */
 
@@ -25,13 +28,17 @@
 #endif /* AARCH32 */
 
 #if CRASH_REPORTING
-#define CPU_DATA_LOG2SIZE		7
 #define CPU_DATA_CRASH_BUF_END		(CPU_DATA_CRASH_BUF_OFFSET + \
 						CPU_DATA_CRASH_BUF_SIZE)
 #else
-#define CPU_DATA_LOG2SIZE		6
 #define CPU_DATA_CRASH_BUF_END		CPU_DATA_CRASH_BUF_OFFSET
 #endif
+
+/* cpu_data size is the data size rounded up to the platform cache line size */
+#define CPU_DATA_SIZE			(((CPU_DATA_CRASH_BUF_END + \
+					CACHE_WRITEBACK_GRANULE - 1) / \
+						CACHE_WRITEBACK_GRANULE) * \
+							CACHE_WRITEBACK_GRANULE)
 
 #if ENABLE_RUNTIME_INSTRUMENTATION
 /* Temporary space to store PMF timestamps from assembly code */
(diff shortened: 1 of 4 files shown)


include/lib/el3_runtime/cpu_data.h
```diff
@@ -24,12 +26,16 @@
 
 #endif /* AARCH32 */
 
+/* CPU_DATA_LOG2SIZE depends on cache line size (FIXME:clean this) */
+#ifndef CACHE_WRITEBACK_SHIFT
+#define CACHE_WRITEBACK_SHIFT		6
+#endif
 #if CRASH_REPORTING
-#define CPU_DATA_LOG2SIZE		7
```
soby-mathew: Hi @etienne-lms, the CACHE_WRITEBACK_SHIFT is not a platform macro whereas CACHE_WRITEBACK_GRANULE is a platform macro. Hence CPU_DATA_LOG2SIZE need to be derived from CACHE_WRITEBACK_GRANULE. But it is not easy to extract the trailing zeroes because this macro is required in assembly as well hence the __builtin_ctzl cannot be used here.

The other alternative is to remove this macro totally and use  `count_tz` macro in `_cpu_data_by_index()`  to extract the trainiling zeroes from CACHE_WRITEBACK_GRANULE.

lib/el3_runtime/aarch32/cpu_data.S
```diff
@@ -34,7 +34,13 @@ endfunc _cpu_data
  * -----------------------------------------------------------------
  */
 func _cpu_data_by_index
+#if CRASH_REPORTING
+	mov	r1, #(CACHE_WRITEBACK_GRANULE * 2)
```
soby-mathew: The only concern is if the size of cpu_data_t changes, the corresponding will not reflect here. How about defining a new CPU_DATA_SIZE macro in the header which also allows to get rid of conditional compilation here?

davidcunado-arm:
@etienne-lms 
Please see comment from Soby.

etienne-lms:
@soby-mathew, i put that on hold for a while but now it's done. Please have a look at this proposal.

include/lib/el3_runtime/cpu_data.h
```diff
@@ -25,14 +28,18 @@
...
 #else
-#define CPU_DATA_LOG2SIZE		6
 #define CPU_DATA_CRASH_BUF_END		CPU_DATA_CRASH_BUF_OFFSET
 #endif
 
+/* cpu_data size is the data size rounded up to the platform cache line size */
+#define CPU_DATA_SIZE			(((CPU_DATA_CRASH_BUF_END + \
+					CACHE_WRITEBACK_GRANULE - 1) / \
+						CACHE_WRITEBACK_GRANULE) * \
+							CACHE_WRITEBACK_GRANULE)
```
danh-arm: minor: could use `round_up()` in utils_def.h

danh-arm:
Hi @soby-mathew & @davidcunado-arm. Do you have anything more to say on this PR?

include/lib/el3_runtime/cpu_data.h
etienne-lms: `round_up()` relies on `__typeof__` which is not supported in assembly.
Implement assembly compliant `round_up()/down()` in utils_def.h ?

davidcunado-arm:
Nothing more from me! @soby-mathew over to you!

include/lib/el3_runtime/cpu_data.h
danh-arm: Hmm, good point. There's another patch in flight that adds an alternative round_up macro as you suggest. That doesn't need to delay this patch though (it can be fixed later). So I'll merge as-is.

danh-arm:
jenkins: test this please

lib/el3_runtime/aarch32/cpu_data.S
```diff
@@ -34,7 +34,9 @@ endfunc _cpu_data
  * -----------------------------------------------------------------
  */
 func _cpu_data_by_index
+	mov_imm	r1, #CPU_DATA_SIZE
```
soby-mathew: The build failure is because `mov_imm` doesn't need the `#` for the immediate data. Same comment for the AArch64 file as well.

etienne-lms: oh, sorry...

danh-arm:
jenkins: retest this please

davidcunado-arm:
jenkins: retest this please

davidcunado-arm:
jenkins: test this please

PR #1020, opened 2017-07-05
optee

pengxj2017:
get pull


vchong:
huh?

danh-arm:
What is this PR about? It looks like it's trying to merge some out of date OP-TEE header parsing patches. Are you aware that a group in ARM is looking at adding this support in a more upstream friendly way?

danh-arm:
Closing since there was no response

PR #1021, opened 2017-07-06
lib: psci: early suspend handler for platforms

vwadekar:
This patch adds an early suspend handler, that executes with
SMP and data cache enabled. This handler allows platforms to
perform any early actions during the CPU suspend entry sequence.

This handler is optional and platforms can choose to implement it
depending on their needs. The `pwr_domain_suspend` handler still
exists and platforms can keep on using it without any side effects.

### DIFF
--- a/docs/porting-guide.rst
+++ b/docs/porting-guide.rst
@@ -2060,6 +2060,23 @@ power down state where as it could be either power down, retention or run state
 for the higher power domain levels depending on the result of state
 coordination. The generic code expects the handler to succeed.
 
+plat\_psci\_ops.pwr\_domain\_suspend\_pwrdown\_early() [optional]
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This optional function may be used as a performance optimization to replace
+or complement pwr_domain_suspend() on some platforms. Its calling semantics
+are identical to pwr_domain_suspend(), except the PSCI implementation only
+calls this function when suspending to a power down state, and it guarantees
+that data caches are enabled.
+
+When HW_ASSISTED_COHERENCY = 0, the PSCI implementation disables data caches
+before calling pwr_domain_suspend(). If the target_state corresponds to a
+power down state and it is safe to perform some or all of the platform
+specific actions in that function with data caches enabled, it may be more
+efficient to move those actions to this function. When HW_ASSISTED_COHERENCY
+= 1, data caches remain enabled throughout, and so there is no advantage to
+moving platform specific actions to this function.
+
 plat\_psci\_ops.pwr\_domain\_suspend()
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/lib/psci/psci_suspend.c
+++ b/lib/psci/psci_suspend.c
@@ -80,6 +80,17 @@ static void psci_suspend_to_pwrdown_start(unsigned int end_pwrlvl,
 	if (psci_spd_pm && psci_spd_pm->svc_suspend)
 		psci_spd_pm->svc_suspend(max_off_lvl);
 
+#if !HW_ASSISTED_COHERENCY
+	/*
+	 * Plat. management: Allow the platform to perform any early
+	 * actions required to power down the CPU. This might be useful for
+	 * HW_ASSISTED_COHERENCY = 0 platforms that can safely perform these
+	 * actions with data caches enabled.
+	 */
+	if (psci_plat_pm_ops->pwr_domain_suspend_pwrdown_early)
+		psci_plat_pm_ops->pwr_domain_suspend_pwrdown_early(state_info);
+#endif
+
 	/*
 	 * Store the re-entry information for the non-secure world.
 	 */
--- a/docs/platform-migration-guide.rst
+++ b/docs/platform-migration-guide.rst
@@ -146,6 +146,7 @@ for the ``plat_psci_ops`` structure which is declared as :
         void (*cpu_standby)(plat_local_state_t cpu_state);
         int (*pwr_domain_on)(u_register_t mpidr);
         void (*pwr_domain_off)(const psci_power_state_t *target_state);
+        void (*pwr_domain_suspend_early)(const psci_power_state_t *target_state);
         void (*pwr_domain_suspend)(const psci_power_state_t *target_state);
         void (*pwr_domain_on_finish)(const psci_power_state_t *target_state);
         void (*pwr_domain_suspend_finish)(
@@ -170,12 +171,12 @@ convert the power-state parameter (possibly encoding a composite power state)
 passed in a PSCI ``CPU_SUSPEND`` to the ``psci_power_state`` format. This handler
 is now mandatory for PSCI ``CPU_SUSPEND`` support.
 
-The ``plat_psci_ops`` handlers, ``pwr_domain_off`` and ``pwr_domain_suspend``, are
-passed the target local state for each affected power domain. The platform
-must execute operations specific to these target states. Similarly,
-``pwr_domain_on_finish`` and ``pwr_domain_suspend_finish`` are passed the local
-states of the affected power domains before wakeup. The platform
-must execute actions to restore these power domains from these specific
+The ``plat_psci_ops`` handlers, ``pwr_domain_off``, ``pwr_domain_suspend_early``
+and ``pwr_domain_suspend``, are passed the target local state for each affected
+power domain. The platform must execute operations specific to these target
+states. Similarly, ``pwr_domain_on_finish`` and ``pwr_domain_suspend_finish``
+are passed the local states of the affected power domains before wakeup. The
+platform must execute actions to restore these power domains from these specific
 local states.
 
 -  Difference in invocation
--- a/include/lib/psci/psci.h
+++ b/include/lib/psci/psci.h
@@ -267,6 +267,8 @@ typedef struct plat_psci_ops {
 	void (*cpu_standby)(plat_local_state_t cpu_state);
 	int (*pwr_domain_on)(u_register_t mpidr);
 	void (*pwr_domain_off)(const psci_power_state_t *target_state);
+	void (*pwr_domain_suspend_pwrdown_early)(
+				const psci_power_state_t *target_state);
 	void (*pwr_domain_suspend)(const psci_power_state_t *target_state);
 	void (*pwr_domain_on_finish)(const psci_power_state_t *target_state);
 	void (*pwr_domain_suspend_finish)(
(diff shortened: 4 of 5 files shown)


vwadekar:
This PR takes care of the feedback provided as part of PR #998

vwadekar:
@danh-arm 

If this change looks OK, I will add commits for the Tegra port and update the PR

vwadekar:
Updated the PR with a commit for Tegra platforms. Once we are done with the reviews, I will pull this downstream to help me with the actual use case. This PR will help an upcoming platform development in the downstream repo.

vwadekar:
David/Dan, any comments?

docs/platform-migration-guide.md
```diff
@@ -152,6 +152,7 @@ for the `plat_psci_ops` structure which is declared as :
         void (*cpu_standby)(plat_local_state_t cpu_state);
         int (*pwr_domain_on)(u_register_t mpidr);
         void (*pwr_domain_off)(const psci_power_state_t *target_state);
+        void (*pwr_domain_suspend_early)(const psci_power_state_t *target_state);
```
davidcunado-arm: Re-base as docs now migrated to .rst format.

docs/porting-guide.md
```diff
@@ -1903,6 +1903,27 @@ power down state where as it could be either power down, retention or run state
 for the higher power domain levels depending on the result of state
 coordination. The generic code expects the handler to succeed.
 
+#### plat_psci_ops.pwr_domain_suspend_early() [optional]
```
davidcunado-arm: Re-base as docs now migrated to .rst format.

davidcunado-arm:
@vwadekar 
From my cursory review, looks like you've addressed the comments raised in #998. But the PR needs to be re-based anyway (due to migration to .rst format for docs). 

@soby-mathew 
Could you review, please?

vwadekar:
@davidcunado-arm 

Thanks. If there are no further comments, I will rebase and submit

davidcunado-arm:
@vwadekar 
Not from me, but Soby / Dan may have some comments.

vwadekar:
@danh-arm @davidcunado-arm @soby-mathew 

Update the PR

soby-mathew:
@vwadekar, Thanks for the PR. I have the following questions in my mind which needs resolution:

1. Currently the newly introduced `pwr_domain_suspend_early` is only invoked for suspend to power down states. Although this doesn't make much difference for pure retention states (because caches are not disabled), we could still choose to have this hook applicable for powerdown and retention states.  In that case, the call site of the handler need to be moved to `psci_cpu_suspend_start()` function.
Or else, we could argue that this is not of much benefit to retention states and hence restrict this hook to powerdown suspend states. In that case, the name of the hook need to be renamed to `pwr_domain_suspend_pwrdwn_early` to clarify this.

2. The argument for adding this hook for optimization to CPU_SUSPEND also holds true for the CPU_OFF code path. In that case we need a `pwr_domain_off_early` hook as well.

3. Going by symmetry of the hooks, we could also add `pwr_domain_suspend_pwrdwn_finish_late` and `pwr_domain_on_finish_late` hooks as well as this would benefit platforms which turn on Caches late in the warm boot sequence.

Please let me know of your opinion on these. Meanwhile I can chat internally to reach a consensus for these.

Best Regards
Soby Mathew

lib/psci/psci_suspend.c
```diff
@@ -85,6 +85,14 @@ static void psci_suspend_to_pwrdown_start(unsigned int end_pwrlvl,
 	 */
 	cm_init_my_context(ep);
 
+	/*
+	 * Plat. management: Allow the platform to perform any early
+	 * actions required to power down the CPU, before we disable
+	 * the data cache.
```
danh-arm: We only disable the data cache in the `HW_ASSISTED_COHERENCY = 0` case. I suggest the comma becomes a full stop and a new sentence added: "This might be useful for `HW_ASSISTED_COHERENCY = 0` platforms that can safely perform these actions with data caches enabled."

lib/psci/psci_suspend.c
```diff
@@ -180,7 +188,8 @@ void psci_cpu_suspend_start(entry_point_info_t *ep,
 	 * Plat. management: Allow the platform to perform the
 	 * necessary actions to turn off this cpu e.g. set the
 	 * platform defined mailbox with the psci entrypoint,
-	 * program the power controller etc.
+	 * program the power controller etc. This handler executes
+	 * with data cache disabled.
```
danh-arm: That is only true for `HW_ASSISTED_COHERENCY = 0` platforms. I don't think this comment needs changing but if you do change it, this needs clarifying.

lib/psci/psci_suspend.c
```diff
@@ -180,7 +188,8 @@ void psci_cpu_suspend_start(entry_point_info_t *ep,
...
 	 */
 	psci_plat_pm_ops->pwr_domain_suspend(state_info);
```
danh-arm: I think this function is no longer mandatory in the presence of `pwr_domain_suspend_early()`. The presence of 1 or both is OK. So:

*   This call should be wrapped in a `if (psci_plat_pm_ops->pwr_domain_suspend)`
*   There should be assert that one of both of these function are present before they are used (further up this function).

docs/porting-guide.rst
```diff
@@ -2055,6 +2055,28 @@ power down state where as it could be either power down, retention or run state
...
+corresponding to the local state at each power level. The generic code
+expects the handler to succeed.
+
+This handler gets called with the data cache enabled to allow platforms
+to perform necessary actions before the CPU actually enters the suspend
+state.
+
+This platform handler is optional and can theoretically replace the
+``pwr_domain_suspend`` handler, when implemented.
+
```
danh-arm: Rather than copying and pasting the text from `pwr_domain_suspend()`, I think it would be clearer to merge these together in the same section. The text needs to make it clear that:

*   One or both of these functions must be present
*   One of the functions is called before CPU specific power down operations are performed, and one after.
*   On `HW_ASSISTED_COHERENCY = 0` platforms, one is called with data caches enabled, the other with data caches disabled. On `HW_ASSISTED_COHERENCY = 1` platforms, the data caches are enabled throughout.

davidcunado-arm:
My comments addressed - pending addressing comments from Dan.

lib/psci/psci_suspend.c
```diff
@@ -85,6 +85,14 @@ static void psci_suspend_to_pwrdown_start(unsigned int end_pwrlvl,
...
 	cm_init_my_context(ep);
 
+	/*
+	 * Plat. management: Allow the platform to perform any early
+	 * actions required to power down the CPU, before we disable
+	 * the data cache.
+	 */
+	if (psci_plat_pm_ops->pwr_domain_suspend_early)
+		psci_plat_pm_ops->pwr_domain_suspend_early(state_info);
+
```
soby-mathew: Sorry, I didn't properly think through  the full implication of moving this to psci_cpu_suspend_start() earlier. Typically we dont want to tear down platform resources for power down without invoking the SPD hook.  So the correct place for the new hook would be after the `psci_spd_pm->svc_suspend()` SPD hook.

Thinking more about this, my opinion is that this new hook should be an optional hook whereas `pwr_domain_suspend()` would still be the mandatory hook for CPU_SUSPEND.  This is because ,

1. Platforms which require interconnect programming would need to do the same after caches are disabled.
2. The sequence limitation of calling SPD hook prior to any platform programming for suspend restricts the new hook to just power down suspend and therefore the `psci_spd_pm->svc_suspend()` needs to be used for `pure` retention suspend. The new hook is not useful for `pure` retention suspends anyway.
3. Having 2 hooks for doing platform suspend operations would be confusing as a Platform API IMO. Hence maintaining the current hook as mandatory while reducing the new hook as an optimization opportunity is less confusing I think.

If we go with what I mentioned, this new hook can be renamed to `psci_plat_pm_ops->pwr_domain_suspend_pwrdown_early` or something similar.

danh-arm: OK, I agree the new hook has to go after `psci_spd_pm->svc_suspend()`. However, I'm not sure about your arguments for making this an optional hook on power down only.

For argument 1 above, platforms that require interconnect programming can continue to use `pwr_domain_suspend()`. Platforms that don't require this can use the new hook instead. Forcing the latter platforms to provide an empty `pwr_domain_suspend()` seems an over-restriction.

For argument 2 above, I think you mean to refer to `pwr_domain_suspend()` in the statement "therefore the `psci_spd_pm->svc_suspend()` needs to be used for pure retention suspend". Anyway, it would be possible to refactor the code to move `psci_do_pwrdown_sequence()` out of `psci_suspend_to_pwrdown_start()`, so that the hook is also called for retention states. But I agree the use-case for this is tenuous.

Argument 3 is purely subjective. I like the symmetry of a matching `pwr_domain_suspend_early()`, but I can see you don't. I think either way can work if it's documented properly.

I think this comes down to whether we think this new hook is all we're going to need for the foreseeable future or whether other matching hooks (e.g. for off and the finishers) are going to be needed. At the moment, I'm finding hard to see a need for other hooks and I can see your proposal would involve less code changes.

Does anyone else have an opinion?

danh-arm: I had a chat with Soby about this and we agreed the easiest thing is to go his way. So that means:

*   Move this call to just after `psci_spd_pm->svc_suspend()`. At the same time this can be wrapped in `#if !HW_ASSISTED_COHERENCY` to make it clear this is only useful in this case.
*   Rename the function to `pwr_domain_suspend_pwrdown_early()`.
*   Document the semantics of the new optional hook properly. Ignore my comments about merging `pwr_domain_suspend()` and `pwr_domain_suspend_early()` documentation. However, it should be possible to for the new function to refer to the existing `pwr_domain_suspend()` documentation rather than copy/pasting.

vwadekar:
@danh-arm @davidcunado-arm @soby-mathew 

Updated the PR. Please let me know if this looks good.

docs/porting-guide.rst
danh-arm: For some reason the GitHub UI is still showing the above outdated comment. That should now be ignored.

I don't think the above text properly captures how `pwr_domain_suspend_pwr_down_early()` should be used. Rather than make specific comments, I'm going to suggest alternative text for this section...

This optional function may be used as a performance optimization to replace or complement `pwr_domain_suspend()` on some platforms. Its calling semantics are identical to `pwr_domain_suspend()`, except the PSCI implementation only calls this function when suspending to a power down state, and it guarantees that data caches are enabled.

When `HW_ASSISTED_COHERENCY = 0`, the PSCI implementation disables data caches before calling `pwr_domain_suspend()`. If the `target_state` corresponds to a power down state and it is safe to perform some or all of the platform specific actions in that function with data caches enabled, it may be more efficient to move those actions to this function. When `HW_ASSISTED_COHERENCY = 1`, data caches remain enabled throughout, and so there is no advantage to moving platform specific actions to this function.

vwadekar:
Thanks @danh-arm. Updated the PR to address your comments.

davidcunado-arm:
jenkins: test this please

PR #1031, opened 2017-07-20
Use standard UNIX file:line format in assert

robertovargas-arm:
This format is understood by almost all the UNIX tools (vi, emacs, acme, ...),
and it allows these tools to jump directly to the line where the assert
failed.

Change-Id: I648fa93c7cc65f911a17dcad5e1a775ac1ae5ed4


davidcunado-arm:
jenkins: test this please

PR #1033, opened 2017-07-21
Address edge case for stale PSCI CPU data in cache

davidcunado-arm:
There is a theoretical edge case during CPU_ON where the cache
may contain stale data for the target CPU data - this can occur
under the following conditions:

- the target CPU is in another cluster from the current
- the target CPU was the last CPU to shutdown on its cluster
- the cluster was removed from coherency as part of the CPU shutdown

In this case the cache maintenace that was performed as part of the
target CPUs shutdown was not seen by the current CPU's cluster. And
so the cache may contain stale data for the target CPU.

This patch adds a cache maintenance operation (flush) for the
cache-line containing the target CPU data - this ensures that the
target CPU data is read from main memory.

Change-Id: If8cfd42639b03174f60669429b7f7a757027d0fb

### DIFF
--- a/lib/psci/psci_on.c
+++ b/lib/psci/psci_on.c
@@ -64,7 +64,20 @@ int psci_cpu_on_start(u_register_t target_cpu,
 	/*
 	 * Generic management: Ensure that the cpu is off to be
 	 * turned on.
+	 * Perform cache maintanence ahead of reading the target CPU state to
+	 * ensure that the data is not stale.
+	 * There is a theoretical edge case where the cache may contain stale
+	 * data for the target CPU data - this can occur under the following
+	 * conditions:
+	 * - the target CPU is in another cluster from the current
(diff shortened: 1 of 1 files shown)


lib/psci/psci_on.c
```diff
@@ -64,7 +64,20 @@ int psci_cpu_on_start(u_register_t target_cpu,
 	/*
 	 * Generic management: Ensure that the cpu is off to be
 	 * turned on.
+	 * Perform cache maintanence ahead of reading the target CPU state to
+	 * ensure that the data is not stale.
+	 * There is a theoretical edge case where the cache may contain stale
+	 * data for the target CPU data - this can occur under the following
+	 * conditions:
+	 * - the target CPU is in another other cluster from the current
```
danh-arm: minor: remove word "other".

PR #1039, opened 2017-07-26
build: reverse header include path priority

Leo-Yan:
When build fip tool, makefile includes the headers from pathes "./"
and "../../include/tools_share". In this two directories both have
header file "uuid.h", by default it includes ./uuid.h header file;
this header file has not defined macro _UUID_STR_LEN so report
building errors when build fip:

In file included from fiptool.c:10:0:
fiptool.c: In function ‘uuid_to_str’:
fiptool.c:244:17: error: ‘_UUID_STR_LEN’ undeclared (first use in this function)
  assert(len >= (_UUID_STR_LEN + 1));
                 ^

This patch reverses include paths, so give higher priority to path
"../../include/tools_share", finally can include correct uuid header
file and dismiss build error.

### DIFF
--- a/tools/fiptool/Makefile
+++ b/tools/fiptool/Makefile
@@ -27,7 +27,7 @@ else
   Q :=
 endif
 
-INCLUDE_PATHS := -I. -I../../include/tools_share
+INCLUDE_PATHS := -I../../include/tools_share -I.
 
 HOSTCC ?= gcc
 


danh-arm:
Where is the 2nd uuid.h you mention coming from? The source tree only has one (the one in ./include/tools_share/uuid.h so I don't understand why you're getting this problem.

Leo-Yan:
Ah, in my local repo I have one stale header file: tools/fiptool/uuid.h, this file should be left by the old version code. Sorry for noise, drop this patch.

masahir0y:
Hi.  This was closed, but I think -I. is unneeded in the first place.

I've just opened a respin as #1041

PR #1040, opened 2017-07-26
Support opteed header

sliai:
Support Trusted OS firmware extra images in TF tools and support extra image parse.


davidcunado-arm:
jenkins: test this please

vchong:
Changes in `arm_io_storage.c` should be in `hikey_io_storage.c`, `hikey960_io_storage.c`, `qemu_io_storage` and `fvp_io_storage.c` as well?

sliai:
@vchong Yes, you should apply the changes to hikey or qemu.

davidcunado-arm:
Actually, this patch needs to address these issues.
@vchong please hold-off and I'll discuss with @sliai what the next steps are, as the PR cannot be merged until the checks pass.

danh-arm:
If the intention is to make use of this header parsing functionality in other platforms, then the new utility code in plat/arm/common/arm_bl2_optee_utils.c should be moved to a more general location, for example plat/common/plat_bl2_optee_utils.c.

services/spd/opteed/opteed_common.c
```diff
@@ -5,6 +5,7 @@
  */
 
 #include <arch_helpers.h>
+#include <arm_def.h>
```
danh-arm: OPTEED should not pull in platform specific headers.

sliai:
@danh-arm The arm_def.h is not used in opteed_common.c anymore, sorry for that, I will delete it. And for parse function, I will move to plat/common, and named as you said.

davidcunado-arm:
jenkins: test this please

davidcunado-arm:
@hzhuang1 @vchong 
FYI - change has been updated.

sliai:
Is this error caused by build system?
ERROR 503: Service Temporarily Unavailable

sandrine-bailleux-arm:
@sliai 
No, it's not a build system issue, it's just a temporary infrastructure issue. Sorry about that.
I'll retrigger the build.

sandrine-bailleux-arm:
jenkins: test this please

davidcunado-arm:
jenkins: test this please

vchong:
This PR only adds opteed header parsing support for ARM platforms? What about the other opteed supported platforms, namely hikey, hikey960, qemu and fvp?

vchong:
Bit of a nit, but the message `This patch also adds support for legacy OP-TEE images that do not include header information` in the 3rd commit might be better in the 2nd commit instead.

vchong:
@jenswi-linaro @jbech-linaro FYI

sliai:
@vchong Yes, this patch only support on ARM Platform. Because we only test it on ARM platform. For others platforms, need whom want to use paging support to merge it.
For the comment, it is main for we add two parameters there. And those two parameters will not affect the legacy image.

vchong:
> Yes, this patch only support on ARM Platform. Because we only test it on ARM platform. For others platforms, need whom want to use paging support to merge it.

Ok. Was under the impression from https://github.com/ARM-software/arm-trusted-firmware/pull/1040#issuecomment-317990498 that other platforms would be updated as well.

> For the comment, it is main for we add two parameters there. And those two parameters will not affect the legacy image.

Sorry for 'nit-ting' but in that case it might be less confusing to just say `Legacy OP-TEE images that do not include header information are not affected`.

danh-arm:
jenkins:test this please

danh-arm:
jenkins:add to whitelist

PR #1041, opened 2017-07-26
fiptool: remove local directory from the header search path

masahir0y:


davidcunado-arm:
jenkins: test this please

PR #1053, opened 2017-08-15
Add new alignment parameter to func assembler macro

jwerner-chromium:
Assembler programmers are used to being able to define functions with a
specific aligment with a pattern like this:

    .align X
  myfunction:

However, this pattern is subtly broken when instead of a direct label
like 'myfunction:', you use the 'func myfunction' macro that's standard
in Trusted Firmware. Since the func macro declares a new section for the
function, the .align directive written above it actually applies to the
*previous* section in the assembly file, and the function it was
supposed to apply to is linked with default alignment.

An extreme case can be seen in Rockchip's plat_helpers.S which contains
this code:

  [...]
  endfunc plat_crash_console_putc

  .align 16
  func platform_cpu_warmboot
  [...]

This assembles into the following plat_helpers.o:

  Sections:
  Idx Name                             Size  [...]  Algn
   9 .text.plat_crash_console_putc 00010000  [...]  2**16
  10 .text.platform_cpu_warmboot   00000080  [...]  2**3

As can be seen, the *previous* function actually got the alignment
constraint, and it is also 64KB big even though it contains only two
instructions, because the .align directive at the end of its section
forces the assembler to insert a giant sled of NOPs. The function we
actually wanted to align has the default constraint. This code only
works at all because the linker just happens to put the two functions
right behind each other when linking the final image, and since the end
of plat_crash_console_putc is aligned the start of platform_cpu_warmboot
will also be. But it still wastes almost 64KB of image space
unnecessarily, and it will break under certain circumstances (e.g. if
the plat_crash_console_putc function becomes unused and its section gets
garbage-collected out).

There's no real way to fix this with the existing func macro. Code like

 func myfunc
 .align X

happens to do the right thing, but is still not really correct code
(because the function label is inserted before the .align directive, so
the assembler is technically allowed to insert padding at the beginning
of the function which would then get executed as instructions if the
function was called). Therefore, this patch adds a new parameter with a
default value to the func macro that allows overriding its alignment.

Also fix up all existing instances of this dangerous antipattern.

### DIFF
--- a/bl32/tsp/aarch64/tsp_entrypoint.S
+++ b/bl32/tsp/aarch64/tsp_entrypoint.S
@@ -43,10 +43,7 @@
 	msr	spsr_el1, \reg2
 	.endm
 
-	.section	.text, "ax"
-	.align 3
-
-func tsp_entrypoint
+func tsp_entrypoint _align=3
 
 	/* ---------------------------------------------
 	 * Set the exception vector to something sane.
--- a/include/common/asm_macros_common.S
+++ b/include/common/asm_macros_common.S
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013-2016, ARM Limited and Contributors. All rights reserved.
+ * Copyright (c) 2013-2017, ARM Limited and Contributors. All rights reserved.
  *
  * SPDX-License-Identifier: BSD-3-Clause
  */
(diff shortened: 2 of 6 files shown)


bl32/tsp/aarch64/tsp_entrypoint.S
```diff
@@ -43,10 +43,7 @@
 	msr	spsr_el1, \reg2
 	.endm
 
-	.section	.text, "ax"
-	.align 3
-
-func tsp_entrypoint
+func tsp_entrypoint _align=3
```
danh-arm: I suspect the .align 3 here is a historical artefact and is not needed. But that's off-topic.

danh-arm:
jenkins: test this please

PR #1054, opened 2017-08-15
Fix x30 reporting for unhandled exceptions

jwerner-chromium:
Some error paths that lead to a crash dump will overwrite the value in
the x30 register by calling functions with the no_ret macro, which
resolves to a BL instruction. This is not very useful and not what the
reader would expect, since a crash dump should usually show all
registers in the state they were in when the exception happened. This
patch replaces the offending function calls with a B instruction to
preserve the value in x30.

Change-Id: I2a3636f2943f79bab0cd911f89d070012e697c2a

### DIFF
--- a/common/aarch64/debug.S
+++ b/common/aarch64/debug.S
@@ -175,6 +175,6 @@ el3_panic:
 _panic_handler:
 	/* Pass to plat_panic_handler the address from where el3_panic was
 	 * called, not the address of the call from el3_panic. */
-	mov	x30,x6
-	no_ret	plat_panic_handler
+	mov	x30, x6
+	b	plat_panic_handler
 endfunc do_panic
--- a/bl31/aarch64/runtime_exceptions.S
+++ b/bl31/aarch64/runtime_exceptions.S
@@ -49,7 +49,8 @@
 	b.eq	smc_handler64
 
 	/* Other kinds of synchronous exceptions are not handled */
-	no_ret	report_unhandled_exception
+	ldr	x30, [sp, #CTX_GPREGS_OFFSET + CTX_GPREG_LR]
+	b	report_unhandled_exception
 	.endm
 
 
@@ -152,7 +153,7 @@ vector_base runtime_exceptions
 	 */
 vector_entry sync_exception_sp_el0
 	/* We don't expect any synchronous exceptions from EL3 */
-	no_ret	report_unhandled_exception
+	b	report_unhandled_exception
 	check_vector_size sync_exception_sp_el0
 
 vector_entry irq_sp_el0
@@ -160,17 +161,17 @@ vector_entry irq_sp_el0
 	 * EL3 code is non-reentrant. Any asynchronous exception is a serious
 	 * error. Loop infinitely.
 	 */
-	no_ret	report_unhandled_interrupt
+	b	report_unhandled_interrupt
 	check_vector_size irq_sp_el0
 
 
 vector_entry fiq_sp_el0
-	no_ret	report_unhandled_interrupt
+	b	report_unhandled_interrupt
 	check_vector_size fiq_sp_el0
 
 
 vector_entry serror_sp_el0
-	no_ret	report_unhandled_exception
+	b	report_unhandled_exception
 	check_vector_size serror_sp_el0
 
 	/* ---------------------------------------------------------------------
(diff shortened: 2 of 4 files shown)


vwadekar:
Looks good. Ack.

jeenu-arm:
I agree `LR` is being clobbered in this case. However, to identify "code that actually caused the exception", in case of a synchronous exception, one should be looking at `ELR_EL3`, not `LR`.

danh-arm:
I'm a bit confused by this. The address of the code that caused the exception is stored in the ELR_ELX register, not x30, so why does it matter if x30 is overwritten? The crash dump already reports the state of the EL3_ELX registers.

Having said that, I agree some of these "no_ret" calls can result in loss of useful information in x30. For example, do_panic ends up calling "no_ret plat_panic_handler", which is the place where panics will usually be debugged from. While in plat_panic_handler(), x30 will point to do_panic, rather than the function which called panic(), which would be more useful.

jeenu-arm:
We just realized that the the exception message is specifically calling out `x30`, which we agree is confusing. Given the scarcity of registers at the time this occurring, it's not easy to preserve all CPU registers whilst keeping the same mechanism intact. So we think part where `x30` being called out should be removed from the messages instead, to avoid a reader being mislead (irrelevance notwithstanding).

soby-mathew:
The intention for x30 during `report_unhandled_exception` was to indicate the vector address it was triggered from which makes it easier to figure out the type of exception without looking at ESR_EL3. Hence the use of `bl` instruction to branch in the exception vectors and IMO they need not be changed.

As Dan pointed out, the do_panic() handler should use `b` instruction (which it was originally before switch to `no_ret` macro)  to indicate the caller of panic handler. This need to be corrected.

jwerner-chromium:
Just to clarify, I know what ELR_EL3 is and the code is working fine for that. With "the x30
register from the code that caused the exception" I don't mean the address of the instruction that caused it (that is in ELR), I mean the state the x30 register was in while executing that instruction.

The point of a crash dump is to show you the exact state of registers when the exception happens, to give you a better idea of what the machine was doing at that time. The current crash handler does exactly that for x0 - x29 (they're all printed with the values they had at the time of the exception, which the exception handler takes a lot of care not to clobber). But for x30, it shows the exception vector address rather than the value x30 had when the exception occurred. This is what I think is both confusing (because it differs from the way the other registers are dumped) and unhelpful.

In my experience, x30 can be one of the most useful registers in a crash dump for environments that don't print a full stack backtrace, because it can give you a clue where the program was when the actual crashing instruction was in a common helper function (e.g. tf_printf()). The type of exception (as Soby mentioned) can also be inferred (in much more detail, even) from ESR_EL3, so I think it's much more useful to print x30 in the way that the programmer would expect and that gives him additional helpful information (which function the crashing function was called from).

soby-mathew:
Yes, I agree.  Only the issue with do_panic() `x30` reporting remains which is good to be resolved as well along with this patch.

jwerner-chromium:
Thanks, Soby. I'll update the PR to fix the x30 passed to plat_panic_handler for the panic() cases. However, note that plat_panic_handler is also called from many more places (over 50!), almost all of them use no_ret, and for many of them it would not be trivial to restore the correct x30 because it's no longer available at that point (e.g. asm_assert). The work to figure out how the register value can be preserved in all those cases should probably be left to a separate patch.

bl31/aarch64/crash_reporting.S
```diff
@@ -327,18 +327,16 @@ func do_crash_reporting
 
 	bl	plat_crash_console_flush
 
-	/* Done reporting */
-	no_ret	plat_panic_handler
+crash_panic:
+	/* Done reporting, restore original x30 */
+	mrs	x0, tpidr_el3
+	ldr	x30, [x0, #REG_SIZE * 7]
```
danh-arm: At this point `[x0, #REG_SIZE * 7]` no longer contains x30 as this location has been re-used for other purposes (e.g. in line 231). It's not going to be easy to restore x30 in this case so I suggest leaving it as it was.

common/aarch64/debug.S
```diff
@@ -175,6 +175,6 @@ el3_panic:
 _panic_handler:
 	/* Pass to plat_panic_handler the address from where el3_panic was
 	 * called, not the address of the call from el3_panic. */
-	mov	x30,x6
-	no_ret	plat_panic_handler
+	mov	x30, x6
+	b	plat_panic_handler
```
danh-arm: An equivalent change could be made to common/aarch32/debug.S

bl31/aarch64/crash_reporting.S
jwerner-chromium: Oh right, sorry, I missed that. I'll leave it out then.

common/aarch64/debug.S
jwerner-chromium: Done, but note that I don't have an AArch32 platform to test it with.

danh-arm: No problem - we'll test our end

danh-arm:
jenkins: test this please

PR #1056, opened 2017-08-17
update the interrupt diagrams

geesun:
The sec-int-handling.png was mismatched with the code, we should change TSP_HANDLED_S_EL1_FIQ to TSP_HANDLED_S_EL1_INTR in sec-int-handling.png. 
   
   - Redraw the interrupt diagrams with dia tool
   - Change TSP_HANDLED_S_EL1_FIQ to TSP_HANDLED_S_EL1_INTR in sec-int-handling.png
   - Use the makefile generate the image to avoid unnessary generate
   - Add dia source code


antonio-nino-diaz-arm:
Why is the resolution of the new png files so small compared to the old ones?

geesun:
I update the script to generate the image the same size as before.

danh-arm:
jenkins: add to whitelist

danh-arm:
jenkins: test this please

PR #1058, opened 2017-08-21
psci_common: Resolve GCC static analysis false positive

alistair23:
Previously commit 555ebb34db8f3424c1b394df2f10ec attmpted to fix this
GCC issue:

services/std_svc/psci/psci_common.c: In function 'psci_do_state_coordination':
services/std_svc/psci/psci_common.c:220:27: error: array subscript is above
array bounds [-Werror=array-bounds]
  psci_req_local_pwr_states[pwrlvl - 1][cpu_idx] = req_pwr_state;

This fix doesn't work as asserts aren't built in non-debug build flows.
Let's ensure this error is fixed for all build cases.

Signed-off-by: Alistair Francis <alistair.francis@xilinx.com>
Signed-off-by: Nathan Rossi <nathan@nathanrossi.com>

### DIFF
--- a/lib/psci/psci_common.c
+++ b/lib/psci/psci_common.c
@@ -194,8 +194,15 @@ static void psci_set_req_local_pwr_state(unsigned int pwrlvl,
 					 unsigned int cpu_idx,
 					 plat_local_state_t req_pwr_state)
 {
+	/*
+	 * This should never happen, we have this here to avoid
+	 * "array subscript is above array bounds" errors in GCC.
+	 */
 	assert(pwrlvl > PSCI_CPU_PWR_LVL);
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Warray-bounds"
 	psci_req_local_pwr_states[pwrlvl - 1][cpu_idx] = req_pwr_state;
+#pragma GCC diagnostic pop
 }
 
 /******************************************************************************


sivadur:
ASSERT was used in patch "555ebb34db8f3424c1b394df2f10ecf9c1f70901" just to 
resolve static analysis  error. But now as per you, it doesn’t really solves it in Non debug flows. Then,
we can completely remove ASSERT and keep this if condition which solves issue
for both Debug and Non debug flows.

lib/psci/psci_common.c
```diff
@@ -395,6 +395,8 @@ void psci_do_state_coordination(unsigned int end_pwrlvl,
 	plat_local_state_t target_state, *req_states;
 
 	assert(end_pwrlvl <= PLAT_MAX_PWR_LVL);
+	if (end_pwrlvl > PLAT_MAX_PWR_LVL)
+		return;
```
danh-arm: As stated [here](https://github.com/ARM-software/arm-trusted-firmware/pull/1058#issuecomment-323919722), the assert can be removed if a runtime check is added. Also, since this case should never happen, it would be better to panic() instead of returning. A comment to explain this would also help.

Actually, given that this is a false positive, I would prefer a solution that didn't result in additional code in the final binary. For example, a compiler hint to suppress the warning. But I don't know if that is possible.

alistair23: The logic for leaving the assert() in is that in the future if someone made a change and there was actually an above array bounds error it would be hard to detect with just a return from a function. An assert() would ensure that the issue was caught (assuming it's compiled in).

If you are happy with using a panic() call instead of a return I'll use that instead and remove the assert().

As for not adding code, I'm not sure of anyway to do that besides disabling the check which I think is a bad idea. I'll re-spin the pull request with a panic() and remove the assert().

alistair23:
Just pushed a new version with the assert() removed and panic() used instead.

danh-arm:
Has anyone investigated using `#pragma GCC diagnostic` to suppress this warning?
https://gcc.gnu.org/onlinedocs/gcc/Diagnostic-Pragmas.html

It looks like this code could simply be wrapped in:

```
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Werror=array-bounds"
parent_idx = psci_cpu_pd_nodes[cpu_idx].parent_node;  foo(b);
#pragma GCC diagnostic pop
```

alistair23:
I tried this and I still see the error:

`commit 7c3fdfbce003713cb6daecf43d88b2187cda0687
Author: Alistair Francis <alistair.francis@xilinx.com>
Date:   Wed Aug 23 13:55:21 2017 -0700

    psci_common: Resolve GCC static analysis false positive
    
    Previously commit 555ebb34db8f3424c1b394df2f10ecf9c1f70901 attmpted to fix this
    GCC issue:
    
    services/std_svc/psci/psci_common.c: In function 'psci_do_state_coordination':
    services/std_svc/psci/psci_common.c:220:27: error: array subscript is above
    array bounds [-Werror=array-bounds]
      psci_req_local_pwr_states[pwrlvl - 1][cpu_idx] = req_pwr_state;
    
    This fix doesn't work as asserts aren't built in non-debug build flows.
    
    Let's use GCCs #pragma option (documented here:
    https://gcc.gnu.org/onlinedocs/gcc/Diagnostic-Pragmas.html) to avoid
    this false positive instead.
    
    Signed-off-by: Alistair Francis <alistair.francis@xilinx.com>

diff --git a/lib/psci/psci_common.c b/lib/psci/psci_common.c
index f31b3238..1a1a37eb 100644
--- a/lib/psci/psci_common.c
+++ b/lib/psci/psci_common.c
@@ -195,7 +195,10 @@ static void psci_set_req_local_pwr_state(unsigned int pwrlvl,
                                         plat_local_state_t req_pwr_state)
 {
        assert(pwrlvl > PSCI_CPU_PWR_LVL);
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Werror=array-bounds"
        psci_req_local_pwr_states[pwrlvl - 1][cpu_idx] = req_pwr_state;
+#pragma GCC diagnostic pop
 }
 
 /******************************************************************************
@@ -394,8 +397,15 @@ void psci_do_state_coordination(unsigned int end_pwrlvl,
        unsigned int start_idx, ncpus;
        plat_local_state_t target_state, *req_states;
 
+       /*
+        * This should never happen, we have this here to avoid
+        * "array subscript is above array bounds" errors in GCC.
+        */
        assert(end_pwrlvl <= PLAT_MAX_PWR_LVL);
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Werror=array-bounds"
        parent_idx = psci_cpu_pd_nodes[cpu_idx].parent_node;
+#pragma GCC diagnostic pop
 
        /* For level 0, the requested state will be equivalent
           to target state */
`

alistair23:
That worked. I just pushed a new patch with the GCC pragmas.

danh-arm:
jenkins: test this please

PR #1059, opened 2017-08-23
fix a typo abort sctlr_el2

kenkuang:
which will cause write_sctlr_el2 use all sctlr_el1 value except the EE bit

the code doesn't "Use SCTLR_EL1.EE value to initialise sctlr_el2"
but, read out SCTLR_EL1 and clear EE bit, then set to sctlr_el2

Signed-off-by: Ken Kuang <ken.kuang@spreadtrum.com>


danh-arm:
jenkins: test this please

PR #1069, opened 2017-08-29
Hikey: enable watchdog reset

Leo-Yan:
At the system boot time we need enable watchdog reset, otherwise after
the watchdog is timeout it cannot reset the SoC. We need set the bit 0
and bit 16 together, the bit 16 is mask bit so after set bit 16 we have
permission to operate bit 0 and bit 0 is watchdog reset enabling bit.


Leo-Yan:
@hzhuang1 could you review this patch?

danh-arm:
jenkins:test this please

hzhuang1:
@Leo-Yan Acked

PR #1072, opened 2017-08-30
ARM platforms: Map TSP only when TSPD is included

sandrine-bailleux-arm:
This patch ensures that the ARM_MAP_TSP_SEC_MEM memory region is mapped
in BL2 only if the TSPD has been included in the build. This saves one
entry in the plat_arm_mmap[] array and avoids to map extra memory when
it's not needed.


vchong:
`#ifdef SPD_tspd` instead of `#if SPD_tspd`?

sandrine-bailleux-arm:
@vchong Yes, that's better, thanks. I've updated the pull request.

PR #1074, opened 2017-08-30
fiptool: Update file open modes with 'b' (for Windows)

EvanLloyd:
Unix does not distinguish binary and text modes.
On Windows the 'b' flag (e.g. "rb" instead of "r") is used to
indicate that files should be opened in binary mode.
This has no impact on Unix, but is needed on Windows to avoid
end-of-line issues.

Change-Id: I69424c55735d5d563d36c50bedd6357b8e05137e
Signed-off-by: Evan Lloyd <evan.lloyd@arm.com>


davidcunado-arm:
jenkins: test this please

PR #1076, opened 2017-08-31
asm_macros: set the default assembly code alignment to 4 byte

masahir0y:


include/common/asm_macros_common.S
```diff
@@ -16,7 +16,7 @@
 	 * use a raw .align directive. Since func switches to a new section,
 	 * this would not have the desired effect.
 	 */
-	.macro func _name, _align=-1
+	.macro func _name, _align=2
```
danh-arm: minor: it would be good to add a comment in line 15 to say the alignment defaults to 2 since both A32 and A64 instructions must be word aligned.

danh-arm:
jenkins: test this please

masahir0y:
> Looks good, although I'm curious to know if you've seen any alignment faults in practice? Although, this code was modified recently, it looks like this is potentially a long standing issue.

I usually use as new compiler as possible.  I noticed this problem just yesterday.
(As mentioned in git-log, I do not observe this problem with GCC 6 or newer.)

I think the condition triggering this problem is complex.

In my platform, plat/socionext/uniphier/uniphier_rotpk.S happens to be located before 
bl1/aarch64/bl1_exceptions.S

So, there is a unaligned symbol before "func smc_handler64"
because the DER header for TBBR is 19 byte.

My platform is an unlucky case, but this potential problem may arise for others.

masahir0y:
> minor: it would be good to add a comment in line 15 to say the alignment defaults to 2 since both A32 and A64 instructions must be word aligned.

I added a comment and pushed.

danh-arm:
jenkins: retest this please

PR #1081, opened 2017-09-01
uniphier: fix-up for PXs3 SoC

masahir0y:


davidcunado-arm:
jenkis: test this please

PR #1083, opened 2017-09-05
Fix BL2 memory map when OP-TEE is the Secure Payload

soby-mathew:
The commit 3eb2d67 optimizes the memory map for BL2 when TSP
is not present. But this also broke OP-TEE as it was reusing
the TSP mapping. This patch fixes this problem by adding a
separate mapping for OP-TEE in the BL2 memory map table.

Change-Id: I130a2ea552b7b62d8478081feb1f4ddf5292a118


davidcunado-arm:
jenkins: test this please

PR #1087, opened 2017-09-06
Reduce time lock in psci_do_cpu_off

robertovargas-arm:
psci_set_power_off_state only initializes a local variable, so there
isn't any reason why it should be done while the lock is held.

Change-Id: I1c62f4cd5d860d102532e5a5350152180d41d127


davidcunado-arm:
jenkins: test this please

PR #1090, opened 2017-09-07
Hikey: enable CPU debug module

Leo-Yan:
Every CPU has its own debug module and this module is used by JTAG
debugging and coresight tracing. If without enabling it, it's easily to
introduce lockup issue when we enable debugging features.

This patch is to enable CPU debug module when power on CPU; this allows
connecting to all cores through JTAG and used by kernel coresight
driver.

Signed-off-by: Matthias Welwarsky <maw@sysgo.com>


davidcunado-arm:
@Leo-Yan 
Suggest that you close this PR and update #1089 with changes here and change to fix the style-checker failure.

PR #1093, opened 2017-09-11
Implement log framework

soby-mathew:


davidcunado-arm:
@sbranden 
FYI - this patch looks to address some of the future-proofing concerns that were in #883.
I'll look to close #883 once this PR is merged.

PR #1097, opened 2017-09-12
plat/arm: Fix BL31_BASE when RESET_TO_BL31=1

davidcunado-arm:
The value of BL31_BASE currently depends on the size of BL31. This
causes problems in the RESET_TO_BL31 case because the value of
BL31_BASE is used in the model launch parameters, which often changes.

Therefore, this patch fixes BL31_BASE to the middle of Trusted SRAM,
to avoid further model parameter changes in future.

Change-Id: I6d7fa4fe293717d84768974679539c0e0cb6d935


davidcunado-arm:
NOTE - expect the checks to fail until the corresponding change in the internal CI is made to use the new BL31_BASE address when using the RESET_TO_BL31 option.

davidcunado-arm:
jenkins: test this please

PR #1100, opened 2017-09-16
trusty: save/restore FPU registers in world switch

ajs-sun:
Currently, Trusty OS/LK implemented FPU context switch in internal
thread switch but does not implement the proper mechanism for world
switch. This commit just simply saves/restores FPU registes in world
switch to prevent FPU context from being currupted when Trusty OS uses
VFP in its applications.

It should be noted that the macro *CTX_INCLUDE_FPREGS* must be defined
in trusty.mk if Trusty OS uses VFP

Signed-off-by: goodbach <goodbach@gmail.com>

### DIFF
--- a/services/spd/trusty/trusty.c
+++ b/services/spd/trusty/trusty.c
@@ -99,6 +99,16 @@ static struct args trusty_context_switch(uint32_t security_state, uint64_t r0,
 	ret.r1 = r1;
 	ret.r0 = r0;
 
+	/*
+	 * To avoid the additional overhead in PSCI flow, skip FP context
+	 * saving/restoring in case of CPU suspend and resume, asssuming that
+	 * when it's needed the PSCI caller has preserved FP context before
+	 * going here.
+	 */
+#if CTX_INCLUDE_FPREGS
+	if (r0 != SMC_FC_CPU_SUSPEND && r0 != SMC_FC_CPU_RESUME)
+		fpregs_context_save(get_fpregs_ctx(cm_get_context(security_state)));
+#endif
 	cm_el1_sysregs_context_save(security_state);
 
 	ctx->saved_security_state = security_state;
@@ -107,6 +117,11 @@ static struct args trusty_context_switch(uint32_t security_state, uint64_t r0,
 	assert(ctx->saved_security_state == !security_state);
 
 	cm_el1_sysregs_context_restore(security_state);
+#if CTX_INCLUDE_FPREGS
+	if (r0 != SMC_FC_CPU_SUSPEND && r0 != SMC_FC_CPU_RESUME)
+		fpregs_context_restore(get_fpregs_ctx(cm_get_context(security_state)));
+#endif
+
 	cm_set_next_eret_context(security_state);
 
 	return ret;


davidcunado-arm:
@ajs-sun 
Ahead of reviewing this patch, there are a few submission process (as per the arm-trusted-firmware/contributing.rst guide) issues to resolve:

- Please target the integration branch rather than master - I'll change this now, but an update is going to be needed and so please ensure correct branch is targeted.
- Please use your real name for the Signed-off field.

Additionally, the sub-maintainer for trusty, @vwadekar will need to approve this change.

services/spd/trusty/trusty.c
```diff
@@ -99,6 +99,9 @@ static struct args trusty_context_switch(uint32_t security_state, uint64_t r0,
 	ret.r1 = r1;
 	ret.r0 = r0;
 
+#if CTX_INCLUDE_FPREGS
+	fpregs_context_save(get_fpregs_ctx(cm_get_context(security_state)));
+#endif
```
danh-arm: Initially when I looked at this, I wondered why we don't already call `fpregs_context_save()` as part of `cm_el1_sysregs_context_save()`, to avoid you having to do this. However, I know the reason why now...

For general world switch cases (e.g. SMCs, interrupts), this code is the right thing to do, since the specs (e.g. SMC Calling Convention) say FP&SIMD state should be preserved by the firmware. However, this is not the case for PSCI operations that result in a loss of power (e.g. CPU_OFF/CPU_SUSPEND). In those cases, it's expected that the caller preserves FP&SIMD state before invoking the PSCI SMCs. Therefore FP&SIMD save/restore in those case is not necessary, and doing it may have a significant performance impact.

I suggest either you move these FP save/restore invocations to the calling code (except the suspend/resume case), or you inspect the value of r0 here to see if it's not equal to `SMC_FC_CPU_SUSPEND`/`SMC_FC_CPU_RESUME` before saving FP&SIMD state.

ajs-sun: Thanks for your review and suggestion. The update has been uploaded with changed name info.
I prefer to inspect the value before calling the FP save/restore routine. It seems like that there is no exactly good place to call these rotines  if I move them out from `trusty_context_switch`..

services/spd/trusty/trusty.c
```diff
@@ -99,6 +99,10 @@ static struct args trusty_context_switch(uint32_t security_state, uint64_t r0,
 	ret.r1 = r1;
 	ret.r0 = r0;
 
+#if CTX_INCLUDE_FPREGS
```
danh-arm: Looks good, although It might be nice to add a comment here to say FP&SIMD state should be preserved by the SMC caller in the CPU suspend/resume case.

ajs-sun: Thanks, uploaded new.

davidcunado-arm:
jenkins: test this please

davidcunado-arm:
@vwadekar 
Can you review these changes to trusty?

vwadekar:
Looks good.

Acked-by: Varun Wadekar

davidcunado-arm:
jenkins: test this please

davidcunado-arm:
jenkins: test this please

davidcunado-arm:
jenkins: test this please

davidcunado-arm:
jenkins: test this please

PR #1103, opened 2017-09-19
New console API and coreboot support

jwerner-chromium:
Hi @danh-arm,

Here's the third iteration of my console PR. I hope I managed to incorporate all your suggestions from the last review, please let me know if I forgot something. Since we decided to guard the new API by a compile-time option now, I had more freedom with the design and didn't need to use all the weak symbol and macro magic. Backwards compatibility in the existing drivers now just comes from #ifdef, and newer drivers would be expected to only implement the new API.

I've decided to solve the problem of multiple instances by just having the caller pass in the storage for the console structure as a pointer to a global or static local variable. I think this works just as well and is much simpler than implementing a whole separate memory pool just for consoles.


docs/porting-guide.rst
```diff
@@ -2456,13 +2456,16 @@ Crash Reporting mechanism (in BL31)
...
-print the register dump.
-
-The following functions must be implemented by the platform if it wants crash
-reporting mechanism in BL31. The functions are implemented in assembly so that
-they can be invoked without a C Runtime stack.
+of the CPU to enable quick crash analysis and debugging. By default, the
+definitions in ``plat/common/aarch64/platform\_helpers.S`` will cause the crash
+output to be routed over the normal console infrastructure and get printed on
+the same consoles used for boot and runtime output. ``console_set_scope()`` can
+be used to control whether a console is used for crash output.
```
danh-arm: Instead of "... and get printed on the same consoles used for boot and runtime output.", would be better to say "and printed on consoles registered to output in crash state".

drivers/arm/pl011/aarch64/pl011_console.S
```diff
@@ -13,15 +13,72 @@
...
+	 * *must* be persistent (i.e. not from the stack).
+	 * In: x0 - pointer to empty console_pl011_t struct
+	 *     x1 - UART register base address
+	 *     w2 - UART clock in Hz
+	 *     w3 - Baud rate
+	 * Out: return 1 on success, 0 on error
+	 * Clobber list : x0, x1, x2, x6, x7, x14
+	 * -----------------------------------------------
+	 */
+func console_pl011_register
```
danh-arm: Again, I suspect this will collapse into a very simple function if implemented in C.

include/drivers/arm/pl011.h
```diff
@@ -79,4 +81,20 @@
...
+
+#include <types.h>
+
+typedef struct {
+	console_t console;
+	uintptr_t base;
+} console_pl011_t;
+
+int console_pl011_register(console_pl011_t *console, uintptr_t baseaddr,
+			   uint32_t clock, uint32_t baud);
```
danh-arm: I think we also need to export console_pl011_init(), or perhaps this can be called console_pl011_early_init()?

drivers/arm/pl011/aarch64/pl011_console.S
```diff
@@ -13,15 +13,72 @@
...
+	mov	x7, x30
+	cbz	x0, register_fail
+	str	x1, [x0, #CONSOLE_T_PL011_BASE]
+
+	/* shift registers around to match ABI of legacy init function */
```
danh-arm: Although we're deprecating the other console_core_xxx functions, I think we'll need to keep the specific driver init function indefinitely. So it's not legacy.

include/drivers/arm/pl011.h
```diff
@@ -79,4 +81,20 @@
...
+#define CONSOLE_T_PL011_BASE	CONSOLE_T_DRVDATA
+
+#ifndef __ASSEMBLY__
+
+#include <types.h>
+
+typedef struct {
+	console_t console;
+	uintptr_t base;
+} console_pl011_t;
```
danh-arm: Won't need this if base moves into console_t.

drivers/arm/pl011/aarch64/pl011_console.S
```diff
@@ -13,15 +13,72 @@
...
+	add	x1, x1, :lo12:console_pl011_putc
+	str	x1, [x0, #CONSOLE_T_PUTC]
+	adrp	x1, console_pl011_getc
+	add	x1, x1, :lo12:console_pl011_getc
+	str	x1, [x0, #CONSOLE_T_GETC]
+	adrp	x1, console_pl011_flush
+	add	x1, x1, :lo12:console_pl011_flush
+	str	x1, [x0, #CONSOLE_T_FLUSH]
+	mov	x1, #(1 << CONSOLE_FLAG_BOOT_SHIFT | \
+		      1 << CONSOLE_FLAG_CRASH_SHIFT)
```
danh-arm: We need the caller to be able to specify the console flags instead of hardcoding this here. Same for the other drivers

include/drivers/console.h
```diff
@@ -7,14 +7,59 @@
...
+#else
+#define REGSZ U(8)
+#endif
+
+#define CONSOLE_T_NEXT			(U(0) * REGSZ)
+#define CONSOLE_T_FLAGS			(U(1) * REGSZ)
+#define CONSOLE_T_PUTC			(U(2) * REGSZ)
+#define CONSOLE_T_GETC			(U(3) * REGSZ)
+#define CONSOLE_T_FLUSH			(U(4) * REGSZ)
+#define CONSOLE_T_DRVDATA		(U(5) * REGSZ)
```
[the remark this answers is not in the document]
danh-arm: OK, I'll buy that. However, it's crucial these values are kept in sync with the console_t field offsets. It would be good to use `CASSERT` and `__builtin_offset_of` in C context to ensure this. ep_info.h has an example of this.

include/drivers/console.h
```diff
@@ -7,14 +7,59 @@
...
 
-#include <stdint.h>
+#include <utils_def.h>
 
-int console_init(uintptr_t base_addr,
-		unsigned int uart_clk, unsigned int baud_rate);
+#ifdef AARCH32
+#define REGSZ U(4)
+#else
+#define REGSZ U(8)
```
danh-arm: minor: I hoped stdlib would have something you can use here but I can't see anything quite right. Perhaps this could be moved to utils_def.h (and called REG_SIZE)?

include/drivers/console.h
```diff
@@ -7,14 +7,59 @@
...
+#ifndef __ASSEMBLY__
+
+#include <types.h>
+
+typedef struct console {
+	struct console *next;
+	u_register_t flags;
+	void (*putc)(int character, uintptr_t baseaddr);
+	int (*getc)(uintptr_t baseaddr);
+	int (*flush)(uintptr_t baseaddr);
```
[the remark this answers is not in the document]
danh-arm: OK.

include/drivers/console.h
```diff
@@ -7,14 +7,59 @@
...
+} console_t;
+
+/* Remove a single console_t instance from the console list. */
+int console_unregister(console_t *console);
+/* Clear the console list (remove all console instances). */
 void console_uninit(void);
+/* Switch to a new global console state (1 << CONSOLE_FLAG_XXX_SHIFT). */
+void console_switch_state(uint32_t new_state);
+/* Set scope mask of a console that determines in what states it is active. */
+void console_set_scope(console_t *console, uint8_t scope);
```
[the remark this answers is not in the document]
danh-arm: I think that would be a bit clearer

drivers/console/aarch64/skeleton_console.S
```diff
@@ -6,97 +6,179 @@
...
-	cbz	x0, core_init_fail
-	/* Check baud rate and uart clock for sanity */
-	cbz	w1, core_init_fail
-	cbz	w2, core_init_fail
-	/* Insert implementation here */
-	mov	w0, #1
-	ret
-core_init_fail:
-	mov	w0, wzr
+func console_xxx_register
```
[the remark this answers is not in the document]
jwerner-chromium: > Perhaps another helper macro could be provided to declare this.

Yeah, good idea... I'll do that and sketch out how to use it in the porting guide.

drivers/console/aarch64/skeleton_console.S
```diff
@@ -6,97 +6,179 @@
...
 	 * Function to get a character from the console.
-	 * It returns the character grabbed on success
-	 * or -1 on error.
-	 * In : x0 - console base address
+	 * It returns the character grabbed on success,
+	 * -1 if no character available, or any value
+	 * between -2 and -127 otherwise
+	 * In : x0 - pointer to console_t struct
+	 * Out: w0 - character on success, -1 if no char
+	             available, < -1 on error
```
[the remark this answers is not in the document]
jwerner-chromium: > So you're introducing a policy that drivers should now not block but return success if they have a char ready in their FIFO, so that the char to return is from the first "ready" driver. I don't see any real need for this change. Drivers could continue blocking until they have a char (or a real error).

But... I don't see how this could work when you have more than one console. Let's say you're receiving characters on the console that comes second in the list... console_getc() would call getc() from the console that's first in the list, and then that one would just block forever (waiting for characters that will never come) while your input on the second console never gets handled. I think the only way you can really make a blocking getc() work across multiple consoles is to make the drivers non-blocking and keep looping over them like I have it here.

> Also this relies on all drivers being updated to the new policy for it to work properly (which isn't yet done).

My patches to the individual console drivers make the getc() implementations non-blocking so they work with the new API.

include/drivers/console.h
jwerner-chromium: Yeah, there are only a few independent implementations of this scattered around, like in crash_reporting.S. I can consolidate them into utils_def.h.

Makefile
```diff
@@ -451,6 +451,7 @@ $(eval $(call assert_boolean,ERROR_DEPRECATED))
 $(eval $(call assert_boolean,GENERATE_COT))
 $(eval $(call assert_boolean,HW_ASSISTED_COHERENCY))
 $(eval $(call assert_boolean,LOAD_IMAGE_V2))
+$(eval $(call assert_boolean,MULTI_CONSOLE_API))
```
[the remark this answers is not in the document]
danh-arm: Sorry, what I meant to say was the new API should be enabled non-configurably. So all the code under `MULTI_CONSOLE_API = true` should always be compiled in, and the other under `MULTI_CONSOLE_API = false` should only be compiled in when `ERROR_DEPRECATED = false`.

For example, in console.S, instead of:

```
#if MULTI_CONSOLE_API
#include "multi_console.S"
#else
#include "deprecated_console.S"
#endif
```

Have this:

```
// put all content of multi_console.S inline here
#if !ERROR_DEPRECATED
#include "deprecated_console.S"
#endif
```

Then the deprecated APIs in deprecated_console.S should be implemented using the new APIs instead. Similarly in console drivers that support the deprecated APIs, the deprecated APIs should be implemented using the new APIs (currently it's the other way around).

drivers/arm/pl011/aarch64/pl011_console.S
danh-arm: I agree it's better to share code but think it's better if the old API is implemented using the new one, since that will make it easier to remove the old one later.

drivers/arm/pl011/aarch64/pl011_console.S
danh-arm: Using the new API in the Arm platform case, we would want to:

* Register one serial console as the boot/crash console and a 2nd serial console as the runtime console.
* Override bl31_plat_runtime_setup() to change the crash console from the 1st serial console to the 2nd (as well as switching to the runtime state).

So I don't think the default setting should be a property of the driver. It would be less confusing from an API perspective if the default were the same for all drivers, or the setting always had to be provided by the client (so that there is no default).

drivers/console/aarch64/multi_console.S
```diff
@@ -0,0 +1,270 @@
...
+	tst	w0, w1
+	b.eq	getc_continue
+	ldr	x1, [x14, #CONSOLE_T_GETC]
+	cbz	x1, getc_continue
+	mov	x0, x14
+	blr	x1
+	cmp	w0, #0				/* if X0 >= 0: return */
+	b.ge	getc_found
+	cmp	w13, #-1			/* update W13 unless it's -1 */
+	csel	w13, w13, w0, eq		/* (ret = -1 has precedence) */
```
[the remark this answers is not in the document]
danh-arm: I hadn't noticed this console_getc() can loop forever. I was suggesting just going through the list once, returning a permanent error if all the console have returned an error (of any sort). Drivers can continue to block until they receive a char (or return a permanent error).

I don't doubt that your alternative implementation is a fairer way of getting a char in the presence of multiple consoles. My point is that this is not currently a requirement and the code is untested, so why bother complicating this PR? It's already complicated enough.

drivers/console/aarch64/skeleton_console.S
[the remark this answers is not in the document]
danh-arm: > console_getc() would call getc() from the console that's first in the list, and then that one would just block forever (waiting for characters that will never come) while your input on the second console never gets handled.

Yes, that would have been acceptable for now. However, given that you've already made this change, and it looks like this is backwards compatible for existing drivers (i.e. they will still work if they are not changed to be non-blocking), you can leave it as it is now.

> My patches to the individual console drivers make the getc() implementations non-blocking so they work with the new API.

Sorry, I missed that.

drivers/console/aarch64/multi_console.S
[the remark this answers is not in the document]
danh-arm: Later: On reflection, I'll accept this new implementation, given that the new semantics are a better design, there isn't that much new code, and the old implementation isn't tested either. However, the comments could document the intended behaviour better.

drivers/arm/pl011/aarch64/pl011_console.S
```diff
@@ -104,41 +164,46 @@ func console_core_putc
...
+	 * or -1 if no character is available.
+	 * In : x0 - pointer to console_t struct
+	 * Out: w0 - character if available, else -1
 	 * Clobber list : x0, x1
 	 * ---------------------------------------------
 	 */
-func console_core_getc
-	cbz	x0, getc_error
-1:
+func console_pl011_getc
```
danh-arm: Seems like we should still be asserting that x0 is not NULL here.

drivers/arm/pl011/aarch64/pl011_console.S
```diff
@@ -104,41 +164,46 @@ func console_core_putc
...
+#if MULTI_CONSOLE_API
+	ldr	x0, [x0, #CONSOLE_T_PL011_BASE]
```
danh-arm: assert here too (and in other drivers)

drivers/arm/pl011/aarch64/pl011_console.S
```diff
@@ -104,41 +164,46 @@ func console_core_putc
...
+#endif
 	/* Check if the receive FIFO is empty */
 	ldr	w1, [x0, #UARTFR]
-	tbnz	w1, #PL011_UARTFR_RXFE_BIT, 1b
+	tbnz	w1, #PL011_UARTFR_RXFE_BIT, no_char
```
danh-arm: minor: relative branches are preferred. So `1f` here.

drivers/arm/pl011/aarch64/pl011_console.S
jwerner-chromium: Will add.

drivers/arm/pl011/aarch64/pl011_console.S
jwerner-chromium: Okay. I wasn't quite clear on which style you prefer... from existing code, I got the impression that you mostly use named labels (except when it's part of a macro).

I feel that named labels can help add some clarity to the code without needing even more explicit comments to describe what a particular check or code section is for everywhere. Doesn't matter much here, but it's more important in more complicated functions like the ones in multi_console.S. Do you want me to change them all to relative labels?

drivers/console/aarch64/multi_console.S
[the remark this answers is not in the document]
jwerner-chromium: Okay, will try to document the design more explicitly.

drivers/console/aarch64/skeleton_console.S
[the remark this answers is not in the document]
jwerner-chromium: > it looks like this is backwards compatible for existing drivers

Existing drivers (e.g. in some non-upstream platform directory) won't be compatible with MULTI_CONSOLE_API at all anyway -- neither for getc, nor for putc. They're supposed to keep using the legacy API until the platform as a whole is updated to use the new API... and when that happens, the driver should also be updated to do a non-blocking getc. (I mean, yeah, it will still kinda work without that but the idea is that non-blocking is the correct way to implement the new API. Should be a simple change for most drivers anyway.)

Makefile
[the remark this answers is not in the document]
danh-arm: Oh yes, I remember now. If we can't make the old APIs co-exist easily, then I agree that keeping this under a separate macro is better than using ERROR_DEPRECATED directly. At the point we remove deprecated APIs (i.e. a future v2 release), I expect we can remove MULTI_CONSOLE_API too and just keep the new code.

So sorry for the confusion and ignore my previous comments here.

drivers/arm/pl011/aarch64/pl011_console.S
danh-arm: I don't feel strongly which style to use in functions; it just seemed strange to change from using a relative label. It's more important to use relative labels in macros, where it's easy to get label collisions.

So it's optional whether you fix this or not.

drivers/console/aarch64/skeleton_console.S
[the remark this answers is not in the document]
danh-arm: Yes, agree.

drivers/console/aarch64/multi_console.S
```diff
@@ -0,0 +1,270 @@
...
+	 * Returns 1 in x0 for convenient use in
+	 * plat_crash_console_init.
+	 * In : x0
+	 * Out: x0 - always 1
+	 * Clobber list: x0, x1
+	 * -----------------------------------------------
+	 */
+func console_switch_state
+	adrp	x1, console_state
+	strb	w0, [x1, :lo12:console_state]
```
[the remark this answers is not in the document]
jwerner-chromium: When implementing this I noticed that it's not really necessary. The value is already implicitly truncated to 8 bits by the strb which is what I picked for CONSOLE_FLAG_SCOPE_MASK (there are only 3 scopes right now, but I wanted to leave room for expansion and clearly mark which bits are for scopes and which could be used for other generic console classification if needed in the future).

I could count the bits but I think it might be useful to leave open the option of doing console_switch_state(0) (effectively turn off all console output for a while) or enable multiple scopes at the same time.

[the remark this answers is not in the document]
danh-arm: You could assert that bits 7:3 are 0 for now and remove/change the assert if new scopes are added later, but I don't feel that strongly about  it. I agree that use of strb at least ensures "non-scope" bits are dropped. I also agree it might be useful to allow turning off all console output.

danh-arm:
Hi @jwerner-chromium 

We had a wider discussion in Arm about this PR. While we support most of the changes, there was a concern around the crash console behaviour. The original motivation for the `plat_crash_console_xxx()` functions was that they should not access BL memory, in case this memory was corrupted or otherwise inaccessible for some reason at crash time (e.g. incorrect mapping). In the case of BL1, it's not possible to allocate anything into the data section until this has been copied from ROM to RAM, so with your changes it becomes impossible to do crash printing before this time.

This is what the `console_core_xxx()' functions were supposed to facilitate, as all the data needed by the driver is provided in the arguments. We'd like to retain this support. Obviously we can't retain these function names in a MULTI_CONSOLE_API environment (due to name clashes), but we could keep a renamed version of these functions for each driver, e.g. `console_<driver>_core_xxx()`. Then your new `console<driver>_xxx()` functions could look up the required arguments in the `console_<driver>_t` pointer and call `console_<driver>_core_xxx()`. This would give more flexibility for platforms than the current conditional behaviour in your `console<driver>_xxx()` functions.

Given the intent behind the crash console behaviour, we think the "crash" state (CONSOLE_FLAG_CRASH_SHIFT) should be removed from the generic API so there are only "boot" and "runtime" states. The old plat/common `plat_crash_console_xxx()` implementations could be retained. Platforms that want to do crash printing on multiple consoles can sequentially invoke `console_<driver>_core_xxx()` for each crash console (rather than iterating the linked list).

I know in coreboot platforms you are more interested in having driver init in a single place, using arguments from a previous boot stage, than having very early crash support. In those platforms you could override the `plat_crash_console_xxx()` functions (except `plat_crash_console_init()`) to call your new `console_<driver>_xxx()` functions instead. Then crash printing would only work on previously initialized consoles (assuming BL memory is valid).

I hope that makes sense

Regards

Dan.

jwerner-chromium:
Hi @danh-arm,

I wasn't aware of the ROM-only use case, I agree that makes it more tricky. I think I can rewrite it so the existing drivers export init, putc and flush functions similar to the old API as you suggest. The callbacks for the new API would then just be small wrappers that tail-call to them.

However, I really would like to keep the concept of the crash state and the integration with the main console API. I understand that there are crash scenarios where this won't work, but I also think that it provides a lot of value for the majority of cases. Not all platforms may really run into crashes in those early stages that make it worth setting up hardcoded crash console support (some of them don't even use BL1), and for the rest I think it's really nice to have crash dumps "just work" in the majority of cases when they haven't yet implemented any special behavior for it. It also allows crash consoles initialized with runtime-decided parameters and things like that which the hardcoded model can't really do.

I don't see a problem with just allowing both models to live side by side. Platforms are still free to override the crash console functions however they want, I'm just providing default behavior that will do the right thing in the majority of cases. If you're worried about it accessing invalid addresses, I can guard it with #ifndef IMAGE_BL1.

So as long as I provide the core init/putc/flush functions to allow platforms to implement hardcoded crash handling where they want it, can I keep the new default behavior?

Thanks,
Julius

danh-arm:
Hi @jwerner-chromium 

Sorry I missed your last comment on this. On reflection, we're happy with what you propose since that should allow all platforms to do what they want. It would be good to at least document the limitations of the default crash behaviour in platform_helpers.S (i.e. that the console must have been registered previously so this won't work for early crashes).

Regards

Dan.

jwerner-chromium:
Superseded by PR #1193.

PR #1104, opened 2017-09-20
Makefile: Add ability to build dtb (v2)

nmenon:
This is a revamp of the original approach in:
https://github.com/ARM-software/arm-trusted-firmware/pull/747

Current build system has no means to automatically generate dtbs from
dts, instead, stores the dtbs in the fdts/ folder. While this makes
perfect sense for many reference platforms, this becomes a minor
breakage in development flow for newer platforms.

However, this can be solved by providing a rule for the dtbs while
building the ATF binaries by purely describing which dts sources we
need.

For example, with this change, we will now be able to describe the
dtbs we need for the platform in the corresponding platform.mk file:
FDT_SOURCES += fdts/abc.dts

This should be able to generate the abc.dtb appropriately.

Since device trees are specification of hardware, we don't tie the rule
to any specific BL, instead a generic rule is introduced.

Further, this approach allows us to generate appropriate dtbs which may be
need to be regenerated when a common dtsi gets updated, by just
restricting changes to the dtsi alone, instead of synchronizing all the
dtbs as well.

If dtc is not available in default paths, but is available in an
alternate location, it can be chosen by overriding the DTC variable
such as 'make DTC=~/dtc/dtc ....`

NOTE: dtbs are built only with the explicit make dtbs command. The rule
is only available if the platform defines a FDT_SOURCES variable.

Signed-off-by: Benjamin Fair <b-fair@ti.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -125,6 +125,7 @@ OC			:=	${CROSS_COMPILE}objcopy
 OD			:=	${CROSS_COMPILE}objdump
 NM			:=	${CROSS_COMPILE}nm
 PP			:=	${CROSS_COMPILE}gcc -E
+DTC			?=	dtc
 
 ifeq ($(notdir $(CC)),armclang)
 TF_CFLAGS_aarch32	=	-target arm-arm-none-eabi -march=armv8-a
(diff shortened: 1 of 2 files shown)


danh-arm:
This subject of secure firmware configuration in general, and device tree in particular, is going to be discussed at the Linaro Connect conference next week in this session:
http://connect.linaro.org/resource/sfo17/sfo17-310/

I'll reserve my comments to this PR until after that time. What I will say now is this I expect something similar to this PR to be OK, although I expect it will need to be made slightly more generic to allow other file formats to be supported.

danh-arm:
jenkins: test this please

davidcunado-arm:
@nmenon 
Apologies for the delay in dealing with this PR - we'll look to merge now.

nmenon:
@davidcunado-arm  much appreciated. thanks. one patch off my upstream list.

PR #1106, opened 2017-09-21
Fix type of `unsigned long` constants

antonio-nino-diaz-arm:
The type `unsigned long` is 32 bit wide in AArch32, but 64 bit wide in AArch64. This is inconsistent and that's why we avoid using it as per the Coding Guidelines. This patch changes all `UL` occurrences to `U` or `ULL` depending on the context so that the size of the constant is clear.

This problem affected the macro `BIT(nr)`. As long as this macro is used to fill fields of registers, that's not a problem, since all registers are 32 bit wide in AArch32 and 64 bit wide in AArch64. However, if the macro is used to fill the fields of a 64-bit integer, it won't be able to set the upper 32 bits in AArch32.

By changing the type of this macro to `unsigned long long` the behaviour is always the same regardless of the architecture, as this type is 64-bit wide in both cases.

Some Tegra platform files have been modified by this patch.

### DIFF
--- a/drivers/arm/ccn/ccn.c
+++ b/drivers/arm/ccn/ccn.c
@@ -236,7 +236,7 @@ static unsigned long long ccn_master_to_rn_id_map(unsigned long long master_map)
 		node_id = ccn_plat_desc->master_to_rn_id_map[iface_id];
 
 		/* Set the bit corresponding to this node ID */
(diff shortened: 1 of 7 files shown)


antonio-nino-diaz-arm:
Could you take a look at this, @vwadekar ?

vwadekar:
Looks good

PR #1110, opened 2017-09-26
Fix MAP_REGION for GCC 4.9

masahir0y:


davidcunado-arm:
jenkins: test this please

PR #1112, opened 2017-09-28
Allow lower ELs to use and access SVE (fixes #520)

stsquad:
Currently booting an SVE enabled kernel through the trusted path ends
up crashing the kernel. To fix I've added a config flag called
ENABLE_SVE_FOR_LOWER_ELS which defaults to 1 and allows the kernel to
access SVE registers. If the trusted firmware ever wants to use SVE
itself additional changes will need to be made.

The changes are based on a similar patch added to the bootwrapper.

Signed-off-by: Alex Bennée <alex.bennee@linaro.org>
Cc: Dave Martin <Dave.Martin@arm.com>
Cc: Mark Rutland <mark.rutland@arm.com>


danh-arm:
Thanks for this. We acknowledge the issue needs addressing but your solution isn't quite acceptable for upstream. The problem is if SVE-enabled hardware contains an existing Trusted OS, or any other Secure Payload using SIMD&FP (or SVE) registers, any normal world SVE payload is likely to fail in an unpredictable way.

We intended to address this soon, by altering the secure world context initialization to disable SVE support *and* SIMD&FP support while executing in the secure world, but only when the SVE extensions are detected in ID registers at runtime.

You can either alter your solution to along these lines, or we (Arm) will do this over the next month or so, possibly based on your solution.

Regards

Dan.

stsquad:
@danh-arm: I'm happy to leave this one to you guys as the domain experts. I'll carry this solution locally as I'm not using a secure world and just wanted to test Dave's SVE enabled kernel but with an unified UEFI based boot process.

PR #1113, opened 2017-09-28
Introduce BOOL macro

sandrine-bailleux-arm:
Introduce the BOOL macro to convert any integer type to an essentially
boolean value. This allows to write boolean expressions using integers
where the conversion between integer and boolean is explicit. For
example,

instead of writing: !!(x)   ^  !!(y)
you would write   : BOOL(x) != BOOL(y)


davidcunado-arm:
Closing - upon further reflection, it was decided that this support is not needed at this time.

PR #1047, opened 2017-08-02
GICv3 context save and restore

douglas-raillard-arm:
Introduce helpers to save and restore the context of GICv3:
* Distributor
* Redistributor
* For Distributor and Redistributor, the helpers also support the implementation-defined part of GIC-500 and GIC-600
* ITS (only the architectural part, most of the work needs to be done according to the ITS implementation of the platform)

PR #1116, opened 2017-10-04
GICv3 context save and restore

soby-mathew:
Introduce helpers to save and restore the context of GICv3:

Distributor
Redistributor
For Distributor and Redistributor, the helpers also support the implementation-defined part of GIC-500 and GIC-600
ITS (only the architectural part, most of the work needs to be done according to the ITS implementation of the platform)

### DIFF
--- a/docs/porting-guide.rst
+++ b/docs/porting-guide.rst
@@ -2107,12 +2107,31 @@ power domain levels. The handler needs to perform power management operation
 corresponding to the local state at each power level. The generic code
 expects the handler to succeed.
 
-The difference between turning a power domain off versus suspending it
-is that in the former case, the power domain is expected to re-initialize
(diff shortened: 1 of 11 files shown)


soby-mathew:
Supercedes PR https://github.com/ARM-software/arm-trusted-firmware/pull/1047

soby-mathew:
Updated the ITS helpers to leave the ITS disabled on resume from system suspend.

dbasehore:
Is there any update on what the kernel team wants for re-init of the ITS collections in the kernel on resume?

soby-mathew:
@dbasehore, we don't deal with the ITS restore issue anymore because , as it stands, this issue is something that needs to be solved within the kernel and we in the firmware don't have any say in it. I will try pinging them again, but suggest quering the ARM support or kernel team again.

PR #1124, opened 2017-10-12
Allow to override toolchain

paulmenzel:


dp-arm:
You can still do make CC=clang and it will work.  I agree however that CC=clang make should also work.

paulmenzel:
Indeed, the user guide says `export CC=…` though.

davidcunado-arm:
jenkins: test this please

davidcunado-arm:
jenkins: retest this please

dp-arm:
> Indeed, the user guide says export CC=… though.

The guide mentions export CROSS_COMPILE though unless I am totally missing something.

davidcunado-arm:
@paulmenzel 
Unfortunately, we're not going to be able to accept this change in its current state. 

The build scripts have been written with this restrictiveness, which is relied upon by our CI environment. 

With this change the checks failures because the build machine environments have CC defined by default to the native compiler and so all the CI builds fail.

davidcunado-arm:
@paulmenzel 
Closing as per earlier comment.

PR #1129, opened 2017-10-16
Fix use of MSR (immediate)

robertovargas-arm:
The macro DEFINE_SYSREG_WRITE_CONST_FUNC defines an inline function
to an assembly statement that uses the MSR (immediate) instruction
to access the PSTATE.  The "i" (immediate) assembly constraint on
the operand was only satisfied when compiling with optimizations
enabled which resulted in the function being optimized out - the
"const uint64_t v" parameter was optimized out and replaced by a
literal value.

When compiling without optimizations, the function call remained and
therefore the parameter is not optimized out - compilation fails as
the constraint is impossible to satisfy by the compiler.

This patch replaces the function encapsulating the use of
the MSR (immediate) with a macro that allows the literal value to be
directly fed to the inline assembly statement

Change-Id: Ib379a7acc48ef3cb83090a680cd8a6ce1a94a9d9

### DIFF
--- a/include/lib/aarch64/arch_helpers.h
+++ b/include/lib/aarch64/arch_helpers.h
@@ -31,11 +31,8 @@ static inline void write_ ## _name(uint64_t v)				\
 	__asm__ volatile ("msr " #_reg_name ", %0" : : "r" (v));	\
 }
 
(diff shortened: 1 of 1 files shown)


davidcunado-arm:
jenkins: test this please

PR #1132, opened 2017-10-16
Publish and Subscribe framework

jeenu-arm:
This light-weight framework allows for calling into EL3 components without having explicit dependency. This works by defining events, registering subscribers, and publishing those events at required code path.


davidcunado-arm:
Note that the checks are failing because checkpatch.pl complains about new, multi-line macro REGISTER_PUBSUB_EVENT not being enclosed in a do-while loop - this is not a valid solution for this macro as the macro is not used in C code.

PR #1136, opened 2017-10-17
Add APIs to get and modify attributes of memory regions

antonio-nino-diaz-arm:
Also:
- Small fix to AArch32 port of the library.
- Add functions to disable MMU in EL1.

For more details, check the individual commits.


davidcunado-arm:
jenkins: test this please

PR #1142, opened 2017-10-24
qemu: update deprecated interrupt registering

etienne-lms:
Registered interrupts are configured in edge detection as the default
previous configuration assumed in previous code.

Signed-off-by: Etienne Carriere <etienne.carriere@linaro.org>

### DIFF
--- a/plat/qemu/qemu_bl31_setup.c
+++ b/plat/qemu/qemu_bl31_setup.c
@@ -7,6 +7,7 @@
 #include <assert.h>
 #include <bl_common.h>
 #include <console.h>
+#include <gic_common.h>
 #include <gicv2.h>
 #include <platform_def.h>
 #include "qemu_private.h"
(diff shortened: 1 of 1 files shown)


plat/qemu/qemu_bl31_setup.c
```diff
@@ -112,22 +113,44 @@ void bl31_plat_arch_setup(void)
...
+static unsigned int target_mask_array[PLATFORM_CORE_COUNT];
+
 static const struct gicv2_driver_data plat_gicv2_driver_data = {
 	.gicd_base = GICD_BASE,
 	.gicc_base = GICC_BASE,
-	.g0_interrupt_num = ARRAY_SIZE(irq_sec_array),
-	.g0_interrupt_array = irq_sec_array,
+	.interrupt_props = qemu_interrupt_props,
+	.interrupt_props_num = ARRAY_SIZE(qemu_interrupt_props),
+	.target_masks = target_mask_array,
```
jeenu-arm: If you intend to send/receive SGIs on the platform, you should also [setup the per-PE target mask for each PE powered up](https://github.com/ARM-software/arm-trusted-firmware/commit/dea417cdd6#diff-8befd0f5eeb9363dc44437633ad46530R79).

soby-mathew: To clarify, @jeenu-arm is referring to the case if you intend to handle/send SGI in EL3. If the handling/sending of Secure SGI is within S-EL1, then the `target_mask_array` is not required for BL31 and the per-PE target mask does not need to be known within EL3.

etienne-lms:
Indeed, the qemu bl31 is not expected to send/route SGIs.
I will update the patch discarding `target_masks` info.

davidcunado-arm:
@etienne-lms 
Apologies for the delay! Could you squash these 2 commits together? I'll then merge...

davidcunado-arm:
jenkins: test this please

PR #1144, opened 2017-10-25
Change the default option of ARM_TSP_RAM_LOCATION and Enlarge the BL2 size on ARM platforms

geesun:
1. Change the default option of ARM_TSP_RAM_LOCATION
    On Arm standard platforms, it runs out of SRAM space when TBB is enabled, so the TSP default location is changed to dram when TBB is enabled.

2. Enlarge the BL2 size on Arm platforms when TBB is enabled
    For Trusted Board Boot, BL2 needs more space to support the ECDSA and ECDSA+RSA algorithms.


davidcunado-arm:
jenkins: test this please

PR #1145, opened 2017-10-25
Support ARMv7 architectures

etienne-lms:
This P-R proposes to add support for ARMv7-A architectures in the ARM trusted firmware.
This P-R is an update of https://github.com/ARM-software/arm-trusted-firmware/pull/1057.

Obviously, only ARMv7 with the Trustzone extensions can be supported by ARM trusted firmware. The proposed changes aims at covering most such ARMv7 cores. However these changes first cover Cortex-A9 and Cortex-A15 based systems since former P-Rs will propose 2 ARMv7 platforms: namely the b2260 board (Cortex-A9) and the Qemu/ARMv7 (Cortex-A15). Each will come with support for sp_min as BL32 or OP-TEE as BL32. These requires few added changes:
- Changes to enable OP-TEE boot on ARM-TF/ARMv7 and add the Qemu/ARMv7 platform.
(see branch armv7-qemu in https://github.com/etienne-lms/arm-trusted-firmware).
- Changes to support platform b2260 (with some Cortex-A9 specificities: pl310, non-lpae).
(see branch armv7-b2260 in https://github.com/etienne-lms/arm-trusted-firmware).


Makefile
```diff
@@ -260,6 +268,48 @@ include lib/stack_protector/stack_protector.mk
...
+# GCC and Clang require -march=armv7-a for C-A9 and -march=armv7ve for C-A15.
+# armClang requires -march=armv7-a for all ARMv7 Cortex-A. To comply with
+# all, just drop -march and supply only -mcpu.
+#
+# Platform can override march32-directive.
+march32-set-${ARM_CORTEX_A5}	:= -mcpu=cortex-a5
+march32-set-${ARM_CORTEX_A7}	:= -mcpu=cortex-a7
+march32-set-${ARM_CORTEX_A9}	:= -mcpu=cortex-a9
+march32-set-${ARM_CORTEX_A15}	:= -mcpu=cortex-a15
+march32-neon-$(ARM_WITH_NEON)	:= -mfpu=neon
```
mmind: What about Cortex-A12/A17 [both called A17 nowadays]? I don't think there should be much difference to the others, but maybe we still want to have separate options?

etienne-lms: Yes, I guess A12/A17 could be supported too.
I will add a change for that.

etienne-lms:
Appended few commits to add cortex-a5/7/12/17 support basics as suggested.

include/lib/aarch32/arch.h
```diff
@@ -431,6 +438,14 @@
 #define PMCR		p15, 0, c9, c12, 0
 #define CNTHP_CTL	p15, 4, c14, c2, 1
 
+/* AArch32 coproc registers for 32bit MMU descriptor support */
+#define PRRR		p15, 0, c10, c2, 0
+#define NMRR		p15, 0, c10, c2, 1
+#define DACR		p15, 0, c3, c0, 0
+
+/* ARMv7 Cortex-A9 specific coproc register */
+#define PCR		p15, 0, c15, c0, 0
```
danh-arm: This should be in cortex_a9.h instead

include/lib/aarch32/arch_helpers.h
```diff
@@ -273,6 +274,16 @@ DEFINE_COPROCR_RW_FUNCS(hdcr, HDCR)
...
 
+DEFINE_COPROCR_RW_FUNCS(nsacr, NSACR)
+
+/* AArch32 coproc registers for 32bit MMU descriptor support */
+DEFINE_COPROCR_RW_FUNCS(prrr, PRRR)
+DEFINE_COPROCR_RW_FUNCS(nmrr, NMRR)
+DEFINE_COPROCR_RW_FUNCS(dacr, DACR)
+
+/* ARMv7 Cortex-A9 specific coproc register */
+DEFINE_COPROCR_RW_FUNCS(pcr, PCR)
```
danh-arm: Move to cortex_a9.h

include/lib/aarch32/arch.h
```diff
@@ -513,4 +528,9 @@
 
 #define MAKE_MAIR_NORMAL_MEMORY(inner, outer)	((inner) | ((outer) << MAIR_NORM_OUTER_SHIFT))
 
+/*******************************************************************************
+ * ACTLR bit fields
+ ******************************************************************************/
+#define ACTLR_FLZW_BIT		(1 << 3)
```
danh-arm: This register is IMP DEF so move this to relevant cortex_aX.h files

docs/firmware-design.rst
```diff
@@ -2433,6 +2433,29 @@ This Architecture Extension is targeted when ``ARM_ARCH_MAJOR`` == 8 and
...
+extension is mandatory to support the ARM Trusted Firmware bootloader
+and runtime services.
+
+Platform implementing an ARMv7 system can to define from its target
+Cortex-A architecture through ``ARMV7_CORTEX_A<X> = yes``. Defined
+targets are:
+* ``ARMV7_CORTEX_A5=yes`` for a Cortex-A5 target.
+* ``ARMV7_CORTEX_A7=yes`` for a Cortex-A7 target.
+* ``ARMV7_CORTEX_A9=yes`` for a Cortex-A9 target.
+* ``ARMV7_CORTEX_A15=yes`` for a Cortex-A15 target.
```
danh-arm: Unless you're going to enumerate all possible CPUs here I suggest you just use one of the above as an example.

common/aarch32/debug.S
```diff
@@ -71,7 +71,15 @@ endfunc report_exception
 assert_msg1:
 	.asciz "ASSERT: File "
 assert_msg2:
+#if !ARCH_SUPPORTS_VIRTUALIZATION
+	/******************************************************************
+	 * Virtualization comes with the UDIV/SDIV instructions. If missing
+	 * write file line number is hexadecimal format.
```
danh-arm: minor: "... in hexadecimal ..."

drivers/arm/gic/v2/gicv2_main.c
```diff
@@ -167,7 +167,19 @@ void gicv2_driver_init(const gicv2_driver_data_t *plat_driver_data)
 	gic_version = gicd_read_pidr2(plat_driver_data->gicd_base);
 	gic_version = (gic_version >> PIDR2_ARCH_REV_SHIFT)
 					& PIDR2_ARCH_REV_MASK;
-	assert(gic_version == ARCH_REV_GICV2);
+
+	/*
+	 * GICv1 with security extension complies with A-TF GICv2 driver
```
danh-arm: minor: Say "Arm TF" or just "TF", not "ATF" or "A-TF"

include/lib/cpus/aarch32/cortex_a15.h
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright (c) 2016, ARM Limited and Contributors. All rights reserved.
```
danh-arm: minor: Can use 2017 for all new files!

bl32/sp_min/aarch32/entrypoint.S
```diff
@@ -147,6 +157,8 @@ func handle_smc
 
 	smcc_save_gp_mode_regs
 
+	clrex_on_monitor_entry
```
soby-mathew: Would be good to use the same macro for the `clrex` in handle_fiq() as well.

include/common/aarch32/asm_macros.S
```diff
@@ -79,6 +79,16 @@
 	ldr r0, =(\_name + \_size)
 	.endm
 
+#if (ARM_ARCH_MAJOR == 7)
+	/* ARMv7 does not support stl instruction */
+	.macro stl _reg, _write_lock
+	dmb
+	str	\_reg, \_write_lock
+	dsb
+	sev
```
soby-mathew: minor: Not sure whether the `sev` is needed.

soby-mathew:
Generally regarding the new ARCH_SUPPORTS_* build macros need to be defined only for ARMv7 builds. The separate makefile for ARMv7 platforms should help in this regard as then they will only be defined for ARMv7.

Within the C code, most of the checks for these defines can be restricted to aarch32 versions of the files if possible (which is the case already except for CASSERT in xlat_tables). 

Also since these defines will only be defined for ARMv7, it is better to use `#ifdef` rather than `#if` for the conditional compilation.

lib/cpus/aarch32/cortex_a5.S
```diff
@@ -0,0 +1,84 @@
...
+
+	assert_cache_enabled
+
+	/* Flush L1 caches */
+	mov	r0, #DC_OP_CISW
+	bl	dcsw_op_level1
+
+	bl	plat_disable_acp
+
+#ifdef PLAT_OUTER_CACHE
```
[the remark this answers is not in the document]
soby-mathew: If `plat_flush_outer_cache` is not specific to this CPU, you could also consider invoking this from within the plat_psci_ops {power_domain_off  & pwr_domain_suspend} hooks. The `cortex_a5_cluster_pwr_dwn` will complete before plat_psci_ops hook is called.

Would prefer it that way unless there is a sequence restriction that `plat_flush_outer_cache` need to be called before the `cortex_a5_disable_smp`.

include/common/aarch32/asm_macros.S
soby-mathew: Edit: When this instruction is used in  spin lock context "In ARMv8, when the global monitor for a PE changes from Exclusive Access state to Open Access state, an event is generated."

Whereas on ARMv7 this is not the case.  `sev` is required for ARMv7 to match the `wfe` in spin_lock().

It would be better to isolate this change to the aarch32 spin_lock.S file similar to the aarch64 `COND_SEV()` implementation.

Makefile
```diff
@@ -532,9 +486,11 @@ $(eval $(call assert_boolean,WARMBOOT_ENABLE_DCACHE_EARLY))
 $(eval $(call assert_numeric,ARM_ARCH_MAJOR))
 $(eval $(call assert_numeric,ARM_ARCH_MINOR))
 
+ifeq ($(ARM_ARCH_MAJOR),7)
```
soby-mathew: The below `assert_boolean` s can be moved to armv7-a-cpus.mk as well.

Makefile
```diff
@@ -573,9 +529,11 @@ $(eval $(call add_define,USE_COHERENT_MEM))
 $(eval $(call add_define,USE_TBBR_DEFS))
 $(eval $(call add_define,WARMBOOT_ENABLE_DCACHE_EARLY))
 
+ifeq ($(ARM_ARCH_MAJOR),7)
```
soby-mathew: The below `add_define` s can be moved to armv7-a-cpu.mk

etienne-lms: ok

Makefile
etienne-lms: ok

etienne-lms:
I squashed the fixup commits and fixed few checkpatch issues.

bl32/sp_min/aarch32/entrypoint.S
```diff
@@ -203,6 +215,8 @@ func handle_fiq
...
+
```
soby-mathew: nit: the clrex at line 224 can be removed.

Makefile
```diff
@@ -340,6 +352,12 @@ ifeq ($(HW_ASSISTED_COHERENCY)-$(USE_COHERENT_MEM),1-1)
 $(error USE_COHERENT_MEM cannot be enabled with HW_ASSISTED_COHERENCY)
 endif
 
+ifeq (${ARM_ARCH_MAJOR},7)
+ifneq (${ARCH},aarch32)
```
soby-mathew: nit: this check can be in armv7-a-cpus.mk

include/lib/aarch32/smcc_macros.S
```diff
@@ -82,6 +121,44 @@
 
 	/* Restore the banked registers including the current SPSR */
 	add	r1, r0, #SMC_CTX_SP_USR
+
+#if ARM_ARCH_MAJOR == 7 && !ARCH_SUPPORTS_VIRTUALIZATION
```
soby-mathew: Please see my comment below regarding this. This can be changed to :
`#if ARM_ARCH_MAJOR == 7 && !defined(ARCH_SUPPORTS_VIRTUALIZATION)`

make_helpers/armv7-a-cpus.mk
```diff
@@ -32,12 +32,17 @@ endif
...
+# ARCH_SUPPORTS_VIRTUALIZATION = 0/1
+# Defines if ARMv7 core supports the Virtualization extension.
 
 define pre-a15-extensions
 ARCH_SUPPORTS_LARGE_PAGE_ADDRESSING	?= 	0
+ARCH_SUPPORTS_VIRTUALIZATION		?= 	0
 endef
 define post-a15-extensions
 ARCH_SUPPORTS_LARGE_PAGE_ADDRESSING	?= 	1
+ARCH_SUPPORTS_VIRTUALIZATION		?= 	1
```
soby-mathew: The use of pre-processor build macros is currently a bit inconsistent in ARM TF (see https://github.com/ARM-software/tf-issues/issues/446). As part of cleaning this up, we would like to move to a model wherein a build macro, if used as a boolean flag, is either defined or not defined. 

Hence it would be good to have all new code follow this semantics and  `$(eval $(call add_define,ARCH_SUPPORTS_VIRTUALIZATION))` should be done here. The code would then change to explicitly check if the macro is defined or not using `defined()` pre-processor directive.

Could you migrate all the new macros as suggested above ?

Also, suggest to rename the macro to indicate these are ARMv7 specific like `ARMV7_VIRTUALIZATION_SUPPORT`

etienne-lms:
Late reply. I appended fix commits to address comments.

etienne-lms:
Squashed and rebased.

davidcunado-arm:
jenkins: test this please

etienne-lms:
Thanks all for this and the time spent.
Next step is booting an op-tee: https://github.com/ARM-software/arm-trusted-firmware/pull/1173.

PR #1149, opened 2017-10-30
Add FWU booting instructions to the user guide

robertovargas-arm:
FWU uses additional images that have to be loaded,
and this patch adds the documentation of how to do
it in FVP and Juno.

Change-Id: I1a40641c11c5a4c8db0aadeaeb2bec30c9279e28

### DIFF
--- a/docs/user-guide.rst
+++ b/docs/user-guide.rst
@@ -1205,6 +1205,56 @@ The resulting BL1 and FIP images may be found in:
     ./build/fvp/release/bl1.bin
     ./build/fvp/release/fip.bin
 
+
+Booting Firmware Update images
+-------------------------------------
+
+When Firmware Update (FWU) is enabled there are at least 2 new images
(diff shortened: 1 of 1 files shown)


docs/user-guide.rst
```diff
@@ -1205,6 +1205,56 @@ The resulting BL1 and FIP images may be found in:
...
+that have to be loaded, the Non-Secure FWU ROM (NS-BL1U), and the
+FWU FIP.
+
+Juno
+~~~~
+
+The new images must be programmed in flash memory by adding
+an entry in the ``SITE1/HBI0262x/images.txt`` configuration file
+on the Juno SD card (where ``x`` depends on the revision of the Juno board).
+Refer to the ``Juno Getting Started Guide`_, section 2.3 "Flash memory
```
sandrine-bailleux-arm: Should remove one back quote, the current text does not display correctly.

PR #1151, opened 2017-10-31
Change sizeof to use type of struct not function

JoelHutton:
Change sizeof call so it references a static type instead of return of
a function in order to be MISRA compliant.

Change-Id: I6f1adb206073d6cd200156e281b8d76249e3af0e
Signed-off-by: Joel Hutton <joel.hutton@arm.com>


davidcunado-arm:
jenkins: add to whitelist

davidcunado-arm:
jenkins: test this please

PR #1153, opened 2017-11-01
Avoid use of undefined macros

robertovargas-arm:
There are several cases where some macros are used in prepocessor conditionals without a previous definition and in some cases it was generating some problems.


davidcunado-arm:
jenkins: test this please

davidcunado-arm:
jenkins: test this please

PR #1154, opened 2017-11-02
Fix PSCI STAT time stamp collection

soby-mathew:
This patch includes various fixes for PSCI STAT functionality
relating to timestamp collection:

1. The PSCI stat accounting for retention states for higher level
power domains were done outside the locks which could lead to
spurious values in some race conditions. This is moved inside
the locks. Also, the call to start the stat accounting was redundant
which is now removed.

2. The timestamp wrap-around case when calculating residency did
not cater for AArch32. This is now fixed.

3. In the warm boot path, `plat_psci_stat_accounting_stop()` was
getting invoked prior to population of target power states. This
is now corrected.

Change-Id: I851526455304fb74ff0a724f4d5318cd89e19589

### DIFF
--- a/lib/psci/psci_common.c
+++ b/lib/psci/psci_common.c
@@ -767,11 +767,11 @@ void psci_warmboot_entrypoint(void)
 	psci_acquire_pwr_domain_locks(end_pwrlvl,
 				      cpu_idx);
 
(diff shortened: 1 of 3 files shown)


davidcunado-arm:
jenkins: test this please

davidcunado-arm:
jenkins: test this please

PR #1155, opened 2017-11-03
Fix build error when creating ROT key for UniPhier platform

masahir0y:


davidcunado-arm:
jenkins: test this please

Makefile
```diff
@@ -208,6 +208,8 @@ include ${MAKE_HELPERS_DIRECTORY}plat_helpers.mk
 BUILD_BASE		:=	./build
 BUILD_PLAT		:=	${BUILD_BASE}/${PLAT}/${BUILD_TYPE}
 
+$(eval $(call MAKE_PREREQ_DIR,${BUILD_PLAT}))
```
sandrine-bailleux-arm: This breaks the Tegra platform port build:
```
make: *** No rule to make target `build/tegra/t210/debug', needed by `build/tegra/t210/debug/bl31'.  Stop.
```

As far as I can see, this is because the Tegra makefile redefines the BUILD_PLAT variable so I think this line needs to be moved after the inclusion of the platform Makefile, i.e. at line 263.

masahir0y: Oh, I did not notice such a naughty Makefile.

I think a more correct way is to revert 1f95e28ce288af52ee4f8544b1456af7fca7e7bd.
I read the commit log, but I still did not get it.
If users can change the output directory, they can specify BUILD_PLAT from the command line.
And it is a commonly-used tip to keep multiple build results.
I see no good reason to change the default from platform Makefile.

Having said that, I do not want to spend time to argue that.
The quickest way to finish this work is to move my eval line.  I will push a new one as you suggested.

davidcunado-arm:
jenkins: test this please

davidcunado-arm:
jenkins: test this please

davidcunado-arm:
jenkins: test this please

PR #1156, opened 2017-11-06
Introduce AArch64 Raspberry Pi 3 port

AntonioND:
This port can be compiled to boot an AArch64 or AArch32 payload with the build option `RPI3_BL33_AARCH32`.

Note: This is not a secure port of the Trusted Firmware. This port is only meant to be a reference implementation to experiment with an inexpensive board in real hardware.

This port is based on the work of Ard Biesheuvel (@ardbiesheuvel), that can be found in http://www.workofard.com/2017/02/uefi-on-the-pi/, see repository in https://github.com/ardbiesheuvel/arm-trusted-firmware/tree/rpi3. That port has been heavily modified. For starters, that port didn't manage to boot Linux or to release the secondary CPUs from their wait loop because the RAM wasn't initialized to 0 before entering the loop.

### DIFF
--- a/docs/plat/rpi3.rst
+++ b/docs/plat/rpi3.rst
@@ -1,0 +1,297 @@
+ARM Trusted Firmware for Raspberry Pi 3
+=======================================
+
+.. section-numbering::
+    :suffix: .
+
+.. contents::
+
+The Raspberry Pi 3 is an inexpensive single-board computer that contains four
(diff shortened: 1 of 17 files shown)


antonio-nino-diaz-arm:
Superseded by https://github.com/ARM-software/arm-trusted-firmware/pull/1157

PR #1158, opened 2017-11-10
SPM: Fix pointer to MP info in boot info struct

antonio-nino-diaz-arm:
The MP info struct is placed right after the boot info struct. However, when calculating the address of the MP info, the size of the boot info struct was being multiplied by the size of the MP boot info. This left a big gap of empty space between the structs.

This didn't break any code because the boot info struct has a pointer to the MP info struct. It was just wasting space.


davidcunado-arm:
Check that failed was an infrastructure issue - re-run of that check passed.

PR #1163, opened 2017-11-20
Add ARMv8.2 ID_AA64MMFR0_EL1.PARange value

antonio-nino-diaz-arm:
If an implementation of ARMv8.2 includes ARMv8.2-LPA, the value 0b0110 is permitted in ID_AA64MMFR0_EL1.PARange, which means that the Physical Address range supported is 52 bits (4 PiB). It is a reserved value otherwise.


davidcunado-arm:
jenkins: test this please

PR #1164, opened 2017-11-20
Flush the affinity data in psci_affinity_info

robertovargas-arm:
There is an edge case where the cache maintaince done in
psci_do_cpu_off may not seen by some cores. This case is handled in
psci_cpu_on_start but it hasn't handled in psci_affinity_info.

Change-Id: I4d64f3d1ca9528e364aea8d04e2d254f201e1702


davidcunado-arm:
jenkins: test this please

PR #1165, opened 2017-11-21
Add support sha512 for hash algorithm

geesun:
add an option -hash-alg for cert_create and add build flag HASH_ALG to let the user to select the SHA

It supports the following options:
- sha256 (default)
- sha384
- sha512


davidcunado-arm:
jenkins: test this please

PR #1167, opened 2017-11-22
Set alignment size to 512B for Hikey/Hikey960

Leo-Yan:
These two patches are to set alignment size to 512B for Hikey/Hikey960, as results we can generate the fip.bin with 512B alignment and ensure it's safe for USB downloading with DMA enabled.


Leo-Yan:
Hi @hzhuang1, please review this patch. Thanks.

hzhuang1:
@Leo-Yan acked

davidcunado-arm:
jenkins: test this please

robertovargas-arm:
I and @geesun are tracking a problem in [https://github.com/ARM-software/tf-issues/issues/528](url), and in some cases it can generate unaligned access and incorrect reads when the size of the read isn't a multiple of the block size. We still have to look if both problems are related, but it is very likely.

Anyway, it seems that the solution in this PR is a workaround and instead of fixing the actual problem is hiding it.

davidcunado-arm:
@Leo-Yan 
I've now merged #1204 into integration - please let me know whether this PR can be merged as is or you're planning on changing the alignment value (as per your comment in #1204).

Leo-Yan:
Hi @davidcunado-arm, I have consolidate with Haojian, we prefer to use 512B as alignment size. So please help merge this PR. Thanks.

davidcunado-arm:
@Leo-Yan 
Ok looking to merge this now that I've merged the fip padding change is now in master.

davidcunado-arm:
jenkins: test this please

PR #1173, opened 2017-11-29
support to boot OP-TEE on AArch32/Armv7+example with Cortex-A15/Qemu

etienne-lms:
These changes follow armv7 support changes (https://github.com/ARM-software/arm-trusted-firmware/pull/1145). They propose some support for booting OP-TEE bl32 images. The Qemu/Cortex-A15 is provided as example. Board b2260 (Cortex-a9) is planned, will be proposed later.


bl1/aarch32/bl1_exceptions.S
```diff
@@ -73,6 +73,12 @@ debug_loop:
 	ldr	r1, [r8, #(ENTRY_POINT_INFO_PC_OFFSET + 4)]
 	msr	spsr, r1
 
+#ifdef ENTRY_POINT_INFO_LR_OFFSET
```
soby-mathew: It seems a bit weird that this needs to be programmed to lr_svc whereas lr_mon is being programmed above.

I think this is somthing that can be abstracted within the platform layer. Perhaps the `bl1_plat_prepare_exit()` can do this selectively if the BL32 is optee or Trusty in AArch32 mode? It doesn't have to be in entry_point_info_t in that case.

etienne-lms: Hi. Sorry for this very late feedback.

`lr_mon` programs the entry point of the next booted image (bl32) when bl1 exits Monitor mode while `lr_svc` is expected by OP-TEE and Trusty (tbc) as the entry point of the non-secure bl33 image. 

When bl1 boots bl32, it needs to set the bl32 entry point AND provide to bl32 the entry point of bl33. In case of bl32=sp_min, the later is available from the structured parameters passed from bl2 to bl1 then to bl32. When bl32 is optee or trusty, these firmware do not handle the structured `bl_params_t`.

soby-mathew: Ok, I see the issue you are trying to solve. I am thinking of an alternative way to do this:

The ep_info.x3 (or any other unused register) of BL32 can be modified to have the entrypoint information of BL33 in BL2. Then in `bl1_plat_prepare_exit` you can read the x3 from ep_info and write it to `lr_svc` . x3 can be made 0 after this if required.

This way we can avoid changing the the entrypoint_info_t structure for AArch32.

etienne-lms: Indeed this would work. I am not fund of this since it will prevent bl32 images from benefiting from a 4th native parameters (carried by register `r3`) since it would be used in trusted firmware for the AArch32 specific "bl33 entry point" boot argument.
Note that `entrypoint_info_t` structure is already different between AArch32 and AArch64.

soby-mathew: Hi @etienne-lms, 
Sorry for the delayed response as I was on holiday.

Hmm,  this would be restricted to Optee in the platform layer and other BL32 images will not be affected (i.e they still would have r0 - r3 available). Ideally Optee should have changed to accept arguments from ARM-TF. Availability of r0-r3 for passing arguments to Optee does not matter because it doesn't consider them anyway. 

We could add more fields to `epinfo` structure but it should be unconditional IMO. But I see this as a Optee specific problem because it doesn't look in `r0-r3` but expects to `lr_svc` to have the address. Hence I would think the solution should be optee specific and within platform code.

etienne-lms: ok, noticed. Thanks. I think this is better as it leaves x0-x3 to any standardization of optee boot arguments. I will update the p-r to remove the condition.

etienne-lms: p-r updated: no real change but renaming the `lr` field into `lr_svc`. As before, this field is included in the entry point description structure only for AArch32 mode.

etienne-lms:
Hello. Any news regarding this change?

davidcunado-arm:
@etienne-lms 
Apologies - @soby-mathew has been looking into this but we've had other internal priorities. Should hget back to it soon...

include/plat/arm/common/aarch32/arm_macros.S
```diff
@@ -0,0 +1,71 @@
...
+
+	/* ---------------------------------------------
+	 * The below utility macro prints out relevant GIC
+	 * registers whenever an unhandled exception is
+	 * taken in BL31 on ARM standard platforms.
+	 * Expects: GICD base in x16, GICC base in x17
+	 * Clobbers: r0 - r10, sp
+	 * ---------------------------------------------
+	 */
+	.macro arm_print_gic_regs
```
soby-mathew: AFAIK, Crash reporting is not enabled from AArch32. I don't think these macros need to be implemented unless crash reporting is implemented for AArch32 in SP_MIN/OPTEE

plat/qemu/qemu_bl1_setup.c
```diff
@@ -57,9 +57,15 @@ void bl1_early_platform_setup(void)
  * does basic initialization. Later architectural setup (bl1_arch_setup())
  * does not do anything platform specific.
  *****************************************************************************/
+#if ARM_ARCH_MAJOR == 7
```
soby-mathew: Is it possible for QEMU to support AArch32 mode on ARMv8? In that case would prefer `#ifdef AARCH32` check here. Similarly other places if is the case.

plat/qemu/platform.mk
```diff
@@ -4,6 +4,24 @@
...
+# Qemu Cortex-A15 model does not implement the virtualization extention.
+# For this reason, we cannot set ARM_CORTEX_A15=yes and must define all
+# the ARMv7 build directives.
+MARCH32_DIRECTIVE 	:= 	-mcpu=cortex-a15
+$(eval $(call add_define,ARMV7_SUPPORTS_LARGE_PAGE_ADDRESSING))
+$(eval $(call add_define,ARMV7_SUPPORTS_GENERIC_TIMER))
+endif # ARMv7
+
+ifeq (${SPD},opteed)
+add-lib-optee 		:= 	yes
```
soby-mathew: This would be to cover the AArch64 mode. Would be worh to call this out in the commit message.

bl32/optee/optee.mk
```diff
@@ -0,0 +1,12 @@
...
+# This makefile only aims at complying with A-TF build process that "optee"
+# is a valid A-TF AArch32 Secure Playload identifier.
+
+ifneq ($(ARCH),aarch32)
+$(error This directory targets AArch32 support)
+endif
+
+$(eval $(call add_define,AARCH32_SP_OPTEE))
+
+$(info A-TF built for OP-TEE payload support)
```
soby-mathew: ARM-TF instead of A-TF

bl1/aarch32/bl1_exceptions.S
```diff
@@ -73,6 +73,12 @@ debug_loop:
 	ldr	r1, [r8, #(ENTRY_POINT_INFO_PC_OFFSET + 4)]
 	msr	spsr, r1
 
+#ifdef ENTRY_POINT_INFO_LR_SVC_OFFSET
```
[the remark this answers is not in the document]
etienne-lms: ok

bl2/bl2_image_load_v2.c
```diff
@@ -87,8 +87,10 @@ entry_point_info_t *bl2_load_images(void)
 	assert(bl2_to_next_bl_params->h.version >= VERSION_2);
 	assert(bl2_to_next_bl_params->head->ep_info);
 
+#ifndef AARCH32_SP_OPTEE
```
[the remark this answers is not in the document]
etienne-lms: ok

bl32/optee/optee.mk
etienne-lms: ok, i will change to "trusted firmware" here and there.

include/common/ep_info.h
```diff
@@ -108,6 +110,12 @@ CASSERT(ENTRY_POINT_INFO_PC_OFFSET ==
 		__builtin_offsetof(entry_point_info_t, pc), \
 		assert_BL31_pc_offset_mismatch);
 
+#ifdef ENTRY_POINT_INFO_LR_SVC_OFFSET
```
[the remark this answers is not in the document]
etienne-lms: I must at least replace it with a `#ifdef AARCH32` as field `lr_svc` is not defined for AArch64.
Actually this is not a build option but a macro defined few lines above.

include/lib/aarch32/arch.h
```diff
@@ -314,6 +314,7 @@
 
 #define MODE_RW_SHIFT		0x4
 #define MODE_RW_MASK		0x1
+#define MODE_RW_64		0x0
 #define MODE_RW_32		0x1
```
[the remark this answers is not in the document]
etienne-lms: **lib/optee/optee_utils.c** requires this macro. Either we define it in both aarch32 and aarch64 **arch.h** or we must add a conditional case in the **optee_utils.c**. I think defining it in both case is simpler.

include/plat/arm/common/aarch32/arm_macros.S
etienne-lms: ok, I will remove this.

plat/qemu/platform.mk
etienne-lms: This does not change the previous implementation which already integrate the optee lib for AArch64 based on `ifeq (${SPD},opteed)`.

plat/qemu/qemu_bl1_setup.c
etienne-lms: ok, I will use AARCH32 rather than ARM_ARCH_MAJOR when possible.

etienne-lms:
I've pushed fixes from comment as appended commit to easy review.
I will squash them if you agree with the changes.

include/common/ep_info.h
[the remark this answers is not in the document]
soby-mathew: Ok

include/lib/aarch32/arch.h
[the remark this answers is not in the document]
soby-mathew: Ah, I see why you added this. But from a architecture point of view, this value for SPSR[4] is not possible for AArch32 mode. Hence would prefer if this is fixed in optee_utils.c.

plat/qemu/platform.mk
soby-mathew: OK.

bl32/optee/optee.mk
soby-mathew: nit: Being pedantic, preferably "ARM Trusted Firmware"

soby-mathew:
@etienne-lms, thanks for the rework. It looks fine and can be squashed. Just minor comments.

include/lib/aarch32/arch.h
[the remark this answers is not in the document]
etienne-lms: Ok, I see you point of view.

davidcunado-arm:
jenkins: test this please

davidcunado-arm:
@jenswi-linaro 
Please ACK this PR?

bl2/bl2_image_load_v2.c
```diff
@@ -87,8 +87,10 @@ entry_point_info_t *bl2_load_images(void)
 	assert(bl2_to_next_bl_params->h.version >= VERSION_2);
 	assert(bl2_to_next_bl_params->head->ep_info);
 
-	/* Populate arg0 for the next BL image */
-	bl2_to_next_bl_params->head->ep_info->args.arg0 = (u_register_t)bl2_to_next_bl_params;
+	/* Populate arg0 for the next BL image if not already provided */
+	if (!bl2_to_next_bl_params->head->ep_info->args.arg0)
```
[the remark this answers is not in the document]
etienne-lms: Ok. Thanks. Good to know.

jenswi-linaro:
Fine with me.

davidcunado-arm:
jenkins; test this please

bl2/bl2_image_load_v2.c
```diff
@@ -87,8 +87,10 @@ entry_point_info_t *bl2_load_images(void)
 	assert(bl2_to_next_bl_params->h.version >= VERSION_2);
 	assert(bl2_to_next_bl_params->head->ep_info);
 
-	/* Populate arg0 for the next BL image */
-	bl2_to_next_bl_params->head->ep_info->args.arg0 = (u_register_t)bl2_to_next_bl_params;
+	/* Populate arg0 for the next BL image if not already provided */
+	if (bl2_to_next_bl_params->head->ep_info->args.arg0 == 0)
```
davidcunado-arm: @etienne-lms 
Apologies - still getting to grips with this checker... minor correction here to resolve this:
MISRA C-2012 Rule 10.4 violation: The operands bl2_to_next_bl_params->head->ep_info->args.arg0 and 0 have essential type categories unsigned 64-bit int and signed 8-bit int, which do not match.

Change 0 to 0ULL - sorry, should have spotted that in my previous comment :(

masahir0y:
> Apologies - still getting to grips with this checker... minor correction here to resolve this:
MISRA C-2012 Rule 10.4 violation: The operands bl2_to_next_bl_params->head->ep_info->args.arg0 and 0 have essential type categories unsigned 64-bit int and signed 8-bit int, which do not match.

This is not true.
In aarch64, `u_register_t` is `uint64_t` a.k.a. `unsigned long`
In aarch32, `u_register_t` is `uint32_t` a.k.a. `unsigned int`

But, the most important is to throw away your silly checker.
What is a real problem in `== 0`?

davidcunado-arm:
@masahir0y 

One of the goals of the MISRA (functional safety) rules is to reduce the risk of implicit types causing unintended behaviours - and as such operands must be of the same essential type, using explicit casts as necessary.

> What is a real problem in == 0?

This does seem unlikely to cause an issue, I agree.

The default type for a decimal is int - signed and unsigned ints are different essential types, hence the check fail and the need for explicit cast to ULL or U.

bl2/bl2_image_load_v2.c
```diff
@@ -88,7 +88,7 @@ entry_point_info_t *bl2_load_images(void)
 	assert(bl2_to_next_bl_params->head->ep_info);
 
 	/* Populate arg0 for the next BL image if not already provided */
-	if (bl2_to_next_bl_params->head->ep_info->args.arg0 == 0)
+	if (bl2_to_next_bl_params->head->ep_info->args.arg0 == (u_register_t)0)
```
etienne-lms: Should fix the issue.

masahir0y: You could also write like follows if you want to avoid the MISRA issue.

    if (!bl2_to_next_bl_params->head->ep_info->args.arg0)

davidcunado-arm: Unfortunately, that would trigger the original issue: 

MISRA C-2012 Rule 10.1 violation: The operand "bl2_to_next_bl_params->head->ep_info->args.arg0" has essential type "unsigned 64-bit int", but should be Essentially Boolean.

davidcunado-arm:
jenkins: test this please

bl2/bl2_image_load_v2.c
masahir0y: Please skip except mandatory.
This check is ridiculous.

davidcunado-arm:
Check fails - false positives from static checks; ignore

bl2/bl2_image_load_v2.c
danh-arm: For what it's worth, I think a better (and more consistent) fix for this would have been to use `U(0)` or `0u` in the comparison instead of the cast.

Both 10.1 and 10.4 are "required" MISRA rules so we'd like to keep them on if possible. However, we haven't properly announced our policy yet. We should do this.

PR #1174, opened 2017-11-29
Replace magic numbers in linkerscripts by PAGE_SIZE

antonio-nino-diaz-arm:
When defining different sections in linker scripts it is needed to align them to multiples of the page size. In most linker scripts this is done by aligning to the hardcoded value 4096 instead of PAGE_SIZE.

This may be confusing when taking a look at all the codebase, as 4096 is used in some parts that aren't meant to be a multiple of the page size.

### DIFF
--- a/bl1/bl1.ld.S
+++ b/bl1/bl1.ld.S
@@ -5,6 +5,7 @@
  */
 
 #include <platform_def.h>
+#include <xlat_tables_defs.h>
 
 OUTPUT_FORMAT(PLATFORM_LINKER_FORMAT)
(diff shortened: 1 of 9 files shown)


antonio-nino-diaz-arm:
By the way, @mtk09422 and @rockchip-linux, could you confirm that the changes are correct?

davidcunado-arm:
@mtk01412 and @rockchip-linux
I'll look to merge this at the end of this week - if you have any concerns, please let me know before then.

davidcunado-arm:
jenkins: test this please

PR #1180, opened 2017-12-05
Rename some macros in SPM code

sandrine-bailleux-arm:


davidcunado-arm:
jenkins: test this please

PR #1181, opened 2017-12-05
ARM Platforms: Change the TZC access permissions for EL3 payload

soby-mathew:
This patch allows non-secure bus masters to access TZC region0 as well
as the EL3 Payload itself.

Change-Id: I7e44f2673a2992920d41503fb4c57bd7fb30747a


davidcunado-arm:
jenkins: test this please

PR #1182, opened 2017-12-05
Unify cache flush code path after image load

soby-mathew:
Previously the cache flush happened in 2 different places in code
depending on whether TRUSTED_BOARD_BOOT is enabled or not. This
patch unifies this code path for both the cases. The `load_image()`
function is now made an internal static function.

Change-Id: I96a1da29d29236bbc34b1c95053e6a9a7fc98a54


davidcunado-arm:
jenkins: test this please

PR #1183, opened 2017-12-06
SDEI: Fix return value of reset calls

jeenu-arm:
At present, both SDEI_PRIVATE_RESET and SDEI_SHARED_RESET returns
SDEI_PENDING if they fail to unregister an event. The SDEI specification
however requires that the APIs return SDEI_EDENY in these cases. This
patch fixes the return codes for the reset APIs.

Change-Id: Ic14484c91fa8396910387196c256d1ff13d03afd


davidcunado-arm:
jenkins: test this please

PR #1184, opened 2017-12-06
fvp: Disable SYSTEM_SUSPEND when ARM_BL31_IN_DRAM 

antonio-nino-diaz-arm:
After returning from SYSTEM_SUSPEND state, BL31 reconfigures the TrustZone Controller during the boot sequence. If BL31 is placed in TZC-secured DRAM, it will try to change the permissions of the memory it is being executed from, causing an exception.

The solution is to disable SYSTEM_SUSPEND when the Trusted Firmware has been compiled with ``ARM_BL31_IN_DRAM=1``.

Also, a small issue regarding an Arm platform header has been fixed.


davidcunado-arm:
jenkins: test this please

PR #1187, opened 2017-12-06
SPM: Move S-EL1/S-EL0 xlat tables to TZC DRAM

antonio-nino-diaz-arm:
A new platform define, `PLAT_SP_IMAGE_XLAT_SECTION_NAME`, has been introduced to select the section where the translation tables used by the S-EL1/S-EL0 are placed.

This define has been used to move the translation tables to DRAM secured by TrustZone.

Most of the extra needed space in BL31 when SPM is enabled is due to the large size of the translation tables. By moving them to this memory region we can save 44 KiB.

A new argument has been added to REGISTER_XLAT_CONTEXT2() to specify the region where the translation tables have to be placed by the linker.


davidcunado-arm:
jenkins: test this please

PR #1188, opened 2017-12-08
SPM: Declare explicit width based types in secure_partition_boot_info…

SNG-ARM:
…_t structure

The secure_partition_boot_info_t structure is used to communicate boot
parameters with the StandaloneMM code executing at S-EL0 through a
shared buffer. Certain data types used for members of this structure
are opaque with their size depending on the toolchain being used.

Declare the members of the structure with explicit width based data
types, which would maintain compatibility across toolchains.

Signed-off-by: Sughosh Ganu <sughosh.ganu@arm.com>

### DIFF
--- a/include/services/secure_partition.h
+++ b/include/services/secure_partition.h
@@ -36,26 +36,26 @@ extern uintptr_t __SP_IMAGE_XLAT_TABLES_END__;
  */
 typedef struct secure_partition_mp_info {
 	u_register_t		mpidr;
-	unsigned int		linear_id;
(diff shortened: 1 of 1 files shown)


davidcunado-arm:
@SNG-ARM 
For future PRs, please target the integration branch rather than master, as per the contributing guidelines.
(I've updated this PR)

davidcunado-arm:
@SNG-ARM 
I'm going to close this PR for the moment. We can re-open at a later date.

PR #1191, opened 2017-12-12
arm-tf: Update scmi driver to support multiple MHU generations

samarthp:
Updated scmi driver to allow support for multiple generations of MHU
controller available in ARM systems.

As per MHUv2 specification implemented in latest ARM sub-systems, the
sender side must wake the receiver before initiating any data transfer
over MHUv2. The same requirement has been incorporated with this patch
for supporting transactions over MHUv2.

Also, defined plat_css_scmi_plat_info as weak structure to allow a
platform to override the way Door Bell parameters of MHU are
initialised for SCMI.

Signed-off-by: Samarth Parikh <samarth.parikh@arm.com>

### DIFF
--- a/plat/arm/css/drivers/scmi/scmi.h
+++ b/plat/arm/css/drivers/scmi/scmi.h
@@ -83,6 +83,12 @@
 #define SCMI_E_OUT_OF_RANGE		-5
 #define SCMI_E_BUSY			-6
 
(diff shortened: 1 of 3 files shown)


davidcunado-arm:
@samarthp 
I'm closing this PR - we can re-open once the changes are ready.

PR #1192, opened 2017-12-12
SPM: Fix MM_COMMUNICATE_AARCH32/64 parameters

sandrine-bailleux-arm:
This partially reverts commit d6b532b50f8, keeping only the fixes to
the assertions. The changes related to the order of arguments passed
to the secure partition were not correct and violated the
specification of the SP_EVENT_COMPLETE SMC.

This patch also improves the MM_COMMUNICATE argument validation.  The
cookie argument, as it comes from normal world, can't be trusted and thus
needs to always be validated at run time rather than using an assertion.

Also validate the communication buffer address and return
INVALID_PARAMETER if it is zero, as per the MM specification.

Fix a few typos in comments and use the "secure partition" terminology
rather than "secure payload".


davidcunado-arm:
jenkins: test this please

PR #1193, opened 2017-12-12
New console API and coreboot support [v4]

jwerner-chromium:
Hi @danh-arm,

I've finally managed to find time to finish up the next console API iteration again. I think this should address all comments you had on the last version and in particular keep the non-structure console core functions in existing drivers alive for use in BL1 as we discussed. Let me know if I missed anything!


danh-arm:
jenkins: test this please

davidcunado-arm:
jenkins: test this please

davidcunado-arm:
jenkins: test this please

plat/common/aarch64/platform_helpers.S
```diff
@@ -57,31 +58,54 @@ func plat_report_exception
...
+	 * valid for the crash state) has been registered with
+	 * the console framework. To debug crashes that occur
+	 * earlier, the platform has to override these functions
+	 * with an implementation that initializes a console
+	 * driver with hardcoded parameters. See
+	 * docs/porting-guide.rst for more information.
 	 * -----------------------------------------------------
 	 */
 func plat_crash_console_init
+#if defined(IMAGE_BL1)
```
davidcunado-arm: The checks failed for several build configurations with the following: 
/work/workspace/trusted-firmware-build/BUILD_CONFIG/hikey960-default/MODE/Debug/label/platforms/plat/common/aarch64/platform_helpers.S:87: undefined reference to `console_switch_state'
/work/workspace/trusted-firmware-build/BUILD_CONFIG/hikey960-default/MODE/Debug/label/platforms/plat/common/aarch64/platform_helpers.S:87:(.text.plat_crash_console_init+0x8): relocation truncated to fit: R_AARCH64_CALL26 against undefined symbol `console_switch_state'

I think the issue is that this line should be:
#if defined(IMAGE_BL1) || defined(IMAGE_BL2)

This fixes the failing hikey builds - build options:
- DEBUG=1 PLAT=hikey960
- DEBUG=1 PLAT=hikey960 SPD=opteed
The impacted build configurations are:
- hikey960: make

include/drivers/console.h
```diff
@@ -7,14 +7,69 @@
...
+/* Read a character (blocking) from any console registered for current state. */
 int console_getc(void);
+/* Flush all consoles registered for the current state. */
 int console_flush(void);
 
+#if !MULTI_CONSOLE_API
+/* DEPRECATED on AArch64 -- use console_<driver>_register() instead! */
+int console_init(uintptr_t base_addr,
+		 unsigned int uart_clk, unsigned int baud_rate) __deprecated;
+void console_uninit(void) __deprecated;
```
davidcunado-arm: Build failures with the uniphier platform configs - e.g. build options DEBUG=1 PLAT=uniphier

/work/workspace/trusted-firmware-build/BUILD_CONFIG/uniphier-default/MODE/Debug/label/platforms/plat/socionext/uniphier/uniphier_console.S:211: undefined reference to `console_init'
/work/workspace/trusted-firmware-build/BUILD_CONFIG/uniphier-default/MODE/Debug/label/platforms/plat/socionext/uniphier/uniphier_console.S:211:(.text.uniphier_console_setup+0x30): relocation truncated to fit: R_AARCH64_JUMP26 against undefined symbol `console_init'
make: *** [build/uniphier/debug/bl1/bl1.elf] Error 1

Adding MULTI_CONSOLE = 1 and also remove the __deprecated from console_uinit resolved the build issue.
(NOTE - the uniphier platform has ERROR_DEPRECATED=1)

plat/rockchip/common/bl31_plat_setup.c
```diff
@@ -69,8 +71,20 @@ void params_early_setup(void *plat_param_from_bl2)
...
+	params_early_setup(plat_params_from_bl2);
+
+#if COREBOOT
+	if (coreboot_serial.type)
+		console_16550_register(coreboot_serial.baseaddr,
+				       coreboot_serial.input_hertz,
+				       coreboot_serial.baud,
+				       &console);
+#else
+	console_16550_register(PLAT_RK_UART_BASE, PLAT_RK_UART_CLOCK,
```
davidcunado-arm: Static check failure:
Return value of non-void function console_16550_register() is not used.

include/drivers/console.h
danh-arm: We need to remove the ERROR_DEPRECATED=1 from the upstream uniphier platform. This flag will always cause problems with deprecated APIs. We have a patch to do this in another in-flight patchset. Platforms should not be forced to use MULTI_CONSOLE=1.

plat/common/aarch64/platform_helpers.S
danh-arm: No, I think the condition should be:

    #if defined (MULTI_CONSOLE) && defined (IMAGE_BL1)

The other changes to the plat_crash_console_xxx functions should also be configured under MULTI_CONSOLE. Platforms shoudn't be forced to change these functions.

include/drivers/console.h
jwerner-chromium: So, should I wait for that patch to land? Is it going to land soon, or if not can they pull it out into a separate PR?

plat/common/aarch64/platform_helpers.S
jwerner-chromium: Yes, sorry, this whole thing needs to be wrapped into an #if MULTI_CONSOLE_API block. Not sure at what point I forgot to do that. Fixed.

plat/rockchip/common/bl31_plat_setup.c
jwerner-chromium: Is this a blocking error? The old code here was also calling console_init() (which returned an int) without checking the result. I'm essentially not changing anything about that, just adjusting this to match the slightly changed API (but the return value still means the same).

Also, I don't think there's anything really useful you can do about an error here.

include/drivers/console.h
danh-arm: You can ignore that error in this PR. We'll deal with it separately

danh-arm:
jenkins: test this please

include/drivers/console.h
davidcunado-arm: @jwerner-chromium 
That PR has now landed and so this issue should be resolved.

plat/rockchip/common/bl31_plat_setup.c
davidcunado-arm: @jwerner-chromium 
Ok - we can ignore this issue.

davidcunado-arm:
jenkins: test this please

davidcunado-arm:
@jwerner-chromium 
For our CI, should we now be building the rockchip platforms with COREBOOT=1? Should this be set in the corresponding platform.mk files?

jwerner-chromium:
@davidcunado-arm 
I think it would be worth building it both with and without COREBOOT=1 if your capacity allows for that, so that the coreboot parts won't break accidentally in the future. But it should not be set in platform.mk because the Rockchip platform may also use other bootloaders (like U-Boot).

davidcunado-arm:
@jwerner-chromium 
Sorted - have added additional configurations that include COREBOOT=1.

PR #1194, opened 2017-12-13
io: block: fix block_read/write may read/write overlap buffer

robertovargas-arm:
The block operations were trying to optimize the number of memory
copies, and it tried to use directly the buffer supplied by the user
to them. This was a mistake because it created too many corner cases:

	1- It was possible to generate unaligned
	   operations to unaligned buffers. Drivers that were using
	   DMA transfer failed in that case.

	2- It was possible to generate read operations
	   with sizes that weren't a multiple of the block size. Some
	   low level drivers assumed that condition and they calculated
	   the number of blocks dividing the number of bytes by the
	   size of the block, without considering the remaining bytes.

	3- The block_* operations didn't control the
	   number of bytes actually copied to memory, because the
	   low level drivers were writing directly to the user buffer.

This patch rewrite block_read and block_write to use always the device
buffer, which the platform ensures that has the correct aligment and
the correct size.

Change-Id: I5e479bb7bc137e6ec205a8573eb250acd5f40420
Signed-off-by: Qixiang Xu <qixiang.xu@arm.com>

### DIFF
--- a/drivers/io/io_block.c
+++ b/drivers/io/io_block.c
@@ -167,15 +167,98 @@ static int block_seek(io_entity_t *entity, int mode, ssize_t offset)
 	return 0;
 }
 
+/*
+ * This function allows the caller to read any number of bytes
+ * from any position. It hides from the caller that the low level
+ * driver only can read aligned blocks of data. For this reason
+ * we need to handle the use case where the first byte to be read is not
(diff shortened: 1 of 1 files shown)


davidcunado-arm:
jenkins: test this please

PR #1195, opened 2017-12-14
Disable PIE compilation option

davidcunado-arm:
ARM TF does not work correctly if built with a version of gcc
that is configured to use PIE by default (e.g. Debian Stretch).

This patch identifies when such a version of gcc is being used
(by searching for --enable-default-pie) and adds -fno-PIE option
to TF_CFLAGS.

fixes arm-software/tf-issues#519

Change-Id: I2322122c49841746d35d152694e14f6f73beb0fd
Co-Authored-by: Evan Lloyd <evan.lloyd@arm.com>
Tested-by: Steve Capper <steve.capper@arm.com>
Tested-by: Alexei Fedorov <alexei.fedorov@arm.com>


davidcunado-arm:
jenkins: test this please

PR #1200, opened 2017-12-15
Add BL2_AT_EL3 build option

robertovargas-arm:
This option enables runnig BL2 at EL3 and it makes possible to run BL2 without having a BL1.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -586,8 +586,12 @@ $(eval $(call MAKE_BL,1))
 endif
 
 ifeq (${NEED_BL2},yes)
-$(if ${BL2}, $(eval $(call MAKE_TOOL_ARGS,2,${BL2},tb-fw)),\
-	$(eval $(call MAKE_BL,2,tb-fw)))
+ifeq (${BL2_AT_EL3}, 0)
+FIP_BL2_ARGS := tb-fw
+endif
+
+$(if ${BL2}, $(eval $(call MAKE_TOOL_ARGS,2,${BL2},${FIP_BL2_ARGS})),\
+	$(eval $(call MAKE_BL,2,${FIP_BL2_ARGS})))
 endif
 
 ifeq (${NEED_SCP_BL2},yes)
--- a/Makefile
+++ b/Makefile
@@ -500,6 +500,7 @@ $(eval $(call assert_boolean,TRUSTED_BOARD_BOOT))
 $(eval $(call assert_boolean,USE_COHERENT_MEM))
 $(eval $(call assert_boolean,USE_TBBR_DEFS))
 $(eval $(call assert_boolean,WARMBOOT_ENABLE_DCACHE_EARLY))
+$(eval $(call assert_boolean,BL2_AT_EL3))
 
 $(eval $(call assert_numeric,ARM_ARCH_MAJOR))
 $(eval $(call assert_numeric,ARM_ARCH_MINOR))
@@ -543,6 +544,7 @@ $(eval $(call add_define,TRUSTED_BOARD_BOOT))
 $(eval $(call add_define,USE_COHERENT_MEM))
 $(eval $(call add_define,USE_TBBR_DEFS))
 $(eval $(call add_define,WARMBOOT_ENABLE_DCACHE_EARLY))
+$(eval $(call add_define,BL2_AT_EL3))
 
 # Define the EL3_PAYLOAD_BASE flag only if it is provided.
 ifdef EL3_PAYLOAD_BASE
--- a/bl1/bl1.mk
+++ b/bl1/bl1.mk
@@ -13,7 +13,10 @@ BL1_SOURCES		+=	bl1/bl1_main.c				\
 				lib/cpus/errata_report.c		\
 				lib/el3_runtime/${ARCH}/context_mgmt.c	\
 				plat/common/plat_bl1_common.c		\
-				plat/common/${ARCH}/platform_up_stack.S
+				plat/common/${ARCH}/platform_up_stack.S \
+				${MBEDTLS_COMMON_SOURCES}		\
+				${MBEDTLS_CRYPTO_SOURCES}		\
+				${MBEDTLS_X509_SOURCES}
 
 ifeq (${ARCH},aarch64)
 BL1_SOURCES		+=	lib/el3_runtime/aarch64/context.S
--- a/bl2/aarch32/bl2_el3_entrypoint.S
+++ b/bl2/aarch32/bl2_el3_entrypoint.S
@@ -1,0 +1,84 @@
+/*
+ * Copyright (c) 2017, ARM Limited and Contributors. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include <arch.h>
+#include <asm_macros.S>
+#include <bl_common.h>
+#include <el3_common_macros.S>
+
+
+	.globl	bl2_entrypoint
+	.globl	bl2_run_next_image
+
+
+func bl2_entrypoint
+	/* Save arguments x0-x3 from previous Boot loader */
+	mov	r9, r0
+	mov	r10, r1
+	mov	r11, r2
+	mov	r12, r3
+
+	el3_entrypoint_common                                   \
+                _init_sctlr=1                                   \
+                _warm_boot_mailbox=!PROGRAMMABLE_RESET_ADDRESS  \
+                _secondary_cold_boot=!COLD_BOOT_SINGLE_CPU      \
+                _init_memory=1                                  \
+                _init_c_runtime=1                               \
+                _exception_vectors=bl2_vector_table
+
+	/*
+	 * Restore parameters of boot rom
+	 */
+	mov	r0, r9
+	mov	r1, r10
+	mov	r2, r11
+	mov	r3, r12
+
+	bl	bl2_el3_early_platform_setup
+	bl	bl2_el3_plat_arch_setup
+
+	/* ---------------------------------------------
+	 * Jump to main function.
+	 * ---------------------------------------------
+	 */
+	bl	bl2_main
+
+	/* ---------------------------------------------
+	 * Should never reach this point.
+	 * ---------------------------------------------
+	 */
+	no_ret	plat_panic_handler
+
+endfunc bl2_entrypoint
+
+func bl2_run_next_image
+	mov	r8,r0
+
+	/*
+	 * MMU needs to be disabled because both BL2 and BL32 execute
+	 * in PL1, and therefore share the same address space.
+	 * BL32 will initialize the address space according to its
+	 * own requirement.
+	 */
+	bl	disable_mmu_icache_secure
+	stcopr	r0, TLBIALL
+	dsb	sy
+	isb
+	mov	r0, r8
+	bl	bl2_el3_plat_prepare_exit
+
+	/*
+	 * Extract PC and SPSR based on struct `entry_point_info_t`
+	 * and load it in LR and SPSR registers respectively.
+	 */
+	ldr	lr, [r8, #ENTRY_POINT_INFO_PC_OFFSET]
+	ldr	r1, [r8, #(ENTRY_POINT_INFO_PC_OFFSET + 4)]
+	msr	spsr, r1
+
+	add	r8, r8, #ENTRY_POINT_INFO_ARGS_OFFSET
+	ldm	r8, {r0, r1, r2, r3}
+	eret
+endfunc bl2_run_next_image
--- a/bl2/aarch32/bl2_el3_exceptions.S
+++ b/bl2/aarch32/bl2_el3_exceptions.S
@@ -1,0 +1,21 @@
+/*
+ * Copyright (c) 2017, ARM Limited and Contributors. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include <arch.h>
+#include <asm_macros.S>
+#include <bl_common.h>
+
+	.globl	bl2_vector_table
+
+vector_base bl2_vector_table
+	b	bl2_entrypoint
+	b	report_exception	/* Undef */
+	b	report_exception	/* SVC call */
+	b	report_exception	/* Prefetch abort */
+	b	report_exception	/* Data abort */
+	b	report_exception	/* Reserved */
+	b	report_exception	/* IRQ */
+	b	report_exception	/* FIQ */
--- a/bl2/aarch64/bl2_el3_entrypoint.S
+++ b/bl2/aarch64/bl2_el3_entrypoint.S
@@ -1,0 +1,77 @@
+/*
+ * Copyright (c) 2017, ARM Limited and Contributors. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include <arch.h>
+#include <asm_macros.S>
+#include <bl_common.h>
+#include <el3_common_macros.S>
+
+	.globl	bl2_entrypoint
+	.globl	bl2_vector_table
+	.globl	bl2_el3_run_image
+	.globl	bl2_run_next_image
+
+func bl2_entrypoint
+	/* Save arguments x0-x3 from previous Boot loader */
+	mov	x20, x0
+	mov	x21, x1
+	mov	x22, x2
+	mov	x23, x3
+
+	el3_entrypoint_common                                   \
+		_init_sctlr=1                                   \
+		_warm_boot_mailbox=!PROGRAMMABLE_RESET_ADDRESS  \
+		_secondary_cold_boot=!COLD_BOOT_SINGLE_CPU      \
+		_init_memory=1                                  \
+		_init_c_runtime=1                               \
+		_exception_vectors=bl2_el3_exceptions
+
+	/*
+	 * Restore parameters of boot rom
+	 */
+	mov	x0, x20
+	mov	x1, x21
+	mov	x2, x22
+	mov	x3, x23
+
+	bl	bl2_el3_early_platform_setup
+	bl	bl2_el3_plat_arch_setup
+
+	/* ---------------------------------------------
+	 * Jump to main function.
+	 * ---------------------------------------------
+	 */
+	bl	bl2_main
+
+	/* ---------------------------------------------
+	 * Should never reach this point.
+	 * ---------------------------------------------
+	 */
+	no_ret	plat_panic_handler
+endfunc bl2_entrypoint
+
+func bl2_run_next_image
+	mov	x20,x0
+        /*
+         * MMU needs to be disabled because both BL2 and BL31 execute
+         * in EL3, and therefore share the same address space.
+         * BL31 will initialize the address space according to its
+         * own requirement.
+         */
+	bl	disable_mmu_icache_el3
+	tlbi	alle3
+	bl	bl2_el3_plat_prepare_exit
+
+	ldp	x0, x1, [x20, #ENTRY_POINT_INFO_PC_OFFSET]
+	msr	elr_el3, x0
+	msr	spsr_el3, x1
+
+	ldp	x6, x7, [x20, #(ENTRY_POINT_INFO_ARGS_OFFSET + 0x30)]
+	ldp	x4, x5, [x20, #(ENTRY_POINT_INFO_ARGS_OFFSET + 0x20)]
+	ldp	x2, x3, [x20, #(ENTRY_POINT_INFO_ARGS_OFFSET + 0x10)]
+	ldp	x0, x1, [x20, #(ENTRY_POINT_INFO_ARGS_OFFSET + 0x0)]
+	eret
+endfunc bl2_run_next_image
--- a/bl2/aarch64/bl2_el3_exceptions.S
+++ b/bl2/aarch64/bl2_el3_exceptions.S
@@ -1,0 +1,131 @@
+/*
+ * Copyright (c) 2017, ARM Limited and Contributors. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include <arch.h>
+#include <asm_macros.S>
+#include <bl1.h>
+#include <bl_common.h>
+#include <context.h>
+
+/* -----------------------------------------------------------------------------
+ * Very simple stackless exception handlers used by BL2.
+ * -----------------------------------------------------------------------------
+ */
+	.globl	bl2_el3_exceptions
+
+vector_base bl2_el3_exceptions
+
+	/* -----------------------------------------------------
+	 * Current EL with SP0 : 0x0 - 0x200
+	 * -----------------------------------------------------
+	 */
+vector_entry SynchronousExceptionSP0
+	mov	x0, #SYNC_EXCEPTION_SP_EL0
+	bl	plat_report_exception
+	no_ret	plat_panic_handler
+	check_vector_size SynchronousExceptionSP0
+
+vector_entry IrqSP0
+	mov	x0, #IRQ_SP_EL0
+	bl	plat_report_exception
+	no_ret	plat_panic_handler
+	check_vector_size IrqSP0
+
+vector_entry FiqSP0
+	mov	x0, #FIQ_SP_EL0
+	bl	plat_report_exception
+	no_ret	plat_panic_handler
+	check_vector_size FiqSP0
+
+vector_entry SErrorSP0
+	mov	x0, #SERROR_SP_EL0
+	bl	plat_report_exception
+	no_ret	plat_panic_handler
+	check_vector_size SErrorSP0
+
+	/* -----------------------------------------------------
+	 * Current EL with SPx: 0x200 - 0x400
+	 * -----------------------------------------------------
+	 */
+vector_entry SynchronousExceptionSPx
+	mov	x0, #SYNC_EXCEPTION_SP_ELX
+	bl	plat_report_exception
+	no_ret	plat_panic_handler
+	check_vector_size SynchronousExceptionSPx
+
+vector_entry IrqSPx
+	mov	x0, #IRQ_SP_ELX
+	bl	plat_report_exception
+	no_ret	plat_panic_handler
+	check_vector_size IrqSPx
+
+vector_entry FiqSPx
+	mov	x0, #FIQ_SP_ELX
+	bl	plat_report_exception
+	no_ret	plat_panic_handler
+	check_vector_size FiqSPx
+
+vector_entry SErrorSPx
+	mov	x0, #SERROR_SP_ELX
+	bl	plat_report_exception
+	no_ret	plat_panic_handler
+	check_vector_size SErrorSPx
+
+	/* -----------------------------------------------------
+	 * Lower EL using AArch64 : 0x400 - 0x600
+	 * -----------------------------------------------------
+	 */
+vector_entry SynchronousExceptionA64
+	mov	x0, #SYNC_EXCEPTION_AARCH64
+	bl	plat_report_exception
+	no_ret	plat_panic_handler
+	check_vector_size SynchronousExceptionA64
+
+vector_entry IrqA64
+	mov	x0, #IRQ_AARCH64
+	bl	plat_report_exception
+	no_ret	plat_panic_handler
+	check_vector_size IrqA64
+
+vector_entry FiqA64
+	mov	x0, #FIQ_AARCH64
+	bl	plat_report_exception
+	no_ret	plat_panic_handler
+	check_vector_size FiqA64
+
+vector_entry SErrorA64
+	mov	x0, #SERROR_AARCH64
+	bl	plat_report_exception
+	no_ret	plat_panic_handler
+	check_vector_size SErrorA64
+
+	/* -----------------------------------------------------
+	 * Lower EL using AArch32 : 0x600 - 0x800
+	 * -----------------------------------------------------
+	 */
+vector_entry SynchronousExceptionA32
+	mov	x0, #SYNC_EXCEPTION_AARCH32
+	bl	plat_report_exception
+	no_ret	plat_panic_handler
+	check_vector_size SynchronousExceptionA32
+
+vector_entry IrqA32
+	mov	x0, #IRQ_AARCH32
+	bl	plat_report_exception
+	no_ret	plat_panic_handler
+	check_vector_size IrqA32
+
+vector_entry FiqA32
+	mov	x0, #FIQ_AARCH32
+	bl	plat_report_exception
+	no_ret	plat_panic_handler
+	check_vector_size FiqA32
+
+vector_entry SErrorA32
+	mov	x0, #SERROR_AARCH32
+	bl	plat_report_exception
+	no_ret	plat_panic_handler
+	check_vector_size SErrorA32
--- a/bl2/bl2.mk
+++ b/bl2/bl2.mk
@@ -7,7 +7,10 @@
 BL2_SOURCES		+=	bl2/bl2_main.c				\
 				bl2/${ARCH}/bl2_arch_setup.c		\
 				lib/locks/exclusive/${ARCH}/spinlock.S	\
-				plat/common/${ARCH}/platform_up_stack.S
+				plat/common/${ARCH}/platform_up_stack.S	\
+				${MBEDTLS_COMMON_SOURCES}               \
+				${MBEDTLS_CRYPTO_SOURCES}		\
+				${MBEDTLS_X509_SOURCES}
 
 ifeq (${ARCH},aarch64)
 BL2_SOURCES		+=	common/aarch64/early_exceptions.S
--- a/bl2/bl2.mk
+++ b/bl2/bl2.mk
@@ -5,7 +5,6 @@
 #
 
 BL2_SOURCES		+=	bl2/bl2_main.c				\
-				bl2/${ARCH}/bl2_entrypoint.S		\
 				bl2/${ARCH}/bl2_arch_setup.c		\
 				lib/locks/exclusive/${ARCH}/spinlock.S	\
 				plat/common/${ARCH}/platform_up_stack.S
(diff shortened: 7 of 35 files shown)


danh-arm:
Hi @masahir0y / @hzhuang1 

This PR might help streamline the boot flow on your platforms. It should make it possible to branch directly from your ROM code to TF BL2 executing at EL3, without the need for a TF "pseudo-BL1". Let us know if you have any feedback.

Regards

Dan.

masahir0y:
I started to take a look.

I tested this on my board.
I needed to fix build error for Trusted Board Boot,
but otherwise, almost working.

 - Do not add BL1_SOURCES for TBB if BL2_AT_EL3
   since we use own boot rom.
 - Remove unneeded blank line
 - Remove blank line at the end of file
 - Fix nits in doc
 - Remove mismatch endif comments

So far for today.  I will probably continue review tomorrow.

Diff is below.


    index 3891ad1..7e01063 100644
    --- a/Makefile
    +++ b/Makefile
    @@ -495,7 +495,6 @@ $(eval $(call assert_boolean,USE_TBBR_DEFS))
     $(eval $(call assert_boolean,BL2_AT_EL3))
     
    -
     $(eval $(call assert_numeric,ARM_ARCH_MAJOR))
     $(eval $(call assert_numeric,ARM_ARCH_MINOR))
     
    diff --git a/bl2/aarch32/bl2_el3_exceptions.S b/bl2/aarch32/bl2_el3_exceptions.S
    index c4bbd37..11ddf37 100644
    --- a/bl2/aarch32/bl2_el3_exceptions.S
    +++ b/bl2/aarch32/bl2_el3_exceptions.S
    @@ -19,4 +19,3 @@ vector_base bl2_vector_table
     	b	report_exception	/* Reserved */
     	b	report_exception	/* IRQ */
     	b	report_exception	/* FIQ */
    -
    diff --git a/docs/porting-guide.rst b/docs/porting-guide.rst
    index 7f2a7e0..2112976 100644
    --- a/docs/porting-guide.rst
    +++ b/docs/porting-guide.rst
    @@ -1624,11 +1624,11 @@ element in the boot sequence. If there are no more boot sources then it
     must return 0, otherwise it must return 1. The default implementation
     of this always returns 0.
     
    -Boot Loader Stage 2 at EL3 (BL2)
    +Boot Loader Stage 2 (BL2) at EL3
     --------------------------------
     
     When the platform has a non-TF Boot ROM it is desirable to jump
    -directly to BL2 instead of TF BL1. In this case BL2 it is expected to
    +directly to BL2 instead of TF BL1. In this case BL2 is expected to
     execute at EL3 instead of executing at EL1. Refer to the `Firmware
     Design`_ for more information.
     
    diff --git a/drivers/auth/mbedtls/mbedtls_common.mk b/drivers/auth/mbedtls/mbedtls_common.mk
    index f2b6f6e..6e08fba 100644
    --- a/drivers/auth/mbedtls/mbedtls_common.mk
    +++ b/drivers/auth/mbedtls/mbedtls_common.mk
    @@ -29,7 +29,9 @@ MBEDTLS_COMMON_SOURCES	:=	drivers/auth/mbedtls/mbedtls_common.c	\
     				platform.c 				\
     				)
     
    +ifneq (${BL2_AT_EL3},1)
     BL1_SOURCES		+=	${MBEDTLS_COMMON_SOURCES}
    +endif
     BL2_SOURCES		+=	${MBEDTLS_COMMON_SOURCES}
     
     endif
    diff --git a/drivers/auth/mbedtls/mbedtls_crypto.mk b/drivers/auth/mbedtls/mbedtls_crypto.mk
    index 8eb4873..34c527b 100644
    --- a/drivers/auth/mbedtls/mbedtls_crypto.mk
    +++ b/drivers/auth/mbedtls/mbedtls_crypto.mk
    @@ -90,5 +90,7 @@ endif
     $(eval $(call add_define,TF_MBEDTLS_KEY_ALG_ID))
     $(eval $(call add_define,TF_MBEDTLS_HASH_ALG_ID))
     
    +ifneq (${BL2_AT_EL3},1)
     BL1_SOURCES			+=	${MBEDTLS_CRYPTO_SOURCES}
    +endif
     BL2_SOURCES			+=	${MBEDTLS_CRYPTO_SOURCES}
    diff --git a/drivers/auth/mbedtls/mbedtls_x509.mk b/drivers/auth/mbedtls/mbedtls_x509.mk
    index 0f28b65..2271706 100644
    --- a/drivers/auth/mbedtls/mbedtls_x509.mk
    +++ b/drivers/auth/mbedtls/mbedtls_x509.mk
    @@ -12,5 +12,7 @@ MBEDTLS_X509_SOURCES	:=	drivers/auth/mbedtls/mbedtls_x509_parser.c	\
     				x509_crt.c 					\
     				)
     
    +ifneq (${BL2_AT_EL3},1)
     BL1_SOURCES		+=	${MBEDTLS_X509_SOURCES}
    +endif
     BL2_SOURCES		+=	${MBEDTLS_X509_SOURCES}
    diff --git a/include/common/aarch32/el3_common_macros.S b/include/common/aarch32/el3_common_macros.S
    index 4c7808d..d654b65 100644
    --- a/include/common/aarch32/el3_common_macros.S
    +++ b/include/common/aarch32/el3_common_macros.S
    @@ -273,7 +273,7 @@
     		ldr	r1, =__RW_END__
     		sub	r1, r1, r0
     		bl	inv_dcache_range
    -#endif /* IMAGE_BL32 */
    +#endif
     
     		ldr	r0, =__BSS_START__
     		ldr	r1, =__BSS_SIZE__
    diff --git a/include/common/aarch64/el3_common_macros.S b/include/common/aarch64/el3_common_macros.S
    index 69c8e15..ebaf7aa 100644
    --- a/include/common/aarch64/el3_common_macros.S
    +++ b/include/common/aarch64/el3_common_macros.S
    @@ -282,7 +282,7 @@
     		adr	x1, __RW_END__
     		sub	x1, x1, x0
     		bl	inv_dcache_range
    -#endif /* IMAGE_BL31 */
    +#endif
     
     		ldr	x0, =__BSS_START__
     		ldr	x1, =__BSS_SIZE__

etienne-lms:
Dear all, here is a comment regarding the AArch32 support for this change.

Current implementation would make BL2 fully executing in the AArch32 Monitor mode of the CPU which is not handy in case of sync exceptions (aborts). Would it be possible to get the AArch32 implementation of this change to have BL2 executing in SVC mode instead?
- AArch32 `el3_entrypoint_common` should not switch to Monitor mode.
- AArch32 `bl2_run_next_image` should end with a branch instead of an `eret`.
- AArch32 `bl2_run_next_image` should also do some tricks regarding `cpsr` value (bl2 effective `cpsr` value versus target next-image `cpsr`value).

danh-arm:
Hi @masahir0y / @etienne-lms 

Thanks for your feedback. All good comments. We're going to into the holiday period here so it might take a while to get these addressed. Just a couple of points:

* I believe Roberto tried removing BL1 sources in this case but ran into some problems. I'll let him comment further.

* The AArch32 support has not been tested and is not the main use-case for this feature. I'm wondering what the value of this is given that AArch32 platforms reset to Secure SVC? Non-TF ROMs could branch to TF BL2 in the normal configuration. Perhaps the only useful thing here is the reset handling support? Anyway, unless these comments can be addressed quickly, I'd rather merge this feature without AArch32 support, and add in AArch32 later if needed.

Regards

Dan.

masahir0y:
Hi.

Please exclude BL2 image from FIP when BL2_AT_EL3 is enabled.

Having BL2 in FIP is pointless because platform's own ROM does not have the parser of FIP.
It will load BL2 by its own way.

You can do it by not giving tb-fw to MAKE_BL macro.
For example,

    index b97c048..a558d21 100644
    --- a/Makefile
    +++ b/Makefile
    @@ -586,8 +586,11 @@ $(eval $(call MAKE_BL,1))
     endif
     
     ifeq (${NEED_BL2},yes)
    +ifneq (${BL2_AT_EL3},1)
    +BL2_FIP_ARG := tb-fw
    +endif
     $(if ${BL2}, $(eval $(call MAKE_TOOL_ARGS,2,${BL2},tb-fw)),\
    -	$(eval $(call MAKE_BL,2,tb-fw)))
    +	$(eval $(call MAKE_BL,2,${BL2_FIP_ARG})))
     endif
     
     ifeq (${NEED_SCP_BL2},yes)

etienne-lms:
@danh-arm, fine with me to get the AArch32 support matured/tested/merged later...
And thanks for the feedback.

masahir0y:
+We assume 3 different types of BootROM support on the platform:
     +
     +#. The Boot ROM always jumps to the same address, for both cold
     +   and warm boot. In this case, we will need to keep a resident part
     +   of BL2 whose memory cannot be reclaimed by any other image. The
     +   linker script defines the symbols __TEXT_RESIDENT_START__ and
     +   __TEXT_RESIDENT_END__ that allows the platform to configure
     +   correctly the memory map.

Setting emulators aside,
do we have *real* hardware for this case?

Every vendor boot ROM should have similar mechanism to handle SMP, like BL1 does
(either secondary_cold_boot or COLD_BOOT_SINGLE_CPU)
Otherwise, the boot ROM is completely useless.

Talking about warn boot,
BL2 is supposed to be the first program loaded by boot ROM.
In modern SoC architecture, it will probably run in internal SRAM, not in DRAM.
In my understanding, the content in SRAM will be lost in low power state,
while DRAM is put into retention (self-refresh mode) during suspend-to-ram.
I do not to know why we need to protect the RESIDENT area.

This use-case sounds hypothetical to me.
I'd like to know the real hardware use-case, if any.

masahir0y:
+	 * We generate a NULL pointer here to allow a void* parameter
     +	 * in bl2_el3_early_platform_setup. It is not used at this moment.
     +	 */
     +	mov	x0, #0

You can save x0, like bl2_entrypoint.S, bl31_entrypoint.S do.

Perhaps, we can save more registers (x0-x3 is enough?)
because we never know the design of vendor boot ROM.

Then, those registers are passed to  bl2_el3_early_platform_setup()
as they are.

    void bl2_el3_early_platform_setup(register_t x0, register_t x1, register_t x2, register_t x3)

As for mbedtls_{common,crypto,x509}.mk
Maybe, we can leave BL1_SOURCES, BL2_SOURCES setup to platform.mk
This is what  xlat_tables.mk does.

robertovargas-arm:
@etienne-lms The problem is that BL2 has to load BL31/BL32, and they are executed at EL3/Monitor mode, so if we execute BL2 at EL2/SVC mode then we will not be able to jump to BL31/BL32 at EL3/Monitor mode. In the normal configuration BL2 executes a SMC that is served by BL1 which executes at EL3/Monitor mode. If we want to avoid having the full BL2 running at EL3/Monitor mode then we have to set a SMC dispatcher and do something similar to what BL1 does at this moment.

@masahir0y We discused about including or not BL2 in the fip for the reason that you comment, but at the end we thought that is better to keep it in the fip because the image is contiguous in the fip, so if the boot ROM can not parse it (what will be the most common case) then you only have to use the address where BL2 image begins, you don't have to parse the fip. If for some reason the platform actually needs BL2 out of the fip then it can do it in its own platform makefile.

I tried to remove BL1 build in the arm platforms removing BL1_SOURCES from arm_common.mk, but the problem is (as you already realized) that mbedtls_common.mk adds files to it. It shouldn't. We can add the ifdef that you added in mbedtls_common.mk because it doesn't break any build, although I think the correct solution should be to remove BL1_SOURCES definition from mbedtls_common.mk, but it would break all the current builds.

At this moment we are setting X0 to 0 because it is impossible to know how the boot ROM is going to pass the parameters. We can assume that it will do it in the first registers (the most logical decision), but we don't know if we can have different configurations. We thought about adding a platform hook like plat_get_params() at the very beginning of bl2_entrypoint, and maybe it is the best option to be sure that the boot parameters are in x0,x1,x2,x3.

We are  using DRAM because it was the easier solution. Since we don't know how the boot ROM is going to work, we have a lot of unknowledge here. If the boot ROM is going to jump always to the same address then we need a resident part of BL2, otherwise we will not be able to support warm_boot because the memory of BL2 would be reclaimed by other images later. It was really hard to have this resident part in SRAM in our platforms, mainly due to our current memory map, and doing a big change to achieve it was pointless because other platforms will not reuse any of our code, so we adopted the easiest solution that was loading it in DRAM.

If the boot ROM has some kind of mailbox where we can indicate where it has to jump for the warm_boot then we don't need the BL2 resident part because it can jump directly to BL31/BL32. In our case we don't have any boot ROM and we simulate this configuration modifying RVBAR.

hzhuang1:
This patch set relies on non-TF BootRom loading BL2 from storage device to SRAM first. But the problem is how to handle installing/upgrading firmware when storage device isn't ready yet. I mean that the storage device is still full of blank.

In HiKey/HiKey960, there're two parts of firmware. One is BL1+NS updater, and the other is FIP. BL2 is always in FIP.

If we want to switch to BL2_EL3, it seems that we need split BL2_EL3 from FIP. Is it right? But I find you keep BL2_EL3 in FIP. It'll break to upgrade firmware on HiKey/HiKey960.

masahir0y:
> We discused about including or not BL2 in the fip for the reason that you comment, but at the end we thought that is better to keep it in the fip because the image is contiguous in the fip, so if the boot ROM can not parse it (what will be the most common case) then you only have to use the address where BL2 image begins, you don't have to parse the fip.

How should the TOC offset be handled?
The boot ROM cannot change the address from which the next image is loaded since it is hard-weird.

> If for some reason the platform actually needs BL2 out of the fip then it can do it in its own platform makefile.

How?  platform.mk cannot filter-out the FIP arguments
because platform.mk is included before $(call MAKE_BL,2,...) is evaluated.


> I tried to remove BL1 build in the arm platforms removing BL1_SOURCES from arm_common.mk, but the problem is (as you already realized) that mbedtls_common.mk adds files to it. It shouldn't. We can add the ifdef that you added in mbedtls_common.mk because it doesn't break any build, although I think the correct solution should be to remove BL1_SOURCES definition from mbedtls_common.mk, but it would break all the current builds.

lib/xlat_tables_v2/xlat_tables.mk does not add source files to BL*_SOURCES.
platform.mk adds $(XLAT_TABLES_LIB_SRCS) by itself.

Do likewise for mbedtls.


> We thought about adding a platform hook like plat_get_params() at the very beginning of bl2_entrypoint,

This is a good solution, and what U-Boot does.
(save_boot_params in https://github.com/u-boot/u-boot/blob/master/arch/arm/cpu/armv8/start.S)


> and maybe it is the best option to be sure that the boot parameters are in x0,x1,x2,x3.

For maximum flexibility,
perhaps this platform hook can take arbitraty number of arguments by <stdarg.h>.


> If the boot ROM has some kind of mailbox where we can indicate where it has to jump for the warm_boot then we don't need the BL2 resident part because it can jump directly to BL31/BL32. In our case we don't have any boot ROM and we simulate this configuration modifying RVBAR.

Your statement "we don't have any boot ROM" is odd here
since the motivation for BL2-at-EL3 is to support non-TF boot ROM.
You can use normal boot sequence BL1 -> BL2 -> ... for your platform.

You are implementing BL1 code into bl2_el3_entrypoint.S   This is odd. 
We have a lot of unknowledge about the non-TF boot ROM, but we need assumption to some extent.
At least, it needs to be able to handle secondary cold boot by itself.
If not (i.e. all CPUs always jump to the same address), it is fine
to implement another image before BL2-at-EL3, but this is
probably the case where we should choose BL1 and BL2.

I still believe this is very unlikely in real products.
Please do not introduce unnecessary complexity.

masahir0y:
@hzhuang1 

> If we want to switch to BL2_EL3, it seems that we need split BL2_EL3 from FIP. Is it right? But I find you keep BL2_EL3 in FIP. It'll break to upgrade firmware on HiKey/HiKey960.

If I understood correctly, ARM folks suggested to have two copies of BL2_EL3.

You can flash a bare BL2_EL3 image to the first part so that the Boot ROM can load it.
You have another copy of BL2_EL3 in FIP, but it is never used.

I do not like it, though.

robertovargas-arm:
> How should the TOC offset be handled?
> The boot ROM cannot change the address from which the next image is loaded since it is hard-weird.

As I explained, this case is one of the cases that we covered. If this is your case then you will need BL2 out of the fip, or maybe you can locate the fip in the position that will place the internal BL2 in the desired position. This is something that should be taken by the platform.

> How? platform.mk cannot filter-out the FIP arguments
> because platform.mk is included before $(call MAKE_BL,2,...) is evaluated.

True. Then, if we want to give to the platform the possibility of removing BL2 from the fip then we have to modify the top level makefiles.

> lib/xlat_tables_v2/xlat_tables.mk does not add source files to BL*_SOURCES.
> platform.mk adds $(XLAT_TABLES_LIB_SRCS) by itself.
> 
> Do likewise for mbedtls.

Yes, but if we move the decision to the platforms then we have to update all the makefiles of the platforms, and we don't have access to all of them.

> For maximum flexibility,
> perhaps this platform hook can take arbitrary number of arguments by <stdarg.h>.

It is really hard to implement it. In that case the hook has to return a va_list object, and it means that the function, that has to be coded in assembler because we don't have a c runtime properly initialized at that moment, has to reproduce the rules of the ABI for va_lists, and these rules are very complex.

> You are implementing BL1 code into bl2_el3_entrypoint.S This is odd.
> We have a lot of unknowledge about the non-TF boot ROM, but we need assumption to some extent.
> At least, it needs to be able to handle secondary cold boot by itself.
> If not (i.e. all CPUs always jump to the same address), it is fine
> to implement another image before BL2-at-EL3, but this is
> probably the case where we should choose BL1 and BL2.

BL2 and BL31 need some initialization that is done by BL1, if you don't have BL1 then this work has to be done by BL2 at EL3.

> If I understood correctly, ARM folks suggested to have two copies of BL2_EL3.
> 
> You can flash a bare BL2_EL3 image to the first part so that the Boot ROM can load it.
> You have another copy of BL2_EL3 in FIP, but it is never used.

No, you understood it wrong. We only said that we don't know how the platform is going to load BL2, and this is why we wanted to let this decision to the platform.

You discovered that with the current makefiles it is not possible to do that. I think it is going to be easier to remove BL2 from the fip. I am going to try that.

soby-mathew:
@etienne-lms, Thanks for your comments.

> Current implementation would make BL2 fully executing in the AArch32 Monitor mode of the CPU which is not handy in case of sync exceptions (aborts). 

Ah, I see the problem you refer to. But isn't this a problem in AArch32 BL1 as well ? We do install the same vector table in MVBAR and S-VBAR and since both monitor mode and S-SVC have the same address space, shouldn't that take care of the issue?

soby-mathew:
@masahir0y, Thanks for your comments.

>Every vendor boot ROM should have similar mechanism to handle SMP, like BL1 does
(either secondary_cold_boot or COLD_BOOT_SINGLE_CPU)

Yes, this is the expected case in which case, the whole of BL2 memory can be reclaimed by the runtime firmware (BL31, BL32). You don't have to keep any part of BL2 resident. 

Keeping BL2 resident is only applicable if mailbox in BL2 is needed for some reason. You are right in that it is more of a hypothetical case and unlikely in a real world BootROM. 

> Perhaps, we can save more registers (x0-x3 is enough?)
because we never know the design of vendor boot ROM.

Yes, this is a useful suggestion and we can look how to make it more flexible.

>You are implementing BL1 code into bl2_el3_entrypoint.S This is odd.

Mailbox is not the only implementation in that code. As Roberto says we need to execute the EL3 set up and the necessary errata workarounds for the primary CPU because we can't rely on the setup done by non-TF Bootrom. Hence some bits of BL1 need to be pulled into BL2.

soby-mathew:
Hi @hzhuang1,

> If we want to switch to BL2_EL3, it seems that we need split BL2_EL3 from FIP. Is it right?

If you using BL1 in your platform, you mustn't use BL2_EL3 config. This is only meant for platforms **not** using BL1.

hzhuang1:
>If you are using BL1 in your platform, you mustn't use BL2_EL3 config. This is only meant for platforms not using BL1.

Yes, BL2_EL3 is plan to replace pseudo BL1 that is used on hikey/hikey960 platform. My question is focus on BL2_EL3 should be moved out of FIP, since BL2_EL3 + NS updater should work when storage device isn't ready. And I think that @robertovargas-arm is working on it now.

>This PR might help streamline the boot flow on your platforms. It should make it possible to branch directly from your ROM code to TF BL2 executing at EL3, without the need for a TF "pseudo-BL1". Let us know if you have any feedback.

Yes, I won't build BL1 & BL2_EL3 together. When BL2_EL3 is ready, I'll replace BL1 by BL2_EL3 on HiKey/HiKey960 platform.

soby-mathew:
@hzhuang1, I understand. Thanks for the clarification. Yes, Roberto is working on the patch for not including BL2_EL3 in FIP.

robertovargas-arm:
@hzhuang1 @masahir0y @soby-mathew 
I have added a  new commit 'bl2-el3: Don't include BL2 in fip for BL2 at EL3' which removes bl2 from fip when BL2_AT_EL3 is defined. I am not addressing the problems with the parameters of the boot rom because it will be addressed later with dynamic config. Does this patch solve your problem?

masahir0y:
> I am not addressing the problems with the parameters of the boot rom because it will be addressed later with dynamic config. Does this patch solve your problem?

I have not tested, but I notice most of my comments were ignored.
What happened to my first comment?  What about the build error of Trusted Board Boot?

I am still not convinced by the resident part added to support a hypothetical case.

davidcunado-arm:
jenkins: test this please

davidcunado-arm:
Checks failure - looks to have been a temp infrastructure issue as re-run passed.

@masahir0y , @hzhuang1 
From my perspective this is ready to merge - please let me know asap if you have any concerns, otherwise I'll aim to merge on FRI 19-JAN.
Thanks!

masahir0y:
+
     + * Mandatory BL2 at EL3 functions: Must be implemented if BL2_AT_EL3 image is
     + * supported
     +void bl2_el3_early_platform_setup(u_register_t arg1, u_register_t arg2,
     +				  u_register_t arg3, u_register_t arg4);

Please arg0-arg3 instead of arg1-arg4.
They correspond to x0-x3 or r0-r3 from the previous boot loader.

Or, reg0-reg3 if you want to self-document that they are registers passed
from the previous loader.


     +Function : bl2\_el3\_early\_platform\_setup() [mandatory]
     +~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     +
     +::
     +	Argument : void *
     +	Return   : void

You need to update domentation accordingly.


     +	/* Save arguments x0-x3 from previous Boot loader */
     +	mov	x20,x0
     +	mov	x21,x1
     +	mov	x22,x2
     +	mov	x23,x3

Nit:
Please add a space after ','
I see more other occurrences.


    @@ -172,5 +172,9 @@ endif
     # Add support for platform supplied linker script for BL31 build
     $(eval $(call add_define,PLAT_EXTRA_LD_SCRIPT))
     
     +ifneq (${BL2_AT_EL3}, 0)
     +    override BL1_SOURCES =
     +endif
     +

Why 'override' directive here?

'override' is used to prevent the command line from overriding variables.
Do you have any chance to give BL1_SOURCE from the command line?



      /* Reset fn is needed in BL at reset vector */
     -#if defined(IMAGE_BL1) || defined(IMAGE_BL31)
     +#if defined(IMAGE_BL1) || defined(IMAGE_BL31) || (defined(IMAGE_BL2) && BL2_AT_EL3)
      	/*
  	 * The reset handler common to all platforms.  After a matching
  	 * cpu_ops structure entry is found, the correponding reset_handler

Unmatch comment in the endif part.
Update or remove the comment.
Please check other occurrences.


     +
     +    ASSERT(__CPU_OPS_END__ > __CPU_OPS_START__,
     +          "cpu_ops not defined for this platform.")
     +

CPU_OPS really mandatory for BL2-EL3?
BL2-EL3 is never run as the first image.

In other words, the CPU has been already initialized enough
to run the vendor ROM and to load BL2-EL3 image into memory.

BL31 will do the same thing after this.
Is it necessary here in BL2-EL3?


    --- a/include/common/asm_macros_common.S
    +++ b/include/common/asm_macros_common.S
    @@ -29,7 +29,7 @@
             * debugging experience.
             */
            .cfi_sections .debug_frame
    -       .section .text.\_name, "ax"
    +       .section .text.asm.\_name, "ax"
            .type \_name, %function
            .func \_name
            /*

All assembly code is resident. This seems too rough assumption to me.
Which code is really needed in the resident part?

robertovargas-arm:
> Please arg0-arg3 instead of arg1-arg4.
> They correspond to x0-x3 or r0-r3 from the previous boot loader.

Done.

> Nit:
> Please add a space after ','

Done.

> Why 'override' directive here?
>
> 'override' is used to prevent the command line from overriding variables.
> Do you have any chance to give BL1_SOURCE from the command line?

It also may be used to override definitions done in other places. BL1_SOURCES has to be removed if we want to avoid building BL1, and adding an ifdef in every place where BL1_SOURCES is defined is not a good idea. TF is using this pattern in several places, for example in ./plat/arm/board/juno/platform.mk:43. The important point here is that the override is using '=' and not ':='.

> Unmatch comment in the endif part.
> Update or remove the comment.
> Please check other occurrences.

Done.

> CPU_OPS really mandatory for BL2-EL3?

Yes, because reset_handler is included, and  for this reason we need the errata framework to be sure what is the state of the CPU at that moment.

> All assembly code is resident. This seems too rough assumption to me.
> Which code is really needed in the resident part?

Yes, you are right. Including all the code defined in assembler files is a bit rough, but tracking the calling graph could be a very complex task and we would need some call graph tool incorporated to our build system. At this moment, all the code defined in assembly functions is less than 1 page (I think it is even less than 2KB), so we are not going to have any benefit trying a finer solution. If in the future we need a better solution then we can come back and add that complexity.

hzhuang1:
Excuse me on response late. I'm fine on this.

davidcunado-arm:
@masahir0y @hzhuang1 
Thanks! Ok looks like we're good-to-go - I'll merge this on FRI morning.

masahir0y:
+	Argument : u_register_t, u_register_t, u_register_t, u_register_t
     +	Return   : void
     +
     +This function executes with the MMU and data caches disabled. It is only called
     +by the primary CPU. The argument to this function is a void pointer which can
     +be used by the platform to pass any needed information from the Boot ROM to
     +BL2.

You also need to update the description.


> It also may be used to override definitions done in other places. BL1_SOURCES has to be removed if we want to avoid building BL1, and adding an ifdef in every place where BL1_SOURCES is defined is not a good idea. TF is using this pattern in several places, for example in ./plat/arm/board/juno/platform.mk:43. 

OK.  Understood.

> The important point here is that the override is using '=' and not ':='.

I agree '=' is better than ':=", but is it so different in terms of performance?
In order to suppress the evaluation of the right-hand side of
following BL1_SOURCES += (some expression)
?

> Yes, because reset_handler is included, and for this reason we need the errata framework to be sure what is the state of the CPU at that moment.

If you re-use el3_entrypoint_common as-is,
right, you cannot skip reset_handler.
But, is the reset (and errata workaround) really necessary here?

As I said, the BL2(-EL3) is not the first loader.
If the CPU has reached BL2(-EL3), it means the CPU is working well enough.

Given that the boot ROM has succeeded in loading BL2-EL3 into memory,
BL2-EL3 can load BL31, 32, 33 as well.



> Yes, you are right. Including all the code defined in assembler files is a bit rough, but tracking the calling graph could be a very complex task and we would need some call graph tool incorporated to our build system. At this moment, all the code defined in assembly functions is less than 1 page (I think it is even less than 2KB), so we are not going to have any benefit trying a finer solution. If in the future we need a better solution then we can come back and add that complexity.

I hope you decide to remove the complexity
by dropping the resident part entirely.

If you want to test BL2-EL3 but your platform does not have own ROM,
can you implement an emulation that loads BL2-EL3 and jumps to it?

robertovargas-arm:
> You also need to update the description.

Done.

> I agree '=' is better than ':=", but is it so different in terms of performance?

I meant that the important part was '=' against ':=' due to the moment where they are evaluated.

> If you re-use el3_entrypoint_common as-is,
> right, you cannot skip reset_handler.
> But, is the reset (and errata workaround) really necessary here?

The errata workarounds are still needed if you want a correct cold boot. We cannot trust the Boot ROM to have done the correct errata workarounds.

> If you want to test BL2-EL3 but your platform does not have own ROM,
> can you implement an emulation that loads BL2-EL3 and jumps to it?

Yes. In our case we have done it with FVP. What we do is to load directly the bl2 image in the position where it is linked and then we change the value of the RVBAR register. After it, the model jumps to our entrypoint in BL2.

masahir0y:
> The errata workarounds are still needed if you want a correct cold boot. We cannot trust the Boot ROM to have done the correct errata workarounds.

The errata workarounds are patched by BL31.
Why do you need to do the same things again and again?

I am copy-pasting my previous comments:
   "As I said, the BL2(-EL3) is not the first loader.
   If the CPU has reached BL2(-EL3), it means the CPU is working well enough.

   Given that the boot ROM has succeeded in loading BL2-EL3 into memory,
   BL2-EL3 can load BL31, 32, 33 as well."


> Yes. In our case we have done it with FVP. What we do is to load directly the bl2 image in the position where it is linked and then we change the value of the RVBAR register. After it, the model jumps to our entrypoint in BL2.

This is not what I wanted you to answer.

Can you emulate the boot ROM _with_ _reality_?
i.e. is it possible to load BL2-EL3 into memory by the Boot ROM
instead of putting it into memory directly by the command line?

    +This functionality can be tested with FVP loading the image directly
    +in memory and changing the address where the system jumps at reset.
    +For example:
    +
    +	-C cluster0.cpu0.RVBAR=0x4014000
    +	--data cluster0.cpu0=bl2.bin@0x4014000
    +
    +With this configuration, FVP is like a platform of the first case,
    +where the Boot ROM jumps always to the same address. For simplification,
    +BL32 is loaded in DRAM in this case, to avoid other images reclaiming
    +BL2 memory.


Real world boot ROM cannot do this.
My gut feeling is that the first case with the resident part
is for the hypothetical case you created.

If you implement more realistic boot flow, can you remove the first case?

hzhuang1:
How could I jump to NS BL2U from BL2_EL3?

When BL1 is working, I could implement it in bl1_plat_set_ep_info(). When BL2_EL3 is enabled, I could only find bl2_load_images(). Does it mean that I have to implement NS BL2U in fip.bin?

davidcunado-arm:
@masahir0y 

> I am copy-pasting my previous comments:
> "As I said, the BL2(-EL3) is not the first loader.
> If the CPU has reached BL2(-EL3), it means the CPU is working well enough.
> Given that the boot ROM has succeeded in loading BL2-EL3 into memory,
> BL2-EL3 can load BL31, 32, 33 as well."

True, but there may be errata that ROM does not apply that BL2-EL3 needs to apply. This is a belts-and-braces approach. There is negligible cost for including this in the binary.

> Can you emulate the boot ROM with reality?
> i.e. is it possible to load BL2-EL3 into memory by the Boot ROM
> instead of putting it into memory directly by the command line?

Our expectation is that for the use case this PR is helping address, the platform has a legacy boot-ROM which some platform specific initialisation and then loads and jumps to BL2-EL3, which then continues the boot flow.

> Real world boot ROM cannot do this.

It's not expected to. We're using the FVP in this way is an alternative to a 'dummy' boot-ROM that would simply jump to the BL2-EL3 address (as we'd have no platform specific initialisation to do). 

It's not clear to me what value there would be in creating such a 'dummy' boot-ROM?

> My gut feeling is that the first case with the resident part
> is for the hypothetical case you created.

The resident part is there for platforms that want to take advantage of BL2-EL3 but does not implement a programmable mailbox for warm-reset - e.g. some AArch32 / ARMv7 platforms - in which case BL2-EL3 needs to remain resident.

If a platform does not need such a resident part, then it doesn't need to use it. But the solution has provided that flexibility.

@hzhuang1 
Unfortunately, the firmware updated support requires use of (ARM TF version) BL1. The BL2-EL3 solution does not support firmware update.

The expectation is that platforms wanting to use BL2-EL3 already have a legacy boot-ROM that provides firmware update support (if that use case is needed).

hzhuang1:
@davidcunado-arm Oh, it means that I have to keep using BL1. Thanks.

masahir0y:
@hzhuang1 

> Oh, it means that I have to keep using BL1. Thanks.

IIUC, the firmware updater in TF
is for platforms whose BL1 is real hard-wired ROM (i.e. NS_BL1U is also hard-wired).

Your platform has the vendor boot ROM, then BL1 is a pseudo ROM that runs in RAM, right?

I am not familiar with hikey board, but if the on-board data storage is blank, even BL1 is not there.
The 96 board specification requires the system can be restored from a complete brick state.
So, you need to flash images by other means.  Why do you need to rely on the TF FW updater?

hzhuang1:
> IIUC, the firmware updater in TF
> is for platforms whose BL1 is real hard-wired ROM (i.e. NS_BL1U is also hard-wired).
>
> Your platform has the vendor boot ROM, then BL1 is a pseudo ROM that runs in RAM, right?

Yes, BL1 is a pseudo ROM that runs in RAM in HiKey.
>
> I am not familiar with hikey board, but if the on-board data storage is blank, even BL1 is not there.
> The 96 board specification requires the system can be restored from a complete brick state.
> So, you need to flash images by other means. Why do you need to rely on the TF FW updater?

The real ROM could only download a few images into RAM with small size. I have to flush it TF FW updater. If I want to download & flush large image (such as kernel, system, rootfs), I have to implement it in TF FW updater.

By the way, the NS BL1U is a separate image from ATF.

soby-mathew:
@hzhuang1, is it possible perhaps to have 2 different boot sequence depending on whether Firmware Update is required or not ?

i.e
Regular Boot : BootROM -> BL2_EL3 -> BL33 -> Linux
Firmware Update : BootROM -> BL1 -> NS_BL1U

danh-arm:
The reason that NS_BL2U functionality is normally kept separate from BL1 Firmware Update (FWU) functionality is to minimise the amount of secure ROM code. If FWU needs to be in BL2 instead (as in the HiKey platform) then there is no strong reason for the separation; NS_BL2U and BL2_EL3 could be merged within HiKey platform code.

Alternatively, given that NS_BL2U has already been separated in HiKey, then the BL1 FWU functionality could potentially be abstracted out into common code, to allow platform specific BL2_EL3 code to pull this in. This would only require a small change to existing generic code. We wouldn't want FWU to be included in generic BL2_EL3 code, though. 

Soby's solution could work too.

PR #1204, opened 2017-12-20
Add padding at the end of the last entry

davidcunado-arm:
This patch adds padding bytes at the end of the last image in the
fip to be able to transfer by DMA the last image.

Change-Id: I8c6f07dee389cb3d1dc919936d9d52841d7e5723
 
docs: Update the ToC end marker description in the document
jtzhou2016 committed with davidcunado-arm on 24 Nov 2017

Change-Id: I2e29a63f08aed3b8ea0bb10170a3d55b8d033e62
Signed-off-by: Jett Zhou <jett.zhou@arm.com>

### DIFF
--- a/tools/fiptool/fiptool.c
+++ b/tools/fiptool/fiptool.c
@@ -492,7 +492,7 @@ static int pack_images(const char *filename, uint64_t toc_flags, unsigned long a
 	fip_toc_header_t *toc_header;
 	fip_toc_entry_t *toc_entry;
 	char *buf;
-	uint64_t entry_offset, buf_size, payload_size = 0;
+	uint64_t entry_offset, buf_size, payload_size = 0, pad_size;
 	size_t nr_images = 0;
 
 	for (desc = image_desc_head; desc != NULL; desc = desc->next)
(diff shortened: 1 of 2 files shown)


davidcunado-arm:
@Leo-Yan @hzhuang1 
Please can you take a look at this PR and confirm that it resolves the issue that #1166 was targeted at?
Thanks!

tools/fiptool/fiptool.c
```diff
@@ -528,7 +528,7 @@ static int pack_images(const char *filename, uint64_t toc_flags, unsigned long a
 
 	/* Append a null uuid entry to mark the end of ToC entries. */
 	memset(toc_entry, 0, sizeof(*toc_entry));
-	toc_entry->offset_address = entry_offset;
+	toc_entry->offset_address = (entry_offset + align - 1) & ~(align - 1);
```
geesun: It's better to add some comments here, the last toc_entry offset address is the value of fip size. It will make sure the next change will not break the rule.
Thanks.

geesun:
It is better to update the document to address the Toc end marker contains the FIP size in entry_offset. 
Thanks.

Leo-Yan:
Hi @davidcunado-arm, I verified this patch and this patch can fix the alignment issue on Hikey.

After this patch has been merged, we could merge the patch for Hikey/Hikey960 FIP_ALIGN configuration. I will consolidate with @hzhuang1 if can reduce the FIP_ALIGN from 512B to 64B.

davidcunado-arm:
@geesun 
 I've updated the PR to address your comment, and also added the update to firmware-design doc that Jett did.

tools/fiptool/fiptool.c
```diff
@@ -555,6 +559,13 @@ static int pack_images(const char *filename, uint64_t toc_flags, unsigned long a
 		xfwrite(image->buffer, image->toc_e.size, fp, filename);
 	}
 
+	if (fseek(fp, entry_offset, SEEK_SET))
+		log_errx("Failed to set file position");
+
+	pad_size = toc_entry->offset_address - entry_offset;
```
afaerber: This introduced a use-after-free, addressed by #1242.
Any chance suitable regression tests can be added to your Jenkins to avoid this happening again?

sandrine-bailleux-arm: Hi @afaerber 

> Any chance suitable regression tests can be added to your Jenkins to avoid this happening again?

Thanks for the suggestion, we'll look into running fiptool and cert_create through Valgrind in Jenkins. We might not get around to do it next month but we'll try to schedule this work soon.

Regards,
Sandrine

PR #1210, opened 2017-12-26
BL31: Output error log for friendly debugging

Leo-Yan:
If use assert() function it's easily to check condition is valid or not,
though the code is simple but we lose some debugging info from log.

This commit detects if it's the invalid index then it outputs more
debugging info, e.g. it can output priority number, index number,
interrupt number, so we can get more useful debugging info.

### DIFF
--- a/bl31/ehf.c
+++ b/bl31/ehf.c
@@ -52,7 +52,6 @@ static int pri_to_idx(unsigned int priority)
 
 	idx = EHF_PRI_TO_IDX(priority, exception_data.pri_bits);
 	assert((idx >= 0) && (idx < exception_data.num_priorities));
-	assert(IS_IDX_VALID(idx));
 
 	return idx;
 }
@@ -116,6 +115,11 @@ void ehf_activate_priority(unsigned int priority)
 	 */
 	cur_pri_idx = get_pe_highest_active_idx(pe_data);
 	idx = pri_to_idx(priority);
+	if (!IS_IDX_VALID(idx)) {
+		ERROR("Invalid index : req=0x%x idx=0x%x\n", priority, idx);
+		panic();
+	}
+
 	if ((cur_pri_idx != EHF_INVALID_IDX) && (idx >= cur_pri_idx)) {
 		ERROR("Activation priority mismatch: req=0x%x current=0x%x\n",
 				priority, IDX_TO_PRI(cur_pri_idx));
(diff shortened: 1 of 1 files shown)


bl31/ehf.c
```diff
@@ -408,14 +417,19 @@ static uint64_t ehf_el3_interrupt_handler(uint32_t id, uint32_t flags,
...
+	}
 
 	/* Validate priority */
 	assert(pri == IDX_TO_PRI(idx));
 
 	handler = RAW_HANDLER(exception_data.ehf_priorities[idx].ehf_handler);
 	if (!handler) {
-		ERROR("No EL3 exception handler for priority 0x%x\n",
-				IDX_TO_PRI(idx));
+		ERROR("No EL3 exception handler for intr 0x%x priority 0x%x\n",
```
jeenu-arm: PR #1209 suggests that the change is prompted by your receiving unconfigured interrupts. For that case, wouldn't this change (i.e. including the interrupt number) only suffice? Why alther explicit activations and deactivations for debugging output?

bl31/ehf.c
```diff
@@ -52,7 +52,6 @@ static int pri_to_idx(unsigned int priority)
 
 	idx = EHF_PRI_TO_IDX(priority, exception_data.pri_bits);
 	assert((idx >= 0) && (idx < exception_data.num_priorities));
-	assert(IS_IDX_VALID(idx));
```
jeenu-arm: At this point, we have both the index an prioirty values here, so if at all, why not print the information here rather than duplicating error messages from the caller?

Leo-Yan: At here we only can print out index value, but cannot print the interrupt number.

bl31/ehf.c
Leo-Yan: Yes, the change in function ehf_el3_interrupt_handler() is the main purpose. But all these three functions have called pri_to_idx(), we can change pri_to_idx() function so can let the program to run to here and print out interrupt number.

davidcunado-arm:
@Leo-Yan 
From @jeenu-arm review, it's not clear what the rationale for this change is? Please could you provide some more background?

If the issue is that you'd like the index as well as the interrupt number logged when the assert in pri_to_idx fires, then the patch should look to update the assert - e.g. could replace the assert with some code guarded by ENABLE_ASSERTIONS.

davidcunado-arm:
@Leo-Yan 
I'm going to close this PR - please re-open if needed.

PR #1214, opened 2018-01-04
Workarounds for CVE-2017-5715 on Cortex A57/A72/A73 and A75

dp-arm:


vwadekar:
Acked-by: Varun Wadekar <vwadekar@nvidia.com>

afaerber:
For reference I have backported the three patches to v1.4 here:
https://build.opensuse.org/package/show/hardware:boot/arm-trusted-firmware

afaerber:
Also possibly helpful to readers, I've begun tracking some of the non-mainline forks of ATF for potential backports here:
https://bugzilla.suse.com/show_bug.cgi?id=1074741

prime-zeng:
Why not disable Dcache to solve this problem?

danh-arm:
> Why not disable Dcache to solve this problem? I don't think we have too much data to process in ATF

Sorry, I'm not sure what you're proposing. I guess you don't mean permanently disable the D$ in TF as that would catastrophically affect performance?

iwishguo:
Hi,

There was another patch to enable invalidating BTB when we saw this pull request several days ago, but we can't find it now. Is this patch dropped? If it is, could you tell us what the reason is?
Thanks,

Gary

danh-arm:
> There was another patch to enable invalidating BTB when we saw this pull request several days ago, but we can't find it now. Is this patch dropped? If it is, could you tell us what the reason is?

Sorry about that. Yes, this patch is dropped, although we didn't intend to do that until the final version of this PR. The reason is that there is an A57 erratum (833069) that when "Disabling MMU Translation with CPUACTLR_EL1 "Enable Invalidates of BTB" bit set can cause Invalidate by PA or VA to fail". Also, we now believe there is no requirement for higher level software to induce BTB invalidation by using the `IC IALLU` instruction, since other methods are available and preferred.

We believe this version of the PR is working correctly, but we are still trying to understand why the context overlaying in the previous version didn't work, and to optimize this PR if possible.

I will update the TFV-6 security advisory so it is aligned with this PR.

vsiles:
> Also, we now believe there is no requirement for higher level software to induce BTB invalidation by using the IC IALLU instruction, since other methods are available and preferred.

If we are talking about Variant 2, could you point me to these other methods that are available and preferred for EL2 (aarch64) and EL1S (aarch32/aarch64) ? Since `BPIALL` is no longer available in aarch64 and I can't easily shut off and on the MMU at EL1S, I'm not sure how to address the problem at these levels.

danh-arm:
> If we are talking about Variant 2, could you point me to these other methods that are available and preferred for EL2 (aarch64) and EL1S (aarch32/aarch64) ? Since BPIALL is no longer available in aarch64 and I can't easily shut off and on the MMU at EL1S, I'm not sure how to address the problem at these levels.

If you can't easily shut off and on the MMU (the preferred solution for A57 and A72), then you can call into firmware or the Trusted OS to do the branch predictor invalidation for you. With this PR, you can call any SMC to induce branch predictor invalidation in TF at EL3. `PSCI_VERSION` is used by the Linux kernel for this purpose.

In future, we are looking to define a specific SMC for this branch predictor invalidation, which will be discoverable by the normal world. Details of this are TBC and will follow merging of this PR.

lib/cpus/aarch64/workaround_cve_2017_5715_bpiall.S
```diff
@@ -0,0 +1,372 @@
...
+	         MODE32_svc << MODE32_SHIFT)
+
+	/*
+	 * VBAR_EL3 points to vbar1 which is the vector table
+	 * used while the workaround is executing.
+	 */
+	.quad	workaround_bpiall_vbar1_runtime_exceptions
+
+	/* Setup SCTLR_EL1 with MMU off and I$ on */
+	.quad	SCTLR_AARCH32_EL1_RES1 | SCTLR_I_BIT
```
dp-arm: Enabling the instruction cache in S-EL1 is the only difference between this patch and the previous one.  This change gave a measurable performance boost of about 10%.

danh-arm:
Note this latest version of the PR adds another patch to use a PFR0 ID field to determine whether to apply the variant 2 workaround on Cortex-A75. This means the workaround will be skipped in a future fixed hardware revision.

> When do you plan to merge this PR?
We intend this to be the final version and will merge later today if there are no other comments.

davidcunado-arm:
jenkins: test this please

davidcunado-arm:
NOTE - check failure is minor issue in static analysis where the quad-cluster-core version of BL31 is exceeding RAM limits; is has been resolved in an earlier patch but it has not been merged into master yet.

davidcunado-arm:
jenkins: test this please

davidcunado-arm:
jenkins: test this please

vwadekar:
Thanks David.

davidcunado-arm:
No probs! Patch is now in master - enjoy! :)

PR #1215, opened 2018-01-09
spd: tlkd: support for "NS memory ranges" function ID

vwadekar:
This patch adds support to receive function ID with NS world's
memory ranges to provide the memory snapshot to TLK.


davidcunado-arm:
jenkins: test this please

PR #1218, opened 2018-01-10
xlat v2: Correctly unmap regions on map error

antonio-nino-diaz-arm:
`mm_cursor` doesn't have the needed data because the `memmove()` that is called right before it overwrites that information. In order to get the information of the region that was being mapped, `mm` has to be used instead (like it is done to fill the fields of `unmap_mm`).

If the incorrect information is read, this check isn't reliable and `xlat_tables_unmap_region` may be requested to unmap memory that isn't mapped at all, triggering assertions.


davidcunado-arm:
jenkins: test this please

PR #1219, opened 2018-01-10
SPM: Fix version header definitions

antonio-nino-diaz-arm:
Rename SP_VERSION macros to MM_VERSION, which is the name used in the MM specification [1]. Also, a few more helper macros have been added.

MM-specific definitions have been moved to their own header file.

[1] http://infocenter.arm.com/help/topic/com.arm.doc.den0060a/DEN0060A_ARM_MM_Interface_Specification.pdf


davidcunado-arm:
jenkins: test this please

PR #1220, opened 2018-01-10
Makefile: Use ld.bfd linker if available

jwerner-chromium:
Some toolchain distributions install both the BFD and GOLD linkers under
the names [target]-ld.bfd and [target]-ld.gold. [target]-ld will then be
a symlink that may point to either one of these.

Trusted Firmware should always be linked with the BFD linker, since GOLD
is meant primarily for userspace programs and doesn't support many of
the more obscure linker script features that may be needed for firmware.
With this patch the Makefile will auto-detect if ld.bfd is available and
use it explicitly in that case.

Change-Id: I7017055f67db3bd57d191d20a7af06ca646937d7

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -126,6 +126,12 @@ OD			:=	${CROSS_COMPILE}objdump
 NM			:=	${CROSS_COMPILE}nm
 PP			:=	${CROSS_COMPILE}gcc -E
 
+# Use ${LD}.bfd instead if it exists (as absolute path or together with $PATH).
+ifneq ($(strip $(wildcard ${LD}.bfd) \
+	$(foreach dir,$(subst :, ,${PATH}),$(wildcard ${dir}/${LD}.bfd))),)
+LD			:=	${LD}.bfd
+endif
+
 ifeq (${ARM_ARCH_MAJOR},7)
 target32-directive	= 	-target arm-none-eabi
 # Will set march32-directive from platform configuration


Makefile
```diff
@@ -145,6 +145,13 @@ TF_CFLAGS_aarch32	=	$(march32-directive)
 TF_CFLAGS_aarch64	=	-march=armv8-a
 endif
 
+# Check if ld.bfd exists directly (if it's an absolute path) or is in PATH (if
+# it's just a basename). whereis needs trailing dot because it cuts extensions.
+ifeq ($(shell test -x ${LD}.bfd || whereis ${LD}.bfd. | grep -q / ; echo $$?),0)
```
danh-arm: We try to avoid direct calls into the shell for compatibility with Windows environments. There's an abstraction layer in build_env.mk to help with this sort of thing.

jwerner-chromium: Oh, okay. It doesn't offer the kind of primitives I need for this, unfortunately. And I think whereis doesn't even exist on Windows.

But gold doesn't support non-ELF platforms anyway, so this patch isn't necessary for any of the non-Unix environments. Would it be okay if I just guard this code with ifdef UNIX_MK?

danh-arm: I'd prefer this check to be abstracted into a new command in build_env.mk, with this implementation moved to unix.mk, and the windows.mk version always returning false.

jwerner-chromium: Actually, I figured out that it's not that hard to just reimplement the functionality of whereis within make, since we have the $PATH variable available from the environment. I'll do that and then it doesn't need any shell commands.

davidcunado-arm:
jenkins: test this please

Makefile
```diff
@@ -126,6 +126,13 @@ OD			:=	${CROSS_COMPILE}objdump
 NM			:=	${CROSS_COMPILE}nm
 PP			:=	${CROSS_COMPILE}gcc -E
 
+# Use ${LD}.bfd instead if it exists (as absolute path or together with $PATH).
+ifneq ($(strip $(wildcard ${LD}.bfd) \
+	$(foreach dir,$(subst :, ,${PATH}),$(wildcard ${dir}/${LD}.bfd))),)
+LD			:=	${LD}.bfd
+TF_CFLAGS		+=	-fuse-ld=bfd
```
davidcunado-arm: @jwerner-chromium 
Apologies for the delay... this is breaking our clang / armclang builds. 

clang: error: argument unused during compilation: '-fuse-ld=bfd' [-Werror,-Wunused-command-line-argument]

I tested a noddy fix which was to wrap lines 129-134 in another ifneq that checks if CC is clang or armclang and that seemed to work. But not sure if this is the 'right' solution!

Build options being used:
clang: CROSS_COMPILE=aarch64-linux-gnu- V=1 DEBUG=1 PLAT=fvp CC=clang
armclang: CROSS_COMPILE=aarch64-linux-gnu- V=1 DEBUG=1 PLAT=fvp CC=<path/to/armclang/6.8/installation>/bin/armclang

jwerner-chromium: Ahh... hmm... if I understand this right, it actually understands the -fuse-ld option but it complains when it is passed in an invocation that doesn't really link the binary (i.e. with -c).

It looks like Trusted Firmware never actually calls $(CC) without -c, it always calls $(LD) directly for linking. So I think I can just remove the -fuse-ld part completely. Will update.

davidcunado-arm:
jenkins: test this please

PR #1221, opened 2018-01-11
emmc/dw_mmc: fix the assert on HLE bit

hzhuang1:
When check HLE bit in interrupt register, it should check whether
HLE bit is set, not clear.


davidcunado-arm:
jenkins: test this please

PR #1226, opened 2018-01-16
utils: rename struct mem_region_t to struct mem_region

masahir0y:


masahir0y:
BTW,  I see "typedef struct foo foo_t;"
everywhere in ATF code.

What's the purpose of it?
Do you want to save some characters?

checkpatch.pl from Linux apparently warns
WARNING: do not add new typedefs

See 
https://github.com/torvalds/linux/blob/master/Documentation/process/coding-style.rst
 5) Typedefs

Let's stop this crap, shall we?

dp-arm:
> Do you want to save some characters?

Yes.

danh-arm:
> BTW, I see "typedef struct foo foo_t;"
> everywhere in ATF code.
> What's the purpose of it?

The main pattern we want to avoid is anonymous struct names, which can result in too much header nesting due to having to pull in the full declaration instead of forward declaring:

https://github.com/ARM-software/arm-trusted-firmware/wiki/ARM-Trusted-Firmware-Coding-Guidelines#avoid-anonymous-typedefs-of-structsenums-in-header-files

We don't feel as strongly as Linux about not using typedefs at all. It's up to the individual whether they want to typedef their structs or not. The pattern that's emerged is the one you see.

robertovargas-arm:
> Obviously, we have a good reason to use "struct foo" in header files.
> On the other hand, saving 5 characters is not a good reason to use "foo_t".
> It is really unreadable.

The readability is something that depends of the eyes that read. For me, having very long declarations (specially in functions) due to structs makes the code totally unreadable. With this kind of stuff, is a question of getting used to it.

davidcunado-arm:
jenkins: test this please

PR #1229, opened 2018-01-19
lib/cpus: fix branching in reset function for cortex-a72 AARCH32 mode

manojkumar-arm:
In AARCH32 mode, cortex_a72_reset_func branches to address in lr
register instead of r5 register. This leads to linux boot failure
of Cortex-A72 cores in AARCH32 mode on Juno-R2 board.

This patch fixes the branching of cortex_a72_reset_func to r5
register as in cortex_a57_reset_func implementation.

Signed-off-by: Manoj Kumar <manoj.kumar3@arm.com>


davidcunado-arm:
jenkins: test this please

PR #1231, opened 2018-01-22
Hikey960: Enable invalid FIQ handling

Leo-Yan:
When some interrupts are configured as group 0 in GICv2, these
interrupts trigger FIQ signal; this results in the Linux kernel panic
by reporting log: "Bad mode in FIQ handler detected on CPU0, code
0x00000000 -- Unknown/Uncategorized".  Unfortunately from kernel side it
has no permission to read the GIC register for group 0 interrupts so we
have no chance to get to know which interrupt is configured as secure
interrupt and cause the kernel panic.

For upper reason, this commit enables FIQ exception handling for
SPD_none case.  If the system has not enabled SPD the FIQ interrupt is
trapped into EL3 and the FIQ handler can report the interrupt number so
we can easily narrow down which FIQ introduce unexpected interrupt.
After enable SPD we can rely on SPD interrupt route model to handle FIQ.

### DIFF
--- a/plat/hisilicon/hikey960/hikey960_bl31_setup.c
+++ b/plat/hisilicon/hikey960/hikey960_bl31_setup.c
@@ -16,6 +16,8 @@
 #include <gicv2.h>
 #include <hi3660.h>
 #include <hisi_ipc.h>
+#include <interrupt_mgmt.h>
+#include <platform.h>
 #include <platform_def.h>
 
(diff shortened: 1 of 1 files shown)


davidcunado-arm:
jenkins: test this please

davidcunado-arm:
jenkins: test this please

PR #1237, opened 2018-01-24
SPM: Map devices in the 1st GB

sandrine-bailleux-arm:
This patch maps the devices in the first GB of the system address map
on the FVP into the S-EL1&0 translation regime when SPM support is
enabled. This grants the Secure Partition access to the devices in
this region, for example the memory-mapped Generic Timer device.


davidcunado-arm:
jenkins: test this please

PR #1238, opened 2018-01-24
A new way to deprecate APIs

masahir0y:


danh-arm:
I don't see the need for this PR. I'll briefly explain why but any discussion of this should really happen in a tf-issue, not via commit messages within PRs:

> If ERROR_DEPRECATED=0, the build log is sprinkled with false  positives that come from unnecessary compat layer compiled. The cluttered build log will hide real problems.

The compat layer may not be necessary for you but it is for others. It's entirely possible to remove those false positives as suggested in https://github.com/ARM-software/tf-issues/issues/407.

The expected usage of this is to use ERROR_DEPRECATED=0 in the default build config, and use ERROR_DEPRECATED=1 in a special build config. When the latter fails, it's an indication that the platform should do some API migration work, but it should not be a critical failure.

Marking deprecated APIs with a date doesn't really help platforms with migration or with API removal. Platforms only need to know whether they need to do migration work or not (which the existing flag enables). API removal will happen in 1 big break, independent to the date the API was deprecated.

danh-arm:
I believe the main concern here is addressed by https://github.com/ARM-software/arm-trusted-firmware/pull/1292

PR #1239, opened 2018-01-25
Trusty fixes

arve-android:
Trusty spd updates. Restore trusty setup code and add some missing patches.

### DIFF
--- a/services/spd/trusty/generic-arm64-smcall.c
+++ b/services/spd/trusty/generic-arm64-smcall.c
@@ -1,31 +1,7 @@
 /*
  * Copyright (c) 2016, ARM Limited and Contributors. All rights reserved.
  *
- * Redistribution and use in source and binary forms, with or without
- * modification, are permitted provided that the following conditions are met:
- *
- * Redistributions of source code must retain the above copyright notice, this
- * list of conditions and the following disclaimer.
- *
- * Redistributions in binary form must reproduce the above copyright notice,
- * this list of conditions and the following disclaimer in the documentation
- * and/or other materials provided with the distribution.
- *
- * Neither the name of ARM nor the names of its contributors may be used
- * to endorse or promote products derived from this software without specific
- * prior written permission.
- *
- * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
- * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
- * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
- * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
- * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
- * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
- * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
- * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
- * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
- * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
- * POSSIBILITY OF SUCH DAMAGE.
+ * SPDX-License-Identifier: BSD-3-Clause
  */
 
 #include <debug.h>
--- a/services/spd/trusty/generic-arm64-smcall.c
+++ b/services/spd/trusty/generic-arm64-smcall.c
@@ -66,7 +66,7 @@ static uint64_t trusty_generic_platform_smc(uint32_t smc_fid,
 			 void *handle,
 			 uint64_t flags)
 {
-	switch(smc_fid) {
+	switch (smc_fid) {
 	case SMC_FC_DEBUG_PUTC:
 		trusty_dputc(x1, is_caller_secure(flags));
 		SMC_RET1(handle, 0);
--- a/services/spd/trusty/generic-arm64-smcall.c
+++ b/services/spd/trusty/generic-arm64-smcall.c
@@ -1,0 +1,118 @@
+/*
+ * Copyright (c) 2016, ARM Limited and Contributors. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * Neither the name of ARM nor the names of its contributors may be used
+ * to endorse or promote products derived from this software without specific
+ * prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <debug.h>
+#include <runtime_svc.h>
+
+#include "generic-arm64-smcall.h"
+
+int trusty_disable_serial_debug;
+
+struct dputc_state {
+	char linebuf[128];
+	unsigned l;
+};
+
+static struct dputc_state dputc_state[2];
+
+static void trusty_dputc(char ch, int secure)
+{
+	unsigned i;
+	struct dputc_state *s = &dputc_state[!secure];
+
+	if (trusty_disable_serial_debug)
+		return;
+
+	s->linebuf[s->l++] = ch;
+	if (s->l == sizeof(s->linebuf) || ch == '\n') {
+		if (secure)
+			printf("secure os: ");
+		else
+			printf("non-secure os: ");
+		for (i = 0; i < s->l; i++) {
+			putchar(s->linebuf[i]);
+		}
+		if (ch != '\n') {
+			printf(" <...>\n");
+		}
+		s->l = 0;
+	}
+}
+
+static uint64_t trusty_get_reg_base(uint32_t reg)
+{
+	switch (reg) {
+	case 0:
+		return PLAT_ARM_GICD_BASE;
+
+	case 1:
+		return PLAT_ARM_GICC_BASE;
+
+	default:
+		NOTICE("%s(0x%x) unknown reg\n", __func__, reg);
+		return SMC_UNK;
+	}
+}
+
+static uint64_t trusty_generic_platform_smc(uint32_t smc_fid,
+			 uint64_t x1,
+			 uint64_t x2,
+			 uint64_t x3,
+			 uint64_t x4,
+			 void *cookie,
+			 void *handle,
+			 uint64_t flags)
+{
+	switch(smc_fid) {
+	case SMC_FC_DEBUG_PUTC:
+		trusty_dputc(x1, is_caller_secure(flags));
+		SMC_RET1(handle, 0);
+
+	case SMC_FC_GET_REG_BASE:
+	case SMC_FC64_GET_REG_BASE:
+		SMC_RET1(handle, trusty_get_reg_base(x1));
+
+	default:
+		NOTICE("%s(0x%x, 0x%lx) unknown smc\n", __func__, smc_fid, x1);
+		SMC_RET1(handle, SMC_UNK);
+	}
+}
+
+/* Define a SPD runtime service descriptor for fast SMC calls */
+DECLARE_RT_SVC(
+	trusty_fast,
+
+	SMC_ENTITY_PLATFORM_MONITOR,
+	SMC_ENTITY_PLATFORM_MONITOR,
+	SMC_TYPE_FAST,
+	NULL,
+	trusty_generic_platform_smc
+);
+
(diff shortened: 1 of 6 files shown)


services/spd/trusty/trusty.c
```diff
@@ -327,22 +327,22 @@ static int32_t trusty_init(void)
 	return 0;
 }
 
-static void trusty_cpu_suspend(void)
+static void trusty_cpu_suspend(uint32_t off)
 {
 	struct args ret;
 
-	ret = trusty_context_switch(NON_SECURE, SMC_FC_CPU_SUSPEND, 0, 0, 0);
+	ret = trusty_context_switch(NON_SECURE, SMC_FC_CPU_SUSPEND, off, 0, 0);
```
vwadekar: How does Trusty handle this code?

services/spd/trusty/trusty.c
```diff
@@ -363,18 +363,18 @@ static void trusty_cpu_on_finish_handler(uint64_t unused)
...
 {
-	trusty_cpu_suspend();
+	trusty_cpu_suspend(0);
 }
 
 static void trusty_cpu_suspend_finish_handler(uint64_t unused)
 {
-	trusty_cpu_resume();
+	trusty_cpu_resume(0);
 }
```
vwadekar: In general does it make sense to pass the actual suspend state to Trusty? I suppose TAs can definitely use the System Suspend state. Im not sure how that case would be handled with just two status values

services/spd/trusty/trusty.c
```diff
@@ -448,6 +448,31 @@ static int32_t trusty_setup(void)
...
+			spsr |= MODE32_svc << MODE32_SHIFT;
+		}
+		if (spsr != ns_ep_info->spsr) {
+			NOTICE("Trusty: Switch bl33 from EL2 to EL1 (spsr 0x%x -> 0x%x)\n",
+			       ns_ep_info->spsr, spsr);
+			ns_ep_info->spsr = spsr;
+		}
+	}
+
 	return 0;
```
vwadekar: Doesnt this break the case when we want to run in EL2 mode? e.g. with a hypervisor?

services/spd/trusty/generic-arm64-smcall.c
```diff
@@ -0,0 +1,118 @@
...
+		for (i = 0; i < s->l; i++) {
+			putchar(s->linebuf[i]);
+		}
+		if (ch != '\n') {
+			printf(" <...>\n");
+		}
+		s->l = 0;
+	}
+}
+
```
vwadekar: Why do we need Trusty to send an SMC for printing to the console?

services/spd/trusty/generic-arm64-smcall.c
```diff
@@ -0,0 +1,118 @@
...
+static uint64_t trusty_get_reg_base(uint32_t reg)
+{
+	switch (reg) {
+	case 0:
+		return PLAT_ARM_GICD_BASE;
+
+	case 1:
+		return PLAT_ARM_GICC_BASE;
+
```
vwadekar: Doesnt the Trusty platform code, running in S-EL1, already provide these settings?

services/spd/trusty/trusty.c
```diff
@@ -385,11 +385,23 @@ static const spd_pm_ops_t trusty_pm = {
 	.svc_suspend_finish = trusty_cpu_suspend_finish_handler,
 };
 
+void plat_trusty_set_boot_args(aapcs64_params_t *args);
+
+#ifdef TSP_SEC_MEM_SIZE
+#pragma weak plat_trusty_set_boot_args
+void plat_trusty_set_boot_args(aapcs64_params_t *args)
+{
+	args->arg0 = TSP_SEC_MEM_SIZE;
```
vwadekar: Why did you remove arg1 and arg2 from the equation?

services/spd/trusty/trusty.c
```diff
@@ -398,17 +410,29 @@ static int32_t trusty_setup(void)
...
-	SET_PARAM_HEAD(ep_info, PARAM_EP, VERSION_1, SECURE | EP_ST_ENABLE);
-	ep_info->spsr = SPSR_64(MODE_EL1, MODE_SP_ELX, DISABLE_ALL_EXCEPTIONS);
+	instr = *(uint32_t *)ep_info->pc;
 
-	/*
-	 * arg0 = TZDRAM aperture available for BL32
-	 * arg1 = BL32 boot params
-	 * arg2 = BL32 boot params length
-	 */
-	ep_info->args.arg1 = ep_info->args.arg2;
```
vwadekar: Are these parameters no longer required by Trusty?

arve-android: These are platform specific. If you need them on your platform, you can set them in plat_trusty_set_boot_args.

services/spd/trusty/trusty.c
arve-android: They are platform specific, I don't know where to get them from in generic code.

services/spd/trusty/trusty.c
arve-android: Do we have access to the system suspend state here? There is a separate psci call for system suspend, but that information seems to be lost before these hooks are called.

services/spd/trusty/trusty.c
arve-android: It is used to disable fiqs when a cpu is unplugged but leave them on when a cpu enter idle.

services/spd/trusty/generic-arm64-smcall.c
arve-android: The generic-arm64 target has no platform specific code.

services/spd/trusty/generic-arm64-smcall.c
arve-android: This is only for the generic-arm64 target. You don't need to enable this if you have a platform specific trusty image.

services/spd/trusty/generic-arm64-smcall.c
vwadekar: What is restricting you from adding that platform code? In general, issuing an SMC for printing to console does not look like an elegant solution.

services/spd/trusty/trusty.c
vwadekar: In some implementations the previous level bootloader passes this information to the Trusted OS - memory aperture, uart port, etc. I think we should have that model and then sanity check the boot params in the SPD, instead of coming with another way of passing boot params.

services/spd/trusty/trusty.c
vwadekar: But now you removed the previous code. Shouldnt we be making changes in such a way that current platforms do not break?

services/spd/trusty/trusty.c
vwadekar: I see that system_suspend eventually calls into this trusty_cpu_suspend() with the max suspend level. So trusty SPD should be able to differentiate easily, AFAICS.

services/spd/trusty/trusty.c
arve-android: Can you add the hooks to your platform code? This patch restores the part of the original trusty patches we need to boot on other platforms. Those platforms stopped working after syncing your changes.  I don't know why you decided to remove the mechanism we had for platform specific initialization and add your init code to the generic trusty code instead. Since I don't know which platform(s) rely on this, and since I cannot test them, I did not touch them.

services/spd/trusty/generic-arm64-smcall.c
arve-android: This code is only used for the generic-arm64 target. Adding platform specific code to that target would make it non-generic.

services/spd/trusty/trusty.c
```diff
@@ -299,6 +295,7 @@ static int32_t trusty_init(void)
 	ep_info = bl31_plat_get_next_image_ep_info(SECURE);
 	assert(ep_info);
 
+	fpregs_context_save(get_fpregs_ctx(cm_get_context(NON_SECURE)));
```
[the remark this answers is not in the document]
arve-android: CTX_INCLUDE_FPREGS is enabled in the make file and required for trusty.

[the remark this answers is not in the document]
davidcunado-arm: Apologies missed that bit! And I see that you've also remove the guards around the existing calls to fpregs_context_restore/save.

davidcunado-arm:
@vwadekar 
Please can you progress this with @arve-android ? I won't look to start the merge process until I get your nod - please ping me when ready?

davidcunado-arm:
jenkins: test this please

vwadekar:
Sorry I couldn't get to this PR earlier. I will try to spend some time
today.

vwadekar:
@arve-android

Can you please respond to the comments that are still open?

services/spd/trusty/trusty.c
arve-android: We can add more values later if we need to know about system suspend. For now I don't want to add a hack to try to guess if we are entering system suspend based on the platform specific power level.

services/spd/trusty/trusty.c
arve-android: see comment below

arve-android:
@vwadekar is there a specific comment you want me to respond to? I replied to the two comments that did not have a direct reply from me, but it is not clear to me why you would be waiting on a reply to those specific comments.

davidcunado-arm:
@arve-android @vwadekar 
Check failures include build errors on following configs:
- PLAT=tegra TARGET_SOC=t186 SPD=trusty
- PLAT=tegra TARGET_SOC=t210 SPD=trusty

Build error (which I've not looked into):

./build/tegra/t186/debug/bl31/trusty.o: In function `trusty_setup':
/work/workspace/trusted-firmware-build/BUILD_CONFIG/tegra-t186-trusty/MODE/Debug/label/platforms/services/spd/trusty/trusty.c:434: undefined reference to `plat_trusty_set_boot_args'
/work/workspace/trusted-firmware-build/BUILD_CONFIG/tegra-t186-trusty/MODE/Debug/label/platforms/services/spd/trusty/trusty.c:434:(.text.trusty_setup+0xa4): relocation truncated to fit: R_AARCH64_CALL26 against undefined symbol `plat_trusty_set_boot_args'
make: *** [build/tegra/t186/debug/bl31/bl31.elf] Error 1
Build failed with code 2

vwadekar:
@arve-android

Since you are already touching the trusty handlers,  makes sense  to add System suspend as well. We can start another PR if you want.

But I think you should definitely add documentation for the newly added platform handler, as part of this PR, so that it gets easy to port.

services/spd/trusty/trusty.c
arve-android: I added a patch to add a plat_trusty_set_boot_args hook to the tegra platform. Please test it as I don't have any hardware to test it on.

services/spd/trusty/trusty.c
```diff
@@ -398,17 +406,29 @@ static int32_t trusty_setup(void)
...
+		INFO("trusty: Found 32 bit image\n");
+		aarch32 = 1;
+	} else if (instr >> 8 == 0xd53810 || instr >> 16 == 0x9400) {
+		INFO("trusty: Found 64 bit image\n");
+	} else {
+		NOTICE("trusty: Found unknown image, 0x%x\n", instr);
+	}
+
+	SET_PARAM_HEAD(ep_info, PARAM_EP, VERSION_1, SECURE | EP_ST_ENABLE);
+	if (!aarch32)
```
davidcunado-arm: misra_violation: [Required] MISRA C-2012 Rule 10.1 violation: The operand aarch32 has essential type signed 32-bit int, but should be Essentially Boolean.

Change to:
if (aarch32 == 0)

services/spd/trusty/trusty.c
```diff
@@ -398,17 +406,29 @@ static int32_t trusty_setup(void)
...
+	instr = *(uint32_t *)ep_info->pc;
 
-	/*
-	 * arg0 = TZDRAM aperture available for BL32
-	 * arg1 = BL32 boot params
-	 * arg2 = BL32 boot params length
-	 */
-	ep_info->args.arg1 = ep_info->args.arg2;
-	ep_info->args.arg2 = TRUSTY_PARAMS_LEN_BYTES;
+	if (instr >> 24 == 0xea) {
```
davidcunado-arm: 1. misra_violation: [Required] MISRA C-2012 Rule 10.4 violation: The operands instr>>24 and 0xea have essential type categories unsigned 32-bit int and signed 16-bit int, which do not match.

Change to:
if (instr >> 24 == 0xeaU)

services/spd/trusty/trusty.c
```diff
@@ -425,6 +445,31 @@ static int32_t trusty_setup(void)
 	if (ret)
 		ERROR("trusty: failed to register fiq handler, ret = %d\n", ret);
 
+	if (aarch32) {
```
davidcunado-arm: misra_violation: [Required] MISRA C-2012 Rule 14.4 violation: Condition aarch32 is not essentially boolean.

Change to:
if (aarch32 != 0)

services/spd/trusty/trusty.c
```diff
@@ -398,17 +406,29 @@ static int32_t trusty_setup(void)
...
+	if (!aarch32)
+		ep_info->spsr = SPSR_64(MODE_EL1, MODE_SP_ELX,
+					DISABLE_ALL_EXCEPTIONS);
+	else
+		ep_info->spsr = SPSR_MODE32(MODE32_svc, SPSR_T_ARM,
+					    SPSR_E_LITTLE,
+					    DAIF_FIQ_BIT |
+					    DAIF_IRQ_BIT |
+					    DAIF_ABT_BIT);
+	memset(&ep_info->args, 0, sizeof(ep_info->args));
```
davidcunado-arm: misra_violation: [Required] MISRA C-2012 Rule 17.7 violation: Return value of non-void function memset() is not used.

The function trusty_setup returns uint_32, so could just return the error code from memset?

services/spd/trusty/generic-arm64-smcall.c
```diff
@@ -0,0 +1,94 @@
...
+static uint64_t trusty_generic_platform_smc(uint32_t smc_fid,
+			 uint64_t x1,
+			 uint64_t x2,
+			 uint64_t x3,
+			 uint64_t x4,
+			 void *cookie,
+			 void *handle,
+			 uint64_t flags)
+{
+	switch(smc_fid) {
```
davidcunado-arm: Style check failure:
services/spd/trusty/generic-arm64-smcall.c:69: ERROR: space required before the open parenthesis '('
+	switch(smc_fid) {

services/spd/trusty/trusty.c
arve-android: changed type

services/spd/trusty/trusty.c
arve-android: changed type

services/spd/trusty/trusty.c
arve-android: memset does not return an error it returns the first argument. I'm not sure how this error can be corrected in a reasonable way

danh-arm: Just cast the return value to void:

`(void)memset(&ep_info->args, 0, sizeof(ep_info->args));`

services/spd/trusty/trusty.c
```diff
@@ -398,17 +407,29 @@ static int32_t trusty_setup(void)
...
-	 * arg0 = TZDRAM aperture available for BL32
-	 * arg1 = BL32 boot params
-	 * arg2 = BL32 boot params length
-	 */
-	ep_info->args.arg1 = ep_info->args.arg2;
-	ep_info->args.arg2 = TRUSTY_PARAMS_LEN_BYTES;
+	if (instr >> 24 == 0xeaU) {
+		INFO("trusty: Found 32 bit image\n");
+		aarch32 = true;
+	} else if (instr >> 8 == 0xd53810 || instr >> 16 == 0x9400) {
```
davidcunado-arm: @arve-android 
Similar to change to line 412, there also need to have U suffix to the RH operands of ==.

(Apologies, not sure why I didn't comment this before)

arve-android:
@davidcunado-arm 
I'm not in a hurry to merge this time wise, but I would prefer to get it into the release. The last release broke everyone using the patches we provided to add trusty support. It would be nice if we can tell them to get the next release instead.

davidcunado-arm:
@vwadekar 
Happy for me to merge this?

vwadekar:
I would still like the newly added platform handler to be documented before
we merge. Arve?

davidcunado-arm:
@vwadekar @arve-android 
Ok, we're doing a RC1 probably on MON / TUE - can you get this missing piece done on MON 5-MAR?

vwadekar:
Thanks David. I am good with the PR once we add some documentation.

Arve, what do you think? Would you be able to push an update before the cut
off date?

arve-android:
@vwadekar 
The platform handler is not new (I don't know why you removed it in your initial upload), but it has been added to the documentation now.

vwadekar:
Looks good. David we can go ahead and merge this one.

davidcunado-arm:
jenkins: test this please

davidcunado-arm:
checks - passed (false positives in the static checks).

PR #1242, opened 2018-01-27
fiptool: Fix use after free

afaerber:
Commit 880b9e8b4c99ad99eee14079d5a6162733ef4931 (Add padding at the end
of the last entry) added code using toc_entry pointer, whose memory is
already freed via variable buf. This causes enormous padding on openSUSE.

Free the memory buffer only after padding is done.

This fixes #1204.


davidcunado-arm:
jenkins: test this please

PR #1245, opened 2018-01-29
Analyze coding style of patches individually

antonio-nino-diaz-arm:
With the old system `checkpatch.pl` gets one sole input that consists of the commit message and commit diff of each commit between BASE_COMMIT and HEAD. It also filters out changes in some files, which makes `git format-patch` completely ignore that commit, even the commit message.

With the new system the commit message and commit diff are analyzed separately. This means that, even if all the files modified by a commit are filtered out, the commit message will still be analyzed.

Also, all commits are analyzed individually. This way it's easier to know which commit caused the problem, and there are no warnings about repeated "Signed-off-by" lines.

### DIFF
--- a/.checkpatch.conf
+++ b/.checkpatch.conf
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2016, ARM Limited and Contributors. All rights reserved.
+# Copyright (c) 2016-2018, ARM Limited and Contributors. All rights reserved.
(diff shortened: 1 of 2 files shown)


davidcunado-arm:
jenkins: test this please

davidcunado-arm:
jenkins: test this please

PR #544, opened 2016-03-08
Support for Rockchip's RK3368 SoCs

rkchrome:
RK3368 is one of the Rockchip family SoCs, which is an
Octa-core ARM Cortex-A53 SoC.

This patch adds support to boot the Trusted Firmware on RK3368
SoCs, and adds support to boot secondary CPUs, enter/exit core
power states for all CPUs in the slow/fast clusters.

This is the initial version for rockchip SoCs.(RK3368 and next SoCs)
* Support arm gicv2 & gicv3.
* Boot up 8 cores.
* Add generic CPU helper functions.
* Support suspend/resume.
* Add system_off & system_reset implementation.
* Add delay timer platform implementation.
* Support the new porting interface for the PSCI implementation.

Change-Id: I704bb3532d65e8c70dbd99b512c5e6e440ea6f43
Signed-off-by: Tony Xie <tony.xie@rock-chips.com>
Signed-off-by: Caesar Wang <wxt@rock-chips.com>
Signed-off-by: Shengfei xu <xsf@rock-chips.com>

PR #1247, opened 2018-01-30
rockchip/rk3399: Fix memory corruptions or illegal memory access

rkchrome:
Coverity scan done for the coreboot project found the issue:
Coverity (*** CID 1385418: Memory - illegal accesses (OVERRUN))
Coverity (*** CID 1385419: Memory - corruptions  (OVERRUN))

Fix the Converity error issue with store_cru[] loop needs to be one
element bigger.

Fixes: ARM-software/tf-issues#544

Change-Id: I420f0a660b24baaa5fc5e78fca242cf750c9bbc7

### DIFF
--- a/plat/rockchip/rk3399/drivers/pmu/pmu.c
+++ b/plat/rockchip/rk3399/drivers/pmu/pmu.c
@@ -32,7 +32,7 @@ DEFINE_BAKERY_LOCK(rockchip_pd_lock);
 
 static uint32_t cpu_warm_boot_addr;
 static char store_sram[SRAM_BIN_LIMIT + SRAM_TEXT_LIMIT + SRAM_DATA_LIMIT];
-static uint32_t store_cru[CRU_SDIO0_CON1 / 4];
+static uint32_t store_cru[CRU_SDIO0_CON1 / 4 + 1];
 static uint32_t store_usbphy0[7];
 static uint32_t store_usbphy1[7];
 static uint32_t store_grf_io_vsel;


plat/rockchip/rk3399/drivers/pmu/pmu.c
```diff
@@ -1261,15 +1261,15 @@ void cru_register_save(void)
 {
 	int i;
 
-	for (i = 0; i <= CRU_SDIO0_CON1; i = i + 4)
+	for (i = 0; i < CRU_SDIO0_CON1; i = i + 4)
```
jwerner-chromium: Is this really the right fix, though? I'm not super deep into this code, but AFAICT the point is to save/restore all CRU registers. With this patch you end up not saving the CRU_SDIO0_CON1 register. Looks like the real fix would be to increase the array size after all.

@dbasehore should also take a look.

dbasehore:
jwerner is correct. You should be changing the size of the array to "(CRU_SDIO0_CON1 / 4) + 1" instead.

paulmenzel:
@Caesar-github, thank you for looking into this. Please note, Coverity scan done for the coreboot project found the issue, so Coverity should be credited. Maybe something like below:

> Found-by: Coverity (** CID 1385418:  Memory - illegal accesses  (OVERRUN))

Also, issue #544 tracks the latest findings, so

> Fixes: https://github.com/ARM-software/tf-issues/issues/544

would also be useful in the commit message.

Personally, I prefer verbs in imperative mood for the commit message and merge/pull request summary: *Fix memory corruptions* or *Fix illegal memory access*. And, having the same message in the merge/pull request summary as in the commit is easy, but probably not needed either.

Caesar-github:
Okay, done.

davidcunado-arm:
@paulmenzel 
I've updated the PR title and summary to match the commit message.

davidcunado-arm:
jenkins: test this please

PR #1248, opened 2018-01-30
Correct the Makefile logic for disabling PIE

stevecapperarm:
In the Makefile we use findstring to locate gcc toolchains
that have PIE enabled by default.

Unfortunately the result of findstring is compared against
an integer, 1, rather than a non-empty string; the logic to
disable PIE then doesn't get applied.

This patch fixes the flag test.


davidcunado-arm:
jenkins: test this please

PR #1249, opened 2018-01-30
uniphier: fix and improve memory layout

masahir0y:


masahir0y:
Please hold this back.

I will discuss about this PR a bit more internally in Socionext.

masahir0y:
After Socionext internal discussion, I fixed this a little bit.
This PR is now ready to be pulled.
Please merge.

davidcunado-arm:
jenkins: test this please

plat/socionext/uniphier/include/platform_def.h
```diff
@@ -27,26 +27,28 @@
 #define PLAT_MAX_OFF_STATE		2
 #define PLAT_MAX_RET_STATE		1
 
-#define UNIPHIER_SEC_DRAM_BASE		0x80000000
+#define BL2_BASE			0x80000000
```
davidcunado-arm: @masahir0y 
Check failures were static analysis issues due to:
misra_violation: [Required] MISRA C-2012 Rule 7.2 violation: Unsigned constants must be declared with U or u suffix.

Could you add ULL-suffix to:
- BL2_XXX
- UNIPHIER_SEC_DRAM_XXX
- BL31_XXX
- BL32_XXX
- UNIPHIER_ROM_REGION_XXX
- UNIPHIER_OCM_REGION_XXX
- UNIPHIER_REG_REGION_XXX

Everything else looked fine.

Thanks!

masahir0y:
> Could you add ULL-suffix to:
>
>BL2_XXX
>UNIPHIER_SEC_DRAM_XXX
>BL31_XXX
>BL32_XXX
>UNIPHIER_ROM_REGION_XXX
>UNIPHIER_OCM_REGION_XXX
>UNIPHIER_REG_REGION_XXX

I did so, but I wondered why you suggested `ULL` instead of `UL`.

IIUC, we assume IPL32 for aarch32 compilers and LP64 for aarch64 compilers.
So, `unsigned long` is as long as as a pointer type, so long enough to contain physical address.

The only exception I know is Large Physical Address Extension,
where CPU is 32bit core, but 36 bit physical address is supported.

Is this what was in your mind?

antonio-nino-diaz-arm:
> I did so, but I wondered why you suggested `ULL` instead of `UL`.
> 
> IIUC, we assume IPL32 for aarch32 compilers and LP64 for aarch64 compilers.
> So, `unsigned long` is as long as as a pointer type, so long enough to contain physical address.
> 
> The only exception I know is Large Physical Address Extension,
> where CPU is 32bit core, but 36 bit physical address is supported.
> 
> Is this what was in your mind?

It's more a matter of consistency across the codebase. For the same code to work on AArch64 and AArch32, it is needed to always use `unsigned long long` when a 64-bit variable is needed. This means that whenever you are using `printf` you need to use `%ull` or similar, etc. It's just easier if we don't use `unsigned long` at all. Right now it is mostly used in platform code, anyway. It's still used in common code, but we are trying to move away from it...

masahir0y:
> It's more a matter of consistency across the codebase. For the same code to work on AArch64 and AArch32, it is needed to always use unsigned long long when a 64-bit variable is needed. This means that whenever you are using printf you need to use %ull or similar, etc. It's just easier if we don't use unsigned long at all. Right now it is mostly used in platform code, anyway. It's still used in common code, but we are trying to move away from it...

I had noticed this evil crap.
I threw my suggestion as PR 1255.
(but cannot be pulled.)
Better to  ask maintainers' opinion before investing efforts.

davidcunado-arm:
jenkins: test this please

davidcunado-arm:
Check fails - issue due to existing code; can be ignored.

PR #1253, opened 2018-01-31
AMUv1 support for AArch32

dp-arm:


davidcunado-arm:
jenkins: test this please

davidcunado-arm:
Check fails - false positives in static anlysis checks.

PR #1255, opened 2018-02-02
Use consistent int-ll64 typedefs for aarch32 and aarch64

masahir0y:

### DIFF
--- a/include/lib/stdlib/machine/_types.h
+++ b/include/lib/stdlib/machine/_types.h
@@ -52,56 +52,19 @@ typedef	short			__int16_t;
 typedef	unsigned short		__uint16_t;
 typedef	int			__int32_t;
 typedef	unsigned int		__uint32_t;
-
-
-/*
- * Standard type definitions which are different in AArch64 and AArch32
- */
-#ifdef	AARCH32
 typedef	long long		__int64_t;
 typedef	unsigned long long	__uint64_t;
-typedef	__int32_t	__critical_t;
-typedef	__int32_t	__intfptr_t;
-typedef	__int32_t	__intptr_t;
-typedef	__int32_t	__ptrdiff_t;		/* ptr1 - ptr2 */
-typedef	__int32_t	__register_t;
-typedef	__int32_t	__segsz_t;		/* segment size (in pages) */
-typedef	__uint32_t	__size_t;		/* sizeof() */
-typedef	__int32_t	__ssize_t;		/* byte count or error */
-typedef	__uint32_t	__uintfptr_t;
-typedef	__uint32_t	__uintptr_t;
-typedef	__uint32_t	__u_register_t;
-typedef	__uint32_t	__vm_offset_t;
-typedef	__uint32_t	__vm_paddr_t;
-typedef	__uint32_t	__vm_size_t;
-#elif defined AARCH64
-typedef	long			__int64_t;
-typedef	unsigned long		__uint64_t;
-typedef	__int64_t	__critical_t;
-typedef	__int64_t	__intfptr_t;
-typedef	__int64_t	__intptr_t;
-typedef	__int64_t	__ptrdiff_t;		/* ptr1 - ptr2 */
-typedef	__int64_t	__register_t;
-typedef	__int64_t	__segsz_t;		/* segment size (in pages) */
-typedef	__uint64_t	__size_t;		/* sizeof() */
-typedef	__int64_t	__ssize_t;		/* byte count or error */
-typedef	__uint64_t	__uintfptr_t;
-typedef	__uint64_t	__uintptr_t;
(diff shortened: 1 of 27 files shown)


masahir0y:
This is related to a discussion in PR 1249.

Here is the repeat of the log of the last commit.

    types: use int-ll64 for both aarch32 and aarch64
    
    The result of commit 031dbb122472 ("AArch32: Add essential Arch
    helpers") is a nightmare.
    
    For example, 'uint64_t' is defined as 'unsigned long long' for aarch32
    and 'unsigned long' for aarch64.
    
    Likewise, 'uintptr_t' is 'unsigned int' for aarch32, 'unsigned long'
    for aarch64.
    
    This is a problem when you use printf() in common code.
    
    Use ifdef?
      #ifdef AARCH32
              printf("val = %llx", val);
      #else
              printf("val = %lx", val);
      #endif
    
    or cast to a type long enough to cover all cases?
      printf("val = %lx\n", (unsigned long long)val);
    
    or even uglier?
      printf("val = %" PRIx64 "\n", val);
    
    No.  Use the same typedefs for all arches.  As you see in Linux commit
    0c79a8e29b5f ("asm/types.h: Remove include/asm-generic/int-l64.h"),
    kernel developers completed the task to unify in-kernel typedefs into
    int-ll64, where fixed integer types can be defined as follows:
    
        typedef signed char           int8_t;
        typedef unsigned char         uint8_t;
    
        typedef signed short          int16_t;
        typedef unsigned short        uint16_t;
    
        typedef signed int            int32_t;
        typedef unsigned int          uint32_t;
    
        typedef signed long long      int64_t;
        typedef unsigned long long    uint64_t;
    
    This makes sense where the codebase is shared between 32bit and 64bit,
    with the data mode ILP32, LP64, respectively.

                       ILP32           LP64
        int            32              32
        long           32              64
        long long      64              64
    
    'long long' is always 64bit so used for defining 'uint64_t'.
    'long' has the same width as pointer, so for 'uintptr_t'.
    
    I left an ifdef conditional for (s)size_t because "%z" is used in some
    places.  (maybe we should change size_t to 'unsigned long' by forbidding
    the use of "%z")

I believe this is the right thing, and will make clean code-base in a long run.

However, this commit can not be applied until typedef distortion is fixed.

There are several mismatch between declaration in header and definition in C file.

I think I can do this work, but I am busy.
If somebody offers to do this tedious but valuable work,
I am happy to hand this over.

Note:
Nothing in this PR is signed-offed yet.

antonio-nino-diaz-arm:
I think that the idea is good, but this is might cause problems for non-upstream platforms depending on where the changes have to be done. Let's wait for more comments.

soby-mathew:
Yes, usage of `unsigned long`  was always problematic between AArch32 and AArch64. The original dilemma when the AArch32 mode was introduced was whether to modify the *standard* headers in `stdlib`. But that dilemma is passed now I think.

include/lib/stdlib/machine/_types.h
```diff
@@ -110,8 +73,23 @@ typedef	__int8_t	__int_least8_t;
 typedef	__int16_t	__int_least16_t;
 typedef	__int32_t	__int_least32_t;
 typedef	__int64_t	__int_least64_t;
+typedef	long		__ptrdiff_t;		/* ptr1 - ptr2 */
+typedef	long		__register_t;
+typedef	long		__segsz_t;		/* segment size (in pages) */
+/* differentiate for "%z" in printf() */
+#ifdef	AARCH32
+typedef	unsigned int	__size_t;		/* sizeof() */
+typedef	int		__ssize_t;		/* byte count or error */
```
danh-arm: Perhaps we can fix tf_printf to always using long for %z rather than have the #ifdef here?

dp-arm:
@masahir0y ping

masahir0y:
Sorry, I did not notice this PR had been replied.
I will find some time to work on it next week.

antonio-nino-diaz-arm:
IP Review +1

antonio-nino-diaz-arm:
I would squash all "add static qualifier to locally used functions" and all "fix pointer type mismtach" commits. Also, you have a typo there: mismtach.

masahir0y:
> I would squash all "add static qualifier to locally used functions" and all "fix pointer type mismtach" commits. Also, you have a typo there: mismtach.

OK.  I squashed them, and fixed the typo.

include/lib/stdlib/machine/_types.h
danh-arm: Yes, good point.

danh-arm:
jenkins: test this please

masahir0y:
> Could you take a look at them?

I fixed them and pushed.
Thanks.

antonio-nino-diaz-arm:
The Coverity tests have failed, but everything else has passed.

PR #1258, opened 2018-02-07
optee: print header info before validate

vchong:
Currently optee header info is only printed after it is validated,
but this does not help with debugging in case of error, so print it
before.

Signed-off-by: Victor Chong <victor.chong@linaro.org>


davidcunado-arm:
jenkins: test this please

PR #1263, opened 2018-02-09
Dynamic Configuration Prototype

soby-mathew:
This patchset enables dynamic configuration in ARM_TF for ARM Platforms. The first few patches make generic code changes to have additional config files packaged to FIP and adds them to Chain of Trust. The BL handover interface is expanded to pass more arguments between different BL stages. The final set of patches enables BL1 to load and authenticate TB_FW_CONFIG and pass it on to BL2 which inturn parses it to load the HW_CONFIG on ARM platforms. This feature is always enabled on ARM platforms and gets ignored if the config files are not provided. Only FVP adds the necessary config files.


davidcunado-arm:
Checks - passed (MISRA issues are for existing code), internal link - http://ssg-sw.cambridge.arm.com/jenkins/view/TF_critical_CI_jobs/job/trusted-firmware-pr/1756/

PR #1265, opened 2018-02-12
aarch32: boot from BL2 in SVC mode

etienne-lms:
Hi, 
Related to recent BL2_AT_EL3 feature and constraint on aarhc32 and armv7 platforms, here is a proposal to move such BL2 execution in SVC mode rather than Monitor mode.
* Entry in BL2 do not need to move execution to Monitor mode. This assumes the early platform bootrom stage boots the arm trusted firmware in Supervisor mode.
*  BL2 (in BL2_AT_EL3 config) exits to BL32 is done through a simple branch instruction. The bootrom arguments are passed through cpu registers r0-r3 and the BL33 entry into register lr_svc.

### DIFF
--- a/bl2/aarch32/bl2_el3_entrypoint.S
+++ b/bl2/aarch32/bl2_el3_entrypoint.S
@@ -71,14 +71,16 @@ func bl2_run_next_image
 	bl	bl2_el3_plat_prepare_exit
 
 	/*
-	 * Extract PC and SPSR based on struct `entry_point_info_t`
-	 * and load it in LR and SPSR registers respectively.
+	 * Extract PC, LR and SPSR based on struct `entry_point_info_t`
+	 * and load it in LR and SPSR registers respectively then
+	 * branch to target entry point.
 	 */
-	ldr	lr, [r8, #ENTRY_POINT_INFO_PC_OFFSET]
+	ldr	lr, [r8, #ENTRY_POINT_INFO_LR_SVC_OFFSET]
+	ldr	r9, [r8, #ENTRY_POINT_INFO_PC_OFFSET]
 	ldr	r1, [r8, #(ENTRY_POINT_INFO_PC_OFFSET + 4)]
-	msr	spsr, r1
+	msr	cpsr_fsxc, r1
 
 	add	r8, r8, #ENTRY_POINT_INFO_ARGS_OFFSET
 	ldm	r8, {r0, r1, r2, r3}
-	eret
+	bx	r9
 endfunc bl2_run_next_image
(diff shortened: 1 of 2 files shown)


soby-mathew:
Hi @etienne-lms,
> Entry in BL2 do not need to move execution to Monitor mode. 

I am trying to understand the reason for BL2_AT_EL3 not running in monitor mode. It avoids special case in el3_common_macros for BL2_AT_EL3 if it can remain in Monitor mode. If BL32 expects entry in SVC mode, then the corresponding mode can be set in the `entry_point_info_t` which will get programmed as part of exit from BL2_AT_EL3.

etienne-lms:
Current bl2 for aarch32 only always boots a BL32 (there is no SCR[NS] switch handle), thus the proposed branch preserve the current sequence.

The reason for running BL2 in Svc rather than Mon is to benefit from abort traps.

soby-mathew:
OK. We install the same vector table in MVBAR and S-VBAR and since both monitor mode and S-SVC have the same address space, wouldn't that be enough for abort handling  ? The anomaly being that the abort handler would execute in SVC mode when triggered whereas the rest of BL2_AT_EL3 was executing in Monitor mode prior to the abort.

soby-mathew:
Also, since BL2_AT_EL3 pulls in some common code which are meant to run at EL3 for AArch64 and Monitor  mode in AArch32, any switch of SCR[NS] bit would have to be special cased if this were to run in SVC mode (There is no such instance currently). Hence it is desirable if the special case can be avoided.

etienne-lms:
Exceptions are masked when executing in monitor mode: i.e mmu operates permission checks but instruction sequence is not halted if an abort is fired. This is aarch32 constraint.

soby-mathew:
OK, I see. Let me disucss and get back on this.

soby-mathew:
Hi @etienne-lms,
From my quick look up [1], as long as SCR.EA=0 and HCR.AMO=0, the CPSR.A controls the aynchronous abort. The synchronous aborts cannot be masked and will be taken to abort mode (VBAR offset 0x10). When CPSR.A is unmasked, then asynchronous abort will take the same vector given SCR.EA=0.

Currently CPSR.A is masked in AArch32 and this should be unmasked to be inline with AArch64 initialization (this needs to be corrected).  I couldn't find any reference suggesting abort being masked in monitor mode and instruction sequence proceeding after abort outside the controls I mentioned. If you could provide any reference it will be useful. Meanwhile I will keep checking this.

[1] ( Figure B1-7, Table B1-11, Table B1-17 in ARMv7 ref manual https://silver.arm.com/download/download.tm?pv=1603196 )

etienne-lms:
well, seems i've misunderstood the execution context. I will close this p-r and study your proposal which looks right.

PR #1268, opened 2018-02-14
EHF: Fix priority check

jeenu-arm:
When deactivating, it's not an error if the priority being deactivating
is equal to the active priority. Fix this.

Change-Id: I66f0e9e775ac9aba8a7cc48cd3ecd3b358be63c0


davidcunado-arm:
jenkins: test this please

PR #1270, opened 2018-02-15
Redefine SMC_UNK as -1 instead of 0xFFFFFFFF

antonio-nino-diaz-arm:
According to the SMC Calling Convention (ARM DEN0028B):

> The Unknown SMC Function Identifier is a sign-extended value of (-1) that is returned in R0, W0 or X0 register.

The value wasn't sign-extended because it was defined as a 32-bit unsigned value (`0xFFFFFFFF`).

`SMC_PREEMPT` has been redefined as -2 for the same reason.

NOTE: This might be a compatibility break for some AArch64 platforms that don't follow the previous version of the SMCCC (ARM DEN0028A) correctly. That document specifies that only the bottom 32 bits of the returned value must be checked. If a platform relies on the top 32 bits of the result being 0 (so that `SMC_UNK` is `0x00000000FFFFFFFF`), it will have to fix its code to comply with the SMCCC.

### DIFF
--- a/bl31/aarch64/runtime_exceptions.S
+++ b/bl31/aarch64/runtime_exceptions.S
@@ -390,12 +390,12 @@ smc_unknown:
 	 * content). Either way, we aren't leaking any secure information
 	 * through them.
 	 */
-	mov	w0, #SMC_UNK
+	mov	x0, #SMC_UNK
 	b	restore_gp_registers_callee_eret
 
 smc_prohibited:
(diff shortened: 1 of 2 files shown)


davidcunado-arm:
jenkins: test this please

davidcunado-arm:
Check fails - due to corresponding change needed in internal test framework; will be applied once PR merged.

PR #1271, opened 2018-02-17
tegra: Fix mmap_region_t struct mismatch

afaerber:
Commit fdb1964c34968921379d3592e7ac6e9a685dbab1 ("xlat: Introduce
MAP_REGION2() macro") added a granularity field to mmap_region_t.

Tegra platforms were using the v2 xlat_tables implementation in
common/tegra_common.mk, but v1 xlat_tables.h headers in soc/*/plat_setup.c
where arrays are being defined. This caused the next physical address to
be read as granularity, causing EINVAL error and triggering an assert.

Consistently use xlat_tables_v2.h header to avoid this.

Fixes ARM-software/tf-issues#548.

### DIFF
--- a/plat/nvidia/tegra/include/tegra_private.h
+++ b/plat/nvidia/tegra/include/tegra_private.h
@@ -10,7 +10,7 @@
 #include <arch.h>
 #include <platform_def.h>
 #include <psci.h>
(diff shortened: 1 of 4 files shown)


davidcunado-arm:
jenkins: test this please

PR #1272, opened 2018-02-21
Refactor SPE/SVE code and fix some bugs in AMUv1 on AArch32

dp-arm:


lib/extensions/sve/sve.c
```diff
@@ -71,10 +71,7 @@ void sve_enable(int el2_unused)
 		return;
 
 #if CTX_INCLUDE_FPREGS
-	/*
-	 * CTX_INCLUDE_FPREGS is not supported on SVE enabled systems.
-	 */
-	assert(0);
+#error CTX_INCLUDE_FPREGS is not supported on SVE enabled systems.
```
dp-arm: this is actually wrong, it should be a runtime check.  This is because we cannot determine at compile time if both the platform has SVE and the firmware is saving/restoring the FP registers.  I will drop this patch from the series.

davidcunado-arm:
Checks - passed (with expected fails in existing code for static checks).

PR #1273, opened 2018-02-21
Ensure the correct execution of TLBI instructions

antonio-nino-diaz-arm:
After executing a TLBI a DSB is needed to ensure completion of the TLBI.

rk3328: The MMU is allowed to load TLB entries for as long as it is enabled. Because of this, the correct place to execute a TLBI is right after disabling the MMU.

@TonyXie06, @Caesar-github, could you explain why this TLBI is needed in the first place?

### DIFF
--- a/bl1/aarch64/bl1_exceptions.S
+++ b/bl1/aarch64/bl1_exceptions.S
@@ -187,6 +187,7 @@ func smc_handler64
 
 	bl	disable_mmu_icache_el3
 	tlbi	alle3
+	dsb	ish /* ERET implies ISB, so it is not needed here */
 
 #if SPIN_ON_BL1_EXIT
 	bl	print_debug_loop_message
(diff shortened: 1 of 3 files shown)


davidcunado-arm:
jenkins: test this please

TonyXie06:
yes ，the correct place to execute a TLBI is right after disabling the MMU，For our code we ignore that。

TonyXie06:
could you explain why this TLBI is needed in the first place?
We check the  codes again, It is not needed  executing TLBI. The reason for this is that if system is resumed，system will enable the mmu and execute TLBI，we execute the TLBI at suspend ，because we want to reduce the time of executing TLBI at resume.

antonio-nino-diaz-arm:
@TonyXie06 The truth is that we didn't know. We just realized that `tlbialle3` before `disable_mmu_icache_el3` was a bug, and we didn't know if there was any specific reason why this platform had it in the first place (actually, I asked in the first message in this PR). That's why we decided to keep it.

TonyXie06:
we agree that place the tlbialle3 behind disable_mmu_icache_el3.

antonio-nino-diaz-arm:
We don't think it is needed. The problem is that we don't have a rk3328 to test, and we are not the maintainers of that platform.

TonyXie06:
We check the code, It is not needed. we will delete the code and test it.

antonio-nino-diaz-arm:
Ok, thanks.

PR #1274, opened 2018-02-21
AMU fixes for Cortex-A75

dp-arm:


davidcunado-arm:
Checks - passed (static analysis fails in existing code).

PR #1276, opened 2018-02-22
Make all build results depend on all makefiles

swarren:
This makes incremental builds work when the only change is to a
definition in a makefile.

Fixes arm-software/tf-issues#551

Signed-off-by: Stephen Warren <swarren@nvidia.com>

### DIFF
--- a/make_helpers/build_macros.mk
+++ b/make_helpers/build_macros.mk
@@ -199,7 +199,7 @@ $(eval OBJ := $(1)/$(patsubst %.c,%.o,$(notdir $(2))))
 $(eval DEP := $(patsubst %.o,%.d,$(OBJ)))
 $(eval IMAGE := IMAGE_BL$(call uppercase,$(3)))
 
-$(OBJ): $(2) | bl$(3)_dirs
+$(OBJ): $(2) $(MAKEFILE_LIST) | bl$(3)_dirs
 	@echo "  CC      $$<"
 	$$(Q)$$(CC) $$(TF_CFLAGS) $$(CFLAGS) -D$(IMAGE) $(MAKE_DEP) -c $$< -o $$@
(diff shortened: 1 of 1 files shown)


masahir0y:
> This makes incremental builds work when the only change is to a
definition in a makefile.

Do you do this?

I guess people tend to give options from the command line
instead of editing their platform.mk, and this PR has no help for that case.

dp-arm:
I agree with @masahir0y, not sure how useful this is in the common case where options are passed via the command line.

Generally speaking the build system could be improved in various ways and I am starting to feel uncomfortable hacking on top of it in its current state.

swarren:
> Do you do this?

Yes. This patch was written specifically because I had this issue. In my case, I was building NVIDIA's downstream branch of ATF which has at least some configuration data in the makefiles, such as the link base address. Someone changed this, which then broke peoples' incremental builds.

swarren:
@masahir0y are you OK with this patch given my answer? I understand it doesn't solve all the issues with the build system, but it does solve a real issue, which at least moves us partially towards a good state, even if there are still hole left. Thanks.

masahir0y:
> @masahir0y are you OK with this patch given my answer? I understand it doesn't solve all the issues with the build system, but it does solve a real issue, which at least moves us partially towards a good state, even if there are still hole left. Thanks.

Then, I do not mind that.

My work-flow passes options from the command line.  So, I will do clean & full-build every time.

vwadekar:
Can we get this one merged? Thanks.

davidcunado-arm:
Checks - passed; internal link http://ssg-sw.cambridge.arm.com/jenkins/view/TF_critical_CI_jobs/job/trusted-firmware-pr/1751/

PR #1277, opened 2018-02-22
hikey: migrate to BL2_EL3

hzhuang1:


hzhuang1:
If this pull request could be merged in this week, it would be great. :)

gitfineon:
(1) Why do you want to rename l-loader.bin to recovery.bin?

`make recovery` is already a flash target in the [build](https://github.com/OP-TEE/build) tools and the target to build l-loader.bin is called `make lloader`. 

(2) Is it still possible to build a lloader image with BL1 included, by setting 
`BL2_AT_EL3` to zero?

(3) I would prefer to see #1280 applied first, to have this change to the memory layout in the history of the [plat/hisilicon/hikey/include/hikey_layout.h](https://github.com/ARM-software/arm-trusted-firmware/pull/1280/files#diff-4a534b6366574e5bce27c8f83839c78c) file.

(4) A lot of functions are shared now in *bl1_setup.c and *bl2_setup.c and I think it would be cleaner to have them in a separate file.

HiKey690
```c
static void hikey960_clk_init(void);
static void hikey960_pmu_init(void);
static void hikey960_enable_ppll3(void);
static void bus_idle_clear(unsigned int value);
static void set_vivobus_power_up(void);
static void set_dss_power_up(void);
static void set_vcodec_power_up(void);
static void set_vdec_power_up(void);
static void set_venc_power_up(void);
static void set_isp_power_up(void);
static void set_ivp_power_up(void);
static void set_audio_power_up(void);
static void set_pcie_power_up(void);
static void ispfunc_enable(void);
static void isps_control_clock(int flag);
static void set_isp_srt_power_up(void);
static void hikey960_regulator_enable(void);
static void hikey960_tzc_init(void);
static void hikey960_peri_init(void);
static void hikey960_pinmux_init(void);
```
HiKey620
```c
static void hikey_sp804_init(void);
static void hikey_gpio_init(void);
static void hikey_pmussi_init(void);
static void hikey_hi6553_init(void);
static void init_mmc0_pll(void);
static void reset_mmc0_clk(void);
static void init_media_clk(void);
static void init_mmc1_pll(void);
static void reset_mmc1_clk(void);
static void hikey_mmc_pll_init(void);
static void hikey_rtc_init(void);
```

davidcunado-arm:
@hzhuang1 
Do you want to address @gitfineon comments before progressing this PR?

vchong:
> make recovery is already a flash target in the build tools and the target to build l-loader.bin is called make lloader.

The 'build tools' is a just convenience provided by the Linaro Security Working Group and shouldn't be a reason to consider for or against any particular arm-tf PR. Granted, this PR might (we don't know yet) break the 'build tools', but it still shouldn't affect arm-tf development. `make recovery` doesn't really correspond to `recovery.bin` in this case. Nothing has been done to the 'build tools' to reflect/adapt to the changes in this PR.

Will let @hzhuang1 answer the rest.

hzhuang1:
> (2) Is it still possible to build a lloader image with BL1 included, by setting BL2_AT_EL3 to zero?

It only makes code too complex.

> (3) I would prefer to see #1280 applied first, to have this change to the memory layout in the history of 
 the plat/hisilicon/hikey/include/hikey_layout.h file.

If code is in confliction, I could rebase it on #1280.

> (4) A lot of functions are shared now in *bl1_setup.c and *bl2_setup.c and I think it would be cleaner  to have them in a separate file.

I can move them in a separate file.

hzhuang1:
@davidcunado-arm @gitfineon done.

davidcunado-arm:
@hzhuang1 looks ok.
@gitfineon can you ACK?
Also, is there a dependency between this PR and #1280 ?

davidcunado-arm:
jenkins: test this please

davidcunado-arm:
Check fails - static analysis issues in existing code; ignore.

vchong:
@gitfineon So the 'build tools' did break, and is currently being updated to reflect this patch. See https://github.com/OP-TEE/build/pull/238 for details.

PR #1282, opened 2018-02-27
Misra changes

robertovargas-arm:
Changes oriented to match the C-MISRA rules about function definitions.


davidcunado-arm:
Checks - passed

PR #1285, opened 2018-02-27
FVP: restrict dynamic config to Unix build environment

soby-mathew:
This patch restricts building the dynamic config DTBs to the Unix
build environment as the Device Tree compiler may not be available
on other build environments.

Change-Id: Ie690e80010a174300e966240fd977b37561156e0


davidcunado-arm:
Checks - passed (expected issues in static analysis and ECDSA build failure due to build script issue).

PR #1287, opened 2018-02-27
Update ULL() macro and instances of ull to comply with MISRA

davidcunado-arm:
MISRA C-2012 Rule 7.3 violation: lowercase l shall not be used as literal suffixes.

This patch resolves this for the ULL() macro by using ULL suffix instead
of the ull suffix.

Change-Id: Ia8183c399e74677e676956e8653e82375d0e0a01


davidcunado-arm:
Checks - passed

PR #1289, opened 2018-02-28
Select SCMI/SDS drivers by default on Juno

sandrine-bailleux-arm:
The SCP binaries provided in the 17.10 Linaro release (and onwards)
have migrated to the SCMI/SDS protocols. Therefore, the ARM TF should
now use the corresponding drivers by default.

This patch changes the default value of the CSS_USE_SCMI_SDS_DRIVER
build option to 1 for Juno.


davidcunado-arm:
Checks - passed (static checks failed on existing code)

PR #1291, opened 2018-02-28
Revert "Make all build results depend on all makefiles"

dp-arm:
Seems to have unintended side-effects on the build system such as
rebuilding certain parts of TF even though nothing has changed.

This reverts commit c6f651f9a3322857c8e1f8250274a0984c024283.

Change-Id: I1472e6c630cb6371ec629b7d97a5748d9a6fd096

### DIFF
--- a/make_helpers/build_macros.mk
+++ b/make_helpers/build_macros.mk
@@ -199,7 +199,7 @@ $(eval OBJ := $(1)/$(patsubst %.c,%.o,$(notdir $(2))))
 $(eval DEP := $(patsubst %.o,%.d,$(OBJ)))
 $(eval IMAGE := IMAGE_BL$(call uppercase,$(3)))
 
-$(OBJ): $(2) $(MAKEFILE_LIST) | bl$(3)_dirs
+$(OBJ): $(2) | bl$(3)_dirs
 	@echo "  CC      $$<"
 	$$(Q)$$(CC) $$(TF_CFLAGS) $$(CFLAGS) -D$(IMAGE) $(MAKE_DEP) -c $$< -o $$@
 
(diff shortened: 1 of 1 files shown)


vwadekar:
@swarren 

Can you please take a look and try to fix this?

swarren:
@dp-arm, what platform are you building for, and which parts of TF are getting rebuilt that shouldn't be?

dp-arm:
@swarren I am using this build configuration, which is for Juno:

CROSS_COMPILE=aarch64-linux-gnu- BL33=<path-to-bl33.bin> SCP_BL2=scp-fw.bin make PLAT=juno DEBUG=1 USE_COHERENT_MEM=0 SPD=tspd TSP_NS_INTR_ASYNC_PREEMPT=1 NS_TIMER_SWITCH=1 all fip

Every time I run make, it rebuilds random parts of the source tree until at some point there is nothing more to build.

swarren:
Ah, I see. Deleting the build directory then building builds everything, then building again builds nothing which is fine, and editing a Makefile rebuilds everything, then building again builds nothing which is fine (those cases are what I tested), but if I then touch a .c file and build, then it keeps building something every time make is run which is bad. I'll look into it.

dp-arm:
@swarren thanks!

swarren:
I believe if you replace all $(MAKEFILE_LIST) with $(filter-out %.d,$(MAKEFILE_LIST)) that should fix it. If so, I'll send a PR.

swarren:
https://github.com/ARM-software/arm-trusted-firmware/pull/1293

davidcunado-arm:
Check - passed

PR #1292, opened 2018-02-28
 Suppress spurious deprecated declaration warnings

danh-arm:


masahir0y:
This PR cleans up my build log, but I still wonder
what the point of `__deprecated` for `plat_get_syscnt_freq` is.

I am afraid you are confused between two cases:
[1] platform is calling deprecated APIs
[2] platform is defining deprecated platform hooks

'deprecated' attribute is only helpful for [1] because it warns callers parts.

I am pasting the description from GCC manual:
------------------------------------------------------>8--------------------------------------------
deprecated
The deprecated attribute results in a warning if the function is used anywhere in the source file. This is useful when identifying functions that are expected to be removed in a future version of a program. The warning also includes the location of the declaration of the deprecated function, to enable users to easily find further information about why the function is deprecated, or what they should do instead. Note that the warnings only occurs for uses:
------------------------------------------------------>8--------------------------------------------

plat_get_syscnt_freq() is case [2].

I replaced my `plat_get_syscnt_freq2()` with old `plat_get_syscnt_freq`,
and I did not see any warning about that.

So, the right thing is not to suppress it,
but to remove the pointless __deprecated.

The following code is equivalent to your luxury GCC diagnostic implementation.

     diff --git a/include/plat/common/platform.h b/include/plat/common/platform.h
     index 411202d..082c0de 100644
     --- a/include/plat/common/platform.h
     +++ b/include/plat/common/platform.h
     @@ -40,7 +40,7 @@ struct secure_partition_boot_info;
     /*******************************************************************************
      * Mandatory common functions
      ******************************************************************************/
    -unsigned long long plat_get_syscnt_freq(void) __deprecated;
    +unsigned long long plat_get_syscnt_freq(void);
     unsigned int plat_get_syscnt_freq2(void);
 
    int plat_get_image_source(unsigned int image_id,

If you warn case [2], you need to invent a different way.

For example, add

    WARN("plat_get_syscnt_freq is deprecated. Please define plat_get_syscnt_freq2.")

to the weak fallback of plat_get_syscnt_freq2().

If you want to check it build-time, run 'nm' and check the symbol, etc.

masahir0y:
> If you warn case [2], you need to invent a different way.

Or, maybe this is the case where we consider a per-feature deprecate flag.
For example,

   USE_PLAT_GET_SYSCNT_FREQ

This is set to '1' in defaults.mk
The weak variant of plat_get_syscnt_freq2 is compiled
with warning.

If a platform implements its own plat_get_syscnt_freq2,
platform.mk can set `USE_PLAT_GET_SYSCNT_FREQ := 0`
to compile out the weak variant.

Another idea is to remove the plat_get_syscnt_freq support entirely after the release cycle.
The upstream code base completed the migration.  We have no way to know about the downstream code, but it would be trivial to fix the code.  Actually, it is painful to keep fallback code, but this is up to your policy.

danh-arm:
Thanks for the feedback

> I am afraid you are confused between two cases:
> [1] platform is calling deprecated APIs
> [2] platform is defining deprecated platform hooks
>
> 'deprecated' attribute is only helpful for [1] because it warns callers parts.
>
> plat_get_syscnt_freq() is case [2].

Actually, plat_get_syscnt_freq() is both case [1] and case [2]. Although no upstream platforms call this API, they are allowed to. But you are right that we have no notification solution for [2].

> So, the right thing is not to suppress it,
> but to remove the pointless __deprecated.

I wouldn't say it's pointless (due to this API theoretically being used) but yes it's low value. However, now it's done I don't think it's worth going back.

> If you warn case [2], you need to invent a different way.
> For example, add
```
WARN("plat_get_syscnt_freq is deprecated. Please define plat_get_syscnt_freq2.")
```
> to the weak fallback of plat_get_syscnt_freq2().

Yes, this might be the most pragmatic solution, although I'd prefer something that works at build time. We have a similar issue for the deprecated fields in the `gicv3_driver_data` and `gicv2_driver_data` structures.

Unless I hear a better solution soon, I'll add that `WARN` to this PR.

> If you want to check it build-time, run 'nm' and check the symbol, etc.

If something along these lines can be automated and made robust, then I'd welcome this as a contribution.

> Or, maybe this is the case where we consider a per-feature deprecate flag.

I think this is a bit heavyweight for a simple API like this. We want to avoid a proliferation of build flags. If this were a collection of APIs or an API with no compatiblity layer, this might be OK.

> Another idea is to remove the plat_get_syscnt_freq support entirely after the release cycle.
The upstream code base completed the migration. We have no way to know about the downstream code, but it would be trivial to fix the code. Actually, it is painful to keep fallback code, but this is up to your policy.

We intend to remove this and all code under !ERROR_DEPRECATED when we move to version 2.0, but not until then. We're not ready to do that yet as we're expecting the dynamic config work to introduce more API changes.

danh-arm:
> Although no upstream platforms call this API...

Actually I've just noticed Hikey960 calls `plat_get_syscnt_freq2()` in its `bl1_plat_set_ep_info()`.

masahir0y:
> Unless I hear a better solution soon, I'll add that WARN to this PR.

Sounds good.

> If something along these lines can be automated and made robust, then I'd welcome this as a contribution.

I am less interested in this.

> Actually I've just noticed Hikey960 calls plat_get_syscnt_freq2() in its bl1_plat_set_ep_info().

You are right.  So, keeping __deprecated is reasonable.

danh-arm:
I've just pushed a new version, including another patch to emit WARNs for deprecated GIC driver init.

masahir0y:
Not fully tested, but looks good to me. Thanks!

davidcunado-arm:
Checks - passed

PR #1293, opened 2018-02-28
Don't make build results depend on dependency files

swarren:
This prevents the build system from continually rebuilding random parts
of the source tree due to timestamps on generated dependency makefiles.

Continues work on arm-software/tf-issues#551

Signed-off-by: Stephen Warren <swarren@nvidia.com>


davidcunado-arm:
@swarren 
We're about to start release activities - can we hold this change back until this has completed? With that in mind, I'll merge @dp-arm PR - this one will then need to be updated.

dp-arm:
@swarren Yes indeed, rebase on current integration branch and we will review/merge this patch after the release.

swarren:
I've pushed a new squashed/fixed patch version; that's not terribly obvious from the github UI in the pull request.

davidcunado-arm:
jenkins: test this please

PR #1294, opened 2018-03-02
Change PLATFORM_ROOT to TF_PLATFORM_ROOT

iwishguo:
Since we use "?=" to set PLATFORM_ROOT, it is better to change the
name to be more special, or else it may override by some environment
variables, such as in some CI build environments.

Signed-off-by: Heyi Guo <heyi.guo@linaro.org>

### DIFF
--- a/make_helpers/plat_helpers.mk
+++ b/make_helpers/plat_helpers.mk
@@ -15,14 +15,14 @@ ifndef PLAT_HELPERS_MK
         $(error "Error: Unknown platform. Please use PLAT=<platform name> to specify the platform")
     endif
 
-    # PLATFORM_ROOT can be overridden for when building tools directly
-    PLATFORM_ROOT               ?= plat/
+    # TF_PLATFORM_ROOT can be overridden for when building tools directly
+    TF_PLATFORM_ROOT               ?= plat/
     PLAT_MAKEFILE               := platform.mk
 
     # Generate the platforms list by recursively searching for all directories
     # under /plat containing a PLAT_MAKEFILE. Append each platform with a `|`
     # char and strip out the final '|'.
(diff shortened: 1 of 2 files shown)


davidcunado-arm:
@iwishguo 
As per the contributing guidelines, please target the integration branch (rather than the master branch). I've changed this PR for now.

iwishguo:
@davidcunado-arm Thanks; will do that next time :)

masahir0y:
Is there a policy for prefix?

I see some instances with TF_, such as `TF_CFLAGS`, `TF_LDFLAGS`.

I see nothing with ATF_ except vendor platform code.

iwishguo:
Sorry, I don't know if there is any policy, but prefix of "TF_" should also be OK for me.
Thanks.

dp-arm:
@iwishguo Have you already seen an issue where this accidental override happened in practice?

iwishguo:
@dp-arm Yes, we saw name conflict in Huawei internal code CI environment. The patch is only a suggestion. If you think it is not good to make such change, we can still change our build scripts (unset PLATFORM_ROOT before running make in TF) to get around of this.
Thanks.

iwishguo:
@dp-arm Yes, we saw name conflict in Huawei internal code CI environment. The patch is only a suggestion. If you think it is not good to make such change, we can still change our build scripts (unset PLATFORM_ROOT before running make in TF) to get around of this.
Thanks.

dp-arm:
@iwishguo ok, it makes sense I think.

dp-arm:
@davidcunado-arm I would hold this change back until after the release.

iwishguo:
@dp-arm Thanks. We can apply to our internal repository first for more test.

davidcunado-arm:
jenkins: test this please

PR #1296, opened 2018-03-02
Fix FVP DRAM2 size

robertovargas-arm:
This was correct according to the model specifications , but it seems
that FVP doesn't implement it. It is safer to use the size exposed by
the DTB which is currently used by Linux.

Change-Id: I9aabe3284a50ec2a36ed94966eb7e4ddf37cec3b


davidcunado-arm:
Checks - passed

PR #1297, opened 2018-03-02
Remove sp_min functions from plat_common.c

soby-mathew:
This patch removes default platform implementations of sp_min
platform APIs from plat/common/aarch32/plat_common.c. The APIs
are now implemented in `plat_sp_min_common.c` file within the
same folder.

The ARM platform layer had a weak definition of sp_min_platform_setup2()
which conflicted with the weak definition in the common file. Hence this
patch fixes that by introducing a `plat_arm_` version of the API thus
allowing individual boards within ARM platforms to override it if they
wish to.

Fixes ARM-Software/tf-issues#559

Change-Id: I11a74ecae8191878ccc7ea03f12bdd5ae88faba5


davidcunado-arm:
Checks - passed

PR #1298, opened 2018-03-04
qemu: Accessing UART1 causes a data abort

michpappas:
The register address range of UART1 (crash console) are outside the
address ranges mapped for MMIO, resulting to an MMU abort when the
device registers are accessed.

Increase the size of DEVICE1 memory to include the range of UART1.

Fixes ARM-software/tf-issues#560

Signed-off-by: Michalis Pappas <mpappas@fastmail.fm>


davidcunado-arm:
@jenswi-linaro 
Looks pretty straight forward - will merge MON / TUE unless I hear otherwise from you.

jenswi-linaro:
Looks good to me.

davidcunado-arm:
Checks - passed

PR #1299, opened 2018-03-04
qemu: Support MULTI_CONSOLE_API

michpappas:
Hi,

I moved all console initialization to C, as I found it cleaner than the way recommended in the Porting Guide, ie
```
.section .data.crash_console      /* Reserve space for console structure */
crash_console:
.zero 6 * 8                       /* console_16550_t has 6 8-byte words */
```

### DIFF
--- a/drivers/arm/pl011/aarch64/pl011_console.S
+++ b/drivers/arm/pl011/aarch64/pl011_console.S
@@ -6,6 +6,7 @@
 #include <arch.h>
 #include <asm_macros.S>
 #include <assert_macros.S>
+#include <console_macros.S>
 #include <pl011.h>
 
 /*
--- a/plat/qemu/qemu_console.c
+++ b/plat/qemu/qemu_console.c
@@ -1,0 +1,37 @@
+/*
+ * Copyright (c) 2015-2018, ARM Limited and Contributors. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+#include <console.h>
+#include <pl011.h>
+#include <platform_def.h>
+
+static console_pl011_t console;
+static console_pl011_t crash_console;
+
+void qemu_console_init(void)
+{
+#if MULTI_CONSOLE_API
+	(void)console_pl011_register(PLAT_QEMU_BOOT_UART_BASE,
+			       PLAT_QEMU_BOOT_UART_CLK_IN_HZ,
+			       PLAT_QEMU_CONSOLE_BAUDRATE, &console);
+#else
+	console_init(PLAT_QEMU_BOOT_UART_BASE,
+		     PLAT_QEMU_BOOT_UART_CLK_IN_HZ,
+		     PLAT_QEMU_CONSOLE_BAUDRATE);
+#endif /* MULTI_CONSOLE_API */
+}
+
+void qemu_crash_console_init(void)
+{
+#if MULTI_CONSOLE_API
+	(void)console_pl011_register(PLAT_QEMU_CRASH_UART_BASE,
+			       PLAT_QEMU_CRASH_UART_CLK_IN_HZ,
+			       PLAT_QEMU_CONSOLE_BAUDRATE, &crash_console);
+#else
+	console_core_init(PLAT_QEMU_CRASH_UART_BASE,
+			  PLAT_QEMU_CRASH_UART_CLK_IN_HZ,
+			  PLAT_QEMU_CONSOLE_BAUDRATE);
+#endif /* MULTI_CONSOLE_API */
+}
--- a/plat/qemu/aarch64/plat_helpers.S
+++ b/plat/qemu/aarch64/plat_helpers.S
@@ -14,11 +14,12 @@
 	.globl	platform_mem_init
 	.globl	plat_qemu_calc_core_pos
 	.globl	plat_crash_console_init
+#if MULTI_CONSOLE_API
 	.globl	plat_crash_console_putc
+#endif /* MULTI_CONSOLE_API */
 	.globl  plat_secondary_cold_boot_setup
 	.globl  plat_get_my_entrypoint
 	.globl  plat_is_my_cpu_primary
-
 
 func plat_my_core_pos
 	mrs	x0, mpidr_el1
@@ -96,10 +97,7 @@ endfunc platform_mem_init
 	 * ---------------------------------------------
 	 */
 func plat_crash_console_init
-	mov_imm	x0, PLAT_QEMU_CRASH_UART_BASE
-	mov_imm	x1, PLAT_QEMU_CRASH_UART_CLK_IN_HZ
-	mov_imm	x2, PLAT_QEMU_CONSOLE_BAUDRATE
-	b	console_core_init
+	b	qemu_crash_console_init
 endfunc plat_crash_console_init
 
 	/* ---------------------------------------------
@@ -109,9 +107,10 @@ endfunc plat_crash_console_init
 	 * Clobber list : x1, x2
 	 * ---------------------------------------------
 	 */
+#if !MULTI_CONSOLE_API
 func plat_crash_console_putc
 	mov_imm	x1, PLAT_QEMU_CRASH_UART_BASE
 	b	console_core_putc
 endfunc plat_crash_console_putc
-
+#endif /* MULTI_CONSOLE_API */
 
--- a/plat/qemu/platform.mk
+++ b/plat/qemu/platform.mk
@@ -47,8 +47,9 @@ $(eval $(call assert_boolean,ARM_XLAT_TABLES_LIB_V1))
 $(eval $(call add_define,ARM_XLAT_TABLES_LIB_V1))
 
 
-PLAT_BL_COMMON_SOURCES	:=	plat/qemu/qemu_common.c			\
-				drivers/arm/pl011/${ARCH}/pl011_console.S
+PLAT_BL_COMMON_SOURCES	:=	plat/qemu/qemu_common.c			  \
+				plat/qemu/qemu_console.c		  \
+				drivers/arm/pl011/${ARCH}/pl011_console.S \
 
 ifeq (${ARM_XLAT_TABLES_LIB_V1}, 1)
 PLAT_BL_COMMON_SOURCES	+=	lib/xlat_tables/xlat_tables_common.c		\
@@ -168,6 +169,7 @@ $(eval $(call TOOL_ADD_IMG,bl32_extra2,--tos-fw-extra2))
 endif
 
 SEPARATE_CODE_AND_RODATA := 1
+MULTI_CONSOLE_API	 := 1
 
 # Disable the PSCI platform compatibility layer
 ENABLE_PLAT_COMPAT	:= 	0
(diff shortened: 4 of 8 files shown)


davidcunado-arm:
@jenswi-linaro 
Please ACK?

jenswi-linaro:
Is this a direction ARM-TF is moving towards and the QEMU port just happens to be the first?

davidcunado-arm:
The plan is to have an Arm platform use this framework - we've just not got round to planning it in.

jenswi-linaro:
@davidcunado-arm, thanks. Fine with me then.

davidcunado-arm:
jenkins: test this please

plat/qemu/qemu_console.c
```diff
@@ -0,0 +1,37 @@
...
+#include <pl011.h>
+#include <platform_def.h>
+
+static console_pl011_t console;
+static console_pl011_t crash_console;
+
+void qemu_console_init(void)
+{
+#if MULTI_CONSOLE_API
+	console_pl011_register(PLAT_QEMU_BOOT_UART_BASE,
```
davidcunado-arm: Static checks fail: 

 [Required] MISRA C-2012 Rule 17.7 violation: Return value of non-void function console_pl011_register() is not used.

Add (void) here should eliminate the fail.

plat/qemu/qemu_console.c
```diff
@@ -0,0 +1,37 @@
...
+	console_init(PLAT_QEMU_BOOT_UART_BASE,
+		     PLAT_QEMU_BOOT_UART_CLK_IN_HZ,
+		     PLAT_QEMU_CONSOLE_BAUDRATE);
+#endif /* MULTI_CONSOLE_API */
+}
+
+void qemu_crash_console_init(void)
+{
+#if MULTI_CONSOLE_API
+	console_pl011_register(PLAT_QEMU_CRASH_UART_BASE,
```
davidcunado-arm: same issue to line 16 - 

 [Required] MISRA C-2012 Rule 17.7 violation: Return value of non-void function console_pl011_register() is not used.

michpappas:
@davidcunado-arm fixed

davidcunado-arm:
jenkins: test this please

davidcunado-arm:
Check fails - issues in existing code; ignore for this PR.

drivers/arm/pl011/aarch64/pl011_console.S
```diff
@@ -6,6 +6,7 @@
 #include <arch.h>
 #include <asm_macros.S>
 #include <assert_macros.S>
+#include <console_macros.S>
```
jwerner-chromium: Whoops, this must have been an oversight in the original multi-console patch. Thanks for the fix!

plat/qemu/qemu_console.c
```diff
@@ -0,0 +1,37 @@
...
+			       PLAT_QEMU_BOOT_UART_CLK_IN_HZ,
+			       PLAT_QEMU_CONSOLE_BAUDRATE, &console);
+#else
+	console_init(PLAT_QEMU_BOOT_UART_BASE,
+		     PLAT_QEMU_BOOT_UART_CLK_IN_HZ,
+		     PLAT_QEMU_CONSOLE_BAUDRATE);
+#endif /* MULTI_CONSOLE_API */
+}
+
+void qemu_crash_console_init(void)
```
jwerner-chromium: Note that the reason all the crash console stuff is implemented in assembly is that you do not have a full C runtime environment available when you're running in one of the crash callbacks (like plat_crash_console_init). For example, you're not guaranteed to have a valid stack pointer. If this code happens to work for you, you're probably just getting lucky that the compiler notices you're doing a tail call from a void function (qemu_crash_console_init) to another void function (console_pl011_register) and eliminates the return address spilling. It's not something you should rely on and it can break at any time (e.g. when changing optimization flags). Or maybe you're just using an uninitialized stack pointer here that just happened to point to valid memory in your case.

The good news is that you don't really need to do this at crash time anyway. With the multi-console framework, you can register a crash-console ahead of time (e.g. during boot) and configure it so that it is only used for crashes. You should be able to just move the console_pl011_register() call from here into qemu_console_init(), and then add another line that does console_set_scope(&crash_console, CONSOLE_FLAG_CRASH); below it. Then your crash console should just work if you don't override plat_crash_console_init at all and just use the default version from plat/common. (You might also want to do a separate console_set_scope(&console, CONSOLE_BOOT), unless you want the crash output on both your crash console and your normal console.)

plat/qemu/qemu_console.c
```diff
@@ -0,0 +1,37 @@
...
+#include <console.h>
+#include <pl011.h>
+#include <platform_def.h>
+
+static console_pl011_t console;
+static console_pl011_t crash_console;
+
+void qemu_console_init(void)
+{
+#if MULTI_CONSOLE_API
```
jwerner-chromium: You shouldn't really need the #if here or in any of the other plat/qemu/* files. If you're defining MULTI_CONSOLE_API to be 1 in the platform.mk, it should be okay to just assume that it's 1 in all corresponding code.

plat/qemu/qemu_console.c
michpappas: > Note that the reason all the crash console stuff is implemented in assembly is that you do not have a full C runtime environment available when you're running in one of the crash callbacks (like plat_crash_console_init). For example, you're not guaranteed to have a valid stack pointer. If this code happens to work for you, you're probably just getting lucky that the compiler notices you're doing a tail call from a void function (qemu_crash_console_init) to another void function (console_pl011_register) and eliminates the return address spilling. It's not something you should rely on and it can break at any time (e.g. when changing optimization flags). Or maybe you're just using an uninitialized stack pointer here that just happened to point to valid memory in your case.

@jwerner-chromium the assumption is that this would work because of the AAPCS64, and the stack would not be involved at all. But you're right about the return address spilling part when performing nested calls. Still, I'm not really keen of the idea of hardcoding the size of console_pl011_t. Some other means to reliably get the size of console_pl011_t (and friends) from assembly is needed, and I am not sure what an elegant way to achieve that would be.

> The good news is that you don't really need to do this at crash time anyway. With the multi-console framework, you can register a crash-console ahead of time (e.g. during boot) and configure it so that it is only used for crashes. You should be able to just move the console_pl011_register() call from here into qemu_console_init(), and then add another line that does console_set_scope(&crash_console, CONSOLE_FLAG_CRASH); below it. Then your crash console should just work if you don't override plat_crash_console_init at all and just use the default version from plat/common. (You might also want to do a separate console_set_scope(&console, CONSOLE_BOOT), unless you want the crash output on both your crash console and your normal console.)

That's all true, but the whole case is about early crash console :-) After all, from what I see in finish_console_register(), the default flags are CONSOLE_FLAG_BOOT | CONSOLE_FLAG_CRASH, so no need to additionally perform a console_set_scope(). 
Another thing to mention here is that I am trying to maintain backwards compatibility with the previous behaviour of QEMU to send crash output to a different console, ie UART1 (nevermind that this never worked due to ARM-software/tf-issues#560). In the current case the crash output is being sent to both UART0 and UART1.

michpappas:
@davidcunado-arm this haven't made it into master yet, but still I'm not sure what's the best way to handle potential changes based on the feedback above as well as https://github.com/ARM-software/tf-issues/issues/571. Revert this from integration and push a new patch or just carry on and address issues individually?

plat/qemu/qemu_console.c
jwerner-chromium: > Some other means to reliably get the size of console_pl011_t (and friends) from assembly is needed, and I am not sure what an elegant way to achieve that would be.

Hmmm... okay, I didn't presume this would be a problem because I didn't think those structures would really change much. But if you think it is, I would suggest you define a PL011_CONSOLE_SIZE in pl011.h and then add a CASSERT() to ensure that it will always equal sizeof(console_pl011_t).

> That's all true, but the whole case is about early crash console :-)

Oh okay, that's a legitimate use case. I would've hoped that most platforms don't usually have to deal with crashes that early (at least once their initial platform port is stable) and could use the simpler path. But if you think it's needed for QEMU it's fine to write it that way.

> In the current case the crash output is being sent to both UART0 and UART1.

Note that your code is also overriding plat_crash_console_init() right now, so the framework doesn't really work quite right because it never switches to the CRASH state. You would only get output on UART0 (or UART1 if you don't console_set_scope() it explicitly) if you're still in the BOOT state, not if it had already moved to RUNTIME. It's not super important, but the really clean way to do it would be to also call console_switch_state() from your custom plat_crash_console_init() implementation.

Also note that if you are worried about *really* early crashes (like "memory might not be writeable"-early), this solution may still not work because you're still using the console framework which needs some global variables to work. So if you want custom plat_crash_console implementations anyway, you might as well ignore the console framework completely and just call console_pl011_core_init/putc/flush directly the way it was done before the multi-console API -- we intentionally kept those functions separately available to support that. This way you can get console output while only operating on registers.

michpappas: > Hmmm... okay, I didn't presume this would be a problem because I didn't think those structures would really change much. But if you think it is, I would suggest you define a PL011_CONSOLE_SIZE in pl011.h and then add a CASSERT() to ensure that it will always equal sizeof(console_pl011_t).

Agreed, I don't think the structures will change much either, but I can foresee the case when someone adds a feature to a console driver and updates the structure without knowing that its size is hardcoded somewhere.

Another concern of mine is that there could also be issues when building for AArch32 and AArch64, as, depending on the types contained within the structs, the size in each case could be different (http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dai0490a/ar01s01.html).

Defining PL011_CONSOLE_SIZE addresses the former in a nice way, but not the latter.

> Also note that if you are worried about really early crashes

TBH I'm not really worried about early crashes altogether. Just thought that if I'd do it I should do it right :-) But if it ends up perplexing things too much, then it might as well be worth to keep things simple and not implement early crash support in qemu altogether, as, being an emulator, equivalent debug information can be obtained through other means anyway.

jwerner-chromium: > Another concern of mine is that there could also be issues when building for AArch32 and AArch64, as, depending on the types contained within the structs, the size in each case could be different (http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dai0490a/ar01s01.html).

I paid attention to make all members of console structures machine-word sized (since most of them are pointers anyway). If we continue this scheme, it should work well once someone ports the API to AArch32. So maybe rather than defining a PL011_CONSOLE_SIZE, it should be PL011_CONSOLE_WORDS, and then CASSERT(PL011_CONSOLE_WORDS == sizeof(console_pl011_t) / sizeof(u_register_t)) works on both architectures.

> But if it ends up perplexing things too much, then it might as well be worth to keep things simple and not implement early crash support in qemu altogether

Agreed... I think if you don't expect you'll need it, you should just stick with the default implementation which was intended to give you 95% of all crash debugging you'll ever need automatically without having to implement all these special platform overrides by hand.

davidcunado-arm:
@michpappas 
Do you have a fix for  ARM-software/tf-issues#571 ready? If you get that PR ready we can prioritise and merge into master together - i'd rather avoid reverting if possible.
The other issues we can deal with individually.

michpappas:
@davidcunado-arm opened a separate PR for the fix

PR #1300, opened 2018-03-05
Dynamic cfg: Do not populate args if already initialized

davidcunado-arm:
This patch modifies the common utility function
`populate_next_bl_params_config()` to only modify the entrypoint arguments
to an executable image only if they are not initialized earlier.
This issue was detected while testing Optee on ARM platforms which needed
the current arguments to be preserved in the absence of corresponding
config files.

Change-Id: I1e3fb4be8176fc173959e72442396dd33a99a316


davidcunado-arm:
Checks - passed

PR #1301, opened 2018-03-05
bl2-el3: Fix bl32 lr_svc used for bl33 entry address

ldebieve:
When using BL2_EL3, we need to ensure that lr_svc is
properly given to bl32 as it was previously made by bl1.

Fixes ARM-Software/tf-issues#562

Signed-off-by: Lionel Debieve <lionel.debieve@st.com>


davidcunado-arm:
Checks - passed

PR #1303, opened 2018-03-07
Juno: Fixes for firmware update

soby-mathew:


davidcunado-arm:
Check fails - static analysis issues in existing code; ignore.

PR #1305, opened 2018-03-14
Implement support for v1.2 of firmware interfaces spec (ARM DEN 0070A)

dp-arm:


davidcunado-arm:
Check failure - static analysis error but on review can be ignored (false positive).

PR #1309, opened 2018-03-15
plat/common: remove fall-through on release build

jonathanwright-ARM:
Removes fall-through in switch statement on unknown interrupt type in
release builds.

Previous behaviour was to assert(0) on default case in debug builds but
fall through and interpret the unknown interrupt type as
INTR_TYPE_EL3 in release builds.

Change-Id: I05fb0299608efda0f9eda2288d3e56e5625e05c9
Signed-off-by: Jonathan Wright <jonathan.wright@arm.com>


davidcunado-arm:
jenkins: add to whitelist please

PR #1311, opened 2018-03-15
stdlib: remove comparison with EOF macro to comply with MISRA 

jonathanwright-ARM:


davidcunado-arm:
jenkins: test this please

PR #1314, opened 2018-03-16
Rename 'smcc' to 'smccc'

antonio-nino-diaz-arm:
When the source code says 'SMCC' it is talking about the SMC Calling Convention. The correct acronym is SMCCC. This affects a few definitions and file names.

Some files have been renamed (smcc.h, smcc_helpers.h and smcc_macros.S) but the old files have been kept for compatibility, they include the new ones with an ERROR_DEPRECATED guard.


davidcunado-arm:
@antonio-nino-diaz-arm 
Could you rebase on latest master? Looks like there's a minor 'issue' where this patch is reverting "Arm" back to "ARM" in psci-lib-integration-guide.rst.

antonio-nino-diaz-arm:
@davidcunado-arm Done!

PR #1321, opened 2018-03-19
Trusty: Fix sanity check on NS entry point

sandrine-bailleux-arm:
This patch fixes the sanity check on the non-secure entrypoint value
returned by bl31_plat_get_next_image_ep_info(). This issue has been
reported by Coverity Scan Online:

  CID 264893 (#1 of 1): Dereference null return value (NULL_RETURNS)
  Dereferencing a null pointer ns_ep_info.

### DIFF
--- a/services/spd/trusty/trusty.c
+++ b/services/spd/trusty/trusty.c
@@ -451,7 +451,7 @@ static int32_t trusty_setup(void)
 		uint32_t spsr;
 
 		ns_ep_info = bl31_plat_get_next_image_ep_info(NON_SECURE);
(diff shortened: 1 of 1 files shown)


sandrine-bailleux-arm:
@vwadekar Could you please have a look at this patch and approve it if you're happy with it?

sandrine-bailleux-arm:
@vwadekar Thanks for the approval. That said, Coverity complained about the ns_ep_info pointer being used as a boolean expression so I've reworked the patch. Could you please have another look?

PR #1323, opened 2018-03-20
rockchip/rk3399: save/restore watchdog register correctly

rkchrome:
there are two fix for save/restore watchdog register:
1. watchdog plck will shutdown after secure_watchdog_disable(), so need
   to save register before it and restore after secure_watchdog_enable().
2. need write 0x76 to cnt_restart to keep watchdog alive when restore
   watchdog register.

Change-Id: I1f6fbceae22186e3b72a87df6332a110adf37479
Signed-off-by: Lin Huang <hl@rock-chips.com>


davidcunado-arm:
jenkins: test this please

PR #1328, opened 2018-03-21
Add support for BL2 in XIP memory

JiafeiPan:
When bl2 runs in el3, it will be the first TF image to
be executed, so in some cases bl2 will be in ROM just
like bl1. This patch enable bl2 in ROM in case of it
runs in el3.

Signed-off-by: Jiafei Pan <Jiafei.Pan@nxp.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -373,6 +373,11 @@ ifneq ($(MULTI_CONSOLE_API), 0)
     endif
 endif
 
+#For now, BL2_IN_XIP_MEM is only supported when BL2_AT_EL3 is 1.
+ifeq ($(BL2_AT_EL3)-$(BL2_IN_XIP_MEM),0-1)
+$(error "BL2_IN_XIP_MEM is only supported when BL2_AT_EL3 is enabled")
+endif
+
 ################################################################################
 # Process platform overrideable behaviour
 ################################################################################
@@ -518,6 +523,7 @@ $(eval $(call assert_boolean,USE_COHERENT_MEM))
 $(eval $(call assert_boolean,USE_TBBR_DEFS))
 $(eval $(call assert_boolean,WARMBOOT_ENABLE_DCACHE_EARLY))
 $(eval $(call assert_boolean,BL2_AT_EL3))
+$(eval $(call assert_boolean,BL2_IN_XIP_MEM))
 
 $(eval $(call assert_numeric,ARM_ARCH_MAJOR))
 $(eval $(call assert_numeric,ARM_ARCH_MINOR))
@@ -564,6 +570,7 @@ $(eval $(call add_define,USE_COHERENT_MEM))
 $(eval $(call add_define,USE_TBBR_DEFS))
 $(eval $(call add_define,WARMBOOT_ENABLE_DCACHE_EARLY))
 $(eval $(call add_define,BL2_AT_EL3))
+$(eval $(call add_define,BL2_IN_XIP_MEM))
 
 # Define the EL3_PAYLOAD_BASE flag only if it is provided.
 ifdef EL3_PAYLOAD_BASE
--- a/bl2/bl2_el3.ld.S
+++ b/bl2/bl2_el3.ld.S
@@ -12,15 +12,26 @@ OUTPUT_ARCH(PLATFORM_LINKER_ARCH)
 ENTRY(bl2_entrypoint)
 
 MEMORY {
+#if BL2_IN_XIP_MEM
+    ROM (rx): ORIGIN = BL2_RO_BASE, LENGTH = BL2_RO_LIMIT - BL2_RO_BASE
+    RAM (rwx): ORIGIN = BL2_RW_BASE, LENGTH = BL2_RW_LIMIT - BL2_RW_BASE
+#else
     RAM (rwx): ORIGIN = BL2_BASE, LENGTH = BL2_LIMIT - BL2_BASE
+#endif
 }
 
 
 SECTIONS
 {
+#if BL2_IN_XIP_MEM
+    . = BL2_RO_BASE;
+    ASSERT(. == ALIGN(PAGE_SIZE),
+           "BL2_RO_BASE address is not aligned on a page boundary.")
+#else
     . = BL2_BASE;
     ASSERT(. == ALIGN(PAGE_SIZE),
            "BL2_BASE address is not aligned on a page boundary.")
+#endif
 
 #if SEPARATE_CODE_AND_RODATA
     .text . : {
@@ -33,7 +44,11 @@ SECTIONS
         *(.vectors)
         . = NEXT(PAGE_SIZE);
         __TEXT_END__ = .;
+#if BL2_IN_XIP_MEM
+     } >ROM
+#else
      } >RAM
+#endif
 
     .rodata . : {
         __RODATA_START__ = .;
@@ -56,7 +71,11 @@ SECTIONS
 
         . = NEXT(PAGE_SIZE);
         __RODATA_END__ = .;
+#if BL2_IN_XIP_MEM
+    } >ROM
+#else
     } >RAM
+#endif
 
     ASSERT(__TEXT_RESIDENT_END__ - __TEXT_RESIDENT_START__ <= PAGE_SIZE,
           "Resident part of BL2 has exceeded its limit.")
@@ -95,11 +114,21 @@ SECTIONS
         . = NEXT(PAGE_SIZE);
 
         __RO_END__ = .;
+#if BL2_IN_XIP_MEM
+    } >ROM
+#else
     } >RAM
+#endif
 #endif
 
     ASSERT(__CPU_OPS_END__ > __CPU_OPS_START__,
           "cpu_ops not defined for this platform.")
+
+#if BL2_IN_XIP_MEM
+    . = BL2_RW_BASE;
+    ASSERT(BL2_RW_BASE == ALIGN(PAGE_SIZE),
+           "BL2_RW_BASE address is not aligned on a page boundary.")
+#endif
 
     /*
      * Define a linker symbol to mark start of the RW memory area for this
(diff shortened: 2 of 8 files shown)


antonio-nino-diaz-arm:
Can't the current linkerscript be modified with some ifdefs with `BL2_AT_EL3_ROM`? I don't like the idea of having 2 different files that are basically the same with some small differences...

JiafeiPan:
Hi, Antonio,
Thanks for your comments, I have updated the patch according to your comments. please review again.

Best Regards,
Jiafei.

antonio-nino-diaz-arm:
This looks good to me, but someone else should take a look.

antonio-nino-diaz-arm:
IP Review +1

JiafeiPan:
Thanks Antonio, have updated copyright years.

danh-arm:
I don't understand the need for this patch. If you want to run in ROM, you should use BL1. If there is some feature of BL2 you want in BL1, we should be looking to move that functionality there. What is it in BL2 that you need?

I know there are some differences between the image loading behaviour in BL1 and BL2; is that the rationale for this?

JiafeiPan:
Hi, danh-arm, thanks for your comments. Layerscape already has an internal bootrom, and it support Nor flash boot and SD, usb and the other boot. For SD, usb and all the other boot, internal bootrom will load the first TF-A images into RAM and then execute it in RAM, so we can use BL2 as the first TF image which runs in RAM. But for nor flash boot, internal boot rom will not load the TF-A imges into RAM and will execute it in nor flash directly. Currently we use BL1 as the first TF-A images, but if BL2 can run in the ROM,  we can also remove BL1 and use BL2 as the first image which runs in ROM directly, so that we can save more internal RAM for the following booting stage. This is our use case which need to run BL2 in ROM. Any comments is welcome, thanks.

danh-arm:
Hi @JiafeiPan.

Thanks for the explanation. I understand your use-case now and it makes sense. However, I think it's confusing to use the term "ROM" in this use-case, because if you were able to program this as a real ROM in the factory, then you would also be able use TF BL1 as your real ROM, instead of your existing internal bootrom. I think it's better to use the term "eXecute In Place (XIP) memory" instead, to make it clear that BL2 is stored in XIP (but updateable) memory.

Other comments to your patch will follow...

Regards

Dan.

Makefile
```diff
@@ -518,6 +518,7 @@ $(eval $(call assert_boolean,USE_COHERENT_MEM))
 $(eval $(call assert_boolean,USE_TBBR_DEFS))
 $(eval $(call assert_boolean,WARMBOOT_ENABLE_DCACHE_EARLY))
 $(eval $(call assert_boolean,BL2_AT_EL3))
+$(eval $(call assert_boolean,BL2_AT_EL3_ROM))
```
danh-arm: I think the new flag should be called BL2_IN_XIP_MEM. The term ROM is confusing in this context. Also, this feature potentially could be used when BL2 is in Secure-EL1, so this should be orthgonal to the BL2_AT_EL3 feature. Having said that, for now you could avoid supporting the case where BL2_AT_EL3 = 0 and BL2_IN_XIP_MEM = 1, by throwing an error in the top level Makefile (in the "Check incompatible options" section).

Given the above, I think your commit message should look something like this:

Add support for BL2 in XIP memory

In some use-cases BL2 will be stored in eXecute In Place (XIP) memory, like BL1. In these use-cases, it is necessary to initialize the RW sections in RAM, while leaving the RO sections in place. This patch enable this use-case with a new build option, BL2_IN_XIP_MEM. For now, this option is only supported when BL2_AT_EL3 is 1.

bl2/bl2_el3.ld.S
```diff
@@ -12,15 +12,26 @@ OUTPUT_ARCH(PLATFORM_LINKER_ARCH)
 ENTRY(bl2_entrypoint)
 
 MEMORY {
+#if BL2_AT_EL3_ROM
+    ROM (rx): ORIGIN = BL2_RO_BASE, LENGTH = BL2_RO_LIMIT - BL2_RO_BASE
+    RAM (rwx): ORIGIN = BL2_RW_BASE, LENGTH = BL2_RW_LIMIT - BL2_RW_BASE
```
danh-arm: The porting guide will need updating to describe these new platform porting constants, and to say that these are only needed when BL2_IN_XIP_MEM is 1. Also, to say that BL2_LIMIT and BL2_BASE are not applicable in this case.

bl2/bl2_el3.ld.S
```diff
@@ -174,12 +207,32 @@ SECTIONS
...
+#if BL2_AT_EL3_ROM
+    __BL2_RAM_START__ = ADDR(.data);
+    __BL2_RAM_END__ = .;
+
+    __DATA_ROM_START__ = LOADADDR(.data);
+    __DATA_SIZE__ = SIZEOF(.data);
+
+    /*
+     * The .data section is the last PROGBITS section so its end marks the end
+     * of BL2's actual content in Trusted ROM.
```
danh-arm: minor: Perhaps say "... of BL2's RO content in XIP memory." here.

bl2/bl2_el3.ld.S
```diff
@@ -174,12 +207,32 @@ SECTIONS
...
+     */
+    __BL2_ROM_END__ =  __DATA_ROM_START__ + __DATA_SIZE__;
+    ASSERT(__BL2_ROM_END__ <= BL2_RO_LIMIT,
+           "BL2's ROM content has exceeded its limit.")
```
danh-arm: minor: s/ROM/RO/

bl2/bl2_el3.ld.S
```diff
@@ -174,12 +207,32 @@ SECTIONS
...
     __BSS_SIZE__ = SIZEOF(.bss);
 
+
 #if USE_COHERENT_MEM
     __COHERENT_RAM_UNALIGNED_SIZE__ =
         __COHERENT_RAM_END_UNALIGNED__ - __COHERENT_RAM_START__;
 #endif
 
+#if BL2_AT_EL3_ROM
+    ASSERT(. <= BL2_RW_LIMIT, "BL2 image has exceeded its limit.")
```
danh-arm: minor: "BL2's RW content has ...."

docs/user-guide.rst
```diff
@@ -246,6 +246,9 @@ Common build options
 - ``BL2_AT_EL3``: This is an optional build option that enables the use of
    BL2 at EL3 execution level.
 
+- ``BL2_AT_EL3_ROM``: This is an optional build option that enables the use of
+   BL2 at EL3 execution level, and BL2 read only section is in ROM.
```
danh-arm: Will need rewording like above.

include/common/aarch64/el3_common_macros.S
```diff
@@ -278,8 +278,8 @@
 		 * an earlier boot loader stage.
 		 * -------------------------------------------------------------
 		 */
-		adr	x0, __RW_START__
-		adr	x1, __RW_END__
+		ldr	x0, =__RW_START__
+		ldr	x1, =__RW_END__
```
danh-arm: What's the rationale for this change? Using adr is preferred where possible. If this really is needed, it will need to be in a separate commit with explanation.

Makefile
JiafeiPan: Great, will modify it, thanks.

bl2/bl2_el3.ld.S
JiafeiPan: Done, thanks.

bl2/bl2_el3.ld.S
JiafeiPan: Done, thanks.

bl2/bl2_el3.ld.S
JiafeiPan: Done, thanks.

bl2/bl2_el3.ld.S
JiafeiPan: Done, thanks.

docs/user-guide.rst
JiafeiPan: Done, thanks.

include/common/aarch64/el3_common_macros.S
JiafeiPan: For the adr instruction, it require the lable's offset from the address of this instruction must be in the range +/-1MB. If the option "BL2_IN_XIP_MEM" is set to '1', in some cases, BL2's RW memory will not in the range of +/-1MB from BL2's RO memory region. so we need to use ldr instruction to cover this case.

JiafeiPan:
Hi, Dan, have updated the patches according to your comments, please review again, thanks.

include/common/aarch64/el3_common_macros.S
danh-arm: OK, makes sense - thanks.
minor: Could you just fix the spelling in the commit message (limiation and lable)?

Makefile
```diff
@@ -564,6 +565,7 @@ $(eval $(call add_define,USE_COHERENT_MEM))
 $(eval $(call add_define,USE_TBBR_DEFS))
 $(eval $(call add_define,WARMBOOT_ENABLE_DCACHE_EARLY))
 $(eval $(call add_define,BL2_AT_EL3))
+$(eval $(call add_define,BL2_IN_XIP_MEM))
```
danh-arm: I don't see the check I suggested in my last review: 
"... you could avoid supporting the case where BL2_AT_EL3 = 0 and BL2_IN_XIP_MEM = 1, by throwing an error in the top level Makefile (in the "Check incompatible options" section)"

JiafeiPan: Sorry, I thought that we need to add build check in platform makefile, will add it to top Makefile, thanks.

JiafeiPan: Hi, Dan, have updated the patch according to your comments, please review again, thanks.

danh-arm:
jenkins: test this please

dp-arm:
@JiafeiPan can you resolve the conflicts and then I can merge?

JiafeiPan:
Hi, @dp-arm, have resolved the conflicts, please merge it, thanks.

PR #1332, opened 2018-03-26
Add interface for hash functions in crypto module

ruchi393:
There may be scenarios where platform specific implementation
need to calculate hash of discontiguous data during image auth.
The patch exposes hash function form the crypto module which
can be used to do such calculations.

Signed-off-by: Ruchika Gupta <ruchika.gupta@nxp.com>


dp-arm:
@ruchi393 Can I ask how the platform knows to invoke these function during image loading/athentication?

dp-arm:
@ruchi393 nevermind, I saw your answer to https://github.com/ARM-software/tf-issues/issues/555

PR #1333, opened 2018-03-26
GIC: Fix interrupt setting interrupt configuration

jeenu-arm:
- Interrupt configuration is a 2-bit field, so the field shift has to
    be double that of the bit number.

  - Interrupt configuration (level- or edge-trigger) is specified in the
    MSB of the field, not LSB.

Fixes applied to both GICv2 and GICv3 drivers.

Fixes ARM-software/tf-issues#570

Change-Id: Ia6ae6ed9ba9fb0e3eb0f921a833af48e365ba359


dp-arm:
Minor MISRA detects, will be fixed in separate patch.

PR #1334, opened 2018-03-27
qemu: don't use C functions for the crash console callbacks

michpappas:
Use the console_pl011_core_* functions directly in the crash console
callbacks.

This bypasses the MULTI_CONSOLE_API for the crash console (UART1), but
allows using the crash console before the C runtime has been initialized
(eg to call ASM_ASSERT). This retains backwards compatibilty with respect
to functionality when the old API is used.

Use the MULTI_CONSOLE_API to register UART0 as the boot and runtime
console.

Fixes ARM-software/tf-issues#572

Signed-off-by: Michalis Pappas <mpappas@fastmail.fm>

### DIFF
--- a/plat/qemu/aarch64/plat_helpers.S
+++ b/plat/qemu/aarch64/plat_helpers.S
@@ -14,9 +14,8 @@
 	.globl	platform_mem_init
 	.globl	plat_qemu_calc_core_pos
 	.globl	plat_crash_console_init
-#if MULTI_CONSOLE_API
 	.globl	plat_crash_console_putc
-#endif /* MULTI_CONSOLE_API */
+	.globl	plat_crash_console_flush
 	.globl  plat_secondary_cold_boot_setup
 	.globl  plat_get_my_entrypoint
 	.globl  plat_is_my_cpu_primary
@@ -97,7 +96,10 @@ endfunc platform_mem_init
 	 * ---------------------------------------------
 	 */
 func plat_crash_console_init
-	b	qemu_crash_console_init
+	mov_imm x0, PLAT_QEMU_CRASH_UART_BASE
+	mov_imm x1, PLAT_QEMU_CRASH_UART_CLK_IN_HZ
+	mov_imm x2, PLAT_QEMU_CONSOLE_BAUDRATE
+	b	console_pl011_core_init
 endfunc plat_crash_console_init
 
 	/* ---------------------------------------------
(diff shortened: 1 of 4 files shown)


michpappas:
@jenswi-linaro, @jonathanwright-ARM  I think I'll go for the same approach as @antonio-nino-diaz-arm on rpi3. According to the comments, ASM_ASSERT is meant to work even before the runtime has been initialized, and this has already been working for qemu while using the old API, so no reason to lose functionality here. Hope this sorts everything out.

plat/qemu/aarch64/plat_helpers.S
```diff
@@ -31,6 +30,7 @@ endfunc plat_my_core_pos
  *  With this function: CorePos = (ClusterId * 4) + CoreId
  */
 func plat_qemu_calc_core_pos
+
```
jenswi-linaro: Empty line not needed

michpappas: Fixed

antonio-nino-diaz-arm:
I've actually realized that your console initialization is the correct one, I didn't need to use the `CONSOLE_FLAG_CRASH` flag.

dp-arm:
jenkins: test this please

plat/qemu/aarch64/plat_helpers.S
```diff
@@ -107,10 +109,21 @@ endfunc plat_crash_console_init
...
+	/* ---------------------------------------------
+	 * int plat_crash_console_flush(int c)
+	 * Function to force a write of all buffered
+	 * data that hasn't been output.
+	 * Out : return -1 on error else return 0.
+	 * Clobber list : x0, x1
+	 * ---------------------------------------------
+	 */
+func plat_crash_console_flush
+	mov_imm	x1, PLAT_QEMU_CRASH_UART_BASE
```
jwerner-chromium: Careful, it's x0 here.

plat/qemu/qemu_console.c
```diff
@@ -18,22 +17,13 @@ void qemu_console_init(void)
...
 }
 
-void qemu_crash_console_init(void)
-{
-#if MULTI_CONSOLE_API
-	(void)console_pl011_register(PLAT_QEMU_CRASH_UART_BASE,
-			       PLAT_QEMU_CRASH_UART_CLK_IN_HZ,
-			       PLAT_QEMU_CONSOLE_BAUDRATE, &crash_console);
-#else
-	console_core_init(PLAT_QEMU_CRASH_UART_BASE,
```
jwerner-chromium: Now that you no longer need it, can you please take the console_core_init() prototype you added to include/drivers/console.h back out? I don't think it belongs there, that function was never meant to be called from C (even before MULTI_CONSOLE_API).

plat/qemu/aarch64/plat_helpers.S
michpappas: Fixed

plat/qemu/qemu_console.c
michpappas: Fixed

michpappas:
@jwerner-chromium thanks for looking into it

dp-arm:
jenkins: test this please

PR #1337, opened 2018-03-29
Replace assert(0) with refactored panic() macro

jonathanwright-ARM:
Replaces all instances of assert(0) with panic(). Assert should only be
used to check program invariants; not signal unconditional failure.

The panic() macro has been refactored such that it prints the file and
line number of the panic() when LOG_LEVEL >= LOG_LEVEL_INFO.

Change-Id: Iaed085b85a7772d855b28f604a343dd7fb8198ec
Signed-off-by: Jonathan Wright <jonathan.wright@arm.com>

### DIFF
--- a/bl1/bl1_fwu.c
+++ b/bl1/bl1_fwu.c
@@ -90,8 +90,7 @@ register_t bl1_fwu_smc_handler(unsigned int smc_fid,
 		break;
 
 	default:
-		assert(0);
-		break;
+		panic();
 	}
 
 	SMC_RET1(handle, SMC_UNK);
@@ -693,7 +692,7 @@ __dead2 static void bl1_fwu_done(void *client_cookie, void *reserved)
 	 * Call platform done function.
 	 */
 	bl1_plat_fwu_done(client_cookie, reserved);
-	assert(0);
+	panic();
 }
 
 /*******************************************************************************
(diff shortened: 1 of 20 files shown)


dp-arm:
> No. As far as I can see all these assert(0) are checking program invariants - it's a programming error if this code is reached.

Exactly, but the condition is always false.  assert() makes sense when the condition is likely to be near 100% true which is useful for catching programming errors.  If you know in advance, that you should not proceed further down a particular path because of an assert(0) then that should be the case for both debug and release builds

danh-arm:
> If you know in advance, that you should not proceed further down a particular path because of an assert(0) then that should be the case for both debug and release builds

I don't get your point. Yes, we know in advance that a particular path should not be followed in either debug or release builds, but why does that mean we should check in release builds? This is no different to a normal assert condition, which should be true in both debug and release builds, even though we only check in debug.

I guess this is causing some Coverity problem but I don't see any explanation.

dp-arm:
> I don't get your point. Yes, we know in advance that a particular path should not be followed in either debug or release builds, but why does that mean we should check in release builds? This is no different to a normal assert condition, which should be true in both debug and release builds, even though we only check in debug.

I see your point.  I guess we can try to tackle some of the more confusing cases of assert(0) and see if we can move them so the condition is explicit in the assert.

PR #1341, opened 2018-04-03
lib: xlat_tables_v2: reduce time required to add a mmap region

vwadekar:
The last entry in the mapping table is not necessarily the same as the
end of the table. This patch loops through the table to find the last
entry marker, on every new mmap addition. The memove operation then
has to only move the memory between current entry and the last entry.
For platforms that arrange their MMIO map properly, this opearation
turns out to be a NOP.

The previous implementation added significant overhead per mmap
addition as the memmove operation always moved the difference between
the current mmap entry and the end of the table.

Tested on Tegra platforms and this new approach improves the memory
mapping time by ~75%, thus significantly reducing boot time on some
platforms.

Change-Id: Ie3478fa5942379282ef58bee2085da799137e2ca

### DIFF
--- a/lib/xlat_tables_v2/xlat_tables_internal.c
+++ b/lib/xlat_tables_v2/xlat_tables_internal.c
@@ -735,7 +735,8 @@ static int mmap_add_region_check(xlat_ctx_t *ctx, const mmap_region_t *mm)
 void mmap_add_region_ctx(xlat_ctx_t *ctx, const mmap_region_t *mm)
 {
 	mmap_region_t *mm_cursor = ctx->mmap;
-	mmap_region_t *mm_last = mm_cursor + ctx->mmap_num;
+	const mmap_region_t *mm_end = ctx->mmap + ctx->mmap_num;
+	mmap_region_t *mm_last;
 	unsigned long long end_pa = mm->base_pa + mm->size - 1;
 	uintptr_t end_va = mm->base_va + mm->size - 1;
 	int ret;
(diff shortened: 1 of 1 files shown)


lib/xlat_tables_v2/xlat_tables_internal.c
```diff
@@ -786,6 +786,19 @@ void mmap_add_region_ctx(xlat_ctx_t *ctx, const mmap_region_t *mm)
...
 
+	/*
+	 * Find the last entry marker in the mmap
+	 */
+	mm_last = ctx->mmap;
+	while ((mm_last->size != 0U) && (mm_last < mm_end)) {
+		++mm_last;
+	}
+
+	if (mm_last->size != 0U) {
```
antonio-nino-diaz-arm: Actually, the function `mmap_add_region_check` checks that there is enough space in the mmap array. This should be an assert.

vwadekar:
@antonio-nino-diaz-arm 

Thanks. I updated the patch to assert instead

vwadekar:
@danh-arm @davidcunado-arm 

Does this change look good? Any comments?

soby-mathew:
jenkins: test this please

antonio-nino-diaz-arm:
The check that fails is the following one:

File | Function | Line number | Checker | Description
-- | -- | -- | -- | --
lib/xlat_tables_v2/xlat_tables_internal.c | mmap_add_region_ctx | 738 | MISRA C-2012 Rule 8.13 | [Advisory]  MISRA C-2012 Rule 8.13 violation: Consider adding "const" qualifier to  the pointer "mm_end" as destination object does not appear to be  modified within this function.

vwadekar:
@antonio-nino-diaz-arm 

The latest patch fixes the MISRA warning

dp-arm:
jenkins: test this please

PR #1344, opened 2018-04-06
DMC500: Add platform support to set system interface count

amitdanielkachhap:
Some low end platforms using DMC500 memory controller do not have
CCI(Cache Coherent Interconnect) interface and only have non-coherent
system interface support. Hence this patch makes the system interface
count configurable from the platforms.

Change-Id: I6d54c90eb72fd18026c6470c1f7fd26c59dc4b9a

### DIFF
--- a/drivers/arm/tzc/tzc_dmc500.c
+++ b/drivers/arm/tzc/tzc_dmc500.c
@@ -24,7 +24,7 @@
 #define TZC_DMC500_ACTION_OFF				0x50
 
 /* Pointer to the tzc_dmc500_driver_data structure populated by the platform */
-static const tzc_dmc500_driver_data_t *g_driver_data;
+static tzc_dmc500_driver_data_t *g_driver_data;
 
 #define verify_region_attr(region, attr)	\
 		((g_conf_regions[(region)].sec_attr ==			\
@@ -88,7 +88,7 @@ void tzc_dmc500_config_complete(void)
 
 	for (dmc_inst = 0; dmc_inst < g_driver_data->dmc_count; dmc_inst++) {
 		assert(DMC_INST_BASE_ADDR(dmc_inst));
-		for (sys_if = 0; sys_if < MAX_SYS_IF_COUNT; sys_if++)
+		for (sys_if = 0; sys_if < g_driver_data->sys_if_count; sys_if++)
 			_tzc_dmc500_write_flush_control(
 					DMC_INST_SI_BASE(dmc_inst, sys_if));
 	}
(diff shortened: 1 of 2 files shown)


soby-mathew:
jenkins: test this please

drivers/arm/tzc/tzc_dmc500.c
```diff
@@ -271,5 +272,13 @@ void tzc_dmc500_driver_init(const tzc_dmc500_driver_data_t *plat_driver_data)
...
-	g_driver_data = plat_driver_data;
+
+	g_driver_data = (tzc_dmc500_driver_data_t *)plat_driver_data;
+
+	/* Check valid system interface count */
+	assert(g_driver_data->sys_if_count <= MAX_SYS_IF_COUNT);
+
+	/* If interface count is not present then assume max */
+	if (g_driver_data->sys_if_count == 0U)
+		g_driver_data->sys_if_count = MAX_SYS_IF_COUNT;
```
soby-mathew: @amitdanielkachhap , I just realized this patch can sublty cause issues if the plat_driver_data is pointing to a const to begin with. The tzc_dmc500_driver_init() prototype is expecting a pointer to a const as the arg, and hence the compiler wouldn't flag error.
The solution preferred is to keep g_driver_data as const tzc_dmc500_driver_data_t *. Define another global variable, say g_sys_if_count, and initialize it to plat_driver_data->sys_if_count. Use that for the checks.

drivers/arm/tzc/tzc_dmc500.c
```diff
@@ -271,5 +272,13 @@ void tzc_dmc500_driver_init(const tzc_dmc500_driver_data_t *plat_driver_data)
 	 */
 	/* Validates the information passed by platform */
 	validate_plat_driver_data(plat_driver_data);
-	g_driver_data = plat_driver_data;
+
+	g_driver_data = (tzc_dmc500_driver_data_t *)plat_driver_data;
```
soby-mathew: This is a problem, see comment below.

drivers/arm/tzc/tzc_dmc500.c
amitdanielkachhap: @soby-mathew: Agreed. Sending a new pull request with this change.

amitdanielkachhap:
Closing this pull request as created a new pull request(1348) with review fix changes.

soby-mathew:
The reworked PR is here: https://github.com/ARM-software/arm-trusted-firmware/pull/1348

PR #1345, opened 2018-04-07
rockchip/rk3399: Fix sram_udelay

dbasehore:
This fixes an off by 576x bug the the sram_udelay code. The wrong
value was multipled by the system ticks per mhz value (which is 24),
so we delayed for 1/576th of the requested time.

Signed-off-by: Derek Basehore <dbasehore@chromium.org>

### DIFF
--- a/plat/rockchip/rk3399/drivers/dram/suspend.c
+++ b/plat/rockchip/rk3399/drivers/dram/suspend.c
@@ -85,10 +85,11 @@ static __pmusramfunc uint32_t sram_get_timer_value(void)
 
 static __pmusramfunc void sram_udelay(uint32_t usec)
 {
-	uint32_t start, cnt, delta, delta_us;
+	uint32_t start, cnt, delta, total_ticks;
 
 	/* counter is decreasing */
 	start = sram_get_timer_value();
+	total_ticks = usec * SYS_COUNTER_FREQ_IN_MHZ;
 	do {
 		cnt = sram_get_timer_value();
 		if (cnt > start) {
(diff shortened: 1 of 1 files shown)


dp-arm:
jenkins: test this please

dp-arm:
@dbasehore what do you think of Antonio's comment?

dbasehore:
I need to write it and see if it works.

This code is running in 8KB of SRAM. UART isn't working at this point either. The best we can do in that situation might just be to while(1);

dbasehore:
Okay, from reading the assert code, the best we can do is just something like:
if (usec > UINT32_MAX / SYS_COUNTER_FREQ_IN_MHZ)
    while (1);

The assert code is stored in SDRAM, which we can't access at this point.

Another possibility is to just merge this and wait for the code that can reset the SoC while running in PMUSRAM. I figure we'll probably write 0xdeadbeef or something in PMUSRAM (which doesn't lose state across a warm reset) for debugging purposes.

antonio-nino-diaz-arm:
If the UART isn't working you could set `PLAT_LOG_LEVEL_ASSERT` to 0. That way the check is there but it just calls `plat_panic_handler` instead of trying to print anything. This function is by default a `wfi` loop.

If the watchdog is going to reboot the CPU, I think that maybe you should write your own version of `plat_panic_handler` to disable it, because all the asserts that are in the common code probably don't work for you either.

dp-arm:
> If it is going to be more than a simple check, I'm not sure if it is worth the effort. Maybe we can just merge it. The function is only going to be called with known values (and most likely small ones) so it is not a big problem.

Yes, I agree.  If it is okay with you @dbasehore I will go ahead and merge this as is.

dp-arm:
jenkins: test this please

PR #1347, opened 2018-04-09
FVP: Fix function for translating MPIDR to linear index

davidcunado-arm:
The current AArch32 version of plat_arm_calc_core_pos uses an incorrect
algorithm to calculate the linear position of a core / PE from its
MPIDR.

This patch corrects the algorithm to:

(ClusterId * FVP_MAX_CPUS_PER_CLUSTER) * FVP_MAX_PE_PER_CPU
+ (CPUId * FVP_MAX_PE_PER_CPU)
+ ThreadId

which supports cores where there are more than 1 PE per CPU.

NOTE: the AArch64 version was fixed in 39b21d1

Change-Id: I72aea89d8f72f8b1fef54e2177a0fa6fef0f5513

### DIFF
--- a/plat/arm/board/fvp/aarch32/fvp_helpers.S
+++ b/plat/arm/board/fvp/aarch32/fvp_helpers.S
@@ -104,15 +104,20 @@ func plat_is_my_cpu_primary
 	bx	lr
 endfunc plat_is_my_cpu_primary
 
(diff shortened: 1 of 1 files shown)


dp-arm:
jenkins: test this please

PR #1348, opened 2018-04-09
DMC500: Add platform support to set system interface count

amitdanielkachhap:
Some low end platforms using DMC500 memory controller do not have
CCI(Cache Coherent Interconnect) interface and only have non-coherent
system interface support. Hence this patch makes the system interface
count configurable from the platforms.

Change-Id: I6d54c90eb72fd18026c6470c1f7fd26c59dc4b9a


dp-arm:
jenkins: test this please

PR #1349, opened 2018-04-09
Juno: Increase bl2 max size to fix build when SPD=opteed

amitdanielkachhap:
Building TBBR(SPD=opteed) and non-TBBR TF-A images is breaking for
Juno for different configurations listed below:

* Overflow error of 4096 bytes for rsa algorithm.
* Overflow error of 8192 bytes for ecdsa algorithm.
* Overflow error of 4096 bytes for rsa+ecdsa algorithm.
* Overflow error of 4096 bytes for non-TBBR case.

So this patch increments macro PLAT_ARM_MAX_BL2_SIZE for all the above
cases accordingly.

Change-Id: I75ec6c0a718181d34553fe55437f0496f467683f


dp-arm:
jenkins: test this please

PR #1350, opened 2018-04-09
Introduce support for RAS extensions

danh-arm:
This patchset adds support for the v8.2 and v8.4 RAS extensions, in particular support for external aborts routed to EL3.

PR #1351, opened 2018-04-09
Introduce EA handling for RAS

soby-mathew:
This patch stack introduces External Abort handling in TF-A for RAS errors.  Support for platforms to register RAS Error Interrupts are also added. 

The patches will undergo some minor rework in the coming days.


soby-mathew:
This is a duplicate PR #1350. Closing this one

PR #1354, opened 2018-04-11
ARM platforms: Demonstrate mem_protect from el3_runtime

robertovargas-arm:
Previously mem_protect used to be only supported from BL2. This is not
helpful in the case when ARM TF-A BL2 is not used. This patch demonstrates
mem_protect from el3_runtime firmware on ARM Platforms specifically
when RESET_TO_BL31 or RESET_TO_SP_MIN flag is set as BL2 may be absent
in these cases. The Non secure DRAM is dynamically mapped into EL3 mmap
tables temporarily and then the protected regions are then cleared. This
avoids the need to map the non secure DRAM permanently to BL31/sp_min.

The stack size is also increased, because DYNAMIC_XLAT_TABLES require
a bigger stack.

Change-Id: Ia44c594192ed5c5adc596c0cff2c7cc18c001fde


dp-arm:
jenkins: test this please

dp-arm:
jenkins: test this please

danh-arm:
Still some Coverity issues to fix on this...

PR #1360, opened 2018-04-23
Add support for the SMC Calling Convention 2.0

antonio-nino-diaz-arm:
Due to differences in the bitfields of the SMC IDs, it is not possible to support SMCCC 1.X and 2.0 at the same time.

The behaviour of `SMCCC_MAJOR_VERSION` has changed. Now, it is a build option that specifies the major version of the SMCCC that the Trusted Firmware supports. The only two allowed values are 1 and 2, and it defaults to 1. The value of `SMCCC_MINOR_VERSION` is derived from it.

Note: Support for SMCCC v2.0 is an experimental feature to enable prototyping of secure partition specifications. Support for this convention is disabled by default and could be removed without notice.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -378,6 +378,16 @@ ifeq ($(BL2_AT_EL3)-$(BL2_IN_XIP_MEM),0-1)
 $(error "BL2_IN_XIP_MEM is only supported when BL2_AT_EL3 is enabled")
 endif
 
+# SMC Calling Convention checks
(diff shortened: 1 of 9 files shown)


dp-arm:
jenkins: test this please

jeenu-arm:
jenkins: test this please

jeenu-arm:
jenkins: retest this please

PR #1362, opened 2018-04-24
Remove dtc warnings

robertovargas-arm:
DTC generates warnings when unit names begin with 0, or
when a node containing a reg or range property doesn't have a unit name
in the node name. This patch fixes those cases.

Change-Id: If24ec68ef3034fb3fcefb96c5625c47a0bbd8474


dp-arm:
jenkins: test this pelase

danh-arm:
jenkins: retest this please

PR #1366, opened 2018-04-30
checkpatch: Ignore SPDX_LICENSE_TAG

antonio-nino-diaz-arm:
The Linux kernel expects the SPDX license tag in the first line of each source code file in a comment.

In the context of the Linux kernel repository this makes sense because they have many different license headers across their codebase. Moving the tag to the first line of the source code files makes it easier for analyzers to see the license of each file.

In the Trusted Firmware, we control all headers and make sure that they follow the same pattern, so this is not needed.

Fixes: https://github.com/ARM-software/tf-issues/issues/581

### DIFF
--- a/.checkpatch.conf
+++ b/.checkpatch.conf
@@ -36,6 +36,10 @@
 # This is not Linux so don't expect a Linux tree!
 --no-tree
 
+# The Linux kernel expects the SPDX license tag in the first line of each file.
+# We don't follow this in the Trusted Firmware.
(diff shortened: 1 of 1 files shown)


antonio-nino-diaz-arm:
The discussion about this patch is in this issue: https://github.com/ARM-software/tf-issues/issues/581

danh-arm:
Build error is due to transient Coverity problem. Otherwise OK.

PR #1371, opened 2018-05-02
smccc: Fix checkpatch error in header file

antonio-nino-diaz-arm:


danh-arm:
The only test failure was a Coverity issue as a result of a previous change, not this one.

PR #1376, opened 2018-05-08
lib: el3_runtime: initialise actlr_el1 to hardware defaults

vwadekar:
The context management library initialises the CPU context for the
secure/non-secure worlds to zero. This leads to zeros being stored
to the actual registers when we restore the CPU context, during a
world switch. Denver CPUs dont expect zero to be written to the
implementation defined, actlr_el1 register, at any point of time.
Writing a zero to some fields of this register, results in an
UNDEFINED exception.

This patch bases the context actlr_el1 value on the actual hardware
register, to maintain parity with the expected settings

Change-Id: I1c806d7ff12daa7fd1e5c72825494b81454948f2

### DIFF
--- a/lib/el3_runtime/aarch64/context_mgmt.c
+++ b/lib/el3_runtime/aarch64/context_mgmt.c
@@ -65,7 +65,7 @@ static void cm_init_context_common(cpu_context_t *ctx, const entry_point_info_t 
 	uint32_t scr_el3, pmcr_el0;
 	el3_state_t *state;
 	gp_regs_t *gp_regs;
-	unsigned long sctlr_elx;
+	unsigned long sctlr_elx, actlr_elx;
 
 	assert(ctx);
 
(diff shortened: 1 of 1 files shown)


danh-arm:
jenkins: test this please

danh-arm:
IP Review: +1

vwadekar:
@danh-arm 

Can you please let me know why we see "All checks have failed" error message?

antonio-nino-diaz-arm:
There's a Coverity defect related to this patch, and some others that aren't. This is the one that you can fix:

File | Function | Line number | Checker | Description 
-- | -- | -- | -- | --
el3_runtime/aarch64/context_mgmt.c | cm_init_context_common | 184 | MISRA C-2012 Rule 20.7 | [Required]  MISRA C-2012 Rule 20.7 violation: When calling macro "write_ctx_reg",  its associated arguments shall be enclosed in parentheses.

The fix is something like this:

https://github.com/ARM-software/arm-trusted-firmware/blob/085e80ec111b2ab3607f0f38f6ef0062922bc196/include/lib/aarch64/smccc_helpers.h#L40

vwadekar:
@antonio-nino-diaz-arm 

Thanks. Pushed updated patch

dp-arm:
jenkins: test this please

PR #1378, opened 2018-05-10
CVE-2017-5715 mitigation for Denver CPUs

vwadekar:


lib/cpus/aarch64/denver.S
```diff
@@ -1,16 +1,152 @@
...
+	 * This register is accessed using MSR 3, 0, 15, 0, 6
+	 * and is only implemeted by Denver CPUs with MIDR
+	 * Primary Part number < 4.
+	 * -------------------------------------------------
+	 */
+	mrs	x0, midr_el1
+	lsr	x0, x0, #MIDR_PN_SHIFT
+	and	x0, x0, #MIDR_PN_MASK
+	tst	x0, #4
+	bne	1f
```
dp-arm: you are doing a branch here before invalidating the BP.  Maybe you can use cmp/csel instead.

dp-arm: If you cannot access this register on revisions that don't implement the mitigation, I guess you would have to move the id check before selecting the workaround vector table.

vwadekar: @dp-arm 

The cpu_check_csv2 macro does something similar, so thought of keeping this logic here for Denver. Please let me know if my understanding is incorrect

vwadekar:
@dp-arm 

Updated the PR to fix review comments.

dp-arm:
jenkins: test this please

dp-arm:
@vwadekar just realized, you need to use the declare_cpu_ops_workaround_cve_2017_5715
macro to set the check function which will report to higher layers whether the core is mitigated or not.

This is necessary to comply with the firmware mitigation specification. Check how it is done for Cortex A72 as an example.

dp-arm:
@vwadekar you can send that as a separate patch of course.

vwadekar:
@dp-arm 

Pushed PR #1388 to use the macro

PR #1379, opened 2018-05-11
Fix incorrect NSRAM memory map region for SGI-575

CJKay:
SGI-575's NSRAM memory map region is currently out of alignment with its architectural specification.


dp-arm:
jenkins: test this please

PR #1380, opened 2018-05-11
Fix incorrect number of reserved memory map entries for Arm platforms

CJKay:


dp-arm:
jenkins: test this please

PR #1384, opened 2018-05-15
for rk3399 suspend/resume

rkchrome:
All the m0 code run in SRAM before, but we need to watch PMU_POWER_ST
when SOC enter into FSM, and SRAM will shutdown during this time, so
this code need run in PMUSRAM. But PMUSRAM only 8K space, we can not
put all the m0 binary into PMUSRAM, Split the M0 binary into two, dram
part still run in SRAM, and suspend part run in PMUSRAM.


paulmenzel:
The merge/pull request summary is only from the second commit, isn’t it?

Also, could you please add problem descriptions? Does supend/resume currently not work?

rkchrome:
Paul, we need all three patches. I have update the comment, if not shutdown Logic power rail, suspend/resume is stable with now process, but if shutdown Logic powr rail, we need these three patches to make suspend/resume working.

PR #1372, opened 2018-05-02
Arm platforms: Migrate to multi console driver

antonio-nino-diaz-arm:
In AArch64 builds, Arm platforms have moved to the multi console driver for regular operation. They still use the raw console driver functions for the crash console.

Also, fix an issue when registering the same console struct twice.

PR #1385, opened 2018-05-15
Revert "plat/arm: Migrate AArch64 port to the multi console driver"

antonio-nino-diaz-arm:
This reverts commit 2f18aa1fa35305f8feec25867473d30975b242fe.

It is causing some tests to fail. Until the cause is found and fixed, it is needed to remove this commit from master.


antonio-nino-diaz-arm:
The reason why this wasn't found before and the other PR (https://github.com/ARM-software/arm-trusted-firmware/pull/1372) was merged is that only the extensive tests that run on integration were able to find the bug.

PR #1388, opened 2018-05-16
cpus: denver: report CVE_2017_5715 mitigation to higher layers

vwadekar:
This patch uses the 'declare_cpu_ops_workaround_cve_2017_5715' macro, to
set the check function, to report that Denver cores are mitigated.

Change-Id: I1bb6eefdec8c01fb8b645e112f8d04d4bb8811ef

### DIFF
--- a/lib/cpus/aarch64/denver.S
+++ b/lib/cpus/aarch64/denver.S
@@ -189,6 +189,25 @@ func denver_disable_dco
 	ret
 endfunc denver_disable_dco
 
+func check_errata_cve_2017_5715
+	mov	x0, #ERRATA_MISSING
+#if WORKAROUND_CVE_2017_5715
+	/*
+	 * Check if the CPU supports the special instruction
+	 * required to flush the indirect branch predictor and
+	 * RSB. Support for this operation can be determined by
+	 * comparing bits 19:16 of ID_AFR0_EL1 with 0b0001.
+	 */
+	mrs	x1, id_afr0_el1
+	mov	x2, #0x10000
+	and	x1, x1, x2
+	cbz	x1, 1f
+	mov	x0, #ERRATA_APPLIES
+1:
+#endif
+	ret
+endfunc check_errata_cve_2017_5715
+
 	/* -------------------------------------------------
 	 * The CPU Ops reset function for Denver.
 	 * -------------------------------------------------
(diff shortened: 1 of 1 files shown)


danh-arm:
jenkins: test this please

danh-arm:
IP review: +1

lib/cpus/aarch64/denver.S
```diff
@@ -267,27 +267,27 @@ func denver_cpu_reg_dump
 	ret
 endfunc denver_cpu_reg_dump
 
-declare_cpu_ops denver, DENVER_MIDR_PN0, \
+declare_cpu_ops_workaround_cve_2017_5715 denver, DENVER_MIDR_PN0, \
 	denver_reset_func, \
 	denver_core_pwr_dwn, \
```
dp-arm: you need to set a check function before core_pwr_dwn that returns #ERRATA_APPLIES when the cpu is vulnerable, #ERRATA_NOT_APPLIES if it is not vulnerable and #ERRATA_MISSING if the cpu is vulnerable but mitigation is not enabled.

ditto for all macros below.

dp-arm: the reason it builds is because the macro has variadic arguments

dp-arm:
@vwadekar ping

antonio-nino-diaz-arm:
@vwadekar ping

dp-arm:
@vwadekar can you rework this?

vwadekar:
@dp-arm @antonio-nino-diaz-arm 

Sorry for the delay. Updated the PR with changes. Please let me know if this looks OK

lib/cpus/aarch64/denver.S
```diff
@@ -189,6 +189,15 @@ func denver_disable_dco
 	ret
 endfunc denver_disable_dco
 
+func check_errata_cve_2017_5715
+#if WORKAROUND_CVE_2017_5715
+	mov	x0, #ERRATA_APPLIES
```
dp-arm: It is not enough to assume that erratum applies if WORKAROUND_CVE_2017_5715 is set.

In denver_reset_func you have this:

`
        /*
         * Check if the CPU supports the special instruction
         * required to flush the indirect branch predictor and
         * RSB. Support for this operation can be determined by
         * comparing bits 19:16 of ID_AFR0_EL1 with 0b0001.
         */
        mrs     x0, id_afr0_el1
        mov     x1, #0x10000
        and     x0, x0, x1
        cmp     x0, #0
        adr     x1, workaround_bpflush_runtime_exceptions
        mrs     x2, vbar_el3
        csel    x0, x1, x2, ne
        msr     vbar_el3, x0
`

I think a similar check is required to really be sure that the erratum workaround will apply.

dp-arm: just to clarify, ERRATA_APPLIES means CPU is vulnerable and the workaround is active, ERRATA_NOT_APPLIES means CPU is not vulnerable and ERRATA_MISSING means CPU is vulnerable but workaround is not compiled in.

soby-mathew:
@vwadekar, could you address comments by @dp-arm ?

lib/cpus/aarch64/denver.S
vwadekar: @dp-arm 

Does it make sense to keep the check if special instruction only at one place? Not sure why we need the same check at two places

dp-arm: @vwadekar yeah you can factor it out and call it from two places

antonio-nino-diaz-arm:
@vwadekar, ping

vwadekar:
@antonio-nino-diaz-arm 

Updated the PR

PR #1389, opened 2018-05-17
Code change to fix small bugs

danielboulby-arm:


plat/arm/common/arm_bl2_setup.c
```diff
@@ -46,7 +46,7 @@ CASSERT(BL2_BASE >= (ARM_BL_RAM_BASE + BL1_MEMINFO_OFFSET), assert_bl2_base_over
 
 #if LOAD_IMAGE_V2
 
-#pragma weak bl2_plat_handle_post_image_load
+#pragma weak arm_bl2_handle_post_image_load
```
danh-arm: You're right that bl2_plat_handle_post_image_load should not be defined weak here but arm_bl2_handle_post_image_load does not need to be weak.

plat/arm/common/arm_bl2_setup.c
```diff
@@ -44,12 +44,6 @@ CASSERT(BL2_BASE >= (ARM_BL_RAM_BASE + BL1_MEMINFO_OFFSET), assert_bl2_base_over
 #pragma weak bl2_plat_arch_setup
 #pragma weak bl2_plat_sec_mem_layout
 
-#if LOAD_IMAGE_V2
-
-#pragma weak bl2_plat_handle_post_image_load
-
-#else /* LOAD_IMAGE_V2 */
-
```
danh-arm: You've now got a mismatched #if/#endif. You want this to become `#if !LOAD_IMAGE_V2` so that all the code to line 171 is removed when LOAD_IMAGE_V2 is true.

dp-arm:
jenkins: test this please

dp-arm:
jenkins: test this please

PR #1391, opened 2018-05-21
MISRA rule 21.15 fix

jts-arm:
Rule 21.15: The pointer arguments to the Standard Library functions
memcpy, memmove and memcmp shall be pointers to qualified or unqualified
versions of compatible types.

Basically that means that both pointer arguments must be of the same
type. However, even if the pointers passed as arguments to the above
functions are of the same type, Coverity still thinks it's a violation
if we do pointer arithmetics directly at the function call. Thus the
pointer arithmetic operations were moved outside of the function
argument.

First detected on the following configuration
        make PLAT=fvp LOG_LEVEL=50


dp-arm:
jenkins: test this please

jts-arm:
Waiting for another patch to be merged in order to rebase and re-check.

jeenu-arm:
jenkins: test this please

dp-arm:
jenkins: test this please

PR #1395, opened 2018-05-23
SPM: Refactor codebase

antonio-nino-diaz-arm:
Several changes to improve the codebase and allow the addition of new features.

Arm platforms: If SPM is enabled, BL31 now has to be placed in DRAM. Previously, this was optional.


antonio-nino-diaz-arm:
The job fails beause of some pre-existing MISRA defects in the context management library.

PR #1396, opened 2018-05-31
Extend FIP io driver to support multiple FIP devices

ruchi393:
Platform can define MAX_NUM_FIP_DEVICES in platform_def.h
to define the number of FIP io devices.

Signed-off-by: Ruchika Gupta <ruchika.gupta@nxp.com>

### DIFF
--- a/drivers/io/io_fip.c
+++ b/drivers/io/io_fip.c
@@ -19,6 +19,10 @@
 #include <utils.h>
 #include <uuid.h>
 
+#ifndef MAX_FIP_DEVICES
+#define MAX_FIP_DEVICES		1
+#endif
+
 /* Useful for printing UUIDs when debugging.*/
 #define PRINT_UUID2(x)								\
 	"%08x-%04hx-%04hx-%02hhx%02hhx-%02hhx%02hhx%02hhx%02hhx%02hhx%02hhx",	\
@@ -32,11 +36,33 @@ typedef struct {
 	fip_toc_entry_t entry;
 } file_state_t;
 
+/*
+ * Maintain dev_spec per FIP Device
+ * TODO - Add backend handles and file state
+ * per FIP device here once backends like io_memmap
+ * can support multiple open files
+ */
+typedef struct {
+	uintptr_t dev_spec;
+} fip_dev_state_t;
+
 static const uuid_t uuid_null = { {0} };
+/*
+ * Only one file can be open across all FIP device
+ * as backends like io_memmap don't support
+ * multiple open files. The file state and
+ * backend handle should be maintained per FIP device
+ * if the same support is available in the backend
+ */
 static file_state_t current_file = {0};
 static uintptr_t backend_dev_handle;
 static uintptr_t backend_image_spec;
 
+static fip_dev_state_t state_pool[MAX_FIP_DEVICES];
+static io_dev_info_t dev_info_pool[MAX_FIP_DEVICES];
+
+/* Track number of allocated fip devices */
+static unsigned int fip_dev_count;
 
 /* Firmware Image Package driver functions */
 static int fip_dev_open(const uintptr_t dev_spec, io_dev_info_t **dev_info);
@@ -92,20 +118,94 @@ static const io_dev_funcs_t fip_dev_funcs = {
 	.dev_close = fip_dev_close,
 };
 
+/* Locate a file state in the pool, specified by address */
+static int find_first_fip_state(const uintptr_t dev_spec,
+				  unsigned int *index_out)
+{
+	int result = -ENOENT;
+	unsigned int index;
 
-/* No state associated with this device so structure can be const */
-static const io_dev_info_t fip_dev_info = {
-	.funcs = &fip_dev_funcs,
-	.info = (uintptr_t)NULL
-};
+	for (index = 0; index < (unsigned int)MAX_FIP_DEVICES; ++index) {
+		/* dev_spec is used as identifier since it's unique */
+		if (state_pool[index].dev_spec == dev_spec) {
+			result = 0;
+			*index_out = index;
+			break;
+		}
+	}
+	return result;
+}
 
 
-/* Open a connection to the FIP device */
-static int fip_dev_open(const uintptr_t dev_spec __unused,
+/* Allocate a device info from the pool and return a pointer to it */
+static int allocate_dev_info(io_dev_info_t **dev_info)
+{
+	int result = -ENOMEM;
+
+	assert(dev_info != NULL);
+
+	if (fip_dev_count < (unsigned int)MAX_FIP_DEVICES) {
+		unsigned int index = 0;
+
+		result = find_first_fip_state(0, &index);
+		assert(result == 0);
+		/* initialize dev_info */
+		dev_info_pool[index].funcs = &fip_dev_funcs;
+		dev_info_pool[index].info =
+				(uintptr_t)&state_pool[index];
+		*dev_info = &dev_info_pool[index];
+		++fip_dev_count;
+	}
+
+	return result;
+}
+
+/* Release a device info to the pool */
+static int free_dev_info(io_dev_info_t *dev_info)
+{
+	int result;
+	unsigned int index = 0;
+	fip_dev_state_t *state;
+
+	assert(dev_info != NULL);
+
+	state = (fip_dev_state_t *)dev_info->info;
+	result = find_first_fip_state(state->dev_spec, &index);
+	if (result ==  0) {
+		/* free if device info is valid */
+		zeromem(state, sizeof(fip_dev_state_t));
+		--fip_dev_count;
+	}
+
+	return result;
+}
+
+/*
+ * Multiple FIP devices can be opened depending on the value of
+ * MAX_FIP_DEVICES. Given that there is only one backend, only a
+ * single file can be open at a time by any FIP device.
+ */
+static int fip_dev_open(const uintptr_t dev_spec,
 			 io_dev_info_t **dev_info)
 {
+	int result;
+	io_dev_info_t *info;
+	fip_dev_state_t *state;
+
 	assert(dev_info != NULL);
-	*dev_info = (io_dev_info_t *)&fip_dev_info; /* cast away const */
+#if MAX_FIP_DEVICES > 1
+	assert(dev_spec != (uintptr_t)NULL);
+#endif
+
+	result = allocate_dev_info(&info);
+	if (result != 0)
+		return -ENOMEM;
+
+	state = (fip_dev_state_t *)info->info;
+
+	state->dev_spec = dev_spec;
+
+	*dev_info = info;
 
 	return 0;
 }
@@ -165,7 +265,7 @@ static int fip_dev_close(io_dev_info_t *dev_info)
 	backend_dev_handle = (uintptr_t)NULL;
 	backend_image_spec = (uintptr_t)NULL;
 
-	return 0;
+	return free_dev_info(dev_info);
 }
 
 
@@ -341,7 +441,11 @@ int register_io_dev_fip(const io_dev_connector_t **dev_con)
 	int result;
 	assert(dev_con != NULL);
 
-	result = io_register_device(&fip_dev_info);
+	/*
+	 * Since dev_info isn't really used in io_register_device, always
+	 * use the same device info at here instead.
+	 */
+	result = io_register_device(&dev_info_pool[0]);
 	if (result == 0)
 		*dev_con = &fip_dev_connector;
 


drivers/io/io_fip.c
```diff
@@ -92,20 +100,66 @@ static const io_dev_funcs_t fip_dev_funcs = {
...
-	*dev_info = (io_dev_info_t *)&fip_dev_info; /* cast away const */
+	result = allocate_dev_info(&info);
+	if (result)
+		return -ENOENT;
+
+	state = (file_state_t *)info->info;
+	zeromem(state, sizeof(file_state_t));
+	state->dev_spec = (void *)dev_spec;
+
+	*dev_info = info;	/* cast away const */
```
robertovargas-arm: The cast away message is not meaningful anymore after your changes.

drivers/io/io_fip.c
```diff
@@ -92,20 +100,66 @@ static const io_dev_funcs_t fip_dev_funcs = {
...
 			 io_dev_info_t **dev_info)
 {
+	int result;
+	io_dev_info_t *info;
+	file_state_t *state;
+
 	assert(dev_info != NULL);
-	*dev_info = (io_dev_info_t *)&fip_dev_info; /* cast away const */
+	result = allocate_dev_info(&info);
+	if (result)
```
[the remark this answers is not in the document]
ruchi393: I will make the changes for result check. 

I agree that ENOMEM would be a better return code.  I chose ENOENT to keep it consistent with dev_open functions in block io device and semihosting drivers.

dp-arm:
IP review: +1

dp-arm:
jenkins: test this please

ruchi393:
I see that patch shows that to has failed some checks. I am not able to access the link which has the details in it. Can you let me know if anything is required from my end.

soby-mathew:
@ruchi393 , It failed some coverity changes. Please find them below

/work/workspace/workspace/tf-coverity-list-defects/TF_CONFIG/fvp-default/label/arch-dev/drivers/io/io_fip.c | find_first_fip_state | 113 | MISRA C-2012 Rule 10.3 | [Required] MISRA C-2012 Rule 10.3 violation: Implicit conversion of "index" from essential type signed 32-bit int to different or narrower essential type unsigned 32-bit int. | 249725
-- | -- | -- | -- | -- | --
/work/workspace/workspace/tf-coverity-list-defects/TF_CONFIG/fvp-default/label/arch-dev/drivers/io/io_fip.c | allocate_dev_info | 128 | MISRA C-2012 Rule 10.4 | [Required] MISRA C-2012 Rule 10.4 violation: The operands "fip_dev_count" and "MAX_FIP_DEVICES" have essential type categories "unsigned 32-bit int" and "signed 8-bit int", which do not match. | 249726
/work/workspace/workspace/tf-coverity-list-defects/TF_CONFIG/fvp-default/label/arch-dev/drivers/io/io_fip.c | fip_file_open | 239 | MISRA C-2012 Rule 11.6 | [Required] MISRA C-2012 Rule 11.6 violation: A cast shall not convert a pointer to void type to an arithmetic type. | 249727
/work/workspace/workspace/tf-coverity-list-defects/TF_CONFIG/fvp-default/label/arch-dev/drivers/io/io_fip.c | fip_dev_open | 160 | MISRA C-2012 Rule 11.6 | [Required] MISRA C-2012 Rule 11.6 violation: A cast shall not convert a pointer to void type to an arithmetic type. | 249728
/work/workspace/workspace/tf-coverity-list-defects/TF_CONFIG/fvp-default/label/arch-dev/drivers/io/io_fip.c | fip_file_close | 381 | MISRA C-2012 Rule 8.13 | [Advisory] MISRA C-2012 Rule 8.13 violation: Consider adding "const" qualifier to the pointer "current_file" as destination object does not appear to be modified within this function. | 249731
/work/workspace/workspace/tf-coverity-list-defects/TF_CONFIG/fvp-default/label/arch-dev/drivers/io/io_fip.c | fip_dev_open | 150 | MISRA C-2012 Rule 8.13 | [Advisory] MISRA C-2012 Rule 8.13 violation: Consider adding "const" qualifier to the pointer "info" as destination object does not appear to be modified within this function. | 249732

drivers/io/io_fip.c
```diff
@@ -176,19 +230,23 @@ static int fip_file_open(io_dev_info_t *dev_info, const uintptr_t spec,
...
 	size_t bytes_read;
 	int found_file = 0;
 
 	assert(uuid_spec != NULL);
 	assert(entity != NULL);
+	assert((dev_info->info != (uintptr_t)NULL));
+
+	current_file = (file_state_t *)dev_info->info;
 
 	/* Can only have one file open at a time for the moment. We need to
```
soby-mathew: Does this comment hold anymore?

drivers/io/io_fip.c
```diff
@@ -176,19 +230,23 @@ static int fip_file_open(io_dev_info_t *dev_info, const uintptr_t spec,
...
 	 * track state like file cursor position. We know the header lives at
 	 * offset zero, so this entry should never be zero for an active file.
 	 * When the system supports dynamic memory allocation we can allow more
 	 * than one open file at a time if needed.
 	 */
-	if (current_file.entry.offset_address != 0) {
+	if (current_file->entry.offset_address != 0) {
 		WARN("fip_file_open : Only one open file at a time.\n");
```
soby-mathew: This error message would need updating.

drivers/io/io_fip.c
```diff
@@ -28,16 +32,22 @@
...
 static const uuid_t uuid_null = {0};
-static file_state_t current_file = {0};
+
+static file_state_t current_file_pool[MAX_FIP_DEVICES];
+static io_dev_info_t dev_info_pool[MAX_FIP_DEVICES];
+
+/* Track number of allocated fip devices */
+static unsigned int fip_dev_count;
+
 static uintptr_t backend_dev_handle;
```
soby-mathew: There is only one instance of backend_dev_handle. Shouldn't there be multiple instances of this as well ?

soby-mathew:
The issue is described in ARM-Software/tf-issues#591

drivers/io/io_fip.c
ruchi393: Sure, I will change the comment to "one file per device"

drivers/io/io_fip.c
ruchi393: I was working on the case where all the fip devices had the same backend handle. But , from driver perspective we should not limit this. I will move these handles into the info structure.

dp-arm:
@ruchi393 ping

ruchi393:
@dp-arm, I have updated the pull request with the comments addressed. Am I missing something?

drivers/io/io_fip.c
```diff
@@ -32,11 +36,22 @@ typedef struct {
 	fip_toc_entry_t entry;
 } file_state_t;
 
+/* Maintain backend handles and file state per FIP device */
+typedef struct {
+	uintptr_t dev_spec;
+	uintptr_t backend_dev_handle;
+	uintptr_t backend_image_spec;
+	/* One file per device is maintained */
+	file_state_t current_file;
```
soby-mathew: @ruchi393, Thanks for the rework. In the current io_memmap driver, I realized that it only allows a single file to be open.  Having several open FIP files at the same time, implies the same from io_memmap driver. Hence supporting multiple files open at the io_fip layer is not very useful without corresponding io_memmap support.

From your description of the ARM-Software/tf-issues#591, You seem to require multiple FIP devices but not necessarily several files open at the same time.  Since the generic load_image() always loads the image in full before moving to the next image (i.e only a single file is `active` at a given time), multiple fip devices can be supported without having multiple file support.  Perhaps the structure can be simplified as : 

```
typedef struct {
	uintptr_t dev_spec;
} fip_dev_state_t;
```
Only a single global file state `current_file` would be maintained which implies that the file has to be opened, read and closed before the next file is accessed (which may be in another FIP device). Would that work for you ?

drivers/io/io_fip.c
```diff
@@ -92,20 +105,87 @@ static const io_dev_funcs_t fip_dev_funcs = {
...
+	fip_dev_state_t *state;
+
 	assert(dev_info != NULL);
-	*dev_info = (io_dev_info_t *)&fip_dev_info; /* cast away const */
+	result = allocate_dev_info(&info);
+	if (result != 0)
+		return -ENOMEM;
+
+	state = (fip_dev_state_t *)info->info;
+	zeromem(state, sizeof(fip_dev_state_t));
```
soby-mathew: I don't think we should zero_mem here as this would clear out the state of a currently open connection the same FIP device. See my comments above for some thoughts regarding the design.

drivers/io/io_fip.c
ruchi393: You are right about the requirement.  At a time I will have only one file open of a single device.  I will explore and see what changes are required to maintain a global file for each device.

ruchi393: @soby-mathew , I checked the memmap code in io_memmap.c . There I see that in memmap_block_open function, there is a check on if file is in use and this gets cleared in memmap_block_close function. In fip driver, the backend's open and close functions are not spread across functions.   In BL2 we are executing on a single core. IIUC, this would mean that even though backend driver has a single file, it shouldn't matter. Please help correct me in case I am missing out on something.

soby-mathew: @ruchi393, Your understanding is correct. TBH, my understanding of the I/O layer is not strong. 

> Block devices seem to have multiple block devices support in it. Would it be better to expand the io_memmap driver to add multiple file support. What do you think ?

AIU, the layering of the I/O layer when accessed for FIP is as below:
```
       io_storage
           |
           v
        io_fip
           |
      -------------
     |             |
io_memmap     io_block
```
The `io_block` layer can be used to access the block devices in addition to via FIP. For the `io_memmap`  they are only used via FIP (the IO memmap drivers are not used for direct access).
This is why I think why io_block has multiple file capability whereas io_memmap didn't.

Preferably we don't want to add more support in IO layer unless needed.

> In fip driver, the backend's open and close functions are not spread across functions. In BL2 we are executing on a single core. IIUC, this would mean that even though backend driver has a single file, it shouldn't matter

Yes, the io_fip driver opens and closes the underlaying io_layer for all io operation. I think having a single backend handle should be alright.

> Basically what you are suggesting doesn't require any change in FIP driver. dev_spec is a dummy entity , not used after device_open function

When you want to open more than 1 FIP device at a time (i.e. multiple calls to fip_dev_open()), you would need to support returning multiple handles. Since internally the io_fip driver can only support one open file at time, yes, the dev_spec would remain a dummy entity not used after fip_dev_open().

soby-mathew: > Do you see any issue with multiple backend handles of memmap driver ? I thought it made the design clearer.

> With above design I thought as per your earlier suggestion having multiple backend driver made the flow very clean for the multiple FIP support.

I initially misunderstood that you needed multiple file support in io_fip driver which is not the case. Multiple open file support in io_fip mandates the same support from the all the backends and this brings in additional memory overhead and complexity which is not required now.  Hence would prefer to keep it simple and just allow muiple fip devices for now.

> Please suggest if you want me to remove the multiple backend support and revert to the first patch which I had sent.

The patch would be similar to the first version except that you don't need to add `current_file` field to  `fip_dev_state_t`. Just have the `dev_spec` as the only field. Possibly suitable comments can be added to indicate what is involved if mutiple open files have to be supported and can be taken up when there is a requirement.

ruchi393: I have updated the pull request with the requested changes. Please review.

drivers/io/io_fip.c
ruchi393: Pull Request updated with change

drivers/io/io_fip.c
```diff
@@ -92,20 +118,67 @@ static const io_dev_funcs_t fip_dev_funcs = {
...
+static int allocate_dev_info(io_dev_info_t **dev_info)
+{
+	int result = -ENOMEM;
+
+	assert(dev_info != NULL);
+
+	if (fip_dev_count < (unsigned int)MAX_FIP_DEVICES) {
+		unsigned int index = 0;
+
+		result = find_first_fip_state(0, &index);
```
soby-mathew: > result = find_first_fip_state(0, &index);

nit: prefer `result = find_first_fip_state(NULL, &index);`

drivers/io/io_fip.c
```diff
@@ -92,20 +118,67 @@ static const io_dev_funcs_t fip_dev_funcs = {
...
+	fip_dev_state_t *state;
+
 	assert(dev_info != NULL);
-	*dev_info = (io_dev_info_t *)&fip_dev_info; /* cast away const */
+	result = allocate_dev_info(&info);
+	if (result != 0)
+		return -ENOMEM;
+
+	state = (fip_dev_state_t *)info->info;
+	zeromem(state, sizeof(fip_dev_state_t));
```
soby-mathew: > zeromem(state, sizeof(fip_dev_state_t));

This can be removed as `dev_spec` is written in the next line.

drivers/io/io_fip.c
```diff
@@ -92,20 +118,67 @@ static const io_dev_funcs_t fip_dev_funcs = {
...
+		dev_info_pool[index].info =
+				(uintptr_t)&state_pool[index];
+		*dev_info = &dev_info_pool[index];
+		++fip_dev_count;
+	}
+
+	return result;
+}
 
 /* Open a connection to the FIP device */
```
soby-mathew: This comment needs update. Something on the lines of 

"Multiple FIP devices can be opened depending on the value of MAX_FIP_DEVICES. Given that there is only one backend, only a single file can be open at a time by any FIP device."

drivers/io/io_fip.c
```diff
@@ -341,7 +414,7 @@ int register_io_dev_fip(const io_dev_connector_t **dev_con)
 	int result;
 	assert(dev_con != NULL);
```
soby-mathew: Need a comment here: 
```
  /*
   * Since dev_info isn't really used in io_register_device, always
   * use the same device info at here instead.
   */
```

drivers/io/io_fip.c
```diff
@@ -92,20 +118,67 @@ static const io_dev_funcs_t fip_dev_funcs = {
...
 
 /* Open a connection to the FIP device */
-static int fip_dev_open(const uintptr_t dev_spec __unused,
+static int fip_dev_open(const uintptr_t dev_spec,
 			 io_dev_info_t **dev_info)
 {
+	int result;
+	io_dev_info_t *info;
+	fip_dev_state_t *state;
+
```
soby-mathew: You need to ensure that dev_spec is not NULL otherwise the allocation will fail. Suggest: 
```
#if MAX_FIP_DEVICES > 1
assert(dev_spec != (uintptr_t)NULL);
#endif
```
This will maintain compatibility of exisiting users of io_fip driver as they tend to use NULL as the dev_spec.

drivers/io/io_fip.c
```diff
@@ -32,11 +36,33 @@ typedef struct {
 	fip_toc_entry_t entry;
 } file_state_t;
 
+/*
+ * Maintain dev_spec per FIP Device
+ * TODO - Add backend handles and file state
+ * per FIP device here once backends like memmap
```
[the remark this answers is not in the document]
ruchi393: Ok

drivers/io/io_fip.c
```diff
@@ -32,11 +36,33 @@ typedef struct {
...
+ * can support multiple open files
+ */
+typedef struct {
+	uintptr_t dev_spec;
+} fip_dev_state_t;
+
 static const uuid_t uuid_null = { {0} };
+/*
+ * Only one file can be open for any FIP device
```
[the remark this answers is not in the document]
ruchi393: Ok

drivers/io/io_fip.c
```diff
@@ -32,11 +36,33 @@ typedef struct {
...
+ * as backends like memmap don't support
```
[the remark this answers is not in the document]
ruchi393: Ok

drivers/io/io_fip.c
ruchi393: zeromem will help when more elements like current_file etc are added in the fip_dev_state as it would nullify these pointers,

drivers/io/io_fip.c
ruchi393: ok

drivers/io/io_fip.c
ruchi393: ok

drivers/io/io_fip.c
ruchi393: Ok

ruchi393:
I have updated the pull request with comments addresses.

I have tested the patch on NXP platform and am able to open multiple FIP devices as described in ARM-Software/tf-issues#591

ruchi393:
Comment on free addressed and pull request updated.

ruchi393:
Comment on free addressed and pull request updated

drivers/io/io_fip.c
ruchi393: Ok. I have removed this zeromem and updated the pull request.

ruchi393:
All comments have been addressed and pull request updated.

soby-mathew:
jenkins: test this please

PR #1397, opened 2018-06-01
Add support for Cortex-A76 and Cortex-Ares

dp-arm:


dp-arm:
jenkins: test this please

PR #1402, opened 2018-06-05
drivers: ti: uart: Add TI specific 16550 initialization

glneo:
On TI platforms the UART is disabled by default and must be explicitly
enabled using the MDR1 register.

NOTE: The original definition of
http://www.ti.com/lit/ds/symlink/pc16550d.pdf has no MDR register, but
many TI SoCs implementing 16550 do have a quirky MDR register
implemented. So, this should be enabled with TI_16550_MDR_QUIRK

NOTE: In such implementation, the CSR register does not exist.

### DIFF
--- a/include/drivers/ti/uart/uart_16550.h
+++ b/include/drivers/ti/uart/uart_16550.h
@@ -23,6 +23,8 @@
 #define UARTMSR			0x18
 #define UARTSPR			0x1c
 #define UARTCSR			0x20
+/* Some instances have MDR1 defined as well */
+#define UARTMDR1		0x20
 #define UARTRXFIFOCFG		0x24
 #define UARTMIE			0x28
 #define UARTVNDR		0x2c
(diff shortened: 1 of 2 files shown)


danh-arm:
IP Review: +1

include/drivers/ti/uart/uart_16550.h
```diff
@@ -23,6 +23,8 @@
 #define UARTMSR			0x18
 #define UARTSPR			0x1c
 #define UARTCSR			0x20
+/* Some instances have MDR1 defined as well */
+#define UARTMDR1		0x20
```
danh-arm: minor: It looks like all these register definitions after 0x1c are not defined in the spec you refer to, nor are they used in the code. Perhaps they should be removed (apart from the UARTMDR1 you added)?

danh-arm:
jenkins: test this please

include/drivers/ti/uart/uart_16550.h
glneo: They are available in some implementations, they most likely will not be used like most of these registers, but for completeness I see no reason to drop them here.

PR #1403, opened 2018-06-05
TI K3 platform support

glneo:
Hello all,

This PR adds support for the K3 platform. The full PSCI implementation
will be posted as a separate PR as is it requires drivers to communicate
with the system controller and so is broken out into smaller parts. This
PR alone allows booting to Linux from ATF on a single core.

See AM65x Technical Reference Manual (SPRUID7, April 2018)
for further details: http://www.ti.com/lit/pdf/spruid7

Thanks,
Andrew


plat/ti/k3/common/bl31.c
```diff
@@ -33,6 +58,30 @@ void bl31_early_platform_setup(bl31_params_t *from_bl2,
 	/* There are no parameters from BL2 if BL31 is a reset vector */
 	assert(from_bl2 == NULL);
 	assert(plat_params_from_bl2 == NULL);
+
+	/* Populate entry point information for BL32 */
```
Amit-Radur: Does your platform always has secure image to be booted alone side non secure image ?
Should this be guarded by PRELOADED_BL32_BASE #ifdef?

glneo: If it does not have a secure image, then ATF should be compiled with "SPD=none". The #ifdef guard would do nothing in this case as the board.mk file will populate PRELOADED_BL32_BASE with a default value if it is empty.

plat/ti/k3/common/bl31.c
```diff
@@ -0,0 +1,76 @@
...
+/*
+ * The next 3 constants identify the extents of the code, RO data region and the
+ * limit of the BL3-1 image.  These addresses are used by the MMU setup code and
+ * therefore they must be page-aligned.  It is the responsibility of the linker
+ * script to ensure that __RO_START__, __RO_END__ & __BL31_END__ linker symbols
+ * refer to page-aligned addresses.
+ */
+#define BL31_RO_BASE (unsigned long)(&__RO_START__)
+#define BL31_RO_LIMIT (unsigned long)(&__RO_END__)
+#define BL31_END (unsigned long)(&__BL31_END__)
```
antonio-nino-diaz-arm: Use:

```
IMPORT_SYM(unsigned long, __RO_START__, BL31_RO_BASE);
IMPORT_SYM(unsigned long, __RO_END__, BL31_RO_LIMIT);
IMPORT_SYM(unsigned long, __BL31_END__, BL31_END);
```

antonio-nino-diaz-arm:
Also, I'm not sure about `PRELOADED_BL32_BASE` as it is right now. Maybe we need to add it to the common Makefile, or to rename it to `TI_PRELOADED_BL32_BASE` or similar while it is an option that is specific to this platform.

plat/ti/k3/common/bl31.c
Amit-Radur: My point was, what if there is no secure image.
 is this code still executes then?

plat/ti/k3/common/bl31.c
glneo: Will fix.

glneo:
I'll add PRELOADED_BL32_BASE to the common Makefile and document it next to the already existing PRELOADED_BL33_BASE.

plat/ti/k3/common/bl31.c
glneo: No, if you do not have a secure image and compile without SPD=something then ATF never fetches this address so it will not be used.

antonio-nino-diaz-arm:
Actually, could you replace all occurrences of `ARM_PRELOADED_DTB_BASE` in the codebase by `PRELOADED_DTB_BASE`? Just create a patch on top of "Add generic pre-loaded BL32 and DTB options". I've done a quick search, and there are only 6 places where it is used, so it would be quite simple. I didn't expect other platforms would want to do it, that's why I kept it as a platform option instead of a global one.

glneo:
No problem, will do.

antonio-nino-diaz-arm:
Actually, now that @afaerber has mentioned it in https://github.com/ARM-software/arm-trusted-firmware/pull/1410, we should have a short document explaining how to build it and deploy it.

afaerber:
Also list it in the main readme.rst please.
And same as #1410 it is against the wrong branch.

antonio-nino-diaz-arm:
Oh, and the maintainers.rst file as well.

glneo:
Okay, updated addressing the comments.

dp-arm:
jenkins: test this please

glneo:
Looks like some check has failed, but I can't see the details?

antonio-nino-diaz-arm:
It looks like infrastructure issues, I've retriggered the CI tests.

glneo:
Thanks, looks like it is all good now.

plat/ti/k3/common/plat_psci.c
```diff
@@ -0,0 +1,95 @@
+/*
```
soby-mathew: suggest: k3_plat_psci.c as file name

plat/ti/k3/common/plat_topology.c
```diff
@@ -0,0 +1,70 @@
...
+#include <psci.h>
+
+/* The power domain tree descriptor */
+static unsigned char power_domain_tree_desc[] = {
+	PLATFORM_CLUSTER_COUNT,
+	ARM_CLUSTER0_CORE_COUNT,
+#if ARM_CLUSTER1_MSMC_PORT != UNUSED
+	ARM_CLUSTER1_CORE_COUNT,
+#endif /* ARM_CLUSTER1_MSMC_PORT != UNUSED */
+#if ARM_CLUSTER2_MSMC_PORT != UNUSED
```
soby-mathew: Again comment on names for macros. Would prefer `K3_` as prefix for these macros.

plat/ti/k3/common/bl31.c
```diff
@@ -22,6 +47,34 @@ void bl31_early_platform_setup(bl31_params_t *from_bl2,
 	/* There are no parameters from BL2 if BL31 is a reset vector */
 	assert(from_bl2 == NULL);
 	assert(plat_params_from_bl2 == NULL);
+
+#ifdef PRELOADED_BL32_BASE
+	/* Populate entry point information for BL32 */
+	SET_PARAM_HEAD(&bl32_image_ep_info, PARAM_EP, VERSION_1, 0);
+	bl32_image_ep_info.pc = PRELOADED_BL32_BASE;
```
soby-mathew: Hmm, it seems BL32_BASE can be directly used as the macro. This can be passed in from the platform.mk based on whether the SPD build flag is set.

It is better to avoid introducing generic build macros if possible.

plat/ti/k3/common/plat_topology.c
```diff
@@ -0,0 +1,70 @@
+/*
```
soby-mathew: Suggest k3_plat_topology.c as the file name

Makefile
```diff
@@ -614,6 +614,11 @@ $(eval $(call add_define,WARMBOOT_ENABLE_DCACHE_EARLY))
 $(eval $(call add_define,BL2_AT_EL3))
 $(eval $(call add_define,BL2_IN_XIP_MEM))
 
+# Define the PRELOADED_BL32_BASE flag only if it is provided.
+ifdef PRELOADED_BL32_BASE
```
soby-mathew: I dont think PRELOADED_BL32_BASE build option is required generically. See my later comment regarding this.

plat/ti/k3/common/bl31.c
```diff
@@ -0,0 +1,65 @@
+/*
```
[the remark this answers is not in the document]
glneo: bl31_setup.c works fine for me, but why the k3_ prefix? It is already in a k3 directory. (same question for all below renames)

plat/ti/k3/common/bl31.c
```diff
@@ -9,9 +9,16 @@
 #include <assert.h>
 #include <bl_common.h>
 #include <debug.h>
+#include <plat_arm.h>
```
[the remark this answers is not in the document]
glneo: I was worried about this too, but other platforms also did this (Hisilicon, Mediatek, Xilinx). It seems to be the only way to get at some useful functions (arm_setup_page_tables), perhaps someone should factor those functions out into plat/common/ sometime.

plat/ti/k3/common/plat_common.mk
```diff
@@ -50,6 +50,10 @@ PLAT_GIC_SOURCES	+=	\
 				plat/common/plat_gicv3.c		\
 				${PLAT_PATH}/common/gicv3.c		\
 
+PLAT_PSCI_SOURCES	+=	\
```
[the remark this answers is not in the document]
glneo: No problem, will fix.

plat/ti/k3/common/plat_topology.c
glneo: Will fix.

plat/ti/k3/common/bl31.c
glneo: The same could be done with BL33_BASE, but it seems the PRELOADED_ version of that was introduced to avoid confusion between the original which is a load address used by BL2 and earlier stages to put the BL33 at, and the address "preloaded" by an external bootloader. So I would like to keep the same naming here as our BL32 is externally loaded. If at some-point we want to have TF-A load BL32 then we will not define this and BL2 will do the loading with the other address.

docs/user-guide.rst
```diff
@@ -716,7 +716,7 @@ Arm development platform specific build options
    to have a Linux kernel image as BL33 by preparing the registers to these
    values before jumping to BL33. This option defaults to 0 (disabled). For now,
    it  only supports AArch64 kernels. ``RESET_TO_BL31`` must be 1 when using it.
-   If this option is set to 1, ``ARM_PRELOADED_DTB_BASE`` must be set to the
+   If this option is set to 1, ``PRELOADED_DTB_BASE`` must be set to the
```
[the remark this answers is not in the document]
glneo: Both options are available. You can load either U-Boot or Linux to the PRELOADED_BL33_BASE address. Both have been tested working with this patch-set.

If that is your preference I will change the name to PRELOADED_HW_CONFIG_BASE.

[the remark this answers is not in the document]
antonio-nino-diaz-arm: Actually, in that case you should document it. The platform documentation right now implies that U-Boot is needed.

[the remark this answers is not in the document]
glneo: Will update document with optional direct Linux boot path.

plat/ti/k3/include/platform_gic.h
```diff
@@ -0,0 +1,16 @@
+/*
```
soby-mathew: Because we have a global INCLUDES namespace for all the headers, the platform specific headers within the global INCLUDES should have a specific prefix, say  `plat_k3_*`. 

I am less concerned about the .c file names though.

glneo: No other platform does this, is this something new?

plat/ti/k3/common/bl31.c
glneo: By the same reasoning, why did BL33_BASE need the PRELOADED_ prefix, it could have also just been "BL33_BASE"? Should we change this? I just want to keep things consistent.

plat/ti/k3/include/platform_gic.h
soby-mathew: The headers expected to be exported to the `INCLUDES` from the platform with generic names are platform_def.h, plat_ld.S (if required) and plat_macros.S. Other headers are supposed to have a platform specific name. ARM platforms do this and most of the platforms do this as well, although there are a few exceptions. This is one of things I am looking out for as the number of platforms upstream increase unless we can namespace the header inclusion based on the directory.

plat/ti/k3/common/bl31.c
soby-mathew: Sorry I didn't elaborate previously. PRELOADED_BL33_BASE was actually meant to indicate to the BL2 (the loader) that BL33 doesn't need loading as it already pre-loaded. So its pre-loaded from BL2 point of view. From BL31 point of view, all the images are pre-loaded and this macro came in handy to override the default Non Secure jump address.

Hence as long as the image does not appear pre-loaded to BL2, it is better to avoid using the PRELOADED_ prefix. BL31 can use the BL32_BASE macro to figure out the BL32 load address.

Thinking more about it, PRELOADED_HW_CONFIG_BASE should be just HW_CONFIG_BASE going by the above reasoning. Also this macro doesn't do anything in generic code. Hence this macro should remain platform specific. Please keep the `ARM_PRELOADED_DTB_BASE` macro[1] and and you can introduce a `TI_K3_*` option.  Sorry for the flip flop on this.

[1] At our end, we can re-name this macro to be more consistent with what I mention and perhaps change it to `ARM_HW_CONFIG_BASE`.  We are less conservative with platform specific build option.

plat/ti/k3/include/platform_gic.h
glneo: Okay, thanks for the explanation, will fix.

plat/ti/k3/common/bl31.c
glneo: Then in our case both BL32 and HW_CONFIG should not be loaded by BL2 and so should *have* the preload prefix. Although we don't use BL2 right now, there may be some need for it at some point.

Anyway, this can all be changed later so I'll change them for now to match your preference.

glneo:
Okay, pushed a new set with changes from comments. I redid the naming to match with the ARM reference platforms except with "arm_" prefixes replaced with "k3_", hopefully this will take care of all the naming issues.

PR #1406, opened 2018-06-07
Make TF UUID RFC 4122 compliant

robertovargas-arm:
RFC4122 defines that fields are stored in network order (big endian),
but TF-A stores them in machine order (little endian by default in TF-A).
We cannot change the UUIDs that are already generated, but we can store
all the bytes using arrays and modify fiptool to generate the future
UUIDs with the correct byte order.

NOTE: this patch may generate build errors for out-of-tree UUID
definitions. The structure for the type uuid_t has a different layout
and any direct initialization with the previous layout will fail. The
macro DEFINE_SVC_UUID is deprecated in favour of an equivalent macro
DEFINE_SVC_UUID2, which generates a UUID with the correct byte order.

Change-Id: I97be2d3168d91f4dee7ccfafc533ea55ff33e46f


dp-arm:
jenkins: test this please

dp-arm:
@hzhuang1 @mtk09422 @TonyXie06 @sivadur can you review the platform specific changes?

dp-arm:
@hzhuang1 @mtk09422 @TonyXie06 @sivadur i will be merging this by the end of tomorrow UK time.

PR #1407, opened 2018-06-07
Juno: Bump up the BL1-RW size

soby-mathew:
This patch bumps up the BL1-RW size for Juno and at the same time reduces the
BL2 size when TBB is enabled, TF_MBEDTLS_KEY_ALG=rsa+ecdsa. We were able to
reduce the BL2 size for this config as it was observed that the peak memory
usage is only reached when SPD=opteed and the dual rsa+ecdsa support
is not needed for this case.

Change-Id: Ia9009771b5cfd805e9cc75410aabb7db99fc2fbc


soby-mathew:
jenkins: test this please

dp-arm:
jenkins: test this please

dp-arm:
Infrastructure failures, ignoring.

PR #566, opened 2016-03-24
Add io_emmc support and lower level emmc prototypes

sbranden:
This submission adds upper level io_emmc support to ATF and
provides the function prototypes for necessary lower level
driver.

PR #1408, opened 2018-06-08
invalid DCache of stack before we use it

kenkuang:
On the warm boot path, when we use HW_ASSISTED_COHERENCY options
We have to invalid DCache of stack before we use it
and, before we enable DCache, we must make sure invalid the stack

Signed-off-by: Ken Kuang <ken.kuang@unisoc.com>

### DIFF
--- a/bl31/aarch64/bl31_entrypoint.S
+++ b/bl31/aarch64/bl31_entrypoint.S
@@ -155,6 +155,39 @@ func bl31_warm_entrypoint
 		_exception_vectors=runtime_exceptions
 
 	/*
+	 * On the warm boot path, when we use HW_ASSISTED_COHERENCY options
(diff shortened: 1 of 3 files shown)


jeenu-arm:
Looks like you've encountered the same problem as described by [issue #566](https://github.com/ARM-software/tf-issues/issues/566). Although stack invalidation is one way to solve the problem, the main selling point of DynamIQ support to avoid doing any kind of cache-maintenance in software.

As commented in the issue I linked above, the approach we thought we'd take was to enable MMU without using stack in the first place. A small series of patches is almost ready; we should be able to get a PR out for that soon next week.

kenkuang:
@jeenu-arm Thank you so much. I'll close this pull request.

PR #1414, opened 2018-06-13
rpi3: Fix kernel boot documentation

antonio-nino-diaz-arm:
The order of the arguments of memmap was swapped. The old command was reserving 256 MiB from the 16 MiB barrier, it should be reserving only 16 MiB at the 256 MiB barrier.

It worked because the memory used by the Trusted Firmware was reserved anyway.


antonio-nino-diaz-arm:
By the way, @grandpaul, this is what you told me on your email.

grandpaul:
Yes. 
Tested-by: Ying-Chun Liu (PaulLiu) <paul.liu@linaro.org>

PR #1416, opened 2018-06-13
Enable MMU without using stack

jeenu-arm:


jeenu-arm:
This patch **has not been** merged yet, but was made available as an early preview. This might undergo some more changes.

jeenu-arm:
See #1429 for an update.

PR #1420, opened 2018-06-14
xlat_v2: add a check on mm_cursor->size to avoid infinite loop

Yann-lms:
The issue can occur if end_va is equal to the max architecture address,
and when mm_cursor point to the last entry of mmap_region_t table: {0}.
The first line of the while will then be true, e.g. on AARCH32, we have:
mm_cursor->base_va (=0) + mm_cursor->size (=0) - 1 == end_va (=0xFFFFFFFF)
And the mm_cursor->size = 0 will be lesser than mm->size

A check on mm_cursor->size != 0 should be done as in the previous while,
to avoid such kind of infinite loop.

fixes arm-software/tf-issues#594


antonio-nino-diaz-arm:
IP Review +1

PR #1421, opened 2018-06-14
Build: add cpp build processing for dtb

Yann-lms:
This is an add-on feature that allows processing
device tree with external includes.

fixes arm-software/tf-issues#595

Signed-off-by: Lionel Debieve <lionel.debieve@st.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -198,7 +198,8 @@ BL_COMMON_SOURCES	+=	common/bl_common.c			\
 				${COMPILER_RT_SRCS}			\
 				${STDLIB_SRCS}
 
-INCLUDES		+=	-Iinclude/bl1				\
+INCLUDES		+=	-Iinclude				\
+				-Iinclude/bl1				\
 				-Iinclude/bl2				\
 				-Iinclude/bl2u				\
 				-Iinclude/bl31				\
--- a/make_helpers/build_macros.mk
+++ b/make_helpers/build_macros.mk
@@ -372,11 +372,14 @@ endef
 define MAKE_DTB
 
 $(eval DOBJ := $(addprefix $(1)/,$(call SOURCES_TO_DTBS,$(2))))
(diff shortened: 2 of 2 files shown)


antonio-nino-diaz-arm:
IP Review +1

soby-mathew:
jenkins: test this please

Makefile
```diff
@@ -198,7 +198,8 @@ BL_COMMON_SOURCES	+=	common/bl_common.c			\
 				${COMPILER_RT_SRCS}			\
 				${STDLIB_SRCS}
 
-INCLUDES		+=	-Iinclude/bl1				\
+INCLUDES		+=	-Iinclude				\
```
antonio-nino-diaz-arm: Why are you adding this line?

Yann-lms: It will then be easier to include the header files either in DT or in drivers, like what is done in kernel for example.
The include line is:
`#include <dt-bindings/clock/plat-clks.h>`

That also mean that all those files would be in include/dt-bindings/.

antonio-nino-diaz-arm:
This patch fails to build because the fdts folder isn't in the list of include paths. For example:
```
dtc -I dts -O dtb -d ./build/fvp/debug/fdts/fvp-base-gicv3-psci.d -o build/fvp/debug/fdts/fvp-base-gicv3-psci.dtb ./build/fvp/debug/fdts/fvp-base-gicv3-psci.pre.dts
FATAL ERROR: Couldn't open "fvp-base-gicv3-psci-common.dtsi": No such file or directory
```

Yann-lms:
Hi,

As I do not use the /include/ directive but directly #include, I didn't see this error, sorry.
To correct that, we can add "-i fdts" to the DTC command line.
I will push a new patch.

soby-mathew:
jenkins: test this please

PR #1424, opened 2018-06-15
cortex-a55: disable Dcache in power-down handler

ajs-sun:
If HW_ASSISTED_COHERENCY is not used, cache maintenance must be
performed in the power-up and power-down sequence.
In the power-down path, the used stack is cleaned and invalidated in
function "psci_do_pwrdown_cache_maintenance". If that function is called
with Dcache enabled, the L1D/L2/L3 cache will be reallocated after
function return due to the follow-up stack operation.

This patch turns off Dcache to avoid that.

Signed-off-by: Aijun Sun <aijun.sun@spreadtrum.com>

### DIFF
--- a/lib/cpus/aarch64/cortex_a55.S
+++ b/lib/cpus/aarch64/cortex_a55.S
@@ -12,10 +12,26 @@
 #include <plat_macros.S>
 
 	/* ---------------------------------------------
+	 * Disable L1 data cache and unified L2/L3 cache
+	 * ---------------------------------------------
+	 */
+func cortex_a55_disable_dcache
+	mrs	x1, sctlr_el3
+	bic	x1, x1, #SCTLR_C_BIT
+	msr	sctlr_el3, x1
+	isb
+	ret
+endfunc cortex_a55_disable_dcache
+	/* ---------------------------------------------
 	 * HW will do the cache maintenance while powering down
 	 * ---------------------------------------------
 	 */
 func cortex_a55_core_pwr_dwn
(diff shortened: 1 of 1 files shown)


antonio-nino-diaz-arm:
IP Review +1

soby-mathew:
Unfortunately, HW_ASSISTED_COHERENCY brings in other assumptions about power management capabilities of the core (like cluster power management). Hence adding cache flush for core power down is not sufficient. 

May I know you why you want to turn of HW_ASSISTED_COHERENCY on a DynamIQ CPU ?

ajs-sun:
Thanks for your comments.
Previously we used  HW_ASSISTED_COHERENCY but got in trouble because of cache coherency issue so I tried to turn it off. Although that issue can be fixed with patch PR#1416, I think turning off HW_ASSISTED_COHERENCY should work also. In fact this feature is not forced to enable for cortex-a55 platform in arm trusted firmware currently.
HW_ASSISTED_COHERENCY optimizes the sequence for CPU power-down and power-up. But not using it should not cause problem.

soby-mathew:
Hmm, I think this may cause problems for regular memory writes (if any) after the D-cache is disabled. Since the writes bypass the cache, on power up, the CPU will see the stale values in cache when read after cache is enabled. 

Hence this is not something we want to merge upstream and hopefully PR #1416 should solve the issue you have.

ajs-sun:
Thanks. Close this PR.

jeenu-arm:
See #1429 for an update.

PR #1427, opened 2018-06-18
Add support for Socionext Synquacer SC2A11 SoC based Developerbox platform.

b49020:
Hi Everyone,

This pull request adds support for Socionext Synquacer SC2A11 SoC based Developerbox platform. This platform is compliant with Linaro’s 96Boards open hardware specification. For more info regarding the platform, please refer to https://www.96boards.org/product/developerbox/.

Also this Synquacer SoC is derived from ARM Ashbrook.

Looking forward to your valuable comments!

Regards,
Sumit


plat/socionext/synquacer/sq_psci.c
```diff
@@ -0,0 +1,188 @@
...
+				 system_state);
+}
+
+void sq_pwr_domain_off(const psci_power_state_t *target_state)
+{
+	sq_power_down_common(target_state);
+}
+
+void __dead2 sq_system_off(void)
+{
```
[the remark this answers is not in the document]
b49020: Sure I will use `udelay()` option instead.

b49020:
@antonio-nino-diaz-arm Updated PR to incorporate your comment. Please review again.

plat/socionext/synquacer/sq_bl31_setup.c
```diff
@@ -44,13 +44,18 @@ uint32_t sq_get_spsr_for_bl33_entry(void)
...
 	/* Figure out what mode we enter the non-secure world in */
 	el_status = read_id_aa64pfr0_el1() >> ID_AA64PFR0_EL2_SHIFT;
 	el_status &= ID_AA64PFR0_ELX_MASK;
 
 	mode = (el_status) ? MODE_EL2 : MODE_EL1;
 
+	/* enable the stage 2 override page tables if PD[1] is set */
+	if (gpio[0] & 0x2)
+		mode = MODE_EL1;
+
```
antonio-nino-diaz-arm: Ok, so you can select EL1 or EL2 based on the signal from a GPIO pin. However, you always set up the stage 2 tables. If BL33 gets control at EL2, it can replace them, making the call to `plat_setup_s2_translation()` useless. Wouldn't it be better to map them only if you are entering BL33 in EL1?

docs/plat/synquacer.rst
```diff
@@ -0,0 +1,114 @@
...
+
+       export WORKSPACE=$HOME/build/developerbox-firmware
+       mkdir -p $WORKSPACE
+
+-  Run the following commands to clone the source code:
+
+   .. code:: shell
+
+       cd $WORKSPACE
+       git clone https://github.com/ARM-software/arm-trusted-firmware -b integration
```
antonio-nino-diaz-arm: Also, I don't think that checking out integration is the right thing to do, you should also checkout master in the case of the TF.

plat/socionext/synquacer/sq_bl31_setup.c
b49020: Thanks for your comment. Actually this is a workaround for PCIe for current version of SC2A11 SoC. After more discussion on this internally, it seems that this bug in hardware will be fixed in upcoming SoC. So internally we reached to a conclusion to not upstream this workaround due to its future maintainability. I will remove these workaround patches from this PR.

docs/plat/synquacer.rst
b49020: Sure I will change it to master branch.

b49020:
@antonio-nino-diaz-arm Updated PR. Please have a look.

plat/socionext/synquacer/sq_bl31_setup.c
antonio-nino-diaz-arm: Ok. Just out of curiosity, did you create a tool to generate the translation tables?

plat/socionext/synquacer/sq_bl31_setup.c
```diff
@@ -0,0 +1,171 @@
...
+#ifdef BL32_BASE
+	/* Populate entry point information for BL32 */
+	SET_PARAM_HEAD(&bl32_image_ep_info,
+				PARAM_EP,
+				VERSION_1,
+				0);
+	SET_SECURITY_STATE(bl32_image_ep_info.h.attr, SECURE);
+	bl32_image_ep_info.pc = BL32_BASE;
+	bl32_image_ep_info.spsr = sq_get_spsr_for_bl32_entry();
```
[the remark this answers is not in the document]
b49020: If you see sq_get_spsr_for_bl32_entry api above, it says SPD setup this service and it returns 0.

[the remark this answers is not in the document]
AtomarMent: But , I was wondering if it https://github.com/ARM-software/arm-trusted-firmware/blob/master/plat/arm/common/arm_common.c#L103 can be re-used ?

For instance, what happens if you don't set  "bl32_image_ep_info.spsr = sq_get_spsr_for_bl32_entry(); " here ?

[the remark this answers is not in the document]
b49020: I have not tried to use functions from ARM platform (plat/arm/ folder). As in case if any function changes for ARM platform should not affect other platforms too.

plat/socionext/synquacer/sq_bl31_setup.c
b49020: @ardbiesheuvel Can you please help answer above query?

ardbiesheuvel: I was very clear when we discussed this offline: this functionality does *not* belong in the upstream port. It works around a PCIe issue in rev 1 silicon, and it has its own share of issues.

ardbiesheuvel: Ah, that wasn't the question - apologies.

The stage 2 tables are created statically in NOR flash

https://git.linaro.org/leg/noupstream/edk2-platforms.git/tree/Silicon/Socionext/SynQuacer/Stage2Tables/Stage2Tables.S?h=developer-box&id=ba32985a0631ef371b06cdbfe53f4dd6d11557ba

b49020: Thanks @ardbiesheuvel

antonio-nino-diaz-arm: Oh, ok, so there's no tool. You simply generate the tables by hand. Still interesting, thanks.

b49020:
Re-based to latest integration branch.

plat/socionext/synquacer/sq_psci.c
```diff
@@ -0,0 +1,188 @@
...
+	volatile uint32_t *gpio = (uint32_t *)PLAT_SQ_GPIO_BASE;
+
+	/* set PD[9] high to power off the system */
+	gpio[5] |= 0x2;		/* set output */
+	gpio[1] |= 0x2;		/* set high */
+	dmbst();
```
ardbiesheuvel: Note that this arrangement of wiring GPIO PD[9] to the ATX power control is specific to Developerbox, which is one particular platform that uses the SynQuacer SoC.

Not sure how this is typically handled in ARM-TF, but perhaps we should at least put it inside and #ifdef block?

b49020: Currently I haven't added board specific granular support and makefile due to only board available being Developerbox based on SynQuacer SoC.
AFAIK, in this case if power-off sequence is board specific, then we should create a common weak "power_off" api that would be implemented in board specific files.
Regarding #ifdef block, I don't see any value add at this point without board specific compilation target.
So I think we could add granular board specific support at later stage when we have more than one board available based on SynQuacer SoC and perhaps more clarity on other board specific functionality. Please share your thoughts on this.

antonio-nino-diaz-arm: In this case I think the documentation should be more clear and specify that it's only the Developerbox that is supported. Right now the docs only talk about the Developerbox, but someone could think that as the folder is called "synquacer" it applies to all the platforms that use it. The document is titled "Trusted Firmware-A for Socionext Synquacer SoCs", which can make someone think it applies to all of them.

Maybe you can write after the title something like:

> Socionext's SC2A11 is a multi-core processor with 24 cores of Arm (R)
> Cortex-A53. The Developerbox, of 96boards, is a platform that contains this
> processor. This port of the Trusted Firmware only supports this platform at
> the moment.

(Adapted from http://www.socionext.com/en/download/catalog/AD04-00111-2E.pdf)

b49020: @antonio-nino-diaz-arm Sure will add this text to documentation.

plat/socionext/synquacer/drivers/scpi/sq_scpi.c
```diff
@@ -0,0 +1,170 @@
+/*
```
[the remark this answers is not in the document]
b49020: In case of SCPI driver, we have a new api to get DRAM info from SCP firmware using SCPI interface: "uint32_t scpi_get_draminfo(struct draminfo *info)". This is part of patch: "synquacer: Retrieve DRAM info from SCP firmware".
I think this could also be accommodated as part of abstraction layer for the shared driver in future.

PR #1432, opened 2018-06-20
[RFC] Add MMC framework

Yann-lms:
It is largely based on existing eMMC framework by Haojian Zhuang
(@hzhuang1).

The MMC framework supports both eMMC and SD card devices. It was
written as a new framework since breaking few eMMC framework APIs.

At card probe and after the reset to idle command (CMD0), a Send
Interface Condition Command is sent (CMD8) to distinguish between
eMMC and SD card devices. eMMC devices go through the same
sequence as in the former eMMC framework. Else the framework
uses commands dedicated to SD-cards for init or frequency switch.

A structure is created to share info with the driver. It stores:
- the MMC type (eMMC, SD or SD HC)
- the device size
- the max frequency supported by the device
- the block size: 512 for eMMC and SD-HC and read from CSD
 structure for older SD-cards

Restriction to align buffers on block size has been removed.

The MMC framework includes some MISRA compliance coding style
maybe not yet ported in the existing eMMC framework.

fixes arm-software/tf-issues#597

### DIFF
--- a/drivers/mmc/mmc.c
+++ b/drivers/mmc/mmc.c
@@ -1,0 +1,714 @@
+/*
+ * Copyright (c) 2018, ARM Limited and Contributors. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/* Define a simple and generic interface to access eMMC and SD-card devices. */
+
+#include <arch_helpers.h>
+#include <assert.h>
+#include <debug.h>
+#include <errno.h>
+#include <mmc.h>
+#include <stdbool.h>
+#include <string.h>
+#include <utils.h>
+
+#define MMC_DEFAULT_MAX_RETRIES		5
+#define SEND_OP_COND_MAX_RETRIES	100
+
+#define MULT_BY_512K_SHIFT		19
+
+static const struct mmc_ops *ops;
+static unsigned int mmc_ocr_value;
+static struct mmc_csd_emmc mmc_csd;
+static unsigned char mmc_ext_csd[512] __aligned(4);
+static unsigned int mmc_flags;
+static struct mmc_device_info *mmc_dev_info;
+static unsigned int rca;
+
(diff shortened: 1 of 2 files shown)


antonio-nino-diaz-arm:
IP Review +1

drivers/mmc/mmc.c
```diff
@@ -0,0 +1,680 @@
+/*
+ * Copyright (c) 2018, ARM Limited and Contributors. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Defines a simple and generic interface to access eMMC and SD-card devices.
```
antonio-nino-diaz-arm: Split this into two comments, the regular one used across the codebase and your other line.

drivers/mmc/mmc.c
```diff
@@ -0,0 +1,712 @@
...
+		}
+
+		ret = mmc_send_cmd(MMC_CMD(13), rca << RCA_SHIFT_OFFSET,
+				   MMC_RESPONSE_R(1), &resp_data[0]);
+		if (ret != 0) {
+			return ret;
+		}
+
+		if ((resp_data[0] & STATUS_SWITCH_ERROR) != 0U) {
+			panic();
```
robertovargas-arm: Do you think we must panic in this case? I can understand that maybe a good approach is to try a second boot source as it is done in https://github.com/ARM-software/arm-trusted-firmware/commit/27b2493c6f4fe32a0fa76dff123d193dad86c315. The same apply to the other places where you are calling panic().

Yann-lms: OK, I will check this other patch and make the required changes

vchong:
@dp-arm @hzhuang1 is on vacation and will be back soon (11th). Please hold.

hzhuang1:
@dp-arm @vchong @Yann-lms I'm fine on this patch. Please go ahead.

PR #1450, opened 2018-06-24
Marvell support for Armada 8K SoC family

kostapr:
This is the first pull request for Marvell Armada 8K SoC family support in ATF mainline.
Tested with u-boot 2017 and UEFI 2.7 on Marvell development boards.


kostapr:
The following additional components are required for building the flash image for Marvell A8K board:
* BLE (early init code executed by the BootROM)
https://github.com/MarvellEmbeddedProcessors/ble-marvell/tree/atf-mainline
* DDR initialization library
https://github.com/MarvellEmbeddedProcessors/mv-ddr-marvell/tree/mv_ddr-armada-atf-mainline

Build details are located in doc/marvell/build.txt

plat/marvell/a8k/common/plat_pm.c
```diff
@@ -0,0 +1,781 @@
...
+				reg |= (1 << shift);
+		}
+
+		/* Set the GPIO register, for last step just store register address and values to system registers */
+		if (idx < pm_cfg->cfg.gpio.step_count - 1) {
+			mmio_write_32(MVEBU_CP_GPIO_DATA_OUT(info->cp_index, info->gpio_index), reg);
+			mdelay(pm_cfg->cfg.gpio.delay_ms);
+		} else {
+			/* Save GPIO register value to X17, and address to X18 */
+			__asm__ volatile (
```
antonio-nino-diaz-arm: How is this supposed to work? Won't the compiler simply remove it?

plat/marvell/a8k/common/plat_pm.c
```diff
@@ -0,0 +1,781 @@
...
+				"mov	x17, %0\n\t"
```
antonio-nino-diaz-arm: How is this supposed to work? Won't the compiler simply remove it?

plat/marvell/a8k/common/plat_pm.c
```diff
@@ -0,0 +1,781 @@
...
+				"mov	x18, %1\n\t"
+				: : "r" (reg), "r" (addr));
```
antonio-nino-diaz-arm: How is this supposed to work? Won't the compiler simply remove it?

plat/marvell/a8k/common/plat_pm.c
```diff
@@ -0,0 +1,781 @@
...
+	srcmd = mmio_read_32(sdram_reg);
+	srcmd &= ~(MVEBU_USER_CMD_CH0_MASK | MVEBU_USER_CMD_CS_MASK | MVEBU_USER_CMD_SR_MASK);
+	srcmd |= (MVEBU_USER_CMD_CH0_EN | MVEBU_USER_CMD_CS_ALL | MVEBU_USER_CMD_SR_ENTER);
+
+	/*
+	 * Wait for DRAM is done using registers access only.
+	 * At this stage any access to DRAM (procedure call) will
+	 * release it from the self-refresh mode
+	 */
+	__asm__ volatile (
```
antonio-nino-diaz-arm: I think this should be in an assembly file as a function and call it from here.

kostapr:
@antonio-nino-diaz-arm 
Thank you very much for the review!
One question - when I finish fixing code - should I create a new pull request or there is a way to update this one?
Thanks you
Konstantin

antonio-nino-diaz-arm:
Normally we ask people to force push to the same branch in their repository, that updates the pull request with the new patches. In your case, we could try that, your PR isn't as huge as the Cavium one. If this becomes unmanageable, maybe we'll need to split the PR into two or more (to add the drivers before the platform port itself).

drivers/ti/uart/aarch64/16550_console.S
```diff
@@ -137,6 +137,9 @@ func console_16550_core_putc
 #if ENABLE_ASSERTIONS
 	cmp	x1, #0
 	ASM_ASSERT(ne)
+#else
```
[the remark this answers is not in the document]
kostapr: Always exit when the base is zero is fine with me. I think it was the older code behavior (maybe 1.2 or 1.3). 
Which information could be provided by ASM_ASSERT if the UART base is zero? I would expect some error message on the console, which is in this case not initialized.
Would it be acceptable removing "#if ENABLE_ASSERTION" block and using the code from this patch (check if zero and return if yes)?

docs/marvell/misc/mvebu-amb.txt
```diff
@@ -0,0 +1,45 @@
...
+Mandatory structures:
+	amb_memory_map - Array that include the configuration of the windows
+	  every window/entry is a struct which has 2 parameters:
+	  - base address of the window
+	  - Attribute of the window
+
+Examples:
+	struct addr_map_win amb_memory_map[] = {
+		{0xf900,	AMB_DEV_CS0_ID},
+	};
```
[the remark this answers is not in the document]
kostapr: I will create a new file named address_decoding.txt, for this diagram.

drivers/io/io_fip.c
```diff
@@ -272,7 +272,9 @@ static int fip_file_read(io_entity_t *entity, uintptr_t buffer, size_t length,
 	uintptr_t backend_handle;
 
 	assert(entity != NULL);
+#ifndef PLAT_ALLOW_ZERO_ADDR_COPY
 	assert(buffer != (uintptr_t)NULL);
+#endif
```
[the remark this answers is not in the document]
robertovargas-arm: This is the first platform that needs load an image at 0. It makes invalid the asserts, and since this code is going to be executed with the MMU enabled, the assert doesn't help too much. I think it is better if you remove the NULL checks in these asserts.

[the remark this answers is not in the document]
kostapr: @robertovargas-arm, Thank you! Should I remove all cases where the buffer pointer is checked for 0? In other words - should I remove the PLAT_ALLOW_ZERO_ADDR_COPY definition along with every call to "assert(buffer != (uintptr_t)NULL)"?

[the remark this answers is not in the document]
robertovargas-arm: Yes. We can see now that this check is invalid in one platform, so it doesn't make sense. As the MMU is already enabled, if 0 is not a valid address then it will generate a fault abort, so the assert is not adding too much. I see a bit dangerous that you load something using the address 0, because it means that any NULL pointer will not be caught, opening a door to silent errors that can be exposed later, but this is a platform decision.

drivers/ti/uart/aarch64/16550_console.S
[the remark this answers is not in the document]
antonio-nino-diaz-arm: >  Which information could be provided by ASM_ASSERT if the UART base is zero? I would expect some error message on the console, which is in this case not initialized.

Heh, that's a good point. However, I think that the assert is still useful. You may be trying to print to an uninitialized console, and assert will print to the plat crash console. The crash console may or may not use the multi console API. Arm platforms, RPi3, QEMU, Allwinner... they all use direct calls to the console core functions instead of going through the multi console API, so you can be sure to get the error message.

Maybe what we need is a dummy console driver.

plat/marvell/a8k/common/plat_pm.c
kostapr: Hmm... Agree, I should check it with the original code author and maybe look though the code disassembly.

drivers/ti/uart/aarch64/16550_console.S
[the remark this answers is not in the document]
antonio-nino-diaz-arm: Maybe what you need to use is the multi console driver. With it you can register multiple/one/no console drivers, and you could get the result you want. Basically, you can run with no console registered (and you can even register a runtime console when you are about to leave BL31, for example). You could also not register the console in release builds, if that's what you want.

If you force push to the PR, the old patches are replaced by the new ones. It makes it hard to compare versions of the same PR, but some comments stay in place, so it's better than nothing.

antonio-nino-diaz-arm:
By the way, could you add yourself to the maintainers.rst file (and acknowledgements.rst if you want)? The maintainers.rst file will need a rebase after https://github.com/ARM-software/arm-trusted-firmware/pull/1457 is merged, but that way we don't forget about it. Now we have a script to notify maintainers automatically when there are changes to their platforms, but that depends on the maintainers file being correct.

kostapr:
Sure I will add myself there. Should I do it in a separate patch before fixing this PR?

antonio-nino-diaz-arm:
It doesn't really matter. We still need to review a bit more in-depth, so don't feel too bad about pushing new patches that replace the current ones with my comments, we will have to look at all of that again.

drivers/ti/uart/aarch64/16550_console.S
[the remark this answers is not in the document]
kostapr: I definitely planned to switch to the new console API. It was not working when I tried it first (stop running somewhere before BL3 load). Since I was busy with the other stuff, I decided to debug this issue and switch to the new API bit later. Currently my target is to create an initial code base for Marvell platforms in TF-A (no trusted boot, single platform only, etc.) and then push additional features and platforms one by one.
For archiving that I can drop some patches if a quick fix is not an option. Specifically, if I drop this patch, I will loose an option to download recovery image over serial line. This is very useful feature allowing to de-brick system with bad boot image without JTAG, but it can be disabled for the first patch series.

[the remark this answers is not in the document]
antonio-nino-diaz-arm: I agree that it's better not to modify more things for the first version.

Then I think it's better if you drop the patch from the upstream port as there is no code to load images through the serial port so the upstream code doesn't benefit from this. After it is merged and you migrate to the multi console driver you won't need this patch.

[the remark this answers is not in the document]
kostapr: The recovery image load is a BootROM feature, so there is no special boot image required. A regular flash image created by TF-A build process can be downloaded over the board serial line for bringing the system up after a bad update.
The helper script for doing so is included in u-boot "tools" folder - https://github.com/u-boot/u-boot/blob/master/tools/mrvl_uart.sh
So this feature is quite useful if included.

plat/marvell/a8k/common/plat_pm.c
kostapr: @antonio-nino-diaz-arm we are trying to keep these values in registers for ensuring there is no DRAM access for this data later. Any access to DRAM will return it from self-refresh mode. Do you have suggestion for reaching the same result without specifying register numbers here?

plat/marvell/a8k/common/plat_pm.c
kostapr: The problem is the DRAM access. We should prevent it on this stage, therefore using inline assembly.

kostapr:
I removed support for uninitialized console, so UART download will not work with the current code.
Tested with latest Linaro GCC 7.3.1 and u-boot 2017 on A8040-DB and MacciatoBIN boards up to the u-boot prompt.
Updated ble-marvell and mv-ddr-marvell repositories to match the current pull request.
Copyright headers are currently under review of our legal department. If I allowed to change all headers to the common ones (ARM and Contributors), I will do so immediately.

docs/marvell/build.txt
```diff
@@ -0,0 +1,182 @@
+TF-A Build Instructions
+======================
+
+This section describes how to compile the ARM Trusted Firmware (TF-A) project for Marvell's platforms.
+
+Build Instructions
+------------------
+(1) Set the cross compiler::
+
+		> export CROSS_COMPILE=/path/to/toolchain/aarch64-marvell-linux-gnu-
```
antonio-nino-diaz-arm: Interesting, is that toolchain used in order to optimize the result or is it that ``aarch64-linux-gnu-`` simply doesn't work for you?

drivers/marvell/cache_llc.c
```diff
@@ -0,0 +1,110 @@
...
+
+	INFO("Enabling LLC\n");
+
+	/*
+	 * Enable L2 UniqueClean evictions
+	 *  Note: this configuration assumes that LLC is configured
+	 *	  in exclusive mode.
+	 *	  Later on in the code this assumption will be validated
+	 */
+	__asm__ volatile ("mrs %0, s3_1_c15_c0_0" : "=r" (reg));
```
antonio-nino-diaz-arm: If I am not mistaken, this platform uses Cortex-A72, right? Then I think you should add to ``cortex_a72.h`` this, in a similar way as it is done in ``cortex_a53.h``:
```
#define CORTEX_A72_L2ACTLR_EL1 S3_1_C15_C0_0
```
And in this file do:
```
#include <cortex_a72.h>
...
__asm__ volatile ("mrs %0, " ## CORTEX_A72_L2ACTLR_EL1 : "=r" (reg));
```
I don't think it's worth the effort creating a helper function for this, but at least it should be possible to tell the register you are accessing without checking the TRM.

drivers/marvell/mochi/apn806_setup.c
```diff
@@ -0,0 +1,251 @@
...
+	/* Set the SMMU page size to 64 KB */
+	reg = mmio_read_32(SMMU_sACR);
+	reg |= SMMU_sACR_PG_64K;
+	mmio_write_32(SMMU_sACR, reg);
+}
+
+static void apn806_errata_wa_init(void)
+{
+	/*
+	 * EERATA ID: RES-3033912 - Internal Address Space Init state causes
```
antonio-nino-diaz-arm: nit: ERRATA

include/plat/marvell/a8k/common/board_marvell_def.h
```diff
@@ -0,0 +1,107 @@
+/*
```
antonio-nino-diaz-arm: nit: When the copyright issue is sorted, just have one header in this file (and the others that have 2 headers).

include/plat/marvell/a8k/common/marvell_def.h
```diff
@@ -0,0 +1,208 @@
...
+#define MAX_MMAP_REGIONS		(PLAT_MARVELL_MMAP_ENTRIES +	\
+					 MARVELL_BL_REGIONS)
+
+#define MARVELL_CONSOLE_BAUDRATE	115200
+
+/******************************************************************************
+ * Required platform porting definitions common to all MARVELL std. platforms
+ *****************************************************************************/
+
+#define ADDR_SPACE_SIZE			(1ull << 32)
```
antonio-nino-diaz-arm: Define ``PLAT_VIRT_ADDR_SPACE_SIZE`` and ``PLAT_PHY_ADDR_SPACE_SIZE`` instead, ``ADDR_SPACE_SIZE`` is deprecated.

plat/marvell/a8k/common/plat_pm.c
antonio-nino-diaz-arm: The problem is that the compiler doesn't know that you want to keep the values in these registers. As soon as you exit this function, they can be used by the compiler. I see that you are using them in ``a8k_pwr_domain_pwr_down_wfi()``. I suspect that right now this works by pure luck because the code in between is relatively simple. Instead of doing this I suggest you to pass the values like this:
```
static void plat_marvell_power_off_gpio(struct power_off_method *pm_cfg, u_register_t *gpio_value, u_register_t *gpio_address)
{
    ...
    *gpio_value = reg;
    *gpio_address = addr;
    ...
}
```
And in ``a8k_pwr_domain_pwr_down_wfi()``:
```
u_register_t gpio_value, gpio_address;
plat_marvell_power_off_prepare(pm_cfg, &gpio_value, &gpio_address);
```
Then use ``gpio_value`` and ``gpio_address`` in your block of inline assembly.

The stack is in SRAM so DRAM accesses shouldn't be a problem with this solution.

plat/marvell/a8k/common/plat_pm.c
```diff
@@ -0,0 +1,825 @@
...
+}
+
+/*
+ * Prepare for the power off of the system
+ */
+static void plat_marvell_power_off_prepare(struct power_off_method *pm_cfg)
+{
+	switch (pm_cfg->type) {
+	case PMIC_GPIO:
+		plat_marvell_power_off_gpio(pm_cfg);
```
antonio-nino-diaz-arm: nit: add break here for clarity.

docs/marvell/build.txt
kostapr: Oops, this is the old one. We do not use any internally built toolchains. Linaro only now.

kostapr: As for UART download with the current code, I need that access to uninitialized console driver will not cause error or exception. Instead all such attempts should be just ignored.
The BLE and DDR init code will try to output some error messages during image download and I have to ignore this attempts in order to comply with xmodem protocol.

include/plat/marvell/a8k/common/marvell_def.h
kostapr: OK, I missed this when ported older code

plat/marvell/a8k/common/plat_pm.c
kostapr: Thank you, I just though to use register_t variables as you suggested. I will re-implement this  PM code.

drivers/marvell/cache_llc.c
kostapr: OK, will do

docs/marvell/build.txt
antonio-nino-diaz-arm: I see. so you need a working console for xmodem, but you want to filter all print commands and only let putc from xmodem work, right? I'm thinking that the multi console driver would work.

The way I'd do it is to unregister the console right before the file transfer routine, then use the direct console functions for the file transfer (any printf will be ignored because there is no registered console) and then register the console when the transfer is finished.

kostapr: That is basically what we do. The BLE (BootROM stage code that uses TF-A code as a library and a build environment) is loaded by SoC BootROM into L3 cache, detects that it runs in UART mode and bypasses the console initialization. Then all code that uses TF_A console APIs effectively skipped. If I recall correctly this BLE code was initially based on ATF-v1.2, where the console driver possible ignored "putc" and "getc" when the console base address was 0. So it worked well until asserts were added.
c13b2e32 (Varun Wadekar 2015-05-19 16:44:17 +0530 103) func console_core_putc
c13b2e32 (Varun Wadekar 2015-05-19 16:44:17 +0530 104)  /* Check the input parameter */
c13b2e32 (Varun Wadekar 2015-05-19 16:44:17 +0530 105)  cbz     x1, putc_error

antonio-nino-diaz-arm: Yes, I see, that's why I say that the multi console driver can help. With the multi console driver, ``putc`` can work without any active console, it simply does nothing. However, ``core_xxxx_putc`` can't. Now it is the multi console driver the one that is supposed to have this high level functionality, not the core console functions.

The idea is to move away from the old console in favor of the multi console, so you will have to migrate eventually. Maybe this is a good reason to do it.

kostapr: Sure, I have planned to do this migration. But probably not in this patch series.

drivers/marvell/cache_llc.c
kostapr: Concatenation will not work in this case (allowed only in macro definitions), so it looks like the only way to implement it with CORTEX_A72_L2ACTLR_EL1 is a helper function.

plat/marvell/a8k/common/plat_pm.c
```diff
@@ -0,0 +1,832 @@
...
+
+/*****************************************************************************
+ * A8K handlers to shutdown/reboot the system
+ *****************************************************************************
+ */
+static void __dead2 a8k_system_off(void)
+{
+	ERROR("%s:  needs to be implemented\n", __func__);
+	panic();
+	wfi();
```
antonio-nino-diaz-arm: panic() isn't supposed to return, you can remove everything after the first one.

include/plat/marvell/a8k/common/board_marvell_def.h
antonio-nino-diaz-arm: Any news regarding the copyright? If it is a problem to change it, we could accept it as it is. We mostly do this for consistency across all codebase (if not we have to start adding "parts copyright Arm.."). Maybe you could remove the link to the SPDX website, though, but that's just a nit if you think it's worth your time.

kostapr: Unfortunately the legal department demanded me to fill a formal request, which will take me some time to create and submit. A simple email query did not work for this case. At this moment I do not expect to close this issue shortly, but doing my best to complete all mandatory steps on  my side.
As for link to SPDX site - in the past it was requested by our legal in exchange to the full BSD 3-clause license text removal from the copyright header. Legal people have different view to all this stuff then us, technical crowd :-)

haklai: Thanks Antonio for the efforts that you put in reviewing this patchset. 
It will be great if you could accept keeping the original Marvell copyright as you suggested.

soby-mathew:
jenkins: test this please

include/plat/marvell/a8k/common/board_marvell_def.h
antonio-nino-diaz-arm: Yeah, I agree that legal stuff can be tricky to get right. We can accept the changes as they are now.

drivers/marvell/mochi/ap807_setup.c
```diff
@@ -0,0 +1,237 @@
+/*
+ * Copyright (C) 2018 Marvell International Ltd.
+ *
+ * SPDX-License-Identifier:	BSD-3-Clause
+ * https://spdx.org/licenses
+ */
+
+definition/* AP807 Marvell SoC driver */
```
soby-mathew: This doesn't look right

include/drivers/marvell/mochi/cp110_setup.h
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright (C) 2018 Marvell International Ltd.
+ *
+ * SPDX-License-Identifier:     BSD-3-Clause
+ * https://spdx.org/licenses
+ */
+
+/* CP110 Marvell SoC driver */
+
+#ifndef __PLAT_CP110_H__
```
soby-mathew: minor: The guard macros need to reflect the header file name

plat/marvell/a8k/a70x0_amc/plat_def.h
```diff
@@ -0,0 +1,31 @@
+/*
```
soby-mathew: Would prefer to rename this file to mvebu_def.h, because plat_def.h is too generic a name.

include/plat/marvell/a8k/common/plat_config.h
```diff
@@ -0,0 +1,128 @@
+/*
```
soby-mathew: would prefer to rename the file to a more specific name line "a8k_common.h" or something.

include/plat/marvell/a8k/common/plat_config.h
```diff
@@ -0,0 +1,128 @@
...
+ * Copyright (C) 2018 Marvell International Ltd.
+ *
+ * SPDX-License-Identifier:     BSD-3-Clause
+ * https://spdx.org/licenses
+ */
+
+#ifndef __BOARD_CONFIG_H__
```
soby-mathew: Please modify guard macros to reflect file name. Also fix other instances of this problem.

include/plat/marvell/a8k/common/plat_marvell.h
```diff
@@ -0,0 +1,109 @@
...
+
+/* BL31 utility functions */
+void marvell_bl31_early_platform_setup(bl31_params_t *from_bl2,
+				void *plat_params_from_bl2);
+void marvell_bl31_platform_setup(void);
+void marvell_bl31_plat_runtime_setup(void);
+void marvell_bl31_plat_arch_setup(void);
+
+/* Power management config to power off the SoC */
+void *plat_get_pm_cfg(void);
```
soby-mathew: Suggest a more `marvell` specfic name for this API. perhaps `plat_marvel_*` ?

include/plat/marvell/a8k/common/plat_marvell.h
```diff
@@ -0,0 +1,109 @@
...
+
+/* Check if MSS AP CM3 firmware contains PM support */
+_Bool is_pm_fw_running(void);
+
+/* Bootrom image recovery utility functions */
+void *plat_get_skip_image_data(void);
```
soby-mathew: same comment as above for this function name.

include/plat/marvell/a8k/common/plat_marvell.h
```diff
@@ -0,0 +1,109 @@
...
+
+/* FIP TOC validity check */
+int marvell_io_is_toc_valid(void);
+
+/*
+ * PSCI functionality
+ */
+void psci_arch_init(int ap_idx);
```
soby-mathew: I suppose this function is specific to marvell platform. Please use `marvel` specific prefix.

include/plat/marvell/a8k/common/plat_marvell.h
```diff
@@ -0,0 +1,109 @@
...
+void plat_marvell_io_setup(void);
+int plat_marvell_get_alt_image_source(
+	unsigned int image_id,
+	uintptr_t *dev_handle,
+	uintptr_t *image_spec);
+unsigned int plat_marvell_calc_core_pos(u_register_t mpidr);
+
+const mmap_region_t *plat_marvell_get_mmap(void);
+void ble_prepare_exit(void);
+void exit_bootrom(uintptr_t base);
```
soby-mathew: Prefer more platform specifc names for these APIs.

include/plat/marvell/common/plat_private.h
```diff
@@ -0,0 +1,32 @@
+/*
```
soby-mathew: Prefer more platform specifc name for header.

make_helpers/build_macros.mk
```diff
@@ -312,7 +313,7 @@ bl${1}_dirs: | ${OBJ_DIRS}
 $(eval $(call MAKE_OBJS,$(BUILD_DIR),$(SOURCES),$(1)))
 $(eval $(call MAKE_LD,$(LINKERFILE),$(BL_LINKERFILE),$(1)))
 
-$(ELF): $(OBJS) $(LINKERFILE) | bl$(1)_dirs
+$(ELF): $(OBJS) $(LINKERFILE) | bl$(1)_dirs $(BL_LIBS)
```
soby-mathew: Hmm, this seems to be colliding with the PR#1394.

plat/marvell/a8k/a70x0/board/dram_port.c
```diff
@@ -0,0 +1,89 @@
...
+#include <debug.h>
+#include <mv_ddr_if.h>
+#include <plat_marvell.h>
+
+/*
+ * This function may modify the default DRAM parameters
+ * based on information received from SPD or bootloader
+ * configuration located on non volatile storage
+ */
+void plat_dram_update_topology(void)
```
soby-mathew: If this function is exported, then prefer more platform specific name.

plat/marvell/a8k/common/aarch64/plat_arch_config.c
```diff
@@ -0,0 +1,46 @@
...
+#include <mmio.h>
+#include <debug.h>
+#include <cache_llc.h>
+
+
+#define CCU_HTC_ASET			(MVEBU_CCU_BASE(MVEBU_AP0) + 0x264)
+#define MVEBU_IO_AFFINITY		(0xF00)
+
+
+void plat_enable_affinity(void)
```
soby-mathew: Make this function static if not exported. Else prefer a more specific name.

make_helpers/build_macros.mk
antonio-nino-diaz-arm: It's not too invasive, I think. and that PR isn't ready anyway. This one is mostly ready.

robertovargas-arm: I was reviewing where these new BL*_LIBS macros are defined, but I couldn't find them. Is this patch preparing other patches?

kostapr: Actually we added this change for supporting library linkage on our own platform. It was added in the past when we started to implement Marvell platform on ATF-1.2. I was not aware of any other patches waiting for this change.

danh-arm: If you don't need this now, it would be better to remove this to avoid conflict with the other work Soby/Roberto refer to.

kostapr: We need it for sure. Without this patch the early boot stage cannot be built. This stage is using DDR initialization library, which is out of TF-A tree. If I remove this patch, Marvell platform support in TF-A will be useless.

danh-arm: OK, fair enough. We will rebase our other work on top of this.

kostapr: I forgot to mention that the actual usage of this library macros in in makefile of BLE:
https://github.com/MarvellEmbeddedProcessors/ble-marvell/blob/atf-mainline/ble.mk
We make additional  usage of library support in our LSP sources, but these changes are not yet mainlined.

danh-arm:
jenkins: test this please

danh-arm:
The only test failures were in the static checks. The copyright/license headers weren't as expected (which we've now agreed to) and the order of some includes is not alphabetical (which is really minor). Therefore merging...

antonio-nino-diaz-arm:
@kostapr I'm adding this to our CI build tests and I've noticed two things:
- The docs say that a3700 is supported but it isn't (or, at least, there is no build config for it).
- It's missing the links to the ble and DDR training lib repositories that you posted in your first comment in this PR.

kostapr:
@antonio-nino-diaz-arm
I am preparing platform support for A3700 and will create a pull request in August.
a3700_cust board was removed from platform folder to make the fist pull request less bulky. 
We were planning to put support for the official development board kit and to a community board (if available for this platform). I will clean-up the document from boards not added to the mainline sources.
As for links in the DDR init pointer, you are right, I will add it to the documentation along with A3700 support or maybe issue a separate fix.

PR #1451, opened 2018-06-25
PSCI locking simplification

glneo:
The first patch removes the need for locking in psci_cpu_on_finish(). The second patch correct the lock type that should be used for PSCI when the system has hardware cache-coherency. The last removes spinlocks from the PSCI implementation which may not work correctly on some platforms. This simplifies the PSCI locking scheme to one table of one type of lock.

Andrew

### DIFF
--- a/lib/psci/psci_on.c
+++ b/lib/psci/psci_on.c
@@ -46,6 +46,7 @@ int psci_cpu_on_start(u_register_t target_cpu,
 {
 	int rc;
 	unsigned int target_idx = plat_core_pos_by_mpidr(target_cpu);
+	cpu_pd_node_t *cpu_pd_node = &psci_cpu_pd_nodes[target_idx];
 	aff_info_state_t target_aff_state;
 
 	/* Calling function must supply valid input arguments */
@@ -60,7 +61,7 @@ int psci_cpu_on_start(u_register_t target_cpu,
 			psci_plat_pm_ops->pwr_domain_on_finish);
 
 	/* Protect against multiple CPUs trying to turn ON the same target CPU */
-	psci_spin_lock_cpu(target_idx);
+	psci_lock_get(cpu_pd_node);
 
 	/*
 	 * Generic management: Ensure that the cpu is off to be
@@ -138,7 +139,7 @@ int psci_cpu_on_start(u_register_t target_cpu,
 	}
 
 exit:
-	psci_spin_unlock_cpu(target_idx);
+	psci_lock_release(cpu_pd_node);
 	return rc;
 }
 
--- a/lib/psci/psci_on.c
+++ b/lib/psci/psci_on.c
@@ -115,6 +115,13 @@ int psci_cpu_on_start(u_register_t target_cpu,
 	}
 
 	/*
+	 * Store the re-entry information for the non-secure world here
+	 * before the core that needs it is started, this way we will not
+	 * need a lock to prevent a race to for this data.
+	 */
+	cm_init_context_by_index(target_idx, ep);
+
+	/*
 	 * Perform generic, architecture and platform specific handling.
 	 */
 	/*
@@ -124,11 +131,7 @@ int psci_cpu_on_start(u_register_t target_cpu,
 	 */
 	rc = psci_plat_pm_ops->pwr_domain_on(target_cpu);
 	assert(rc == PSCI_E_SUCCESS || rc == PSCI_E_INTERN_FAIL);
-
-	if (rc == PSCI_E_SUCCESS)
-		/* Store the re-entry information for the non-secure world. */
-		cm_init_context_by_index(target_idx, ep);
-	else {
+	if (rc != PSCI_E_SUCCESS) {
 		/* Restore the state on error. */
 		psci_set_aff_info_state_by_idx(target_idx, AFF_STATE_OFF);
 		flush_cpu_data_by_index(target_idx, psci_svc_cpu_data.aff_info_state);
@@ -168,15 +171,6 @@ void psci_cpu_on_finish(unsigned int cpu_idx,
 	 * to run in the non-secure address space.
 	 */
 	psci_arch_setup();
-
-	/*
-	 * Lock the CPU spin lock to make sure that the context initialization
-	 * is done. Since the lock is only used in this function to create
-	 * a synchronization point with cpu_on_start(), it can be released
-	 * immediately.
-	 */
-	psci_spin_lock_cpu(cpu_idx);
-	psci_spin_unlock_cpu(cpu_idx);
 
 	/* Ensure we have been explicitly woken up by another cpu */
 	assert(psci_get_aff_info_state() == AFF_STATE_ON_PENDING);
--- a/lib/psci/psci_private.h
+++ b/lib/psci/psci_private.h
@@ -27,18 +27,6 @@
 
 #define psci_dsbish()
 
-/*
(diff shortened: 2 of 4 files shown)


lib/psci/psci_on.c
```diff
@@ -114,6 +114,13 @@ int psci_cpu_on_start(u_register_t target_cpu,
 		assert(psci_get_aff_info_state_by_idx(target_idx) == AFF_STATE_ON_PENDING);
 	}
 
+	/*
+	 * Store the re-entry information for the non-secure world here
+	 * before the core that needs it is started, this way we will not
+	 * need a lock to prevent a race to for this data.
+	 */
+	cm_init_context_by_index(target_idx, ep);
+
```
danh-arm: But this de-optimises the code by doing more work before the new core is started, which is not what we want generally. I can see you're doing this to remove spinlock usage but that's a separate issue...

lib/psci/psci_private.h
```diff
@@ -27,18 +27,6 @@
...
- * instead of bakery locks.
- */
-#define DEFINE_PSCI_LOCK(_name)		spinlock_t _name
-#define DECLARE_PSCI_LOCK(_name)	extern DEFINE_PSCI_LOCK(_name)
-
-#define psci_lock_get(non_cpu_pd_node)				\
-	spin_lock(&psci_locks[(non_cpu_pd_node)->lock_index])
-#define psci_lock_release(non_cpu_pd_node)			\
-	spin_unlock(&psci_locks[(non_cpu_pd_node)->lock_index])
-
```
danh-arm: The HW_ASSISTED_COHERENCY flag is intended for use in Arm DynamIQ platforms (or similar), which are v8.2+. On these platforms, it is expected that the Compare and Swap (CAS) instruction is available and that this is used in the underlying spinlock implementation, by ensuring ARM_ARCH_MINOR is at least 1.

In any case, a great deal effort has gone into ensuring spinlocks will work correctly in the situations they are used in the codebase. It is highly undesirable to remove them in favour of much less efficient bakery locks, unless this is absolutely necessary. We need to understand the specific problem in your system (and try to fix that) rather than hitting performance in all systems.

lib/psci/psci_on.c
```diff
@@ -60,7 +61,7 @@ int psci_cpu_on_start(u_register_t target_cpu,
 			psci_plat_pm_ops->pwr_domain_on_finish);
 
 	/* Protect against multiple CPUs trying to turn ON the same target CPU */
-	psci_spin_lock_cpu(target_idx);
+	psci_lock_get(cpu_pd_node);
```
danh-arm: The commit message misquotes the Arm ARM regarding synchronization primitives. The Arm ARM specifically talks about the LoadEx/StoreEx mechanism when the MMU is disabled, which is not the case here.

This patch also (unnecessarily as fas I can tell) adds locking to CPU nodes in addition to non-CPU-nodes, thereby blowing some more performance we've worked hard to achieve.

lib/psci/psci_on.c
glneo: The comment states this lock is only used for local CPU coordination which makes spinlock safe. The use of being a sync point by another CPU was added later. The issue is spinlocks do not work in many cases across cores and clusters. Because of this all the non-cpu-pd locks are bakery locks (at least on non-HW-coherent platforms). So why then why is a spinlock suddenly okay here? My guess is it is not, I think most platforms get lucky and finish writing the context before the target CPU needs it so the lock is not used. If you for testing add a delay loop before context initialization then you will start seeing many other platforms start failing as the lock will fail and the target CPU will run past and read uninitialized data.

I am not the first to run into this: https://github.com/ARM-software/tf-issues/issues/451

Lets take the verifiable correct thing and just have the data ready for the target before releasing it, micro-optimizing the ordering is causing random failures for us and others.

lib/psci/psci_private.h
glneo: Then the flag is a misnomer, as we have HW assisted coherency but are not using v8.2+ cores. Perhaps we can break this into two flags, one that means what it says and deals with cache, and a new one that defines whether to use spinlocks vs bakery locks.

The comment above the code is also wrong, CPUs being cache-coherent does not guarantee we can safely use these spinlocks, it only removes one of many pitfalls described in the ARM ARM that will break them. So as above I think we need a new flag for using spinlocks, but for now this patch is correct in removing the lock type selection from under the scope of the unrelated flag HW_ASSISTED_COHERENCY.

lib/psci/psci_on.c
glneo: I did point out that the quote was for the non MMU case, not necessarily this instance.

This patch does not add locking to CPU nodes, they where locking before, just using spinlocks unconditionally. All the other PSCI locking was done with either spinlocks or bakery locks depending on whether HW_ASSISTED_COHERENCY is set or not. All this patch does is cause this CPU lock to do the same, it will still use spinlocks when HW_ASSISTED_COHERENCY is set. (although other patches in this set do change that behavior as I believe that is not the right flag for this)

lib/psci/psci_on.c
danh-arm: The reason bakery locks are used in some places (on non HW_ASSISTED_COHERENCY platforms) is that in those cases, the locks are released while caches are disabled and so it's not safe to use spinlocks. Here, the locks are acquired and released on both the running and target CPUs while caches are enabled, using a memory type that achitecturally guarantees that a global Exclusives monitor is implemented. Hence it is architecturally safe to use spinlocks.

However, it's possible that there is a bug, or that not all platforms implement the architecture correctly. If there's a bug, we should identify and fix it. If we need to help broken hardware, then we should add abstractions for that, rather than hitting performance for everyone.

> If you for testing add a delay loop before context initialization then you will start seeing many other platforms start failing as the lock will fail and the target CPU will run past and read uninitialized data.

Can you provide more information here? If there's a test case and a list of platforms for which this fails, that may help identify the root problem.

lib/psci/psci_private.h
danh-arm: In our definition of HW_ASSISTED_COHERENCY, data caches can be enabled very early on boot (i.e. without requiring CCI or other platform programming) and remain enabled right up to final power off. This makes it possible for us to comply with the rules around safe usage of exclusive access instructions, which in turn means it's architecturally safe to use spinlocks. I don't know if your definition of HW asssisted coherency is the same. If you think we are not complying with the Arm ARM in some way, then please state the specific issue. 

The option of using CAS instead of exclusive acccess instructions is just an optimisation; it should work either way.

lib/psci/psci_on.c
danh-arm: > I did point out that the quote was for the non MMU case, not necessarily this instance.

You pointed it out in a PR comment but you haven't updated the commit message, and that misquote is the rational for this change.

> This patch does not add locking to CPU nodes...

The PSCI locks are meant to be used for state coordination, for which locking at the CPU level is not required. This patch add locks at that level (although they won't be used during state coordination). The other spinlocks (like here) are used for a different purpose so should be kept separate from the PSCI locks. As I said in another comment, spinlocks are architecturally guaranteed to work here, which is not the case for the PSCI locks on non-HW_ASSISTED_COHERENCY platforms.

lib/psci/psci_on.c
glneo: Hmm, well let me go back and take another look at our locks and see if I can root cause the issues I'm seeing locally, then lets go from there. This PR can be closed for now, I'll re-submit a different fix depending on what I find.

PR #1452, opened 2018-06-25
Couple of fixups for K3 platform

glneo:
This set is a couple trivial fixups that were found during internal review.


antonio-nino-diaz-arm:
IP Review +1

PR #1454, opened 2018-06-26
Update runtime_svc.c

ljf1239848066:
Maybe there should use or "||" operator as the annotation says  in:
arm-trusted-firmware/common/runtime_svc.c line 83.

> /* A runtime service having no init or handle function doesn't make sense */


antonio-nino-diaz-arm:
IP Review +1

antonio-nino-diaz-arm:
Actually, it says that not having init and not having handle at the same time doesn't make sense. You just need to have one.

ljf1239848066:
That is to say a valid runtime service need at least one of init and handle.

PR #1457, opened 2018-06-27
maintainers: Modify format of file

antonio-nino-diaz-arm:
The previous format was not very practical and hard to parse by scripts. The current format is easier as it uses more tokens that can be identified by scripts, while maintaining a reasonably good result when it is rendered.

Some maintainers maintain more than one subsystem but they were all part of the same entry. In cases like this the entry has been split into two to clarify what file belongs to which subsystem.

The list of maintainers of the Trusted Firmware has been updated.

### DIFF
--- a/maintainers.rst
+++ b/maintainers.rst
@@ -7,190 +7,168 @@ parts of the codebase is delegated to the sub-maintainers listed below. An
 acknowledgement from these sub-maintainers may be required before the
 maintainers merge a contribution.
 
-Maintainers
------------
+Main maintainers
+----------------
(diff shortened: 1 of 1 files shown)


antonio-nino-diaz-arm:
Note that GitHub doesn't render rst files in the same way as other viewers such as http://rst.ninjs.org/

antonio-nino-diaz-arm:
@hzhuang1, @vwadekar, @jenswi-linaro, you maintain more than one subsystem so I've split your entries into two. I can put them back into one if you want, but I think it's cleaner this way.

jenswi-linaro:
Looks good to me.

PR #1447, opened 2018-06-22
allwinner: Add BL32 (corresponds to Trusted OS) support

Amit-Radur:
This pull request is an attempt to run Trusted OS (OP-TEE OS being one of them) along
side BL31 image.

This is the second version of PR  #1431, @Andre-ARM can you please have a look.

Trusted OS image can be fetched from below link:
https://github.com/Amit-Radur/Images.git

Signed-off-by: Amit Singh Tomar <amittomer25@gmail.com>

PR #1458, opened 2018-06-28
allwinner: various smaller fixes

Andre-ARM:
This is a collection of fixes, mostly addressing comments that @smaeul had on PR #1400.
The last patch adds code to output the SoC name, which proved to be very handy for error reports and for debugging. Also it allows to use SoC specific code paths, which we probably need in the future (especially with the upcoming H6 support).
I added @Amit-Radur 's BL32 patch from PR #1447 in this PR, to avoid merge conflicts, as I expect it to be merged sooner.


plat/allwinner/common/plat_helpers.S
```diff
@@ -9,29 +9,10 @@
 #include <sunxi_def.h>
 #include <sunxi_mmap.h>
 
-	.globl	plat_crash_console_init
-	.globl	plat_crash_console_putc
-	.globl	plat_crash_console_flush
 	.globl	plat_my_core_pos
 	.globl	platform_mem_init
 	.globl	plat_report_exception
```
antonio-nino-diaz-arm: If you remove this, you won't get crash messages anywhere, your multi console isn't setup to print crash messages. The default implementation doesn't do anything (check ``plat/common/aarch64/platform_helpers.S``).

plat/allwinner/common/sunxi_bl31_setup.c
```diff
@@ -64,6 +64,19 @@ void bl31_plat_arch_setup(void)
...
+	const char *soc_name = "unknown SoC";
+	uint16_t soc_id = sunxi_read_soc_id();
+
+	switch (soc_id) {
+	case 0x1689:
+		soc_name = "A64/H64/R18";
+		break;
+	case 0x1718:
+		soc_name = "H5";
+		break;
```
antonio-nino-diaz-arm: nit: you could maybe have a default clause here.

plat/allwinner/common/sunxi_common.c
```diff
@@ -54,3 +55,19 @@ void sunxi_configure_mmu_el3(int flags)
 
 	enable_mmu_el3(0);
 }
+
+#define SRAM_VER_REG (SUNXI_SYSCON_BASE + 0x24)
```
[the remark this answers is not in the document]
Andre-ARM: It was there (see 1f4b638), but looked weird there as well ;-) The thing is that this is really an offset into this "syscon" device, and thus device related, not memmap related.
The idea is that all device base addresses are in sunxi_mmap.h, in a per-target directory. This allows SoCs with a different memory map, but the same devices to blend in nicely (see 61eedb07963e). I made [one exception](../blob/master/plat/allwinner/sun50i_a64/include/sunxi_mmap.h#L39-L40), because the clock devices are different between SoCs, but for ATF's purposes it's not worth to model this completely. So this one line saves tons of code.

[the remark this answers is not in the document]
antonio-nino-diaz-arm: Ok, fair enough.

docs/plat/allwinner.rst
```diff
@@ -27,3 +29,13 @@ To build:
     make CROSS_COMPILE=aarch64-linux-gnu- PLAT=sun50i_a64 DEBUG=1 bl31
 
 .. _U-Boot documentation: http://git.denx.de/?p=u-boot.git;f=board/sunxi/README.sunxi64;hb=HEAD
+
+Trusted OS dispatcher
+=====================
+
+One can boot Trusted OS(OP-TEE OS, bl32 image) along side bl31 image on Allwinner A64.
```
antonio-nino-diaz-arm: nit: Image names in uppercase (BL31, BL32).

PR #1460, opened 2018-07-03
Make TF compatible with Clang assembler and linker

robertovargas-arm:
These changes are oriented to minimize the need of using GNU tools to build TF. After this change all the major tools involved (compiler, assembler and linker) can be substituted by the  equivalent Clang tools.

### DIFF
--- a/bl1/aarch64/bl1_exceptions.S
+++ b/bl1/aarch64/bl1_exceptions.S
@@ -26,25 +26,25 @@ vector_entry SynchronousExceptionSP0
 	mov	x0, #SYNC_EXCEPTION_SP_EL0
 	bl	plat_report_exception
 	no_ret	plat_panic_handler
-	check_vector_size SynchronousExceptionSP0
+end_vector_entry SynchronousExceptionSP0
 
 vector_entry IrqSP0
 	mov	x0, #IRQ_SP_EL0
 	bl	plat_report_exception
 	no_ret	plat_panic_handler
-	check_vector_size IrqSP0
+end_vector_entry IrqSP0
 
 vector_entry FiqSP0
 	mov	x0, #FIQ_SP_EL0
 	bl	plat_report_exception
 	no_ret	plat_panic_handler
-	check_vector_size FiqSP0
+end_vector_entry FiqSP0
 
 vector_entry SErrorSP0
 	mov	x0, #SERROR_SP_EL0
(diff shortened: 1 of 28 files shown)


antonio-nino-diaz-arm:
Fixes https://github.com/ARM-software/tf-issues/issues/554

bl1/aarch64/bl1_exceptions.S
```diff
@@ -26,25 +26,25 @@ vector_entry SynchronousExceptionSP0
 	mov	x0, #SYNC_EXCEPTION_SP_EL0
 	bl	plat_report_exception
 	no_ret	plat_panic_handler
-	check_vector_size SynchronousExceptionSP0
+end_vector_entry SynchronousExceptionSP0
```
danh-arm: minor: It would be good if the commit message explained why this new macro is needed (i.e. that the Clang linker cannot resolve the calculation in check_vector_size), Also that this macro now doesn't do anything and is deprecated.

docs/user-guide.rst
```diff
@@ -104,9 +104,12 @@ Building TF-A
        export CROSS_COMPILE=<path-to-aarch32-gcc>/bin/arm-linux-gnueabihf-
 
    It is possible to build TF-A using clang or Arm Compiler 6. To do so
-   ``CC`` needs to point to the clang or armclang binary. Only the compiler
-   is switched; the assembler and linker need to be provided by the GNU
-   toolchain, thus ``CROSS_COMPILE`` should be set as described above.
+   ``CC`` needs to point to the clang or armclang binary, but be aware
+   that the GNU linker is used by default. In case of being needed the
```
danh-arm: minor: It would be good to state that when CC is changed, the clang assembler (as well as the compiler) is used.

dp-arm:
I will be merging this PR on Wednesday the 11th.

dp-arm:
jenkins: test this please

paulmenzel:
Nice. Is Jenkins going to test with Clang from now on too?

robertovargas-arm:
The Clang tests that are currently executed are going to use the Clang assembler (as it is used by default now). We are going to keep using GNU linker instead of Clang linker because the version needed to link TF-A (version 6) creates some dependency problems in our machines. We expect to change this in the future once that Clang linker becomes stable enough to be included in our machines.

PR #1464, opened 2018-07-04
rpi3: Implement simple interrupt routing

antonio-nino-diaz-arm:
The Raspberry Pi 3 doesn't have secure peripherals, so all interrupts are treated as non-secure interrupts to be handled by the non-secure world.

### DIFF
--- a/docs/plat/rpi3.rst
+++ b/docs/plat/rpi3.rst
@@ -243,6 +243,12 @@ The following build options are supported:
   BL32_EXTRA1=tee-pager_v2.bin  BL32_EXTRA2=tee-pageable_v2.bin``
   to put the binaries into the FIP.
 
+  Note: If OP-TEE is used it may be needed to add the following options to the
+  Linux command line so that the USB driver doesn't use FIQs:
+  ``dwc_otg.fiq_enable=0 dwc_otg.fiq_fsm_enable=0 dwc_otg.nak_holdoff=0``.
+  This will unfortunately reduce the performance of the USB driver. It is needed
+  when using Raspbian, for example.
+
 - ``TRUSTED_BOARD_BOOT``: This port supports TBB. Set this option
   ``TRUSTED_BOARD_BOOT=1`` to enable it. In order to use TBB, you might
   want to set ``GENERATE_COT=1`` to let the contents of the FIP automatically
--- a/plat/rpi3/rpi3_common.c
+++ b/plat/rpi3/rpi3_common.c
@@ -5,6 +5,7 @@
  */
 
 #include <arch_helpers.h>
+#include <assert.h>
 #include <bl_common.h>
 #include <console.h>
 #include <debug.h>
@@ -198,15 +199,21 @@ unsigned int plat_get_syscnt_freq2(void)
 
 uint32_t plat_ic_get_pending_interrupt_type(void)
 {
+	ERROR("rpi3: Interrupt routed to EL3.\n");
 	return INTR_TYPE_INVAL;
 }
 
-uint32_t plat_interrupt_type_to_line(uint32_t type,
-				     uint32_t security_state)
+uint32_t plat_interrupt_type_to_line(uint32_t type, uint32_t security_state)
 {
-	/* It is not expected to receive an interrupt route to EL3.
-	 * Hence panic() to flag error.
-	 */
-	ERROR("Interrupt not expected to be routed to EL3");
-	panic();
+	assert((type == INTR_TYPE_S_EL1) || (type == INTR_TYPE_EL3) ||
+	       (type == INTR_TYPE_NS));
+
+	assert(sec_state_is_valid(security_state));
+
+	/* Non-secure interrupts are signalled on the IRQ line always. */
+	if (type == INTR_TYPE_NS)
+		return __builtin_ctz(SCR_IRQ_BIT);
+
+	/* Secure interrupts are signalled on the FIQ line always. */
+	return  __builtin_ctz(SCR_FIQ_BIT);
 }


antonio-nino-diaz-arm:
Can you tell me if this works for you, @grandpaul?

grandpaul:
Tested-by: Ying-Chun Liu (PaulLiu) <paul.liu@linaro.org>

Hi. It works for me. 
Please read the boot log I have currently.
https://paste.ubuntu.com/p/MsRXY7YVbd/

antonio-nino-diaz-arm:
The boot log looks promising! I guess that now you are trying to load the kernel from the SD card, right?

grandpaul:
Hi Antonio,

I think kernel dislike "return  __builtin_ctz(SCR_FIQ_BIT);". I need to change it to "return 0" to get kernel working.

https://paste.ubuntu.com/p/XDhVFjdfNJ/

antonio-nino-diaz-arm:
Hmmm does it work with ``SCR_IRQ_BIT``? This problem has to do with the setup that opteed does, I've managed to boot the kernel just fine with this (without optee).

grandpaul:
Yes, Without SPD=opteed, it works. 
But with SPD=opteed, SCR_IRQ_BIT doesn't work either. But "return 0" works.
When kernel doesn't work, it failed to boot at very early stage. No outputs at all even with earlyprintk.

antonio-nino-diaz-arm:
If I am not mistaken, what your "return 0" is doing is not setting up secure interrupts when opteed tries to do it. I don't know much about optee, is it using the secure timer? If so, your return 0 will simply ignore it, while my ``SCR_FIQ_BIT`` will treat it as a FIQ and that's probably causing the problems. If that is the case, ``plat_ic_get_pending_interrupt_type`` will have to be changed to detect if the interrupt has been caused by the secure timer or anything else. If optee is using the secure timer, you could try to disable it to see if the problems disappear.

Also, could you try with my patch but replacing ``plat_ic_get_pending_interrupt_type()`` by this?
```
uint32_t plat_ic_get_pending_interrupt_type(void)
{
if (get_cntp_ctl_istatus(read_cntps_ctl_el1())) != 0)
    return INTR_TYPE_S_EL1;
return INTR_TYPE_NS;
}
```

grandpaul:
Wow. That's great. It goes further. But still hangs. Now I have 
https://paste.ubuntu.com/p/DXMdqQWFhf/
Hangs after USB inits.

antonio-nino-diaz-arm:
After a bit of googling, I've confirmed what I was fearing. Linux is using FIQs. For the USB, to be precise: https://raspberrypi.stackexchange.com/questions/8661/interrupts-on-the-raspberry-pi

When Linux tries to register the port an FIQ is triggered, and the TF routes it to the secure world. Linux never sees it so it hangs while waiting for it.

Unfortunately, I can't see if there is a way to detect external FIQs vs internal FIQs in the interrupt controller that the Raspberry Pi 3 has. [1][2] The one of the RPi 1 had did have some status bit, I believe. Note that the documentation says that it's for the RPi 2, but 3 and 2 have the same one.

There are 3 ways to solve this:
- Modify the kernel so that it doesn't use FIQs. I'm curious about this option, because maybe 64-bit kernels don't use FIQs for the USB and they can run with this PR.
- See if optee can work without secure timer. Not sure if this is possible.
- Somehow detect if the FIQ is caused by an external peripheral or by the secure timer. So far I haven't been able to see how to do it, but I'm still looking at the interrupt controller Linux driver. It might be possible.

[1] https://www.raspberrypi.org/documentation/hardware/raspberrypi/bcm2836/README.md
[2] https://www.raspberrypi.org/documentation/hardware/raspberrypi/bcm2836/QA7_rev3.4.pdf

antonio-nino-diaz-arm:
Could you try this, @grandpaul?

https://xenomai.org/pipermail/xenomai/2017-April/037281.html

> I think that I achieved to solve the problem by disabling the FIQ
> capabilities of the USB driver. There are problems by using the FIQ USB
> driver with RT patches as shown in:
> https://community.emlid.com/t/rpi-2-fiq-usb-driver-problems/508/8
> https://www.raspberrypi.org/forums/viewtopic.php?f=29&t=39951&start=75

> The solution is as simple as adding the following parameters in the
> "cmdline.txt" file in the boot partition:

``dwc_otg.fiq_enable=0 dwc_otg.fiq_fsm_enable=0 dwc_otg.nak_holdoff=0``

igoropaniuk:
@antonio-nino-diaz-arm 

> After a bit of googling, I've confirmed what I was fearing. Linux is using FIQs. For the USB, to be precise: https://raspberrypi.stackexchange.com/questions/8661/interrupts-on-the-raspberry-pi
When Linux tries to register the port an FIQ is triggered and the TF routes it to the secure world. Linux never sees it so it hangs while waiting for it.
Unfortunately, I can't see if there is a way to detect external FIQs vs internal FIQs in the interrupt controller that the Raspberry Pi 3 has. [1][2] The one of the RPi 1 had did have some status bit, I believe. Note that the documentation says that it's for the RPi 2, but 3 and 2 have the same one.

A little bit off topic. Seems that I've just found an explanation for the strange behavior of DWC2 (hub enumeration issues and strange hangs), which I faced a year ago when I used ATF (our own fork, it also doesn't have a proper implementation of interrupt handling).

igoropaniuk:
@antonio-nino-diaz-arm 
https://github.com/raspberrypi/linux/issues/1779#issuecomment-270635808

grandpaul:
I've confirmed that adding "dwc_otg.fiq_enable=0 dwc_otg.fiq_fsm_enable=0 dwc_otg.nak_holdoff=0" to bootargs fixes the problem. Thanks.

antonio-nino-diaz-arm:
Good to know that we are not the first ones to find problems with this, @igoropaniuk.

Thanks for testing, @grandpaul. Can you point me at the instructions to build optee for the rpi3? I'd like to test this a bit before it is merged, I'm a bit worried about what the options may be doing. Specially now that @igoropaniuk has posted the link to that issue.

I don't like that they are using FIQ for the USB, but well, given that the official Raspbian is 32-bit, they can actually make the USB driver a bit faster by using FIQs (the context switches are faster and they don't need to identify the cause of the interrupt because there is only one trigger of FIQ).

igoropaniuk:
@antonio-nino-diaz-arm 
We do have a prepared repo manifest for bringing up full setup in a couple of steps, I've also added additional changes to be able to build with upstream ATF+U-boot

```bash
$ repo init -u https://github.com/OP-TEE/manifest.git -m rpi3.xml
$ repo sync -j20
```
Add upstream remotes to arm-trusted-firmware and u-boot (currently, we're using our own forks for both ATF and U-boot).

Apply my patches from https://github.com/OP-TEE/build/pull/269/ in build rep.

antonio-nino-diaz-arm:
Thanks, I'll try it as soon as I get a chance.

grandpaul:
Here is the xtest log currently we have. (Run on our distro).
https://paste.ubuntu.com/p/3Vmtg6J3f8/
Some test failed. But mostly passed.

antonio-nino-diaz-arm:
As far as I see, all the tests that fail complain about not having enough memory. Maybe instead of reducing the size allocated to optee in https://github.com/OP-TEE/optee_os/pull/2394 you should increase the size here. I haven't checked all tests, of course... Some of them may fail for other reasons.

vchong:
Try increasing the heap size to see if it helps resolve the errors. Re: https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/plat-poplar/conf.mk#L26.

grandpaul:
I've modified in ATF:

> --- a/plat/rpi3/include/platform_def.h
> +++ b/plat/rpi3/include/platform_def.h
> @@ -75,10 +75,10 @@
>  #define SEC_SRAM_SIZE                  ULL(0x00100000)
>  
>  #define SEC_DRAM0_BASE                 ULL(0x10100000)
> -#define SEC_DRAM0_SIZE                 ULL(0x00200000)
> +#define SEC_DRAM0_SIZE                 ULL(0x00700000)
>  
> -#define NS_DRAM0_BASE                  ULL(0x10300000)
> -#define NS_DRAM0_SIZE                  ULL(0x00D00000)
> +#define NS_DRAM0_BASE                  ULL(0x10800000)
> +#define NS_DRAM0_SIZE                  ULL(0x00800000)
> 

Which means now we have 7MB SEC_DRAM0.

And in OPTEE:

> --- a/core/arch/arm/plat-rpi3/platform_config.h
> +++ b/core/arch/arm/plat-rpi3/platform_config.h
> @@ -70,12 +70,12 @@
>  #define TEE_SHMEM_START                (0x08000000)
>  #define TEE_SHMEM_SIZE         (4 * 1024 * 1024)
>  
> -#define TZDRAM_BASE            (TEE_SHMEM_START + TEE_SHMEM_SIZE)
> -#define TZDRAM_SIZE            (32 * 1024 * 1024)
> +#define TZDRAM_BASE            (0x10100000)
> +#define TZDRAM_SIZE            (7 * 1024 * 1024)
>  
> -#define TEE_RAM_VA_SIZE                (4 * 1024 * 1024)
> +#define TEE_RAM_VA_SIZE                (3 * 1024 * 1024)
>  
> -#define TEE_LOAD_ADDR          (TZDRAM_BASE + 0x20000)
> +#define TEE_LOAD_ADDR          TZDRAM_BASE
>  
>  #define TEE_RAM_PH_SIZE                TEE_RAM_VA_SIZE
>  #define TEE_RAM_START          TZDRAM_BASE
> @@ -83,6 +83,6 @@
>  #define TA_RAM_START           ROUNDUP((TZDRAM_BASE + TEE_RAM_VA_SIZE), \
>                                         CORE_MMU_DEVICE_SIZE)
>  
> -# define TA_RAM_SIZE           (16 * 1024 * 1024)
> +# define TA_RAM_SIZE           (4 * 1024 * 1024)
> 

Which means 3 MB for TEE, and 4 MB to TA.
And the fail become much more less.
Please see the log  http://paste.ubuntu.com/p/GQvdRDybQ5/

With or without CFG_CORE_HEAP_SIZE?=98304 doesn't affect the result.
I'd like to know what is the minimum memory required to pass "regression_1013.2 Using large concurrency TA" ??

grandpaul:
OK. I've tried several time and the following works.

> --- a/plat/rpi3/include/platform_def.h
> +++ b/plat/rpi3/include/platform_def.h
> @@ -75,10 +75,10 @@
>  #define SEC_SRAM_SIZE                  ULL(0x00100000)
>  
>  #define SEC_DRAM0_BASE                 ULL(0x10100000)
> -#define SEC_DRAM0_SIZE                 ULL(0x00200000)
> +#define SEC_DRAM0_SIZE                 ULL(0x00B00000)
>  
> -#define NS_DRAM0_BASE                  ULL(0x10300000)
> -#define NS_DRAM0_SIZE                  ULL(0x00D00000)
> +#define NS_DRAM0_BASE                  ULL(0x10C00000)
> +#define NS_DRAM0_SIZE                  ULL(0x00400000)
> 

With optee settings for TEE 3MB, TA 8MB. The test passed without error. 
I'll prepare a PR soon.

antonio-nino-diaz-arm:
@grandpaul ok, perfect.

I'm testing this patch and the command line options and everything seems to work just fine. However, there is a performance hit, the whole point of moving the USB driver to use FIQs was to improve performance. I haven't done any test to see how noticeable it is, though.

In any case, there is no point in worrying about performance if it doesn't work, so it's better to have this than nothing.

antonio-nino-diaz-arm:
I've just tested TF + OP-TEE + AArch32 bootstrap + Raspbian (I didn't feel like waiting for Linux to be built...). I've done it without the memory map changes proposed above.

Everything seems to be working fine (well, it boots and the console output looks sane, I haven't run any test). This patch is needed, as well as the options to disable FIQs in Linux. I've tried a few things but the Linux command line options seem to be really needed.

antonio-nino-diaz-arm:
@grandpaul, I assume you are still ok with the changes, right? If so, we can merge it.

grandpaul:
Yes. I've test this patch already. It is necessary for optee to run. and optee_test (xtest) to pass.
Without this PR it is impossible.

antonio-nino-diaz-arm:
Ok, I've decided to drop the secure timer code. OP-TEE doesn't seem to use the secure timer to generate interrupts. It's better to leave my previous check out because it's code that we don't know if it is correct.

grandpaul:
Still good to me. I've test everything again. Pass the tests.  Thanks.
http://paste.ubuntu.com/p/w6ZDg2CDqR/

dp-arm:
jenkins: test this please

PR #1465, opened 2018-07-04
allwinner: Add H6 SoC support

Andre-ARM:
This adds support for this relatively new SoC. The actual devices are mostly the same, but the memory map has been shuffled, so we have separate header files defining the same symbols.
platform.mk itself is identical, but I use a separate copy. We might see changes in the future (different PMIC), also I couldn't be bothered to refactor this. Maybe a third SoC will trigger this.
The new platform name is "sun50i_h6".
@Icenowy, could you please have a brief look at this? I fixed the issues you raised on the original patch.


Andre-ARM:
@Icenowy, many thanks!

PR #1468, opened 2018-07-06
tools/cert_create: fix makefile to build build_msg.o by HOSTCC

grandpaul:
Previously build_msg.o is built by CC. It causes FTBFS when CC
is not equal to HOSTCC.

Signed-off-by: Ying-Chun Liu (PaulLiu) <paulliu@debian.org>


dp-arm:
IP review +1

dp-arm:
jenkins: test this please

PR #1472, opened 2018-07-11
Rework page table setup for varying number of mem regions

danielboulby-arm:
Change arm_setup_page_tables() to take a variable number of memory
regions

Change-Id: I790054e3b20b056dda1043a4a67bd7ac2d6a3bc0

### DIFF
--- a/include/plat/arm/css/common/css_def.h
+++ b/include/plat/arm/css/common/css_def.h
@@ -101,6 +101,14 @@
 						NSRAM_SIZE,	\
 						MT_DEVICE | MT_RW | MT_NS)
 
+#if defined(IMAGE_BL2U)
+#define CSS_MAP_SCP_BL2U		MAP_REGION_FLAT(		\
+						SCP_BL2U_BASE,		\
+						SCP_BL2U_LIMIT		\
+							- SCP_BL2U_BASE,\
+						MT_RW_DATA | MT_SECURE)
+#endif
+
 /* Platform ID address */
 #define SSC_VERSION_OFFSET			0x040
 
--- a/plat/arm/common/arm_bl1_setup.c
+++ b/plat/arm/common/arm_bl1_setup.c
@@ -7,6 +7,7 @@
 #include <arch.h>
 #include <arm_def.h>
 #include <arm_xlat_tables.h>
+#include <assert.h>
 #include <bl1.h>
 #include <bl_common.h>
 #include <plat_arm.h>
@@ -23,6 +24,19 @@
 #pragma weak bl1_plat_sec_mem_layout
 #pragma weak bl1_plat_prepare_exit
 
+#define MAP_BL1_TOTAL		MAP_REGION_FLAT(			\
+					bl1_tzram_layout.total_base,	\
+					bl1_tzram_layout.total_size,	\
+					MT_MEMORY | MT_RW | MT_SECURE)
+#define MAP_BL1_CODE		MAP_REGION_FLAT(			\
+					BL_CODE_BASE,			\
+					BL1_CODE_END - BL_CODE_BASE,	\
+					MT_CODE | MT_SECURE)
+#define MAP_BL1_RO_DATA		MAP_REGION_FLAT(			\
+					BL1_RO_DATA_BASE,		\
+					BL1_RO_DATA_END			\
+						- BL_RO_DATA_BASE,	\
+					MT_RO_DATA | MT_SECURE)
 
 /* Data structure which holds the extents of the trusted SRAM for BL1*/
 static meminfo_t bl1_tzram_layout;
@@ -84,17 +98,19 @@ void bl1_early_platform_setup(void)
  *****************************************************************************/
 void arm_bl1_plat_arch_setup(void)
 {
-	arm_setup_page_tables(bl1_tzram_layout.total_base,
-			      bl1_tzram_layout.total_size,
-			      BL_CODE_BASE,
-			      BL1_CODE_END,
-			      BL1_RO_DATA_BASE,
-			      BL1_RO_DATA_END
 #if USE_COHERENT_MEM
-			      , BL_COHERENT_RAM_BASE,
-			      BL_COHERENT_RAM_END
+	/* ARM platforms dont use coherent memory in BL1 */
+	assert((BL_COHERENT_RAM_END - BL_COHERENT_RAM_BASE) == 0U);
 #endif
-			     );
+
+	const mmap_region_t bl_regions[] = {
+		MAP_BL1_TOTAL,
+		MAP_BL1_CODE,
+		MAP_BL1_RO_DATA,
+		{0}
+	};
+
+	arm_setup_page_tables(bl_regions, plat_arm_get_mmap());
 #ifdef AARCH32
 	enable_mmu_secure(0);
 #else
--- a/plat/arm/common/arm_bl2_el3_setup.c
+++ b/plat/arm/common/arm_bl2_el3_setup.c
@@ -3,6 +3,8 @@
  *
  * SPDX-License-Identifier: BSD-3-Clause
  */
+#include <arm_def.h>
+#include <assert.h>
 #include <generic_delay_timer.h>
 #include <plat_arm.h>
 #include <platform.h>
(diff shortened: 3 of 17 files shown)


plat/arm/common/arm_common.c
```diff
@@ -28,61 +28,34 @@ extern const mmap_region_t plat_arm_mmap[];
...
+				(void *) (regions->base_va + regions->size),
+				regions->attr);
+		regions++;
+	}
+#endif
 	/*
 	 * Map the Trusted SRAM with appropriate memory attributes.
 	 * Subsequent mappings will adjust the attributes for specific regions.
 	 */
-	VERBOSE("Trusted SRAM seen by this BL image: %p - %p\n",
```
glneo: We lose these printouts which can be useful for debug. Dropping these really seems to be the only functional thing this patch does.

plat/ti/k3/common/k3_bl31_setup.c
```diff
@@ -99,12 +99,28 @@ void bl31_early_platform_setup2(u_register_t arg0, u_register_t arg1,
 
 void bl31_plat_arch_setup(void)
 {
-	arm_setup_page_tables(BL31_BASE,
-			      BL31_END - BL31_BASE,
-			      BL_CODE_BASE,
-			      BL_CODE_END,
-			      BL_RO_DATA_BASE,
-			      BL_RO_DATA_END);
+#define K3_MAP_BL31_TOTAL	MAP_REGION_FLAT(			\
```
glneo: This does not look good to me, why make definitions just to then expand them only once in the very next line? Just add them to the array directly, or since the second parameter already takes a variable number of mappings you can add these to that array and drop this first parameter.

plat/arm/common/arm_common.c
danielboulby-arm: In a couple of the other tasks we are working on we have found the need to be able to map a varying number of memory regions. The old implementation of the function would require variable argument list to the function which we decided was not a clean way of solving the problem. This does however mean we lose some specific information such as the name of the region however I tried to reduce the impact of this by printing out the attributes which can be used to work out the regions being mapped.

plat/ti/k3/common/k3_bl31_setup.c
danielboulby-arm: I think this is just a matter of preference. The macros were defined above to make it clear what the regions are. In the arm platform code there are some common regions as well that are defined in arm_defs.h so the macros are only defined here for the special cases for that image. If you would prefer to remove the macro definitions and add them directly to the array for your platform I am happy to rework this for you. 

These mappings cannot be added to the second array as they have different types. The array in the second parameter can be generated statically however since the first array requires values only resolved during linking then it must be dynamically allocated.

plat/arm/common/arm_common.c
glneo: What tasks, variable number of memory regions are supported by the second parameter.

danielboulby-arm: One of the tasks I'm working on is to reclaim initialisation code in BL31 for secondary CPU stack usage. To this I need to move the init code section to below the data section. We need to map a region for this section however this cannot be added to the array in second parameter as the location of the section is only known after linking.

robertovargas-arm: I am also working in a task to share part of the libraries between different boot loaders. It means that 2 additional regions can be needed.

plat/ti/k3/common/k3_bl31_setup.c
glneo: IMHO good coding style is more than just personal preference. Anyway, yes, it would be nice to have a simple array like we do in [ti/k3/common/k3_bl31_setup.c](https://github.com/ARM-software/arm-trusted-firmware/blob/master/plat/ti/k3/common/k3_bl31_setup.c#L19). Then just use comments to make it clear what the regions are.

include/plat/arm/css/common/css_def.h
```diff
@@ -101,6 +101,14 @@
 						NSRAM_SIZE,	\
 						MT_DEVICE | MT_RW | MT_NS)
 
+#if defined(IMAGE_BL2U)
+#define CSS_MAP_SCP_BL2U		MAP_REGION_FLAT(		\
+						SCP_BL2U_BASE,		\
+						SCP_BL2U_LIMIT		\
+							- SCP_BL2U_BASE,\
+						MT_CODE | MT_SECURE)
```
soby-mathew: This region is used to load and authenticate SCP_BL2U. Hence the attributes of this region should be RW rather Executable.

danielboulby-arm: I have changed this and checked that firmware update still works as expected which it does

danielboulby-arm:
The CI only fails due to coverity violations in some of the macros used which will be addressed with all other cases of the rule later

plat/arm/common/arm_bl1_setup.c
```diff
@@ -84,17 +85,32 @@ void bl1_early_platform_setup(void)
...
+#define ARM_MAP_BL1_RO_DATA	MAP_REGION_FLAT(			\
+					BL1_RO_DATA_BASE,		\
+					BL1_RO_DATA_END			\
+						- BL_RO_DATA_BASE,	\
+					MT_RO_DATA | MT_SECURE)
 #if USE_COHERENT_MEM
-			      , BL_COHERENT_RAM_BASE,
-			      BL_COHERENT_RAM_END
+	/* ARM platforms dont use coherent memory in BL1 */
+	assert((BL_COHERENT_RAM_END - BL_COHERENT_RAM_BASE) == 0U);
```
danh-arm: I don't think this assert is ever going to be executed. It's only used in the definition of the bl_regions array below and will be ignored by the compiler. I think you should pull this out of the macro definition and use CASSERT instead.

plat/arm/common/arm_bl2_el3_setup.c
```diff
@@ -60,17 +62,24 @@ void bl2_el3_early_platform_setup(u_register_t arg0 __unused,
...
+#define ARM_MAP_BL2_EL3_TOTAL	MAP_REGION_FLAT(				\
+					bl2_el3_tzram_layout.total_base,	\
+					bl2_el3_tzram_layout.total_size,	\
+					MT_MEMORY | MT_RW | MT_SECURE)
+
 #if USE_COHERENT_MEM
-			      , BL_COHERENT_RAM_BASE,
-			      BL_COHERENT_RAM_END
+	/* ARM platforms dont use coherent memory in BL2_AT_EL3 */
+	assert(BL_COHERENT_RAM_END - BL_COHERENT_RAM_BASE == 0U);
```
danh-arm: assert has no effect

plat/arm/common/arm_bl2_setup.c
```diff
@@ -231,17 +231,24 @@ void bl2_platform_setup(void)
...
+#define ARM_MAP_BL2_TOTAL	MAP_REGION_FLAT(			\
+					bl2_tzram_layout.total_base,	\
+					bl2_tzram_layout.total_size,	\
+					MT_MEMORY | MT_RW | MT_SECURE)
+
 #if USE_COHERENT_MEM
-			      , BL_COHERENT_RAM_BASE,
-			      BL_COHERENT_RAM_END
+	/* ARM platforms dont use coherent memory in BL2 */
+	assert((BL_COHERENT_RAM_END - BL_COHERENT_RAM_BASE) == 0U);
```
danh-arm: assert has no effect

plat/arm/common/arm_bl2u_setup.c
```diff
@@ -58,18 +59,25 @@ void bl2u_early_platform_setup(meminfo_t *mem_layout, void *plat_info)
...
+					BL2U_BASE,			\
+					BL2U_LIMIT - BL2U_BASE,		\
+					MT_MEMORY | MT_RW | MT_SECURE)
+
 #if USE_COHERENT_MEM
-			      ,
-			      BL_COHERENT_RAM_BASE,
-			      BL_COHERENT_RAM_END
+	/* ARM platforms dont use coherent memory in BL2U */
+	assert((BL_COHERENT_RAM_END - BL_COHERENT_RAM_BASE) == 0U);
```
danh-arm: assert has no effect

plat/arm/common/arm_bl1_setup.c
danielboulby-arm: I have moved the macros to the top of the file and removed the ARM prefix. I didn't move the definitions into the bl_regions array as for some images we use both regions common to many images (defined in arm_defs.h) and regions specific to the current image, therefore I thought it was clearer to define the macros at the top of the file

danh-arm:
The only build issues are advisory Coverity rules that we will tackle later.

PR #1479, opened 2018-07-12
Fix arch.h headers

antonio-nino-diaz-arm:
There are several issues with the headers as they are right now. This pull request fix them to try to prevent subtle programming errors.

### DIFF
--- a/include/lib/aarch64/arch.h
+++ b/include/lib/aarch64/arch.h
@@ -647,7 +647,13 @@
 #define PAR_F_SHIFT	0
 #define PAR_F_MASK	1
 #define PAR_ADDR_SHIFT	12
-#define PAR_ADDR_MASK	(BIT(40) - 1) /* 40-bits-wide page address */
+/*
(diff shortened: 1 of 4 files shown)


include/lib/aarch64/arch.h
```diff
@@ -647,7 +647,7 @@
 #define PAR_F_SHIFT	0
 #define PAR_F_MASK	1
 #define PAR_ADDR_SHIFT	12
-#define PAR_ADDR_MASK	(BIT(40) - 1) /* 40-bits-wide page address */
+#define PAR_ADDR_MASK	(BIT(52) - ULL(1))
```
soby-mathew: nit: Would be good to make comment similar to the commit message here.

antonio-nino-diaz-arm: Done.

antonio-nino-diaz-arm:
Superseded by https://github.com/ARM-software/arm-trusted-firmware/pull/1482

PR #1480, opened 2018-07-13
Always compile debug macros

sandrine-bailleux-arm:
These patches modify the debug macros (INFO, NOTICE, ...) to get the compiler to evaluate them all the time. They also fix the build issues that this uncovered.


soby-mathew:
@vwadekar, Could you please review the change ?

soby-mathew:
@vwadekar, thanks for the review.

PR #1483, opened 2018-07-16
rpi3: PSCI and Linux boot improvements

antonio-nino-diaz-arm:
- ``RESET_TO_BL31`` was broken. It doesn't make sense to keep it as this port is using BL1 and BL2 as well, so it has been removed.
- Added support for the hardware RNG. It is used for the stack protector.
- ``PSCI_CPU_ON`` was broken, but it somehow worked in some cases. It has been fixed and tested in an AArch64 Linux kernel.
- Improve support for direct boot of Linux kernel:
  - Switched dtb and kernel addresses to leave more room for growth for the kernel (it only had 16 MB, and that size is easy to overflow).
  - Add ``RPI3_DIRECT_LINUX_BOOT`` to support direct-to-kernel boot.
- Documentation updated to reflect the changes.

### DIFF
--- a/docs/plat/rpi3.rst
+++ b/docs/plat/rpi3.rst
@@ -216,12 +216,6 @@ The following build options are supported:
   be loaded anywhere by modifying the file ``config.txt``. It doesn't have to
   contain a kernel, it could have any arbitrary payload.
 
-- ``RESET_TO_BL31``: Set to 1 by default. If using a 32-bit kernel like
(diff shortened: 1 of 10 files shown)


soby-mathew:
IP-Review +1

antonio-nino-diaz-arm:
Oh, @grandpaul, I forgot to ping you. There are some changes to the memory map to free up more space for BL32.

PR #1489, opened 2018-07-23
doimage: get rid of non null terminated strings by strncpy

teknoraver:
Signed-off-by: Matteo Croce <mcroce@redhat.com>

### DIFF
--- a/tools/doimage/doimage.c
+++ b/tools/doimage/doimage.c
@@ -216,7 +216,7 @@ void usage(void)
 }
 
 /* globals */
-options_t opts = {
+static options_t opts = {
 	.bin_ext_file = "NA",
 	.sec_cfg_file = "NA",
 	.sec_opts = 0,
@@ -1578,9 +1578,9 @@ error:
 
 int main(int argc, char *argv[])
 {
-	char in_file[MAX_FILENAME+1];
-	char out_file[MAX_FILENAME+1];
-	char ext_file[MAX_FILENAME+1];
+	char in_file[MAX_FILENAME+1] = { 0 };
+	char out_file[MAX_FILENAME+1] = { 0 };
+	char ext_file[MAX_FILENAME+1] = { 0 };
 	FILE *in_fd = NULL;
 	FILE *out_fd = NULL;
 	int parse = 0;
@@ -1590,6 +1590,7 @@ int main(int argc, char *argv[])
 	int image_size;
 	uint8_t *image_buf = NULL;
 	int read;
+	size_t len;
 	uint32_t nand_block_size_kb, mlc_nand;
 
 	/* Create temporary file for building extensions
@@ -1660,13 +1661,19 @@ int main(int argc, char *argv[])
 	if (optind >= argc)
 		usage_err("missing input file name");
 
-	strncpy(in_file, argv[optind], MAX_FILENAME);
+	len = strlen(argv[optind]);
+	if (len > MAX_FILENAME)
+		usage_err("file name too long");
+	memcpy(in_file, argv[optind], len);
 	optind++;
 
 	/* Output file must exist in non parse mode */
-	if (optind < argc)
-		strncpy(out_file, argv[optind], MAX_FILENAME);
-	else if (!parse)
+	if (optind < argc) {
+		len = strlen(argv[optind]);
+		if (len > MAX_FILENAME)
+			usage_err("file name too long");
+		memcpy(out_file, argv[optind], len);
+	} else if (!parse)
 		usage_err("missing output file name");
 
 	/* open the input file */


antonio-nino-diaz-arm:
Maybe we should be using [strlcpy](https://www.freebsd.org/cgi/man.cgi?query=strlcpy&sektion=3) instead of strncpy. Or just snprintf. spnprintf is safe because it includes the terminator: https://linux.die.net/man/3/snprintf

IP Review +1

teknoraver:
strlcpy is an option but must be copied because is not in libc.
snprintf is a bit overkill just for argument copying

antonio-nino-diaz-arm:
Do you mean it's not in the TF's libc? If so, that's not an issue, the tools don't use what we have in the lib folder, they are compiled using the system's libraries. However, it's not standard, I believe. That's why I think that snprintf is a better idea.

Initializing the strings to 0 when they are allocated solves the problem in the wrong place. It is not clear why it is needed to do it, and someone could think that they aren't needed and remove them at some point in the future.

strncpy is quite problematic anyway because of not adding the terminator if the string overflows, so we shouldn't use it.

teknoraver:
No I'm referring to host libraries, eg. GNU libc. They don't have strlcpy, I think it's BSD only.
In the linux kernel [strscpy](https://lxr.missinglinkelectronics.com/linux/lib/string.c#L154) was written to circumvent some strlcpy limitations:
```
Preferred to strlcpy() since the API doesn't require reading memory
from the src string beyond the specified "count" bytes, and since
the return value is easier to error-check than strlcpy()'s.
In addition, the implementation is robust to the string changing out
from underneath it, unlike the current strlcpy() implementation.

Preferred to strncpy() since it always returns a valid string, and
doesn't unnecessarily force the tail of the destination buffer to be
zeroed.  If the zeroing is desired, it's likely cleaner to use strscpy()
with an overflow test, then just memset() the tail of the dest buffer.
```
If we have to embed strlcpy, strscpy could be a better option (with kernel specific code removed)

Ideas?

antonio-nino-diaz-arm:
Oh I see. Then I think that using sprintf is the best option, as we don't need to add more external code.

This is the only file that uses strncpy, and only four times, so it's doesn't require a lot of effort to do it.

tools/doimage/doimage.c
```diff
@@ -1578,9 +1578,9 @@ int write_boot_image(uint8_t *buf, uint32_t image_size, FILE *out_fd)
 
 int main(int argc, char *argv[])
 {
-	char in_file[MAX_FILENAME+1];
-	char out_file[MAX_FILENAME+1];
-	char ext_file[MAX_FILENAME+1];
+	char in_file[MAX_FILENAME+1] = { 0 };
+	char out_file[MAX_FILENAME+1] = { 0 };
+	char ext_file[MAX_FILENAME+1] = { 0 };
```
robertovargas-arm: I don't think this is a correct change. 

This code is using snprintf in the line 1598, and the only problem that I could say is that it doesn't check the value returned by snprintf, but as MAX_FILENAME is defined to be 255, then it is impossible to have a truncation (even in the case getpid returns a 64 bit value). I wouldn't see a bad thing to add that test even in the case that we know it is impossible to have a truncation.

It  uses strncpy in the line 1618, 1636, 1663 and 1668,  ant that code is wrong even with your patch. The correct solution is to check that the lenght of the parameter is smaller than MAX_FILENAME, and after that you can use memcpy because you know the size. Using something like strlcpy doesn't add anything in this case, you still have to check the returned value, because you can have a string '\0' terminated but truncated, that can generate other security problems.

kostapr: I agree with your last statement. The code should check the actual length of optagr, make sure it's smaller than the copy buffer and only then copy the optagr string to the destination.

dp-arm:
@teknoraver can you address @robertovargas-arm's comments?

teknoraver:
will do

teknoraver:
@dp-arm and @robertovargas-arm sorry for the delay, done.

Both solutions are fine to me, I just want to get rid of this annoying warning, but I don't fully understand the concerns about the first imlementation.
`char out_file[MAX_FILENAME+1] = { 0 }` will allocate an all zeroed 255+1 buffer, and `strncpy(out_file, argv[optind], MAX_FILENAME)` will copy at most 255 byte into it, leaving the string null terminated.

Looking at the code, there is no need at all to allocate `in_file` and `out_file` on the stack, you can just keep a pointer to the argv relevant entry, eg.
```
char *in_file;
char *out_file;
[...]
in_file = argv[optind];
[...]
out_file = argv[optind];
```

Also, why defining `MAX_FILENAME` when there are already `NAME_MAX` and `PATH_MAX` defined in `<limits.h>`?

Anyway, I've created a branch https://github.com/teknoraver/arm-trusted-firmware/tree/small_fixes with a reworked version and other fixes.

Cheers,
Matteo

robertovargas-arm:
Using strncpy doesn't solve all the problems. If the parameter is too long it is going to be truncated, which can generate problems. You have to check the size and then using memcpy. In this way you are sure that the parameter is correct and you give a proper message error. @kostapr @dp-arm  agreed in this solution.

teknoraver:
Ok, go for strlen()+memcpy then.

Regards,
Matteo

soby-mathew:
@teknoraver, any update for this PR?

teknoraver:
> @teknoraver, any update for this PR?

Hi,

I tought that fd10a0a34e93270731136e7d3f8dbe3fc2dba281 was good to merge.
it uses strlen() and memcpy() as suggested by @robertovargas-arm

soby-mathew:
Ah, sorry, didn't realize rework is already done. Looks good to me.

PR #1490, opened 2018-07-23
Atf master+linaro warp7 pr0 precursor branch

bryanodonoghue:
Hello

This is the first in a set of 13 pull requests I plan to make against this tree

https://github.com/bryanodonoghue/arm-trusted-firmware/tree/atf-master%2Blinaro-warp7

This first pull request makes two changes - first is adding platform_def.h into misc_helpers.S in order to enable individual platforms to declare the ability to do "smc #0" and secondly a change to the NXP/IMX directory structure to something that looks a bit more logical to me.

You can see the full list of intended pull requests by checking the tags attached to the master branch
atf-master+linaro-warp7-pr0-precursor
atf-master+linaro-warp7-pr1-stub-build
atf-master+linaro-warp7-pr2-clocks
atf-master+linaro-warp7-pr3-uart
atf-master+linaro-warp7-pr4-storage
atf-master+linaro-warp7-pr5-permissions-muxes
atf-master+linaro-warp7-pr6-snvs
atf-master+linaro-warp7-pr7-timers
atf-master+linaro-warp7-pr8-io-storage
atf-master+linaro-warp7-pr9-hab
atf-master+linaro-warp7-pr10-system-counter
atf-master+linaro-warp7-pr11-caam
atf-master+linaro-warp7-pr11-wdog
atf-master+linaro-warp7-pr12-usb
atf-master+linaro-warp7-pr13-docs

Given it's approximately  150 commits I was thinking to break it down into a number of PRs like the above ?


bryanodonoghue:
@omnium21

antonio-nino-diaz-arm:
IP Review +1

antonio-nino-diaz-arm:
Well, it will certainly be less overwhelming to review small PRs and patches than big ones. I think we can try and see how it goes.

danh-arm:
I don't think this project needs to see the full 150 commit history for a single platform port. These should be squashed into logical blocks for easier upstreaming to this project. After that it might be easier to see whether that should be upstreamed in separate PRs or one big one.

bryanodonoghue:
OK, gotcha.

I'll leave the general ethos in-place but trawl through the commits and squash as many as I can.

danh-arm:
Also, one you've have a logical commit history, this can be submitted as a single PR. The only reason to separate into separate PRs is if some of the commits are a work in progress. Each PR should do something useful as a whole.

bryanodonoghue:
OK.

I'll try to whittle this down some more and then submit as one.

I would like to treat this PR separately though 

1. It touches a file outside of plat/
2. It modifies the directory structure of an existing platform

Later patches are entirely standalone to the platform we are adding if that makes sense to you two guys and is acceptable to @Anson-Huang

antonio-nino-diaz-arm:
Yes, I agree, this PR as it is is fine.

lib/aarch32/misc_helpers.S
```diff
@@ -7,6 +7,7 @@
 #include <arch.h>
 #include <asm_macros.S>
 #include <assert_macros.S>
+#include <platform_def.h>
```
danh-arm: I disagree with the solution you've chosen. This file should not require anything platform specific. To avoid having compiler specific directives in the source code, I think it would be better to modify the compiler options in armv7-a-cpus.mk. It looks like you can alter the `-mcpu` and `-march` options with `+sec`.

danh-arm:
Hi @Anson-Huang. Could you ack the proposed directory structure change in this PR?

Anson-Huang:
Do we have to change the directory of imx to nxp/imx? The reason why we do NOT put imx in nxp directory is because sometimes, company name changes, like previous we use fsl directory, but freescale merged with nxp, so freescale is NO longer exist. And NXP also may be merged with other company...

Generally, I am OK with the change if we have to, otherwise, I would say we can keep it for now, maybe the pull request for directory change can be merged later when other NXP products being added? Does it make sense?

danh-arm:
> Do we have to change the directory of imx to nxp/imx?

I guess the problem is that Bryan wants to add imx7 and the existing plat/imx directory is specific to imx8, which could be confusing. So even if we don't create an nxp directory, it would make sense to rename plat/imx to plat/imx8.

Anson-Huang:
Is it OK to put all i.MX in the same imx directory including imx7 and imx8? As we prefer to have imx directory for all i.MX SoCs.

lib/aarch32/misc_helpers.S
bryanodonoghue: I'll check that. If it can be done that way I'll make the change in the next PR.

Anson-Huang:
Sounds good, let's do it in this way.

bryanodonoghue:
@danh-arm 

Agreed. I can do what I need to do by setting this on the build command line
ARM_CORTEX_A7=yes

bryanodonoghue:
Closing in favour of : https://github.com/bryanodonoghue/arm-trusted-firmware/tree/atf-master%2Blinaro-warp7-squash

PR #1491, opened 2018-07-24
Arm platforms: Fix type mismatch for arm_pm_idle_states

jeenu-arm:
This also gets rid of MISRA violations for Rule 8.3 and 8.4.

Change-Id: I45bba011b16f90953dd4b260fcd58381f978eedc


danh-arm:
This change flags another MISRA issue (which was already there):

[MISRA C-2012 Rule 8.11]<271399> The size of array "arm_pm_idle_states" is not stated explicitly.

Would be good to fix this at the same time or at least make an explicit decision about the applicability of this rule.

jeenu-arm:
We agreed to fix this (8.11 - advisory) later as there are already existing instances.

PR #1494, opened 2018-07-25
Hikey960: configure pins for PCIe controller

hzhuang1:
GPIO_089 connects to PCIE_PERST_N. It needs to be configured as
output low.

Signed-off-by: Kaihua Zhong <zhongkaihua@huawei.com>
Signed-off-by: Xiaowei Song <songxiaowei@hisilicon.com>


antonio-nino-diaz-arm:
IP Review +1

antonio-nino-diaz-arm:
Note that you could ignore them as they are not related to this PR. If you care about them you can always raise a PR in the future to fix them.

antonio-nino-diaz-arm:
My bad, they are related!

hzhuang1:
It seems that it's an issue at build time. I'm using GCC without this issue. How to reproduce this issue?

antonio-nino-diaz-arm:
These are MISRA defects, not build problems. In this case the fix would be to add `UL()` to the constants: They are addresses and you will access them as virtual addresses. The type we use for VAs is `uintptr_t`, which internally is `unsigned long` (you need to rebase your patches to get the `UL()` macro, though).

If you don't mind the MISRA defects in your platform we can merge your PR and even disable the checks.

jeenu-arm:
@hzhuang1 Coverity is a commercial tool, and to the best of my knowledge, you ought to have a licensed server instance to run scans. If you do have such a setup, we're happy share the workflow that we use.

PR #1496, opened 2018-07-26
Code change to fix small bugs 

danielboulby-arm:


plat/arm/common/arm_bl2_setup.c
```diff
@@ -37,7 +37,7 @@ CASSERT(BL2_BASE >= ARM_TB_FW_CONFIG_LIMIT, assert_bl2_base_overflows);
 
 #if LOAD_IMAGE_V2
 
-#pragma weak bl2_plat_handle_post_image_load
+#pragma weak arm_bl2_plat_handle_post_image_load
```
danh-arm: This function should not be weak - we do not expect it to be overridden by a specific platform.

danielboulby-arm:
Ok, I will close this pull request and submit them as 2 separate ones with the coverity rules fixed

PR #1504, opened 2018-07-31
Migrate mmc

hzhuang1:


hzhuang1:
@vchong Please help to review and verify it on poplar platform.
@Yann-lms Please help to review the changes on mmc.

shawnguo2:
I tested the branch on Poplar and found it breaks the boot as below.

```
INFO:    BL1: 0x2016000 - 0x201c000 [size = 24576]
NOTICE:  Booting Trusted Firmware
NOTICE:  BL1: v1.5(debug):v1.5-600-gaa643a8b8afc
NOTICE:  BL1: Built : 15:00:04, Aug  1 2018
INFO:    BL1: RAM 0x2016000 - 0x201c000
INFO:    BL1: cortex_a53: CPU workaround for 855873 was applied
INFO:    BL1: initializing emmc
ERROR:   dw_send_cmd, RINTSTS:0x0
```

Yann-lms:
CMD1 is not supported for SD-cards, so I have an error message when booting on those devices.

Doing a CMD8 before that is used in U-Boot to distinguish between an eMMC and an SD-card, and is expected to fail on eMMC devices. It can also be done like this in kernel if the capabilities are not set. 

But maybe that is a wrong approach too. As the boot device is known when initializing the driver, maybe it could be a parameter to the init function, and then we'll have a dedicated init for eMMC devices and SD-cards? Maybe with the flags parameter.

The 2 other patches are OK for me.

I've also made another patch, not yet pushed, that adds delays between retries, as done in U-Boot. It can also improve the boot on some devices, and avoids spamming the card for commands where there are retries. I can push it today.


Another thing, I'll be out of office at the end of the week for 3 weeks.

Yann-lms:
I've made the PR for the delays:
https://github.com/ARM-software/arm-trusted-firmware/pull/1505

hzhuang1:
@shawnguo2 Could you help to verify this on poplar?

Yann-lms:
@shawnguo2 maybe you could try the patch I made with the delays for PR #1505: Yann-lms@0c7f8f1
It solved some of my issues on some configurations of my platform. Could you also add the failing command number to the error trace? It could help solving the issue.

shawnguo2:
@hzhuang1 I was questioning why the medaly(5) below was added into dw_send_cmd().

```
if ((cmd->cmd_idx == 1) || (cmd->cmd_idx == 8)) {
        mdelay(5);
}
```
Although I do not understand it, I see Poplar boot move one step further if I revert this mdelay change.  And it can pass 'BL1: initializing emmc', but stops at 'BL1: Loading BL2' then. 
```
INFO:    BL1: 0x2016000 - 0x201c000 [size = 24576]
NOTICE:  Booting Trusted Firmware
NOTICE:  BL1: v1.5(debug):v1.5-601-g2fa60222bb0f
NOTICE:  BL1: Built : 15:18:58, Aug  3 2018
INFO:    BL1: RAM 0x2016000 - 0x201c000
INFO:    BL1: cortex_a53: CPU workaround for 855873 was applied
INFO:    BL1: initializing emmc
INFO:    BL1: Loading BL2
INFO:    Using eMMC
ERROR:   error initializing fip
ASSERT: plat/hisilicon/poplar/plat_storage.c:212
```

Yann-lms:
I do not use the synopsis IP and driver, so I don't know much about it. @hzhuang1 should know it better.

But maybe you could change this line:
`ERROR("%s, RINTSTS:0x%x\n", __func__, data)`
with
 `ERROR("%s, RINTSTS:0x%x, CMD%d\n", __func__, data, cmd->cmd_idx)`

And add a trace before the return -EIO; 4 lines before, to see if some command failed.

shawnguo2:
@Yann-lms Thanks for the comments.  I tried your delay commit and unfortunately it doesn't help.  I traced the failing command as you suggested.  When it stops at 'initializing emmc' with mdelay(5), dw_send_cmd() fails like below.

```
ERROR:   dw_send_cmd, RINTSTS:0x0, CMD13
```
With mdelay(5) reverted, it stops at 'Loading BL2' as following.
```
INFO:    BL1: 0x2016000 - 0x201c000 [size = 24576]
NOTICE:  Booting Trusted Firmware
NOTICE:  BL1: v1.5(debug):v1.5-603-g4c9583df7fa5
NOTICE:  BL1: Built : 16:02:38, Aug  3 2018
INFO:    BL1: RAM 0x2016000 - 0x201c000
INFO:    BL1: cortex_a53: CPU workaround for 855873 was applied
INFO:    BL1: initializing emmc
ERROR:   dw_send_cmd, EIO RINTSTS:0x2004, CMD6
INFO:    BL1: Loading BL2
INFO:    Using eMMC
ERROR:   dw_send_cmd, EIO RINTSTS:0x104, CMD23
ERROR:   error initializing fip
ASSERT: plat/hisilicon/poplar/plat_storage.c:212
```

shawnguo2:
@hzhuang1 No, adding print for each mmc command in dw_send_cmd() doesn't help me.
```
INFO:    BL1: 0x2016000 - 0x201c000 [size = 24576]
NOTICE:  Booting Trusted Firmware
NOTICE:  BL1: v1.5(debug):v1.5-600-g8dffc52b3a06-dirty
NOTICE:  BL1: Built : 14:43:14, Aug  4 2018
INFO:    BL1: RAM 0x2016000 - 0x201c000
INFO:    BL1: cortex_a53: CPU workaround for 855873 was applied
INFO:    BL1: initializing emmc
INFO:    *** CMD0
INFO:    *** CMD1
INFO:    *** CMD1
INFO:    *** CMD1
INFO:    *** CMD1
INFO:    *** CMD1
INFO:    *** CMD1
INFO:    *** CMD1
INFO:    *** CMD1
INFO:    *** CMD1
INFO:    *** CMD2
INFO:    *** CMD3
INFO:    *** CMD9
INFO:    *** CMD7
INFO:    *** CMD13
INFO:    *** CMD8
INFO:    BL1: Loading BL2
INFO:    Using eMMC
INFO:    *** CMD23
ERROR:   error initializing fip
ASSERT: plat/hisilicon/poplar/plat_storage.c:212
```

shawnguo2:
@hzhuang1 I got some findings by running master and your branch in comparison.  I see different mmc command sequence for enumerating.  With your branch, we get CMD8 right after CMD13, while we get CMD6 after CMD13 with master branch.  So I tried to move mmc_set_ios() which calls mmc_set_ext_csd() prior to  mmc_fill_device_info() like below.  And it just fixes my problem.  Please comment if it is the right fix.
```
diff --git a/drivers/mmc/mmc.c b/drivers/mmc/mmc.c
index b44d4a19bc2a..34df00b39195 100644
--- a/drivers/mmc/mmc.c
+++ b/drivers/mmc/mmc.c
@@ -462,12 +462,17 @@ static int mmc_enumerate(unsigned int clk, unsigned int bus_width)
                }
        } while (ret != MMC_STATE_TRAN);
 
+       ret = mmc_set_ios(clk, bus_width);
+       if (ret != 0) {
+               return ret;
+       }
+
        ret = mmc_fill_device_info();
        if (ret != 0) {
                return ret;
        }
 
-       return mmc_set_ios(clk, bus_width);
+       return 0;
 }
 
 size_t mmc_read_blocks(int lba, uintptr_t buf, size_t size)
```

hzhuang1:
@shawnguo2 Good catch. I'll prepare the 3rd pull request.

hzhuang1:
And I find that mdelay() could be removed now.

hzhuang1:
@antonio-nino-diaz-arm Could you help to merge this pull request?

antonio-nino-diaz-arm:
IP Review +1

dp-arm:
@hzhuang1 can you resolve the conflicts?  I will merge the PR once that is done.

PR #1506, opened 2018-08-01
Fix build for SEPARATE_CODE_AND_RODATA=0

danielboulby-arm:
TF won't build since no memory region is specified
for when SEPARATE_CODE_AND_RODATA=0 it still relies on
the ARM_MAP_BL_RO_DATA region which is never defined for
this case. Create memory region combining code and RO data for
when the build flag SEPARATE_CODE_AND_RODATA=0 to fix this

Change-Id: I6c129eb0833497710cce55e76b8908ce03e0a638


dp-arm:
jenkins: test this please

danielboulby-arm:
The checks fail due to MISRA violations that will be addressed as a whole at a later date

PR #1508, opened 2018-08-01
Fix handler prototype in Trusty generic dispatcher

sandrine-bailleux-arm:
Fix the types of the arguments of trusty_generic_platform_smc()
to match the expected prototype of a runtime service handler
(see rt_svc_handle_t type).


sandrine-bailleux-arm:
There was a single test configuration that failed in the CI but the failure was probably due to some network issues on our side yesterday. I've re-triggered the same test configuration this morning and this time it passed.

PR #1509, opened 2018-08-02
xlat: psci: Use bool instead of int

antonio-nino-diaz-arm:
This is done to comply with MISRA rules about essential types.


dp-arm:
jenkins: test this please

PR #1498, opened 2018-07-26
Early cache enable and coherency fixes

glneo:

PR #1513, opened 2018-08-06
xlat v2: Cleanup and dcache coherency bug fix

antonio-nino-diaz-arm:
- The functions to get and change attributes could only work on a xlat context passed as a pointer. This is not practical if you want to modify the current translation tables as it is needed to export the xlat tables context, which is not a clean fix. The old functions have been renamed and new functions have been created to not need to export the context. Trying to use the old names will now result in a compilation error so that it is obvious that the new functions exist and the user can choose the correct one, potentially making the code cleaner.

- The following bug has been fixed:

```
During cold boot, the initial translation tables are created with data
caches disabled, so all modifications go to memory directly. After the
MMU is enabled and data cache is enabled, any modification to the tables
goes to data cache, and eventually may get flushed to memory.

If CPU0 modifies the tables while CPU1 is off, CPU0 will have the
modified tables in its data cache. When CPU1 is powered on, the MMU is
enabled, then it enables coherency, and then it enables the data cache.
Until this is done, CPU1 isn't in coherency, and the translation tables
it sees can be outdated if CPU0 still has some modified entries in its
data cache.

This can be a problem in some cases. For example, the warm boot code
uses only the tables mapped during cold boot, which don't normally
change. However, if they are modified (and a RO page is made RW, or a XN
page is made executable) the CPU will see the old attributes and crash
when it tries to access it.

This doesn't happen in systems with HW_ASSISTED_COHERENCY. In these
systems, the data cache is enabled at the same time as the MMU. As soon
as this happens, the CPU is in coherency.

There was an attempt of a fix in psci_helpers.S, but it didn't solve the
problem. That code has been deleted. The code was introduced in commit
<264410306381> ("Invalidate TLB entries during warm boot").

Now, during a map or unmap operation, the memory associated to each
modified table is flushed. Traversing a table will also flush it's
memory, as there is no way to tell in the current implementation if the
table that has been traversed has also been modified.
```

Link: 264410306381

### DIFF
--- a/lib/psci/aarch32/psci_helpers.S
+++ b/lib/psci/aarch32/psci_helpers.S
@@ -91,28 +91,6 @@ func psci_do_pwrup_cache_maintenance
 	stcopr	r0, SCTLR
 	isb
 
-#if PLAT_XLAT_TABLES_DYNAMIC
-	/* ---------------------------------------------
-	 * During warm boot the MMU is enabled with data
-	 * cache disabled, then the interconnect is set
-	 * up and finally the data cache is enabled.
-	 *
-	 * During this period, if another CPU modifies
-	 * the translation tables, the MMU table walker
-	 * may read the old entries. This is only a
-	 * problem for dynamic regions, the warm boot
-	 * code isn't affected because it is static.
-	 *
-	 * Invalidate all TLB entries loaded while the
-	 * CPU wasn't coherent with the rest of the
-	 * system.
-	 * ---------------------------------------------
-	 */
-	stcopr	r0, TLBIALL
-	dsb	ish
-	isb
-#endif
-
 	pop	{r12, pc}
 endfunc psci_do_pwrup_cache_maintenance
 
--- a/lib/xlat_tables_v2/xlat_tables_core.c
+++ b/lib/xlat_tables_v2/xlat_tables_core.c
@@ -18,6 +18,13 @@
 
 #include "xlat_tables_private.h"
 
+/* Helper function that cleans the data cache only if it is enabled. */
+static inline void xlat_clean_dcache_range(uintptr_t addr, size_t size)
+{
+	if (is_dcache_enabled())
+		clean_dcache_range(addr, size);
+}
+
 #if PLAT_XLAT_TABLES_DYNAMIC
 
 /*
@@ -329,7 +336,10 @@ static void xlat_tables_unmap_region(xlat_ctx_t *ctx, mmap_region_t *mm,
 			xlat_tables_unmap_region(ctx, mm, table_idx_va,
 						 subtable, XLAT_TABLE_ENTRIES,
 						 level + 1U);
-
+#if !(HW_ASSISTED_COHERENCY || WARMBOOT_ENABLE_DCACHE_EARLY)
+			xlat_clean_dcache_range((uintptr_t)subtable,
+				XLAT_TABLE_ENTRIES * sizeof(uint64_t));
+#endif
 			/*
 			 * If the subtable is now empty, remove its reference.
 			 */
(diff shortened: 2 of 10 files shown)


lib/xlat_tables_v2/xlat_tables_core.c
```diff
@@ -329,7 +329,10 @@ static void xlat_tables_unmap_region(xlat_ctx_t *ctx, mmap_region_t *mm,
 			xlat_tables_unmap_region(ctx, mm, table_idx_va,
 						 subtable, XLAT_TABLE_ENTRIES,
 						 level + 1U);
-
+#if !HW_ASSISTED_COHERENCY
+			flush_dcache_range((uintptr_t)subtable,
```
danh-arm: Isn't `clean_dache_range enough`?

lib/xlat_tables_v2/xlat_tables_core.c
```diff
@@ -563,6 +566,10 @@ static uintptr_t xlat_tables_map_region(xlat_ctx_t *ctx, mmap_region_t *mm,
 			end_va = xlat_tables_map_region(ctx, mm, table_idx_va,
 					       subtable, XLAT_TABLE_ENTRIES,
 					       level + 1U);
+#if !HW_ASSISTED_COHERENCY
+			flush_dcache_range((uintptr_t)subtable,
+				XLAT_TABLE_ENTRIES * sizeof(uint64_t));
```
danh-arm: Why are these calls added after each call to `xlat_tables_map_region()` rather than in that function?

lib/xlat_tables_v2/xlat_tables_utils.c
```diff
@@ -539,7 +539,9 @@ int xlat_change_mem_attributes_ctx(const xlat_ctx_t *ctx, uintptr_t base_va,
 		 * before writing the new descriptor.
 		 */
 		*entry = INVALID_DESC;
-
+#if !HW_ASSISTED_COHERENCY
+		dccvac((uintptr_t)entry);
```
danh-arm: I note you're doing a clean here rather than a clean+invalidate like the other places. I think a clean is enough though.

lib/xlat_tables_v2/xlat_tables_core.c
```diff
@@ -329,7 +329,10 @@ static void xlat_tables_unmap_region(xlat_ctx_t *ctx, mmap_region_t *mm,
 			xlat_tables_unmap_region(ctx, mm, table_idx_va,
 						 subtable, XLAT_TABLE_ENTRIES,
 						 level + 1U);
-
+#if !HW_ASSISTED_COHERENCY
```
danh-arm: I think this needs to be `#if !(HW_ASSISTED_COHERENCY || WARMBOOT_ENABLE_DCACHE_EARLY)`. See https://github.com/ARM-software/arm-trusted-firmware/pull/1498

lib/psci/aarch32/psci_helpers.S
```diff
@@ -91,28 +91,6 @@ func psci_do_pwrup_cache_maintenance
 	stcopr	r0, SCTLR
 	isb
 
-#if PLAT_XLAT_TABLES_DYNAMIC
```
danh-arm: minor: The 264410306381 in the commit message doesn't get displayed as a link in GitHub

antonio-nino-diaz-arm: I think GitHub doesn't like this system, but it's the one that checkpatch.pl likes. It's a problem of the UI anyway.

lib/xlat_tables_v2/xlat_tables_core.c
antonio-nino-diaz-arm: Done.

lib/xlat_tables_v2/xlat_tables_core.c
antonio-nino-diaz-arm: It is. Changed.

lib/xlat_tables_v2/xlat_tables_core.c
antonio-nino-diaz-arm: There are two options:

1 - Clean the whole table after (which will only affect the modified cache lines anyway).
2 - Clean every entry as they are modified, which will increase the code complexity and clean the same cache lines multiple times.

For mapping/unmapping I've chosen 1 because we only need to do the change once. For  ``get_mem_attributes`` and ``change_mem_attributes`` I've chosen 2 because we need to clean once when invalidating the table entry and once when it is set to the new attributes.

lib/xlat_tables_v2/xlat_tables_utils.c
antonio-nino-diaz-arm: I've changed the rest.

PR #1522, opened 2018-08-14
Remove cyclic inclusion of platform_def.h in common_def.h

deepan02:
By removing the inclusion of platform_def.h in common_def.h and
arm_def.h, it helps to avoid some of the cyclic inclusion problems in
platforms. This change mostly affects ARM platforms and it can be fixed
by directly including the required header.

Change-Id: I77421cdfc72d9fc906089b3d8571441bba473dca
Signed-off-by: Deepak Pandey <Deepak.Pandey@arm.com>

### DIFF
--- a/include/plat/arm/common/arm_def.h
+++ b/include/plat/arm/common/arm_def.h
@@ -10,7 +10,6 @@
 #include <common_def.h>
 #include <gic_common.h>
 #include <interrupt_props.h>
-#include <platform_def.h>
 #include <tbbr_img_def.h>
 #include <utils_def.h>
 #include <xlat_tables_defs.h>
--- a/include/plat/common/common_def.h
+++ b/include/plat/common/common_def.h
@@ -7,7 +7,6 @@
 #define __COMMON_DEF_H__
 
 #include <bl_common.h>
-#include <platform_def.h>
 
 /******************************************************************************
  * Required platform porting definitions that are expected to be common to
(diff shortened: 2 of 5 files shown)


jeenu-arm:
jenkins: add to whitelist

jeenu-arm:
jenkins: test this please

soby-mathew:
Although the change itself is not drastic, please hold off merging this till the original problem is understood.

antonio-nino-diaz-arm:
This needs to be rebased.

include/plat/arm/common/arm_def.h
```diff
@@ -10,7 +10,6 @@
 #include <common_def.h>
```
sandrine-bailleux-arm: As far as I can see, ``arm_def.h`` uses no definition from ``common_def.h`` so I think we could/should remove this inclusion.

include/plat/common/common_def.h
```diff
@@ -7,7 +7,6 @@
 #define __COMMON_DEF_H__
 
 #include <bl_common.h>
-#include <platform_def.h>
```
sandrine-bailleux-arm: I think this is not right. ``common_def.h`` uses ``BL2_BASE``, which is provided by ``platform_def.h`` so we have to maintain this inclusion.

soby-mathew:
Closing this PR as no longer needed.

PR #1523, opened 2018-08-14
DSU erratum 936184 workaround

jts-arm:
If the system is in near idle conditions, this erratum could cause a
deadlock or data corruption. This patch applies the workaround that
prevents this.

This DSU erratum affects only the DSUs that contain the ACP interface
and it was fixed in r2p0. The workaround is applied only to the DSUs
that are actually affected.

Link to respective Arm documentation:
http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.epm138168/index.html

Change-Id: I033213b3077685130fc1e3f4f79c4d15d7483ec9
Signed-off-by: John Tsichritzis <john.tsichritzis@arm.com>

### DIFF
--- a/bl31/bl31.mk
+++ b/bl31/bl31.mk
@@ -24,6 +24,7 @@ BL31_SOURCES		+=	bl31/bl31_main.c				\
 				bl31/bl31_context_mgmt.c			\
 				common/runtime_svc.c				\
 				lib/aarch64/setjmp.S				\
+				lib/cpus/aarch64/dsu_helpers.S			\
 				plat/common/aarch64/platform_mp_stack.S		\
 				services/arm_arch_svc/arm_arch_svc_setup.c	\
 				services/std_svc/std_svc_setup.c		\
--- a/bl1/bl1.mk
+++ b/bl1/bl1.mk
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2013-2017, ARM Limited and Contributors. All rights reserved.
+# Copyright (c) 2013-2018, ARM Limited and Contributors. All rights reserved.
 #
 # SPDX-License-Identifier: BSD-3-Clause
 #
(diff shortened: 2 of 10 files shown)


jeenu-arm:
jenkins: test this please

include/lib/cpus/aarch64/dsu_arch.h
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright (c) 2018, ARM Limited and Contributors. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef DSU_ARCH_H
+#define DSU_ARCH_H
```
danh-arm: minor: strictly speaking, the DSU is mircoarchitecture, not architecture, which might affect this filename

bl31/bl31.mk
```diff
@@ -24,6 +24,7 @@ BL31_SOURCES		+=	bl31/bl31_main.c				\
 				bl31/bl31_context_mgmt.c			\
 				common/runtime_svc.c				\
 				lib/aarch64/setjmp.S				\
+				lib/cpus/aarch64/dsu_helpers.S			\
```
danh-arm: CPU libs are usually pulled in by the platforms that need them, not generic code. Also, what about the BL2_AT_EL3 case?

include/lib/cpus/aarch64/dsu_arch.h
jts-arm: I will rename that to dsu_def.h

bl31/bl31.mk
jts-arm: 1) Firstly, the DSU erratum does not apply only to a specific CPU. Secondly, the workaround is the same for every CPU it applies to. For these reasons it was implemented in a separate file, as a helper function.
Subsequently, the "dsu_helpers.S" file is included by generic code, similar to how we include "cpu_helpers.S".
However, the strategy of when to *apply* the workaround is exactly the same as with other workarounds and indeed is controlled by the platform.mk files. Conditional branches inside the relevant cortex_aXX.S files have been added so as to control when the workaround code will be executed.

2) For the BL2_AT_EL3 case, you are right, will be addressed.

soby-mathew: @jts-arm, OK

dp-arm:
jenkins: test this please

PR #1524, opened 2018-08-14
rockchip: Add plat_is_my_cpu_primary function

danielboulby-arm:
This function is required for platforms where
COLD_BOOT_SINGLE_CPU=0 however it was missing from rockchip
platforms

Change-Id: I32a85f226a4f22085a27113903f34bdb6f28dbcc


TonyXie06:
It is OK for our platform。

PR #1526, opened 2018-08-15
memprotect: Move files to specific platform makefiles

robertovargas-arm:
All the arm platforms were including the files related to
mem-protect. This configuration generates some problems
with new platforms that don't support such functionality,
and for that reason this patch moves these files to the
platform specific makefiles.

Change-Id: I6923e5224668b76667795d8e11723cede7979b1e


dp-arm:
jenkins: test this please

dp-arm:
jenkins: test this please

PR #1528, opened 2018-08-17
libc: Cleanup library

antonio-nino-diaz-arm:
This patch does a cleanup of the C library code. For that, it takes code from [SCC](http://www.simple-cc.org/) and [FreeBSD](https://www.freebsd.org/). There are still some files that have been written specifically for the Trusted Firmware and don't come from these projects.

``tf_printf`` is now the default ``printf``, and the previous ``printf`` has been removed: It used a lot of stack and was too big for this project. It has been modified to return the number of characters it prints.

Similarly, ``tf_snprintf`` is now the default ``snprintf`` as it is smaller.

All the codebase has been updated to use the new headers instead of the old non-standard ones, and to use ``printf`` and ``snprintf``.

Documentation updated.


b49020:
For plat/socionext/synquacer/, `Acked-by: Sumit Garg <sumit.garg@linaro.org>`

jenswi-linaro:
`plat/qemu` changes OK

PR #1534, opened 2018-08-21
Add support for Allwinner H6 + X-Powers AXP805 PMIC combination

Icenowy:
Boards with Allwinner H6 SoC, which is a SoC that newly gets supported by mainline TF-A recently, are usually combined with a X-Powers AXP805 PMIC according to the reference design from Allwinner. The PMIC is connected to the SoC via I2C at PL0/PL1 pins (called R_I2C, R_ is the prefix on Allwinner SOCs for RTC power domain). The I2C controller uses the same IP core with Marvell A8K, but with different integration.

As PSCI is going to be implemented by the TF-A, some power-related functions will need to communicate with the PMIC. Boot time power state initialization also needs PSCI to have access to the PMIC.

Add support for the I2C controller and the initialization of the PMIC. PSCI system power down implementation is also added as an example.


plat/allwinner/common/include/spl_head.h
```diff
@@ -0,0 +1,26 @@
+#ifndef __SPL_HEAD_H__
```
antonio-nino-diaz-arm: Missing copyright and SPDX identifier.

antonio-nino-diaz-arm:
I think it would be a good idea to split the changes to the Marvell drivers and to Allwinner platforms into two pull requests. The one that affects the Marvell drivers can be merged as soon as @kostapr is happy with the changes. I know that @Andre-ARM won't be able to take a look at this in a few days, so the changes to Allwinner may take longer, but that also depends on what @smaeul thinks.

plat/allwinner/common/sunxi_common.c
```diff
@@ -71,3 +74,25 @@ uint16_t sunxi_read_soc_id(void)
 
 	return reg >> 16;
 }
+
+const char *sunxi_get_dt_name(void)
+{
+	const struct spl_head *spl_head =
+		(const struct spl_head *) SUNXI_SRAM_A1_BASE;
```
smaeul: U-Boot SPL already has a mechanism (CONFIG_SPL_ATF) for passing the device tree and other data to ATF. We should use that instead of hardcoding addresses and checking for magic values.

Icenowy: But the SPL header method is already well defined, and being used by SPL and U-Boot.

It's not a magic.

antonio-nino-diaz-arm:
Also, you need to modify the paths in maintainers.rst.

plat/allwinner/common/sunxi_common.c
Andre-ARM: I agree with @smaeul that we should do it properly now. While the SPL header method might be somewhat "well defined", checking against all known DT names is not going to fly. I just added this hack to the older ATF port because it provided a pragmatic solution, and the code base was hacked anyway.

So what about the following:
1) We drop the whole SPL header DT string code from this PR.
2) We assume an AXP805 if an H6 is detected, keeping the warning or maybe tone it down to just an announcement that we will use an AXP805.
2a) We could possibly try to detect the AXP805, either using I2C methods to check for a device on the bus or by querying and verifying the ID from the AXP805.
3) Later we add DT support to this ATF port, piggy backing on the Rockchip work, probably.
4) We then scan the DT for the AXP805 compatible string or something equally easy, and use this information to enable the PMIC code.

So 1) and 2) means actually removing code from this PR, and we address 3) and 4) separately later.

Does that make sense?

Icenowy: I choose to follow 2a, although 2 will be used temporarily.

I have already seen H6+AXP802 sample board photo (although no real board).

BTW AXP802 is customized AXP803.

dp-arm:
@Icenowy can you rework and rebase the patch?  We have a code freeze on Friday the 7th of September, so if you want this to be part of the 1.6 release it would have to be merged before then.

@Andre-ARM any comments from you on this PR?

Andre-ARM:
@dp-arm: I replied already to @smaeul's review above. But apart from dropping the DT stub name part and dropping the i2c/ directory level from the driver's path name I am fine with those changes. I will tag a reworked version as reviewed.

Icenowy:
Rebased and fixed build issues on a8k platform.

Icenowy:
Oh forgot to say that DT name reading is currently dropped.

Icenowy:
@Andre-ARM The path is fixed.

For code sharing I will prefer to fix this up later.

Andre-ARM:
@Icenowy: many thanks for the quick fixes!
I can't test this until later today, but the general approach seems fine to me. Special kudos for the Marvell driver reusing!
@soby-mathew: The Allwinner platform part looks very good to me. I am not sure how important this is to get into the next release, but if it is, I am happy with this being merged now.

soby-mathew:
jenkins: test this please

plat/allwinner/sun50i_h6/sunxi_power.c
```diff
@@ -0,0 +1,143 @@
...
+
+	return i2c_read(chip, 0, 0, val, 1);
+}
+
+int axp_i2c_write(uint8_t chip, uint8_t reg, uint8_t val)
+{
+	return i2c_write(chip, reg, 1, &val, 1);
+}
+
+static int axp805_probe()
```
soby-mathew: Just one comment regarding coding style.
```plat/allwinner/sun50i_h6/sunxi_power.c:79: ERROR: Bad function definition - int axp805_probe() should probably be int axp805_probe(void)```

**Comment by soby-mathew** [comment]
jenkins: test this please

**Comment by soby-mathew** [comment]
IP-Review +1

**Comment by kostapr** [comment]
@Icenowy I am OK with the anonymous union change. As for moving the i2c driver code to a new location, I would like to run a test on my own system for making sure everything works as expected. I also want to query chip design team regarding the i2c IP ownership - this information is not always known to SW developers, so I was not aware of it.

File: drivers/marvell/i2c/a8k_i2c.c [hunk 4]
```diff
@@ -7,51 +7,15 @@
...
 #include <mmio.h>
-#include <mvebu_def.h>
-
-#if LOG_LEVEL >= LOG_LEVEL_VERBOSE
-#define DEBUG_I2C
-#endif
 
 #define CONFIG_SYS_TCLK			250000000
 #define CONFIG_SYS_I2C_SPEED		100000
 #define CONFIG_SYS_I2C_SLAVE		0x0
```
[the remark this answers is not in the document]
Andre-ARM: I agree with @kostapr here. I tried this, by renaming the .c stub files to mentor_i2c_plat.h and moving them into plat/{allwinner,marvell/a8k}/common/include (removing the last line, of course). This allows the driver to just `#include <mentor_i2c_plat.h>` and things "magically" fall into place. At least for Allwinner that worked smoothly. So I think we should fix this.

However I don't see this as a blocker to this series, so can we still merge this right now and fix this up later? The PR has been around for a while and there is this code freeze, so it wouldn't be fair to hold it back from the release. I am happy to provide fixup patches for this.

plat/allwinner/sun50i_h6/sunxi_power.c
```diff
@@ -119,3 +120,24 @@ int sunxi_pmic_setup(void)
 
 	return 0;
 }
+
+void __dead2 sunxi_power_down(void)
+{
+	uint8_t val;
+
+	switch (pmic) {
+	case AXP805:
```
Andre-ARM: I think we need to initialise the I2C driver here again. The non-secure OS could have configured this entirely differently or even disabled I2C (especially the platform part), so we can't rely on our initialisation at boot time still being valid. This makes the initialisation at boot time somewhat questionable, but it's still useful for detecting the PMIC, selecting and reporting it.

Can be easily fixed with a follow up patch.

soby-mathew:
As @Andre-ARM commented, It is better to merge this PR and other fixes can follow later.

PR #1540, opened 2018-08-26
Marvell updates 18.09

kostapr:
Some fixes and updates to align the current mainline code with Marvell upcoming 18.09 release. The A3700 support patches are delayed, but the documentation still mentions this platform along with A8K. The A3700 support patches are planned to be pushed to the mainline in September-October this year.


plat/marvell/a8k/common/plat_bl31_setup.c
```diff
@@ -75,6 +76,24 @@ _Bool is_pm_fw_running(void)
 	return pm_fw_running;
 }
 
+/* For TrusTzone we treat the "target" filed of addr_map_win
```
antonio-nino-diaz-arm: nit: Typo

antonio-nino-diaz-arm:
IP Review +1

drivers/arm/gic/v2/gicv2_main.c
```diff
@@ -593,3 +593,11 @@ unsigned int gicv2_set_pmr(unsigned int mask)
 
 	return old_mask;
 }
+
+/*******************************************************************************
+ * This functions configures the single interrupt to be edge triggered
+ ******************************************************************************/
+void gicv2_interrupt_set_edge_triggered(unsigned int id)
```
jeenu-arm: This seems to be useful during init-time only. In that case, the correct way to do this would be interrupt properties--edge/level trigger is one of the interrupt properties passed to various GIC init. See include/common/interrupt_props.h. The old method of GIC driver initialization is deprecated, and will be removed soon.

soby-mathew:
Please target the PR to integration branch rather than master.

kostapr:
@soby-mathew OK, I will update  this PR once we agree on all code changes requested by reviewers.

drivers/arm/gic/v2/gicv2_main.c
soby-mathew: Hi Marcin,
> It must be set each time we want to trigger another SPI via dedicated SoC control block. 

Okay, it seems you are triggering a non-secure interrupt from EL3. The expectation is that the default setup for NS interrupts done by firmware will be overwritten by the OS.  

Be aware that if the interrupt is pending when the property is changed, then the GIC specification says that the behaviour is undefined. Hence the recommendation to set this up one time.

>Otherwise, one-time config will be overriden e.g. during OS boot.

Ideally the OS should do the config for the NS interrupts. Is there some reason why the OS is not doing this correctly ?

> And I don't see any infrastructure to override this setting (if one-time config was the case).

Yes, there is no infrastructure to override the defaults setup for non secure interrupts due to the expectation I mention above.

> OK, I will update this PR once we agree on all code changes requested by reviewers.

Okay.

Best Regards
Soby Mathew

semihalf-wojtas-marcin: Hi @soby-mathew 

I found out, that for PMU IRQ, the ACPI_MADT_PERFORMANCE_IRQ_MODE flag should enable setting edge trigger in OS. This will allow to remove calling gicv2_interrupt_set_edge_triggered() from the IRQ handler. Just need to verify it.

I hope the above works, but still I'd need to set the edge trigger for this IRQ one-time during init (in case UEFI wants to use this IRQ). In such case, would you accept using gicv2_interrupt_set_edge_triggered() in its current form or you prefer to override initial setting differently?

Thanks,
Marcin

semihalf-wojtas-marcin: @soby-mathew @jeenu-arm 

Modifying ACPI MADT table succeeded and now we need only one time configuration of the SPI to be edge triggered. Please confirm if you are fine with:
  - removing gicv2_interrupt_set_edge_triggered() from the handler
  - calling one time during init in mvebu_pmu_interrupt_enable() a following:
```
       /* Configure ODMI Frame IRQs as edge triggered */
       for (idx = 0; idx < PLATFORM_CORE_COUNT; idx++)
               gicv2_interrupt_set_edge_triggered(A7K8K_ODMI_PMU_GIC_IRQ(idx));
```
If you don't have objections, Konstantin will submit a new version of the patchset.

jeenu-arm: @semihalf-wojtas-marcin we think that it's better if the API is named gicv2_interrupt_set_cfg(), and take the interrupt number and configuration (edge/level).

kostapr:
I have re-based PR on top of mainline integration branch, but it looks like the PR itself remains submitted to master branch. Should I re-create this PR and drop the old version completely? Last time I just force-pushed my changes.

soby-mathew:
@kostapr, I have changed the target to integration branch.

soby-mathew:
jenkins: test this please

drivers/arm/gic/v2/gicv2_main.c
```diff
@@ -593,3 +593,11 @@ unsigned int gicv2_set_pmr(unsigned int mask)
 
 	return old_mask;
 }
+
+/*******************************************************************************
+ * This functions configures the single interrupt trigger type
+ ******************************************************************************/
+void gicv2_interrupt_set_cfg(unsigned int id, unsigned int trigger_type)
```
jeenu-arm: nit: would prefer cfg instead of trigger_type to match the API name. If you're reworking that, please also fix the comment.

drivers/arm/gic/v2/gicv2_main.c
```diff
@@ -593,3 +593,12 @@ unsigned int gicv2_set_pmr(unsigned int mask)
 
 	return old_mask;
 }
+
```
soby-mathew: Just a comment regarding the commit message
suggest to change this line from
"This is helpful to modify the default configuration of the GIC SPI's, which..."
to
"This is helpful to modify the default configuration of the non-secure GIC SPI's, which..."

kostapr: Done

soby-mathew:
jenkins: test this please

PR #1541, opened 2018-08-27
zynqmp: pm: Add API to get number of clocks

rajanv-xilinx:
Add new query data parameter to get actual number of
clocks so Linux can get actual clock numbers in advance
and allocate memory accordingly.

Signed-off-by: Rajan Vaja <rajan.vaja@xilinx.com>


antonio-nino-diaz-arm:
IP Review +1

PR #1542, opened 2018-08-28
Some MISRA fixes in BL31, cci and smmu

antonio-nino-diaz-arm:


b49020:
Ok for plat/socionext/synquacer/.

PR #1544, opened 2018-08-28
context_mgmt: Fix HANDLE_EA_EL3_FIRST implementation

jwerner-chromium:
This patch fixes a bug in the context management code that causes it to
ignore the HANDLE_EA_EL3_FIRST compile-time option and instead always
configure SCR_EL3 to force all external aborts to trap into EL3. The
code used #ifdef to read compile-time option declared with add_define in
the Makefile... however, those options are always defined, they're just
defined to either 0 or 1, so #if is the correct syntax to check for
them.

This bug has existed since the Nov 2017 commit 76454ab (AArch64:
Introduce External Abort handling), which changed the
HANDLE_EA_EL3_FIRST option to use add_define.

Change-Id: I7189f41d0daee78fa2fcf4066323e663e1e04d3d

### DIFF
--- a/docs/user-guide.rst
+++ b/docs/user-guide.rst
@@ -457,8 +457,10 @@ Common build options
    .. __: `platform-interrupt-controller-API.rst`
    .. __: `interrupt-framework-design.rst`
 
--  ``HANDLE_EA_EL3_FIRST``: When defined External Aborts and SError Interrupts
-   will be always trapped in EL3 i.e. in BL31 at runtime.
+-  ``HANDLE_EA_EL3_FIRST``: When set to ``1``, External Aborts and SError
+   Interrupts will be always trapped in EL3 i.e. in BL31 at runtime. When set to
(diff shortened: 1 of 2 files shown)


soby-mathew:
Yes, this was a bug introduced by adding the define always. Thanks for the fix.

soby-mathew:
The documentation for the build option in the user-guide says :
"HANDLE_EA_EL3_FIRST: When defined External Aborts and SError Interrupts
   will be always trapped in EL3 i.e. in BL31 at runtime."

Would you be able to update the documentation as well to reflect the new mechanics ? i.e change to 

"HANDLE_EA_EL3_FIRST: When set to 1, External Aborts and SError Interrupts
   will be always trapped in EL3 i.e. in BL31 at runtime. Default value is 0."

jwerner-chromium:
> Would you be able to update the documentation as well to reflect the new mechanics ?

Sure... updated!

PR #1552, opened 2018-08-30
GIC: Fix build error

glneo:
Pointers should be comparied to NULL.

Fixes: 3fea9c8b8e8e ("gic: Fix types")
Signed-off-by: Andrew F. Davis <afd@ti.com>

### DIFF
--- a/drivers/arm/gic/v3/gicv3_main.c
+++ b/drivers/arm/gic/v3/gicv3_main.c
@@ -484,7 +484,7 @@ unsigned int gicv3_get_interrupt_type(unsigned int id,
 		return INTR_GROUP1NS;
 
 	if (id < MIN_SPI_ID) {
-		assert(gicv3_driver_data->rdistif_base_addrs != 0U);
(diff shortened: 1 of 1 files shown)


antonio-nino-diaz-arm:
IP Review +1

antonio-nino-diaz-arm:
Just out of curiosity, is this breaking your build? If so, what compiler are you using? Was it detected by a static analysis tool? It is not causing any problem for us.

I agree with the change of this PR, I would just like to know why you found the problem and we didn't.

glneo:
Yes, I get:

```
drivers/arm/gic/v3/gicv3_main.c:487:48: error: comparison between pointer and zero character constant [-Werror=pointer-compare]
   assert(gicv3_driver_data->rdistif_base_addrs != 0U);
```

I'm using gcc-linaro-7.2.1-2017.11-x86_64_aarch64-linux-gnu

antonio-nino-diaz-arm:
Ok, so that's the problem: we are using GCC 6.3 because that's the one that comes with the Linaro release we claim to support.

PR #1559, opened 2018-09-04
Disable early wake during CPU suspend entry

vwadekar:


lib/psci/psci_suspend.c
```diff
@@ -170,6 +170,7 @@ void psci_cpu_suspend_start(const entry_point_info_t *ep,
 	psci_acquire_pwr_domain_locks(end_pwrlvl,
 				      idx);
 
+#define ENABLE_EARLY_WAKEUP_DETECTION
```
Yann-lms: s/define/if/ ?

vwadekar: Thanks for catching the typo. Fixed this issue.

lib/psci/psci_suspend.c
```diff
@@ -170,15 +169,17 @@ void psci_cpu_suspend_start(const entry_point_info_t *ep,
 	psci_acquire_pwr_domain_locks(end_pwrlvl,
 				      idx);
 
+#if ENABLE_EARLY_WAKEUP_DETECTION
```
soby-mathew: @vwadekar , Could you describe the reason for disabling this optimization ?

vwadekar: This handler does not return any error for the wake up. This means that the kernel thinks that we entered suspend and woke up. Tegra platforms set a timer as the wake source and expect the platform to wake up only after that time. But with this logic the platform wakes up before that time leading to false results.

The pending interrupt is not a wake up source for Tegra, which means it cannot wake Tegra from suspend mode. But this logic contradicts that configuration, leading to confusion.

soby-mathew: OK, from your explanation, the PSCI Client seems to have non-compliant behaviour. The PSCI specification says the following :
 `The powerdown request might not complete due, for example, to pending interrupts. It is also possible that, because of coordination with other cores, the actual state entered is shallower than the one requested.`

Also :
`In the case of an early return due to a pending wakeup event, the implementation can return at the next instruction, with a return code of SUCCESS, or resume at the specified entry point address.`

Part of the PSCI Client (caller) responsibility is to ensure that only appropriate wakeup events are enabled. So if there is a pending interrupt due to unexpected interrupt, it is wrong behaviour from client side. The PSCI implementation shouldn't be patched up to make up for non-compliant clients.

vwadekar: @soby-mathew 

Actually, the linux kernel expects an error code indicating suspend abort. But TF-A does not return any error at this point in the suspend sequence.

One other problem with the PSCI implementation in TF-A, is that it assumes all pending interrupts are wake sources. Tegra at least has a limited set of interrupts marked as wake sources. So platforms should be allowed to decide if the pending interrupt should abort the suspend sequence.

soby-mathew: > Actually, the linux kernel expects an error code indicating suspend abort.
Hmm, As far as I can see the `IDLE` framework in Linux does not have any assumption that the suspend request cannot return early due to a pending interupt request. Could you give more details regarding this ?

If there are some expectations from Linux as you say, this needs to be fixed in Linux as this contradicts the PSCI specification.

> One other problem with the PSCI implementation in TF-A, is that it assumes all pending interrupts are wake sources..
As I highlighted earlier, this is usually the case. The onus is on the PSCI client to enable only the required wake-up sources.

vwadekar: > If there are some expectations from Linux as you say, this needs to be fixed in Linux as this contradicts the PSCI specification.

The psci wrapper (drivers/firmware/psci.c) has code to check for error cases. But since the wake in this case does not return an error, the kernel treats this as a successful suspend/resume.

> As I highlighted earlier, this is usually the case. The onus is on the PSCI client to enable only the required wake-up sources.

I think the PSCI code in TF-A should allow platforms to make this choice, instead of treating all interrupts as wake sources.

soby-mathew:
@vwadekar , We intend to close this PR this week as there is no action pending on this. Let me know if you think otherwise.

vwadekar:
I agree. We can close this one.

PR #1566, opened 2018-09-07
ARM Platforms:Enable non-secure access to UART1

EvanLloyd:
Adds an undocumented build option that enables non-secure access to
the PL011 UART1.
This allows a custom build where the UART can be used as a serial debug
port for WinDbg (or other debugger) connection.

This option is not documented in the user guide, as it is provided as a
convenience for Windows debugging, and not intended for general use.
In particular, enabling non-secure access to the UART might allow
a denial of service attack!

Change-Id: I4cd7d59c2cac897cc654ab5e1188ff031114ed3c
Signed-off-by: Alexei Fedorov <Alexei.Fedorov@arm.com>
Signed-off-by: Evan Lloyd <evan.lloyd@arm.com>


soby-mathew:
Hi Evan, PRs should be targeted to integration branch. I have changed it.

PR #1569, opened 2018-09-10
CSS: Fix overrun if system power level is not available

soby-mathew:
This patch fixes an array overrun in CSS scmi driver if the
system power domain level is less than 2. This was reported from
https://scan.coverity.com/projects/arm-software-arm-trusted-firmware

CID 308492

Change-Id: I3a59c700490816718d20c71141281f19b2b7e7f7


soby-mathew:
Coverity flags up error for using function like macro (assert()). Otherwise all other tests pass.

PR #1572, opened 2018-09-11
Allow setting log level back to compile time value

iq250vip:
When using the tf_log_set_max_level() function, one can dynamically
set the log level to a value smaller than then compile time specified
one, but not equal. This means that when the log level have been
lowered, it can't be reset to the previous value. This commit modifies
this function to allow setting the log level back to the compile time
value.

Fixes ARM-software/tf-issues#624

Change-Id: Ib157715c8835982ce4977ba67a48e18ff23d5a61
Signed-off-by: Junhan Zhou <Junhan@mellanox.com>


dp-arm:
jenkins: test this please

PR #1573, opened 2018-09-16
psci: allow the platform to validate the power operation early

ldts:
This hooks allows any platform to implement an early validation of
every single PSCI request.

For instance this could be used to prevent any CPUs from requesting
suspend entry or cpu offlinining.

Signed-off-by: Jorge Ramirez-Ortiz <jramirez@baylibre.com>

### DIFF
--- a/docs/platform-migration-guide.rst
+++ b/docs/platform-migration-guide.rst
@@ -155,6 +155,7 @@ for the ``plat_psci_ops`` structure which is declared as :
         void (*system_reset)(void) __dead2;
         int (*validate_power_state)(unsigned int power_state,
                         psci_power_state_t *req_state);
+        int (*validate_power_operation)(uint32_t smc_fid);
         int (*validate_ns_entrypoint)(unsigned long ns_entrypoint);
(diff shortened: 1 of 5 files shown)


antonio-nino-diaz-arm:
Just out of curiosity, what were you trying to do with this?

ldts:
just following the renesas code...this was aimged at preventing the boot cpu from being hotplug (something stupid since the kernel assumes that calls to cpu off must always succeed); not sure why renesas have this hack in place, is useless.  in fact what they also do (and this does owrk) is to hijack the call to opteed_cpu_migrate_info to return the boot cpu mpidr plus the non-migratable flag. Then when the user tries to request cpu off on the boot cpu the operation is rejected with operation not permited. so need to figure out how to do this in a proper way now; to start with opteed_cpu_migrate_info should query the TrustedOS instead of returning that the os is migratable by default

antonio-nino-diaz-arm:
Ok, thanks for the explanation.

ldts:
that patch might still be applicable with for system suspend though...definitively not for cpu_off.

PR #1574, opened 2018-09-17
psci: early op validation and migrate info

ldts:
The following pull request implements two different hooks
1. early platform validation: allows the platform to deny PSCI operations early in the process. The hook is used in the system suspend scenario since there are platforms (rcar-gen3) that can only enter suspend from the boot cpu (whichever that might happen to be).
2. platform hook for migrate info: current migrate capabilities fully depend on the SPD. In the case of OPTEED, there are no migrate restrictions (so all CPUs could be hotplugged). This creates a problem for platforms in which some CPUs cant be hotplugged (rcar-gen3 for instance can't hotplug the boot cpu). with this commit a platform hook is provided to allow the platform to communicate its restrictions when the SPD has none.

### DIFF
--- a/include/lib/psci/psci.h
+++ b/include/lib/psci/psci.h
@@ -326,6 +326,7 @@ typedef struct plat_psci_ops {
 	int (*write_mem_protect)(int val);
 	int (*system_reset2)(int is_vendor,
 				int reset_type, u_register_t cookie);
+	int (*migrate_info)(u_register_t *mpidr);
 } plat_psci_ops_t;
 
 /*******************************************************************************
@@ -333,6 +334,7 @@ typedef struct plat_psci_ops {
  ******************************************************************************/
 unsigned int psci_version(void);
 int psci_op_allowed(uint32_t smc_fid);
+int psci_migrate_info(u_register_t *mpidr);
 int psci_cpu_on(u_register_t target_cpu,
 		uintptr_t entrypoint,
 		u_register_t context_id);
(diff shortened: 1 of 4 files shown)


ldts:
client code example using these new platform hooks:
https://github.com/BayLibre/arm-trusted-firmware/commit/5e3a5ce48155d67ec65f99b0e6a34debc7205c29#diff-d2f2a934de6c551fc84ee6e13e7a2327R405

soby-mathew:
Just to understand better,  essentially the boot CPU cannot be `offlined`. One of the criteria for SYSTEM_SUSPEND call is all the other CPUs other than the one issuing the call must be `offline` (See [here](https://github.com/ARM-software/arm-trusted-firmware/blob/master/lib/psci/psci_main.c#L159)).  So, if boot CPU cannot be offlined, then it naturally follows that SYSTEM_SUSPEND must be issued from that CPU. 

So I think satisfying the 2nd point in your description will implicitly satisfy the 1st point.

Regarding the hotplug restriction on the boot CPU, is this originating due to migration restrictions of the Trusted OS ? If the platform is built without a Trusted OS (i.e SPD=none) , does it still have the hotplug restriction on Boot CPU ?

ldts:
hello @soby-mathew: yes agreed, so only the second commit might be necessary.
it is unclear whether the hotplug restriction is due to the TrustedOS but I think we can safely go with that assumption since the boot cpu can be selected using switches on some boards.  So yes, we cant hotplug the CPU that boots the TrustedOS,

PR #1576, opened 2018-09-18
BL31: Fix warning about BL32 init function

antonio-nino-diaz-arm:
The expected value for failure is 0, so the warning only has to be shown in that case. This is the way the TSPD has done it since it was introduced, and the way SPM and OP-TEE do it.

Trusty wrongly returns 0 on success.

In the case of TLK, the return value of tlkd_init() is passed from the secure world in register X1 when calling the SMC TLK_ENTRY_DONE.

### DIFF
--- a/bl31/bl31_main.c
+++ b/bl31/bl31_main.c
@@ -108,10 +108,8 @@ void bl31_main(void)
 
 		int32_t rc = (*bl32_init)();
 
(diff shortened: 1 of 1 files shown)


soby-mathew:
All checks fine except the coverity MISRA check which complains about function like macro being used.

PR #1577, opened 2018-09-18
trusty: Fix return value of trusty_init()

antonio-nino-diaz-arm:
The value used to signal failure is 0. It is needed to return a different value on success.


antonio-nino-diaz-arm:
For consistency with https://github.com/ARM-software/arm-trusted-firmware/pull/1576

PR #436, opened 2015-11-20
Mediatek soc v3

mtk09422:
Fix compilation error

plat/mediatek/board/common/param_init.c:44:13: error: ISO C forbids empty initializer braces [-Werror=pedantic]
} bucks[] = {
^
plat/mediatek/board/common/param_init.c:44:3: error: zero or negative size array 'bucks'
} bucks[] = {
^
cc1: all warnings being treated as errors
make: *** [build/mt8173/debug/bl31/param_init.o] Error 1

PR #384, opened 2015-09-04
Some fixes for suspend and reboot issue

mtk09422:
- update suspend pcm code
- support big core buck/PLL control in system suspend flow
- fix watchdog register setting

PR #360, opened 2015-08-11
Tegra: fix PLATFORM_{MAX_AFFLVL|CORE_COUNT|NUM_AFFS} macros

vwadekar:
Tegra: fix PLATFORM_{CORE_COUNT|NUM_AFFS} macros

This patch fixes the following macros for Tegra SoCs.

* PLATFORM_CORE_COUNT: PLATFORM_CLUSTER_COUNT * PLATFORM_MAX_CPUS_PER_CLUSTER
* PLATFORM_NUM_AFFS: PLATFORM_CORE_COUNT + PLATFORM_CLUSTER_COUNT + 1

Signed-off-by: Varun Wadekar <vwadekar@nvidia.com>

PR #1582, opened 2018-09-21
rcar_gen3: initial support

ldts:
Reference code:
==============

rar_gen3: IPL and Secure Monitor Rev1.0.22
https://github.com/renesas-rcar/arm-trusted-firmware [rcar_gen3]

commit d4a607b7ea3886d8d1098192d2c657face631202
Author: Takuya Sakata <takuya.sakata.wz@bp.renesas.com>
Date:   Thu Aug 30 21:26:41 2018 +0900
	Update IPL and Secure Monitor Rev1.0.22

General Information:
===================

This port has been tested on the Salvator-X Soc_id r8a7795 revision
ES1.1 (uses an SPD) and on the Starter Kit Pro (M3ULCB)

Build Tested:
-------------
ATFW_OPT="LSI=H3 RCAR_DRAM_SPLIT=1 RCAR_LOSSY_ENABLE=1"
MBEDTLS_DIR=$mbedtls

$ make clean bl2 bl31 rcar PLAT=rcar ${ATFW_OPT} SPD=opteed

Other dependencies:
------------------
* mbed_tls:
  git@github.com:ARMmbed/mbedtls.git [devel]

  commit 552754a6ee82bab25d1bdf28c8261a4518e65e4d
  Merge: 68dbc94 f34a4c1
  Author: Simon Butcher <simon.butcher@arm.com>
  Date:   Thu Aug 30 00:57:28 2018 +0100

* optee_os:
  https://github.com/BayLibre/optee_os

  Until it gets merged into OP-TEE, the port requires Renesas' Trusted
  Environment with a modification to support power management.
  commit 80105192cba9e704ebe8df7ab84095edc2922f84

  Author: Jorge Ramirez-Ortiz <jramirez@baylibre.com>
  Date:   Thu Aug 30 16:49:49 2018 +0200
    plat-rcar: cpu-suspend: handle the power level
    Signed-off-by: Jorge Ramirez-Ortiz <jramirez@baylibre.com>

* u-boot:
  The port has beent tested using mainline uboot.

  commit 4cdeda511f8037015b568396e6dcc3d8fb41e8c0
  Author: Fabio Estevam <festevam@gmail.com>
  Date:   Tue Sep 4 10:23:12 2018 -0300

* linux:
  The port has beent tested using mainline kernel.

  commit 7876320f88802b22d4e2daf7eb027dd14175a0f8
  Author: Linus Torvalds <torvalds@linux-foundation.org>
  Date:   Sun Sep 16 11:52:37 2018 -0700
      Linux 4.19-rc4
   
    Overview
    ---------
    
    BOOTROM starts the cpu at EL3; In this port BL2 will therefore be entered
    at this exception level (the Renesas' ATF reference tree [1] resets into
    EL1 before entering BL2 - see its bl2.ld.S)
    
    BL2 initializes DDR (and i2c to talk to the PMIC on some platforms)
    before determining the boot reason (cold or warm).
    
    During suspend all CPUs are switched off and the DDR is put in
    backup mode (some kind of self-refresh mode). This means that BL2 is
    always entered in a cold boot scenario.
    
    Once BL2 boots, it determines the boot reason, writes it to shared
    memory (BOOT_KIND_BASE) together with the BL31 parameters
    (PARAMS_BASE) and jumps to BL31.
    
    To all effects, BL31 is as if it is being entered in reset mode since
    it still needs to initialize the rest of the cores; this is the reason
    behind using direct shared memory access to  BOOT_KIND_BASE and
    PARAMS_BASE instead of using registers to get to those locations (see
    el3_common_macros.S and bl31_entrypoint.S for the RESET_TO_BL31 use
    case).
    
    Depending on the boot reason BL31 initializes the rest of the cores:
    in case of suspend, it uses a MBOX memory region to recover the
    program counters.
    
    [1] https://github.com/renesas-rcar/arm-trusted-firmware
    Tests
    -----
    * cpuidle
       ----------
       use a cpuidle enabled kernel tree [*]
       configure the kernel with the arm_idle default driver
       boot the board.
    
    * system suspend
      ----------------------
      $ cat suspend.sh
        #!/bin/bash
        i2cset -f -y 7 0x30 0x20 0x0F
        read -p "Switch off SW23 and press return " foo
        echo mem > /sys/power/state
    
    * cpu hotplug:
      -----------------
      $ cat offline.sh
        #!/bin/bash
        nbr=$1
        echo 0 > /sys/devices/system/cpu/cpu$nbr/online
        printf "ONLINE:  " && cat /sys/devices/system/cpu/online
        printf "OFFLINE: " && cat /sys/devices/system/cpu/offline
    
      $ cat online.sh
        #!/bin/bash
        nbr=$1
        echo 1 > /sys/devices/system/cpu/cpu$nbr/online
        printf "ONLINE:  " && cat /sys/devices/system/cpu/online
        printf "OFFLINE: " && cat /sys/devices/system/cpu/offline
    
[*] git://git.kernel.org/pub/scm/linux/kernel/git/horms/renesas-bsp.git


plat/renesas/rcar/qos/E3/qos_init_e3_v10_mstat390.h
```diff
@@ -0,0 +1,235 @@
+static uint64_t mstat_fix[] = {
```
antonio-nino-diaz-arm: Copyright header.

plat/renesas/rcar/qos/M3N/qos_init_m3n_v10_qoswt390.h
```diff
@@ -0,0 +1,235 @@
+static uint64_t qoswt_fix[] = {
```
antonio-nino-diaz-arm: Copyright header.

plat/renesas/rcar/qos/M3N/qos_init_m3n_v10_qoswt195.h
```diff
@@ -0,0 +1,235 @@
+static uint64_t qoswt_fix[] = {
```
antonio-nino-diaz-arm: Copyright header.

plat/renesas/rcar/qos/M3N/qos_init_m3n_v10_mstat390.h
```diff
@@ -0,0 +1,235 @@
+static uint64_t mstat_fix[] = {
```
antonio-nino-diaz-arm: Copyright header.

plat/renesas/rcar/qos/M3N/qos_init_m3n_v10_mstat195.h
```diff
@@ -0,0 +1,235 @@
+static uint64_t mstat_fix[] = {
```
antonio-nino-diaz-arm: Copyright header.

plat/renesas/rcar/qos/M3/qos_init_m3_v11_qoswt390.h
```diff
@@ -0,0 +1,219 @@
+static uint64_t qoswt_fix[] = {
```
antonio-nino-diaz-arm: Copyright header.

ldts:
ack

ldts:
@antonio-nino-diaz-arm  I still need to update all the copyright issues you mentioned (however I wanted to enable you to continue with the review, I know is a big pull request)

ldts:
updated the copyrights and a few other files. I'll now wait for the review before making any further changes.

ldts:
@soby-mathew yes I can move the change to optee_os (I wasnt sure if it would be a feature that might be useful to other platforms though); will do that .

sure will do two pull requests. everytime I rebase and force push to the pull request branch however there is a risk that we lose the history of this pull request (comments and all). is this your understanding as well?

AlexeiFedorov:
Hello,
When building for R-Car H3 board with RCAR_GEN3_ULCB=1, the following error is generated:
./build/rcar/debug/bl31/plat_pm.o: In function `rcar_system_reset':
L:\/plat/renesas/rcar/plat_pm.c:374: undefined reference to `cpld_reset_cpu'
L:\/plat/renesas/rcar/plat_pm.c:374:(.text.rcar_system_reset+0x8): relocation truncated to fit: R_AARCH64_CALL26 against undefined symbol `cpld_reset_cpu'
make: *** [Makefile:744: build/rcar/debug/bl31/bl31.elf] Error 1

This can be fixed by:
1.  In \drivers\renesas\rcar\cpld\ulcb_cpld.c change function name and add #if/#endif:

void rcar_cpld_reset_cpu(void)
to
#if (RCAR_GEN3_ULCB == 1)
void cpld_reset_cpu(void)
{
	cpld_init();

	cpld_write(CPLD_ADDR_RESET, 1);
}
#endif

2. In \plat\renesas\rcar\plat_pm.c add function declaration:
#if (RCAR_GEN3_ULCB == 1)
extern void cpld_reset_cpu(void);
#endif

Alexei.

ldts:
thanks. will fix. did you test on that board?

AlexeiFedorov:
Just built the image, will flash in a moment.

ldts:
ok cool. I am in IRCs #trusted-firmware-a in case you want to discuss there

ldts:
that is rebased on Renesa's 1.0.22 and containing @AlexeiFedorov bug report fixed

AlexeiFedorov:
OK. The image boots U-boot & Linux. I'm on IRC channel

ldts:
cool. did you try suspend and cpu hotplug? to make sure psci is working?

soby-mathew:
Hi @ldts,
> everytime I rebase and force push to the pull request branch however there is a risk that we lose the history of this pull request (comments and all). is this your understanding as well?

Yes, that's my understanding too. Github will try to preserve the comments, sometimes without the code context incase the code has undergone drastic restructuring.

drivers/renesas/rcar/qos/M3/qos_init_m3_v11.c
```diff
@@ -0,0 +1,243 @@
...
+		}
+		for (i = 0U; i < ARRAY_SIZE(qoswt_be); i++) {
+			io_write_64(QOSWT_BE_WTQOS_BANK0 + i * 8, qoswt_be[i]);
+			io_write_64(QOSWT_BE_WTQOS_BANK1 + i * 8, qoswt_be[i]);
+		}
+#endif /* RCAR_REWT_TRAINING != RCAR_REWT_TRAINING_DISABLE */
+	}
+
+	/* AXI-IF arbitration setting */
+	//delete
```
antonio-nino-diaz-arm: Delete?

drivers/renesas/rcar/qos/qos_init.c
```diff
@@ -0,0 +1,324 @@
...
+uint32_t qos_init_ddr_ch;
+uint8_t qos_init_ddr_phyvalid;
+
+#endif
+
+#define PRR_PRODUCT_ERR(reg)	do{\
+				ERROR("LSI Product ID(PRR=0x%x) QoS "\
+				"initialize not supported.\n",reg);\
+				panic();\
+				}while(0)
```
antonio-nino-diaz-arm: In here you need to add spaces to follow the Linux coding style.

drivers/renesas/rcar/pwrc/call_sram.S
```diff
@@ -0,0 +1,89 @@
...
+.global rcar_pwrc_switch_stack
+.global rcar_pwrc_save_generic_timer
+.global rcar_pwrc_restore_generic_timer
+
+#define OFFSET_SP_X9_X10	(0x00)
+#define OFFSET_CNTFID0		(0x10)
+#define OFFSET_CNTPCT_EL0	(0x18)
+#define OFFSET_TIMER_COUNT	(0x20)
+
+/*
```
antonio-nino-diaz-arm: Fix comment style.

drivers/renesas/rcar/pwrc/pwrc.c
```diff
@@ -0,0 +1,770 @@
...
+	if (!IS_A53A57(c) && cm != 0) {
+		ret = RCAR_INVALID;
+		goto done;
+	}
+
+	reg = mmio_read_32(RCAR_PRR);
+	cpu = mpidr & MPIDR_CPU_MASK;
+
+	if (IS_CA53(c))
+		if (reg & 1 << (STATE_CA53_CPU + cpu))
```
antonio-nino-diaz-arm: Maybe it's worth adding parentheses to the right operand of ``&``.

drivers/renesas/rcar/pwrc/pwrc.h
```diff
@@ -0,0 +1,80 @@
...
+void rcar_pwrc_init_suspend_to_ram(void);
+void rcar_pwrc_suspend_to_ram(void);
+#endif
+
+extern void rcar_pwrc_save_generic_timer(uint64_t * rcar_stack_generic_timer);
+extern uint32_t rcar_pwrc_switch_stack(uintptr_t jump, uintptr_t stack,
+				       void *arg);
+
+extern uint64_t rcar_stack_generic_timer[5];
+extern uint8_t __system_ram_start__[1];
```
antonio-nino-diaz-arm: You can use the macro ``IMPORT_SYM`` for ``__system_ram_start__``, ``__SRAM_COPY_START__`` and ``__system_ram_end__``.

drivers/renesas/rcar/io/io_private.h
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef IO_PRIVATE_H_
+#define IO_PRIVATE_H_
+
+/* Return codes reported by 'io_*' APIs
+ * The value of fail should not overlap with define of the errno.
+ * The errno is in "include/lib/stdlib/sys/errno.h". */
```
antonio-nino-diaz-arm: The ``*/`` needs to be in the following line, and the first line should be empty.

common/bl_common.c
```diff
@@ -36,7 +36,7 @@ void dyn_disable_auth(void)
 /******************************************************************************
  * Function to determine whether the authentication is disabled dynamically.
  *****************************************************************************/
-static int dyn_is_auth_disabled(void)
+static int __attribute__((unused)) dyn_is_auth_disabled(void)
```
antonio-nino-diaz-arm: Use ``__unused`` instead.

drivers/renesas/rcar/scif/scif.S
```diff
@@ -0,0 +1,325 @@
...
+	strh	w1, [x0, #SCIF_DL]
+	mov	w1, #CKS_INIT_DATA
+	strh	w1, [x0, #SCIF_CKS]
+#endif
+	/* 1-bit interval elapsed */
+	mov	w1, #100
+1:
+	subs	w1, w1, #1
+	cbnz	w1, 1b
+	/* Set bits RTRG[1:0], TTRG[1:0], and MCE in SCFCR, and clear bits
```
antonio-nino-diaz-arm: Fix comment.

make_helpers/build_macros.mk
```diff
@@ -414,8 +421,12 @@ $(BIN): $(ELF)
 	@echo "Built $$@ successfully"
 	@${ECHO_BLANK_LINE}
 
+$(SREC): $(ELF)
+	@echo "  SREC    $$@"
+	$$(Q)$$(OC) -O srec $$< $$@
+
 .PHONY: bl$(1)
-bl$(1): $(BIN) $(DUMP)
+bl$(1): $(SREC) $(BIN) $(DUMP)
```
antonio-nino-diaz-arm: We can keep the SREC rules here, but the dependency of BL1 on SREC is specific to your platform and it should be in ``platform.mk``.

drivers/renesas/rcar/ddr/ddr_a/ddr_init_e3.c
```diff
@@ -0,0 +1,1657 @@
+/*
```
antonio-nino-diaz-arm: Missing SPDX identifier.

drivers/renesas/rcar/ddr/ddr_b/boot_init_dram.c
```diff
@@ -0,0 +1,4467 @@
...
+	/* SCFCTST1 WR-RD */
+	tmp[2] =
+	    1UL * (mmio_read_32(DBSC_DBTR(12)) & 0xff) * 800 * ddr_mbpsdiv /
+	    ddr_mbps;
+	/* SCFCTST1 ACT-RD/WR */
+	tmp[1] = 1UL * js2[JS2_tRCD] * 800 * ddr_mbpsdiv / ddr_mbps;
+	/* SCFCTST1 ASYNCOFS */
+	tmp[0] = 12;
+	mmio_write_32(DBSC_SCFCTST1,
+		      (tmp[3] << 24) | (tmp[2] << 16) | (tmp[1]
```
antonio-nino-diaz-arm: Improve this comment saying that it can be enabled or disabled. Right now it looks like it's leftover code.

antonio-nino-diaz-arm: ?

plat/renesas/rcar/aarch64/plat_helpers.S
```diff
@@ -0,0 +1,397 @@
...
+	 */
+	ldr	x10, =MBOX_BASE
+	bl	platform_get_core_pos
+	lsl	x0, x0, #CACHE_WRITEBACK_SHIFT
+				/* because mailbox size is 0x40 byte per cpu	*/
+	ldr	x0, [x10, x0]	/* top address of mailbox is entrypoint		*/
+	cbz	x0, _panic	/* If warm boot and entrypoint is 0 then panic	*/
+exit:
+	ret	x9
+_panic:	b	_panic
```
antonio-nino-diaz-arm: Can't you call ``do_panic`` here instead of having an infinite loop?

plat/renesas/rcar/aarch64/platform_common.c
```diff
@@ -0,0 +1,265 @@
...
+					BL32_LIMIT - BL32_BASE,		\
+					MT_MEMORY | MT_RW | MT_SECURE)
+#endif
+
+/*
+ * Table of regions for various BL stages to map using the MMU.
+ * This doesn't include TZRAM as the 'mem_layout' argument passed to
+ * configure_mmu_elx() will give the available subset of that,
+ */
+#if IMAGE_BL1
```
antonio-nino-diaz-arm: Are you using BL1? I thought this only uses BL2 and BL31.

plat/renesas/rcar/aarch64/platform_common.c
```diff
@@ -0,0 +1,265 @@
...
+	MAP_ATFW_LOG,
+	MAP_DEVICE_RCAR,
+	MAP_DEVICE_RCAR2,
+	MAP_SRAM,
+	MAP_SRAM_STACK,
+	{0}
+};
+#endif
+
+#if IMAGE_BL32
```
antonio-nino-diaz-arm: Is this enough for OP-TEE?

plat/renesas/rcar/aarch64/platform_common.c
```diff
@@ -0,0 +1,265 @@
...
+	unsigned int freq;
+
+	freq = mmio_read_32(ARM_SYS_CNTCTL_BASE + CNTFID_OFF);
+	if (freq == 0)
+		panic();
+
+	return freq;
+}
+
+void plat_arm_gic_init(void)
```
antonio-nino-diaz-arm: ``plat_rcar_gic_init``

plat/renesas/rcar/aarch64/platform_common.c
```diff
@@ -0,0 +1,265 @@
...
+	ARM_IRQ_SEC_RPC,	/* 70           */
+	ARM_IRQ_SEC_TIMER,	/* 166          */
+	ARM_IRQ_SEC_TIMER_UP,	/* 171          */
+	ARM_IRQ_SEC_WDT,	/* 173          */
+	ARM_IRQ_SEC_CRYPT,	/* 102          */
+	ARM_IRQ_SEC_CRYPT_SecPKA,	/* 97           */
+	ARM_IRQ_SEC_CRYPT_PubPKA	/* 98           */
+#endif
+};
+
```
antonio-nino-diaz-arm: You'll need to migrate this to the new arrays as done in https://github.com/ARM-software/arm-trusted-firmware/pull/1587

plat/renesas/rcar/bl2_plat_setup.c
```diff
@@ -0,0 +1,749 @@
...
+#endif
+
+#define BL2_RO_BASE (unsigned long)(&__RO_START__)
+#define BL2_RO_LIMIT (unsigned long)(&__RO_END__)
+
+CASSERT((PARAMS_BASE + sizeof(bl2_to_bl31_params_mem_t) + 0x100)
+	 < (RCAR_SHARED_MEM_BASE + RCAR_SHARED_MEM_SIZE),
+	assert_bl31_params_do_not_fit_in_shared_memory);
+
+static meminfo_t bl2_tzram_layout __attribute__ ((aligned(PLATFORM_CACHE_LINE_SIZE)));
```
antonio-nino-diaz-arm: Use ``__aligned``

plat/renesas/rcar/bl2_plat_setup.c
```diff
@@ -0,0 +1,749 @@
...
+extern void rcar_io_emmc_setup(void);
+extern void rcar_io_setup(void);
+extern void rcar_swdt_release(void);
+extern void rcar_swdt_init(void);
+extern void rcar_rpc_init(void);
+extern void rcar_pfc_init(void);
+extern void rcar_dma_init(void);
+
+#if USE_COHERENT_MEM
+#define BL2_COHERENT_RAM_BASE  (unsigned long)(&__COHERENT_RAM_START__)
```
antonio-nino-diaz-arm: Use ``IMPORT_SYM``.

plat/renesas/rcar/bl2_plat_setup.c
```diff
@@ -0,0 +1,749 @@
...
+
+	rcar_swdt_release();
+	bl2_system_cpg_init();
+
+#if RCAR_BL2_DCACHE == 1
+	/* Disable data cache (clean and invalidate) */
+	val = read_sctlr_el1() & ~(SCTLR_C_BIT | SCTLR_M_BIT);
+	write_sctlr_el1(val);
+	dcsw_op_all(DCCISW);
+	tlbivmalle1();
```
antonio-nino-diaz-arm: Maybe you can use ``disable_mmu_el1`` here instead the 4 lines here.

plat/renesas/rcar/include/rcar_private.h
```diff
@@ -0,0 +1,107 @@
...
+#define RCAR_PRIVATE_H__
+
+#include <bakery_lock.h>
+#include <bl_common.h>
+#include <cpu_data.h>
+#include <platform_def.h>
+
+typedef volatile struct mailbox {
+	unsigned long value
+	    __attribute__ ((__aligned__(CACHE_WRITEBACK_GRANULE)));
```
antonio-nino-diaz-arm: Use ``__aligned``.

plat/renesas/rcar/plat_pm.c
```diff
@@ -0,0 +1,449 @@
...
+
+uint64_t rcar_stack_generic_timer[5] __attribute__ ((section("data")));
+unsigned int rcar_get_sys_suspend_power_state(void);
+extern void rcar_pwrc_restore_generic_timer(uint64_t *stack);
+extern u_register_t rcar_boot_mpidr;
+#if (RCAR_GEN3_ULCB == 1)
+extern void rcar_cpld_reset_cpu(void);
+#endif
+
+
```
antonio-nino-diaz-arm: This file is using the deprecated APIs and relies on the PSCI compatibility layer. Unfortunately, it has been removed in  https://github.com/ARM-software/arm-trusted-firmware/pull/1587, so you will need to port it to the new interfaces...

plat/renesas/rcar/platform.mk
```diff
@@ -0,0 +1,378 @@
...
+#
+
+ARM_CCI_PRODUCT_ID      	:= 500
+BL2_AT_EL3                      := 1
+RESET_TO_BL31                   := 1
+PROGRAMMABLE_RESET_ADDRESS      := 0
+COLD_BOOT_SINGLE_CPU            := 0
+TRUSTED_BOARD_BOOT	        := 1
+GENERATE_COT		        := 1
+ERROR_DEPRECATED	        := 1
```
antonio-nino-diaz-arm: Remove ``ERROR_DEPRECATED=1``.

common/bl_common.c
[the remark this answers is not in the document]
antonio-nino-diaz-arm: Oh, I've left a comment here because I didn't see this comment. Yes, @filiperinaldi is right.

drivers/renesas/rcar/board/board.c
```diff
@@ -0,0 +1,88 @@
...
+				      0xFFU, 0xFFU, 0xFFU, 0xFFU},
+		[BOARD_KRIEK] = {0x10U, 0xFFU, 0xFFU, 0xFFU,
+				 0xFFU, 0xFFU, 0xFFU, 0xFFU},
+		[BOARD_STARTER_KIT] = {0x10U, 0xFFU, 0xFFU, 0xFFU,
+				       0xFFU, 0xFFU, 0xFFU, 0xFFU},
+		[BOARD_SALVATOR_XS] = {0x10U, 0xFFU, 0xFFU, 0xFFU,
+				       0xFFU, 0xFFU, 0xFFU, 0xFFU},
+		[BOARD_EBISU] = {0x10U, 0xFFU, 0xFFU, 0xFFU,
+				 0xFFU, 0xFFU, 0xFFU, 0xFFU},
+		[BOARD_STARTER_KIT_PRE] = {0x10U, 0x10U, 0xFFU, 0xFFU,
```
filiperinaldi: Are the **two** first entries in the _KIT_PRE meant to be 0x10U? Just checking in case it is a typo and the second was meant to be 0x11 or 0x20.

drivers/renesas/rcar/board/board.h
```diff
@@ -0,0 +1,34 @@
...
+extern const char *g_board_tbl[];
+
+/************************************************************************
+ * Revisions are expressed in 8 bits.
+ *  The upper 4 bits are major version.
+ *  The lower 4 bits are minor version.
+ ************************************************************************/
+#define GET_BOARD_MAJOR(a)	((uint32_t)(a) >> 0x4)
+#define GET_BOARD_MINOR(a)	((uint32_t)(a) &  0xF)
+#define GET_BOARD_NAME(a)	(g_board_tbl[(a)])
```
filiperinaldi: Just a suggestion (could be a future improvement)... Add proper functions to retrieve version and name just like the one used to retrieve type. g_board_tbl[] can be local to board.c.
You may also pick relevant names for the parameters which I believe are: rev and type (?)

antonio-nino-diaz-arm:
IP Review +1

plat/renesas/rcar/platform.mk
```diff
@@ -0,0 +1,378 @@
...
+			drivers/renesas/rcar/emmc/emmc_read.c		\
+			drivers/renesas/rcar/emmc/emmc_cmd.c		\
+			drivers/renesas/rcar/watchdog/swdt.c		\
+			drivers/renesas/rcar/rom/rom_api.c		\
+			drivers/renesas/rcar/board/board.c		\
+			drivers/io/io_storage.c
+
+
+BL31_SOURCES	+=	${RCAR_GIC_SOURCES}				\
+			lib/cpus/aarch64/aem_generic.S			\
```
AlexeiFedorov: The file below is intended for FVP models:
lib/cpus/aarch64/aem_generic.S	
and should be removed from BL31_SOURCES

drivers/renesas/rcar/cpld/ulcb_cpld.c
```diff
@@ -0,0 +1,109 @@
...
+	gpio_set_value(GPIO_OUTDT6, SCLK, 0);
+	gpio_set_value(GPIO_OUTDT2, SSTBZ, 1);
+	gpio_set_value(GPIO_OUTDT6, MOSI, 0);
+
+	gpio_direction_output(GPIO_INOUTSEL6, SCLK);
+	gpio_direction_output(GPIO_INOUTSEL2, SSTBZ);
+	gpio_direction_output(GPIO_INOUTSEL6, MOSI);
+}
+
+void rcar_cpld_reset_cpu(void)
```
AlexeiFedorov: Function rcar_cpld_reset_cpu() is called only in (RCAR_GEN3_ULCB == 1) case,
see plat\renesas\rcar\plat_pm.c:
#if (RCAR_GEN3_ULCB == 1)
extern void rcar_cpld_reset_cpu(void);
#endif
...
#if (RCAR_GEN3_ULCB==1)
	rcar_cpld_reset_cpu();
#endif

So, in drivers\renesas\rcar\cpld\ulcb_cpld.c it should be also defined inside #if (RCAR_GEN3_ULCB == 1) / #endif  pair.

drivers/renesas/rcar/pwrc/pwrc.c
```diff
@@ -0,0 +1,754 @@
...
+#define	BIT_BKUP_CTRL_OUT			((uint8_t)(1U << 4))
+#define	PMIC_SLAVE_ADDR				(0x30U)
+#define	PMIC_BKUP_MODE_CNT			(0x20U)
+#define	PMIC_QLLM_CNT				(0x27U)
+#define	PMIC_RETRY_MAX				(100U)
+#endif
+#define	SCTLR_EL3_M_BIT				((uint32_t)1U << 0)
+#define	RCAR_CA53CPU_NUM_MAX			(4U)
+#define	RCAR_CA57CPU_NUM_MAX			(4U)
+#define IS_A53A57(c) 	(c == RCAR_CLUSTER_A53A57)
```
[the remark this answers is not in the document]
ldts: ok

drivers/renesas/rcar/pwrc/pwrc.c
```diff
@@ -0,0 +1,754 @@
...
+#include <xlat_tables_v2.h>
+#include "iic_dvfs.h"
+#include "micro_delay.h"
+#include "rcar_def.h"
+#include "rcar_private.h"
+#include "pwrc.h"
+
+/* Someday there will be a generic power controller api.
+ * At the moment each platform has its own pwrc so just exporting functions
+ * should be acceptable. */
```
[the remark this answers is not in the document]
ldts: ok

drivers/renesas/rcar/dma/dma_driver.c
```diff
@@ -0,0 +1,119 @@
...
+}
+
+void rcar_dma_exec(uintptr_t dst, uint32_t src, uint32_t len)
+{
+	uint32_t dma_len = len;
+
+	if (len & DMA_FRACTION_MASK)
+		dma_len = (len + DMA_SIZE_UNIT) & ~DMA_FRACTION_MASK;
+
+	if (((src + dma_len - 1U) < src) || ((src & DMA_FRACTION_MASK) != 0U)) {
```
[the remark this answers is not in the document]
ldts: yes, there is a check right below that would catch that: (dst + dma_len > DMA_DST_LIMIT) I'll clean up this code anyway.

drivers/renesas/rcar/qos/E3/qos_init_e3_v10.c
```diff
@@ -0,0 +1,162 @@
...
+
+#endif
+
+static void dbsc_setting(void)
+{
+	/* Register write enable */
+	io_write_32(DBSC_DBSYSCNT0, 0x00001234U);
+
+	/* BUFCAM settings */
+	io_write_32(DBSC_DBCAM0CNF1, 0x00043218);	//dbcam0cnf1
```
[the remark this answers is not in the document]
ldts: um I run ident on this code. I thought it would have taken care of this. my fault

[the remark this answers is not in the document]
ldts: these comments are useless anyway, will just remove

drivers/renesas/rcar/qos/E3/qos_init_e3_v10.c
```diff
@@ -0,0 +1,162 @@
...
+			io_write_64(QOSBW_FIX_QOS_BANK1 + i * 8, mstat_fix[i]);
+		}
+		for (i = 0U; i < ARRAY_SIZE(mstat_be); i++) {
+			io_write_64(QOSBW_BE_QOS_BANK0 + i * 8, mstat_be[i]);
+			io_write_64(QOSBW_BE_QOS_BANK1 + i * 8, mstat_be[i]);
+		}
+	}
+
+	/* AXI-IF arbitration setting */
+	//delete
```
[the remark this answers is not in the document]
ldts: ok

plat/renesas/rcar/bl2_plat_setup.c
ldts: ok

plat/renesas/rcar/bl2_plat_setup.c
ldts: ok. I think this also should be el3 since bl2 no longer starts at el1 now; so the dsb sy operation can replace the tlb and dcsw?

plat/renesas/rcar/include/rcar_private.h
ldts: ok

plat/renesas/rcar/platform.mk
ldts: ok

plat/renesas/rcar/bl2_plat_setup.c
antonio-nino-diaz-arm: Actually, do we really need this? Why is it needed to disable the data cache anyway? This flag is disabled in the makefile unless you manually set it to 1.

Which reminds me that you haven't actually created documentation in ``docs/plat`` and you haven't updated the file ``maintainers.rst``.

common/bl_common.c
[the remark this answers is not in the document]
ldts: ok

plat/renesas/rcar/platform.mk
ldts: @AlexeiFedorov were you going to send a patch (about the gic?))

ldts: ah the first issue it pops up when I disable the deprecated interfaces. anyway dont worry, will take care of it.

AlexeiFedorov: @ldts 
I can send 2 modified files platform_common.c & platform_def.h to you to take a look at.

ldts: No dont worry. that is fine. just doing the image load fixes next and that should be it for the deprecated interfaces.

plat/renesas/rcar/bl2_plat_setup.c
```diff
@@ -0,0 +1,749 @@
...
+void bl2_plat_set_bl33_ep_info(image_info_t * image, entry_point_info_t * ep)
+{
+	SET_SECURITY_STATE(ep->h.attr, NON_SECURE);
+	ep->spsr = SPSR_64(BL33_MODE, MODE_SP_ELX, DISABLE_ALL_EXCEPTIONS);
+#ifdef RCAR_BL33_ARG0
+	ep->args.arg0 = RCAR_BL33_ARG0;
+#endif
+}
+
+void bl2_init_generic_timer(void)
```
AlexeiFedorov: Function bl2_init_generic_timer() used to be called from bl2_reset.S in original Renesas ATF code, see:

func bl2_reset
...
	/* --------------------------------------------
	 * Enable the generic timer via memory mapped
	 * access.
	 * Enable it here quite early in the boot
	 * phase as long as we are still in EL3
	 * --------------------------------------------
	 */
	bl	bl2_init_generic_timer

bl2_reset.S is not present in this implementation and bl2_init_generic_timer() is not being called from anywhere, so this function should be removed from this source file.

plat/renesas/rcar/bl2_plat_setup.c
```diff
@@ -0,0 +1,749 @@
...
+	write_sctlr_el1(val);
+	dcsw_op_all(DCCISW);
+	tlbivmalle1();
+#endif
+}
+
+/*
+ * BL1 has passed the extents of the trusted SRAM that should be visible to BL2
+ * in x0. This memory layout is sitting at the base of the free trusted SRAM.
+ * Copy it to a safe loaction before its reclaimed by later BL2 functionality.
```
AlexeiFedorov: Please fix typos:
"loaction" -> location
"its" -> it's (it is)

drivers/renesas/rcar/board/board.c
```diff
@@ -0,0 +1,88 @@
...
+#define BOARD_CODE_SHIFT	(0x03)
+#define BOARD_ID_UNKNOWN	(0xFF)
+
+const char *g_board_tbl[] = {
+	[BOARD_SALVATOR_X] = "Salvator-X",
+	[BOARD_KRIEK] = "Kriek",
+	[BOARD_STARTER_KIT] = "Starter Kit",
+	[BOARD_SALVATOR_XS] = "Salvator-XS",
+	[BOARD_EBISU] = "Ebisu",
+	[BOARD_STARTER_KIT_PRE] = "Starter Kit",
```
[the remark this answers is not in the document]
filiperinaldi: Well, I think the original code was wrong. Note there is already a "Starter Kit" in line 31 for the non-Premium board. Also, the early firmware in the Premium board does print "Starter Kit Premium" so it is consistent for TF-A to also show the same board string.

plat/renesas/rcar/aarch64/plat_helpers.S
```diff
@@ -0,0 +1,397 @@
...
+
+	/* ---------------------------------------------
+	 * platform_is_primary_cpu
+	 *
+	 * ---------------------------------------------
+	 */
+func platform_is_primary_cpu
+	and	x0, x0, #MPIDR_CPU_MASK	/* Get the bit7-0 of MPIDR_EL1 -> x0	*/
+	cmp	x0, #RCAR_PRIMARY_CPU	/* x0 = 0 (primary CPU)?		*/
+	cset	x0, eq			/* If x0 is 0 carry bit set		*/
```
AlexeiFedorov: This instruction has nothing to do with carry bit, so the comment is incorrect.

plat/renesas/rcar/aarch64/plat_helpers.S
```diff
@@ -0,0 +1,397 @@
...
+	ret
+endfunc plat_report_exception
+
+	/* ---------------------------------------------
+	 * platform_is_primary_cpu
+	 *
+	 * ---------------------------------------------
+	 */
+func platform_is_primary_cpu
+	and	x0, x0, #MPIDR_CPU_MASK	/* Get the bit7-0 of MPIDR_EL1 -> x0	*/
```
AlexeiFedorov: This doesn't take in account MPIDR_CLUSTER_MASK, so Cortex-A53 #0 from the secondary cluster with its MPIDR 0x8000_0100 will be detected as a primary CPU.

plat/renesas/rcar/include/rcar_private.h
```diff
@@ -0,0 +1,107 @@
...
+ */
+void rcar_configure_mmu_el3(unsigned long total_base,
+			    unsigned long total_size,
+			    unsigned long ro_start, unsigned long ro_limit
+#if USE_COHERENT_MEM
+			    , unsigned long coh_start, unsigned long coh_limit
+#endif
+    );
+
+uint32_t rcar_get_spsr_for_bl32_entry(void);
```
AlexeiFedorov: These two functions 
rcar_get_spsr_for_bl32_entry()
and
rcar_get_spsr_for_bl33_entry() 
aren't defined anywhere in the source code, so their declarations should be removed.

plat/renesas/rcar/aarch64/plat_helpers.S
AlexeiFedorov: See **Boot** section of **H3/M3 Device Manual** regarding selection of a master boot processor through the MD7 and MD6 pins.
R-Car H3 Salvator-SX board supports selection between A57/A53 boot CPU using the board's DIP switch and in case of A53 boot the primary CPU #0 will have MPIDR=0x80000100.
Note that this switch is not present on H3/M3 Starter Kit boards, and boot CPU is A57 #0 with MPIDR=0x80000000.
For correct detection of primary CPU, please consider the following modifications:
1) in plat\renesas\rcar\include\rcar_def.h add:
`/* Cluster IDs read from MPIDR */`
`#define RCAR_CLUSTER_A57		0`
`#define RCAR_CLUSTER_A53		1`
2) Modify platform_is_primary_cpu function as:
`func platform_is_primary_cpu`
`    ldr	x1, =RCAR_MODEMR			/* Mode pin register */`
`    and	x0, x0, #(MPIDR_CLUSTER_MASK | MPIDR_CPU_MASK)`
`    ldr	x1, [x1]				/* Read Mode pin register */`
`    and	x1, x1, #MODEMR_BOOT_CPU_MASK		/* Detect master boot processor in MD[7:6] */`
`    mov	x2, #RCAR_CLUSTER_A53			/* x2 = A53 cluster #1 */`
`    cmp	x1, #MODEMR_BOOT_CPU_CA57		/* A57 boot cluster?   */`
`    mov	x1, #RCAR_CLUSTER_A57			/* x1 = A57 cluster #0 */`
`    csel	x1, x1, x2, eq				/* Select boot cluster */` 		
`    cmp	x0, x1, lsl #MPIDR_AFFINITY_BITS	/* Master boot processor #0 ? */`
`    cset	x0, eq					/* If true, x0 = 1, otherwise 0 */`
`    ret`
`endfunc platform_is_primary_cpu`

plat/renesas/rcar/bl2_plat_setup.c
ldts: ack

drivers/renesas/rcar/cpld/ulcb_cpld.c
ldts: ack

plat/renesas/rcar/bl2_plat_setup.c
ldts: ack

plat/renesas/rcar/aarch64/plat_helpers.S
ldts: ack

plat/renesas/rcar/aarch64/plat_helpers.S
ldts: yes however doing otherwise would mean that we  are breaking the dip switch user selectable booting cluster. Although I am having a hard time understanding how can the current code work (I don't have any documentation from Renesas) I haven't encountered any problems having the code boot this way so I propose - unless you are observing otherwise in your functional tests- that  the current code is kept as per Renesas' original code.

plat/renesas/rcar/aarch64/plat_helpers.S
ldts: yes I have seen this but wasnt this code removed from earlier Renesas's versions? did you find out why?

antonio-nino-diaz-arm: >  It's also interesting to note that the old code tries to detect if boot CPU is ARM R7 by execution AArch64 code.

That's certainly interesting...

ldts: @AlexeiFedorov um but whoever replaced that function for its current implementation made a conscious decision (I dont think it was a mistake). The current code 'somehow' works as expected so I still think we should merge the implementation coming from Renesas until we have further clarification or we have a bug report (as I said I cant make it fail on my Salvator-x, can you?)

AlexeiFedorov: @ldts. I'll have a look to work out why this code doesn't fail.

ldts: awesome. I am working on the documentation and instructions. this is the tree that I will use to update the pull request tomorrow after I finished with the docs. https://github.com/ldts/arm-trusted-firmware (branch rcar_gen3/devel)

drivers/renesas/rcar/cpld/ulcb_cpld.c
ldts: fixed it in the makefile

drivers/renesas/rcar/pwrc/pwrc.c
```diff
@@ -0,0 +1,771 @@
...
+	} sram = {
+	.base = (uintptr_t) DEVICE_SRAM_BASE,.len = DEVICE_SRAM_SIZE,};
+	struct ddr_code_t {
+		void *base;
+		size_t len;
+	} code = {
+	.base = (void *)&__SRAM_COPY_START__,.len =
+		    (uint64_t) __system_ram_end__ -
+		    (uint64_t) __system_ram_start__,};
+
```
antonio-nino-diaz-arm: This step is not needed if the SRAM is mapped initially as RW. You could just leave a comment instead of changing th attributes twice.

plat/renesas/rcar/aarch64/plat_helpers.S
AlexeiFedorov: @ldts 
Is it "branch **rcar_gen3/devel**" but not  **rcar_gen3/upstream**?

drivers/renesas/rcar/board/board.c
```diff
@@ -0,0 +1,87 @@
...
+#define BOARD_CODE_SHIFT	(0x03)
+#define BOARD_ID_UNKNOWN	(0xFF)
+
+const char *g_board_tbl[] = {
+	[BOARD_SALVATOR_X] = "Salvator-X",
+	[BOARD_KRIEK] = "Kriek",
+	[BOARD_STARTER_KIT] = "Starter Kit",
+	[BOARD_SALVATOR_XS] = "Salvator-XS",
+	[BOARD_EBISU] = "Ebisu",
+	[BOARD_STARTER_KIT_PRE] = "Starter Kit Premium",
```
AlexeiFedorov: @ldts 
The correct board's name is "Starter Kit Premier"

maintainers.rst
```diff
@@ -139,6 +139,14 @@ Raspberry Pi 3 platform port
 :F: docs/plat/rpi3.rst
 :F: plat/rpi3/
 
+Renesas rcar-gen3 platform port
+-------------------------------
+:M: Jorge Ramirez-Ortiz  <jramirez@baylibre.com>
+:G: `ldts`_
+:F: docs/plat/rcar-gen3.rst
+:F: plat/renesas/rcar
+:F: drivers/renesas/rcar
```
antonio-nino-diaz-arm: Add dummy_create here.

plat/renesas/rcar/bl2_plat_setup.c
```diff
@@ -0,0 +1,630 @@
...
+
+#include <desc_image_load.h>
+#include <arch_helpers.h>
+#include <bl_common.h>
+#include <bl1.h>
+#include <console.h>
+#include <debug.h>
+#include <mmio.h>
+#include <platform.h>
+#include <plat_arm.h>
```
antonio-nino-diaz-arm: Don't include plat_arm.h

plat/renesas/rcar/bl31_plat_setup.c
```diff
@@ -0,0 +1,122 @@
...
+#include <arch_helpers.h>
+#include <bl_common.h>
+#include <bl31.h>
+#include <cci.h>
+#include <console.h>
+#include <mmio.h>
+#include <platform.h>
+#include <stddef.h>
+#include <debug.h>
+#include <plat_arm.h>
```
antonio-nino-diaz-arm: Don't include plat_arm.

plat/renesas/rcar/plat_pm.c
```diff
@@ -0,0 +1,327 @@
...
+#include <bakery_lock.h>
+#include <bl_common.h>
+#include <cci.h>
+#include <debug.h>
+#include <errno.h>
+#include <gicv2.h>
+#include <mmio.h>
+#include <platform.h>
+#include <platform_def.h>
+#include <plat_arm.h>
```
antonio-nino-diaz-arm: Don't include plat_arm.

plat/renesas/rcar/platform.mk
```diff
@@ -0,0 +1,404 @@
...
+# Enable workarounds for selected Cortex-A57 erratas.
+ERRATA_A57_859972  := 1
+ERRATA_A57_813419  := 1
+
+include drivers/renesas/rcar/ddr/ddr.mk
+include drivers/renesas/rcar/qos/qos.mk
+include drivers/renesas/rcar/pfc/pfc.mk
+
+PLAT_INCLUDES	:=	-Iinclude/common/tbbr			\
+			-Iinclude/plat/arm/common		\
```
antonio-nino-diaz-arm: Don't include plat/arm

plat/renesas/rcar/bl31_plat_setup.c
```diff
@@ -0,0 +1,122 @@
...
+{
+	plat_rcar_gic_driver_init();
+	plat_rcar_gic_init();
+
+	/* enable the system level generic timer */
+	mmio_write_32(RCAR_CNTC_BASE + CNTCR_OFF, CNTCR_FCREQ(U(0)) | CNTCR_EN);
+
+	rcar_pwrc_setup();
+#if 0
+	/* todo: dynamic toplogy detection using rcar_pwrc */
```
AlexeiFedorov: "toplogy" -> topology

drivers/renesas/rcar/ddr/ddr_b/boot_init_dram.c
```diff
@@ -0,0 +1,4467 @@
...
+	tmp[3] =
+	    1UL * (mmio_read_32(DBSC_DBTR(11)) & 0xff) * 800 * ddr_mbpsdiv /
+	    ddr_mbps;
+	/* SCFCTST1 WR-RD */
+	tmp[2] =
+	    1UL * (mmio_read_32(DBSC_DBTR(12)) & 0xff) * 800 * ddr_mbpsdiv /
+	    ddr_mbps;
+	/* SCFCTST1 ACT-RD/WR */
+	tmp[1] = 1UL * js2[JS2_tRCD] * 800 * ddr_mbpsdiv / ddr_mbps;
+	/* SC
```
AlexeiFedorov: @ldts 
Please add '\n' at the end of line: 
"BL2: DDR%d(%s)\n"

AlexeiFedorov: Please add '\n' at the end of line.

AlexeiFedorov: the same

AlexeiFedorov: the same here

plat/renesas/rcar/bl31_plat_setup.c
antonio-nino-diaz-arm: Actually, this "setup topology" step is part of old PSCI, I don't think we need it at all. This function isn't even defined anywhere. You can remove it from here and from the header.

lib/psci/psci_main.c
```diff
@@ -44,6 +44,23 @@ int psci_cpu_on(u_register_t target_cpu,
 	return psci_cpu_on_start(target_cpu, &ep);
 }
 
+int psci_platform_migrate_info(u_register_t *mpidr)
```
soby-mathew: @ldts, as I commented earlier, these generic code changes need to be pulled out as a separate PR and we can discuss as how to implement your requirement for preventing CPU_OFF on primary CPU.

ldts:
@soby-mathew yes I can do that; however my concern wss that the rcar-gen3 will be broken (reboot will be needed) if the system enters suspend with cpu0 off. is that ok then?

ldts:
@antonio-nino-diaz-arm @soby-mathew ok will do that.

drivers/renesas/rcar/console/rcar_printf.c
```diff
@@ -0,0 +1,161 @@
...
+	uint8_t res[4];
+} loghead_t;
+
+typedef struct log_map {
+	loghead_t header;
+	uint8_t log_data[RCAR_BL31_LOG_MAX];
+	uint8_t res_data[RCAR_LOG_RES_SIZE];
+} logmap_t;
+
+void rcar_set_log_time(void)
```
AlexeiFedorov: This function is not called from anywhere and should be removed.

drivers/renesas/rcar/console/rcar_printf.h
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef	RCAR_PRINTF_H__
+#define	RCAR_PRINTF_H__
+
+#include <string.h>
+
+void rcar_set_log_time(void);
```
AlexeiFedorov: This function is not called from anywhere and its declaration should be removed.

drivers/renesas/rcar/board/board.c
```diff
@@ -0,0 +1,87 @@
...
+#define SXS_ID	{ 0x10U, 0xFFU, 0xFFU, 0xFFU, 0xFFU, 0xFFU, 0xFFU, 0xFFU }
+#define SX_ID	{ 0x10U, 0x11U, 0xFFU, 0xFFU, 0xFFU, 0xFFU, 0xFFU, 0xFFU }
+#define SKP_ID	{ 0x10U, 0x10U, 0xFFU, 0xFFU, 0xFFU, 0xFFU, 0xFFU, 0xFFU }
+#define SK_ID	{ 0x10U, 0xFFU, 0xFFU, 0xFFU, 0xFFU, 0xFFU, 0xFFU, 0xFFU }
+#define EB4_ID	{ 0x10U, 0xFFU, 0xFFU, 0xFFU, 0xFFU, 0xFFU, 0xFFU, 0xFFU }
+#define EB_ID	{ 0x10U, 0xFFU, 0xFFU, 0xFFU, 0xFFU, 0xFFU, 0xFFU, 0xFFU }
+#define KK_ID	{ 0x10U, 0xFFU, 0xFFU, 0xFFU, 0xFFU, 0xFFU, 0xFFU, 0xFFU }
+
+const char *g_board_tbl[] = {
+	[BOARD_STARTER_KIT_PRE] = "Starter Kit Premium",
```
AlexeiFedorov: The correct board's name is "Starter Kit Premier"

ldts:
updated after #1625 got merged

plat/renesas/rcar/bl2_plat_setup.c
```diff
@@ -0,0 +1,626 @@
...
+{
+	u_register_t *boot_kind = (void *) BOOT_KIND_BASE;
+	bl_mem_params_node_t *bl_mem_params;
+
+	if (image_id != BL31_IMAGE_ID)
+		return 0;
+
+	bl_mem_params = get_bl_mem_params_node(image_id);
+
+	if (is_ddr_backup_mode() == 0)
```
AlexeiFedorov: This should be changed to 
if (is_ddr_backup_mode() == RCAR_COLD_BOOT)

plat/renesas/rcar/bl2_plat_setup.c
```diff
@@ -0,0 +1,626 @@
...
+#endif
+	if (once)
+		return reason;
+
+	once = 1;
+	if ((mmio_read_32(GPIO_INDT) & GPIO_BKUP_TRG_SHIFT) == 0)
+		return reason;
+
+#if PMIC_ROHM_BD9571 && RCAR_SYSTEM_RESET_KEEPON_DDR
+	if (iic_dvfs_receive(PMIC, PMIC_REG_KEEP10, &data)) {
```
AlexeiFedorov: If RCAR_SYSTEM_RESET_KEEPON_DDR is set to 1, this code generates the error:
plat/renesas/rcar/bl2_plat_setup.c:2256: error: implicit declaration of function 'iic_dvfs_receive'; did you mean 'rcar_iic_dvfs_receive'? [-Werror=implicit-function-declaration]

iic_dvfs_receive() should be changed to rcar_iic_dvfs_receive()

plat/renesas/rcar/include/rcar_def.h
```diff
@@ -0,0 +1,295 @@
...
+#define MIDR_CA57			(0x0D07U << MIDR_PN_SHIFT)
+#define MIDR_CA53			(0x0D03U << MIDR_PN_SHIFT)
+/* for SuspendToRAM */
+#define	GPIO_BASE			(0xE6050000U)
+#define	GPIO_INDT1			(GPIO_BASE + 0x100CU)
+#define GPIO_INDT6			(GPIO_BASE + 0x540CU)
+#define	RCAR_COLD_BOOT			(0x00U)
+#define	RCAR_WARM_BOOT			(0x01U)
+#if PMIC_ROHM_BD9571 && RCAR_SYSTEM_RESET_KEEPON_DDR
+#define	PMIC			(0x30U)
```
AlexeiFedorov: When compiled with RCAR_SYSTEM_RESET_KEEPON_DDR = 1
This generates the error:
drivers/renesas/rcar/iic_dvfs/iic_dvfs.h:11:0: error: "PMIC" redefined [-Werror]
 #define PMIC   (0x30)
plat/renesas/rcar/include/rcar_def.h:269:0: note: this is the location of the previous definition
 #define PMIC   (0x30U)

drivers/renesas/rcar/pwrc/pwrc.c
```diff
@@ -0,0 +1,771 @@
...
+	if (rcar_iic_dvfs_send(PMIC, PMIC_BKUP_MODE_CNT, mode))
+		panic();
+#endif
+}
+
+void rcar_pwrc_suspend_to_ram(void)
+{
+#if RCAR_SYSTEM_RESET_KEEPON_DDR
+	rcar_pwrc_code_copy_to_system_ram();
+	if (rcar_iic_dvfs_send(SLAVE_ADDR_PMIC, REG_ADDR_REG_KEEP10, 0)) {
```
AlexeiFedorov: With  RCAR_SYSTEM_RESET_KEEPON_DDR =1 this code generates 3 compilation errors:
rivers/renesas/rcar/pwrc/pwrc.c:669:25: error: 'SLAVE_ADDR_PMIC' undeclared (first use in this function)
 if (rcar_iic_dvfs_send(SLAVE_ADDR_PMIC, REG_ADDR_REG_KEEP10, 0)) {
drivers/renesas/rcar/pwrc/pwrc.c:669:42: error: 'REG_ADDR_REG_KEEP10' undeclared (first use in this function); did you mean 'PMIC_REG_KEEP10'?
  if (rcar_iic_dvfs_send(SLAVE_ADDR_PMIC, REG_ADDR_REG_KEEP10, 0)) {
drivers/renesas/rcar/pwrc/pwrc.c:670:46: error: 'rc' undeclared (first use in this function)
   ERROR("Failed send KEEP10 init ret=%d \n", rc);

plat/renesas/rcar/aarch64/plat_helpers.S
```diff
@@ -0,0 +1,421 @@
...
+endfunc plat_report_exception
+
+	/* ---------------------------------------------
+	 * plat_is_my_cpu_primary
+	 *
+	 * ---------------------------------------------
+	 */
+func plat_is_my_cpu_primary
+	mrs	x0, mpidr_el1
+	and	x0, x0, #MPIDR_CPU_MASK
```
AlexeiFedorov: As I've already noted there is a bug in this function which doesn't take in account MPIDR_CLUSTER_MASK and will cause a secondary CPU #0 from cluster #1 to be detected as a primary CPU.
The only reason why this code doesn't cause a problem when booting CPU #0 from cluster #1, is that it is never executed for any secondary CPU. Please see detailed explanation below.
In _bl31\aarch64\bl31_entrypoint.S_ macro _el3_entrypoint_common_ is called as:
	el3_entrypoint_common					\
		_init_sctlr=1					\
		_warm_boot_mailbox=!PROGRAMMABLE_RESET_ADDRESS	\
		_secondary_cold_boot=!COLD_BOOT_SINGLE_CPU	\
		_init_memory=1					\
		_init_c_runtime=1				\
		_exception_vectors=runtime_exceptions
with _warm_boot_mailbox=1 and secondary_cold_boot=1, because _plat\renesas\rcar\platform.mk_ defines
PROGRAMMABLE_RESET_ADDRESS	:= 0
COLD_BOOT_SINGLE_CPU		        := 0
RESET_TO_BL31			                := 1
This will expand _el3_entrypoint_common_ macro in _include\common\aarch64\el3_common_macros.S_ as
	.if \_warm_boot_mailbox
		/* -------------------------------------------------------------
		 * This code will be executed for both warm and cold resets.
		 * Now is the time to distinguish between the two.
		 * Query the platform entrypoint address and if it is not zero
		 * then it means it is a warm boot so jump to this address.
		 * -------------------------------------------------------------
		 */
		bl	plat_get_my_entrypoint
		cbz	x0, do_cold_boot
		br	x0

	do_cold_boot:
	.endif /* _warm_boot_mailbox */
Because code in _plat_secondary_reset()_ (_plat\renesas\rcar\aarch64\plat_helpers.S_)
       	ldr x3, =BOOT_KIND_BASE
	mov x1, #0x1
	str x1, [x3]
will always set warm reset type (RCAR_WARM_BOOT = 1), _plat_get_my_entrypoint()_ which checks boot type in X1
	/* warm boot or cold boot */
	and	x1, x21, #1
	cmp	x1, #0
	bne	warm_reset
will return X0 value != 0 read from the mail box:
warm_reset:
	/* --------------------------------------------------------------------
	 * A per-cpu mailbox is maintained in the trusted SDRAM. Its flushed out
	 * of the caches after every update using normal memory so its safe to
	 * read it here with SO attributes
	 * ---------------------------------------------------------------------
	 */
	ldr	x10, =MBOX_BASE
	bl	platform_get_core_pos
	lsl	x0, x0, #CACHE_WRITEBACK_SHIFT
	ldr	x0, [x10, x0]
	cbz	x0, _panic

In that case _do_cold_boot_ branch won't be taken: 
		bl	plat_get_my_entrypoint
		cbz	x0, do_cold_boot
		br	x0
do_cold_boot:

and code which checks for primary CPU:
		bl	plat_is_my_cpu_primary
		cbnz	w0, do_primary_cold_boot

won't be executed for any secondary.  Because secondary CPUs are always execute warm boot path and only single CPU comes from cold reset , COLD_BOOT_SINGLE_CPU should be set to 1. This option is also documented in docs\user-guide.rst:

-  **``COLD_BOOT_SINGLE_CPU``: This option indicates whether the platform may
   release several CPUs out of reset. It can take either 0 (several CPUs may be
   brought up) or 1 (only one CPU will ever be brought up during cold reset).
   Default is 0. If the platform always brings up a single CPU, there is no
   need to distinguish between primary and secondary CPUs and the boot path can
   be optimised. The ``plat_is_my_cpu_primary()`` and
   ``plat_secondary_cold_boot_setup()`` platform porting interfaces do not need
   to be implemented in this case."**

ldts: But from what you are describing the warm path does just as expected - hence why I suppose there are no functional issues leaving system suspend, booting or hot-pluging cores. Is there a functional problem that we are trying to fix? 

Also do not forget that the cold path (ie, bl2) enters using exactly the same el3 macro and since only one core comes out of reset (from either cluster one or cluster two depending on a DIP switch setting), we need to be able to boot on any of those cases with the same TF-A; hence why we do not want to filter based on the cluster.

So yeah, the actual _functionality_ seems correct to me. Also worth noticing that if this was indeed a functional bug Renesas would have had many issues reported on their many platforms since this implementation is a 1:1 map of theirs (see [1])

As we discussed, I can of course see that there is a real semantic bug, that is for sure...I mean that function would identify as primary the same cpu on either core; however, for this platform, this seems to be intended.

[1] https://github.com/renesas-rcar/arm-trusted-firmware/blob/rcar_gen3/plat/renesas/rcar/aarch64/rcar_helpers.S#L220

ldts: and yes setting COLD_BOOT_SINGLE_CPU removes  the semantic issue and therefore the unnecessary check.  Much better like this. cool.

ldts: tests are ok. will push an update now

drivers/renesas/rcar/ddr/ddr_b/boot_init_dram.c
```diff
@@ -0,0 +1,4467 @@
+/*
```
soby-mathew: The checkpatch errors need to be corrected. Perhaps `--fix` can rectify most of the problems since they are whitespace issues.

drivers/renesas/rcar/console/rcar_console.S
```diff
@@ -0,0 +1,89 @@
+/*
```
soby-mathew: Need to use MULTI_CONSOLE_API for the console as the previous one is deprecated.

soby-mathew:
jenkins: test this please

drivers/renesas/rcar/console/rcar_console.S
ldts: can you check this again please? from the 2.0 release note I can read the following:

" All deprecated interfaces except the MULTI_CONSOLE_API have been removed from the code 
base."

drivers/renesas/rcar/ddr/ddr_b/boot_init_dram.c
ldts: I think it is unrealistic that I will be able to fix the ~1600 errors being reported (the different tools generate more errors themselves in a catch22 sequence). Would it be enough if I were to focus on the platform/ and skim over the drivers - particularly the larger ones like DDR ?

drivers/renesas/rcar/console/rcar_console.S
ldts: @soby-mathew I am not sure this API makes sense on the rcar  since rcar does not use any of the uarts available...and nobody will ever use the rcar uart. so could you maybe explain the benefits of this interface (and extra code) for platforms not using generic uarts please?

ldts:
addressed the srecord build and fixed the checkpatch errors in the platform

drivers/renesas/rcar/console/rcar_console.S
soby-mathew: Hi @ldts,
The old console API are deprecated in the sense they are marked for removal soon (see [here](https://github.com/ARM-software/arm-trusted-firmware/wiki/TF-A-Release-information#2removal-of-deprecated-interfaces)).  The expectation is all platforms will migrate to the new framework before removal.

There is a bit of extra code as a result of the new framework, but it allows piping the output to several console devices if required (in fact the console can be a memory buffer like the cbmem driver). It also allows easily switching the console depending on the state of the firmware like boot -> runtime -> crash.

We could merge the PR without the MUTLI_CONSOLE, but be aware that deprecated console is scheduled for removal soon.

plat/renesas/rcar/bl2_plat_setup.c
```diff
@@ -0,0 +1,626 @@
...
+cold_boot:
+	*boot_kind  = RCAR_COLD_BOOT;
+	flush_dcache_range(BOOT_KIND_BASE, sizeof(*boot_kind));
+
+	return 0;
+}
+
+int bl2_plat_handle_post_image_load(unsigned int image_id)
+{
+	static bl2_to_bl31_params_mem_t *params;;
```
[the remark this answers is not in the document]
ldts: ok

ldts:
[jramirez@igloo linux.git (master $%)]$ for i in /tmp/patches/*; do  echo "patch: $i"; scripts/checkpatch.pl $i | grep -A1 "ERROR: "; done
patch: /tmp/patches/0001-rcar-gen3-initial-commit-for-the-rcar-gen3-boards.patch
patch: /tmp/patches/0003-rcar_gen3-drivers-authentication.patch
patch: /tmp/patches/0004-rcar_gen3-drivers-avs-adaptive-voltage-scaling.patch
patch: /tmp/patches/0005-rcar_gen3-drivers-board-identification.patch
patch: /tmp/patches/0006-rcar_gen3-drivers-cpld.patch
patch: /tmp/patches/0007-rcar_gen3-drivers-micro-delay-generator.patch
patch: /tmp/patches/0008-rcar_gen3-drivers-dma.patch
patch: /tmp/patches/0009-rcar_gen3-drivers-emmc.patch
patch: /tmp/patches/0010-rcar_gen3-drivers-i2c-dvfs.patch
ERROR: need consistent spacing around '*' (ctx:WxV)
#115: FILE: drivers/renesas/rcar/iic_dvfs/iic_dvfs.c:96:
--
ERROR: need consistent spacing around '*' (ctx:WxV)
#274: FILE: drivers/renesas/rcar/iic_dvfs/iic_dvfs.c:255:
--
ERROR: need consistent spacing around '*' (ctx:WxV)
#297: FILE: drivers/renesas/rcar/iic_dvfs/iic_dvfs.c:278:
--
ERROR: need consistent spacing around '*' (ctx:WxV)
#320: FILE: drivers/renesas/rcar/iic_dvfs/iic_dvfs.c:301:
--
ERROR: need consistent spacing around '*' (ctx:WxV)
#360: FILE: drivers/renesas/rcar/iic_dvfs/iic_dvfs.c:341:
--
ERROR: need consistent spacing around '*' (ctx:WxV)
#384: FILE: drivers/renesas/rcar/iic_dvfs/iic_dvfs.c:365:
--
ERROR: need consistent spacing around '*' (ctx:WxV)
#410: FILE: drivers/renesas/rcar/iic_dvfs/iic_dvfs.c:391:
--
ERROR: need consistent spacing around '*' (ctx:WxV)
#436: FILE: drivers/renesas/rcar/iic_dvfs/iic_dvfs.c:417:
--
ERROR: need consistent spacing around '*' (ctx:WxV)
#459: FILE: drivers/renesas/rcar/iic_dvfs/iic_dvfs.c:440:
--
ERROR: need consistent spacing around '*' (ctx:WxV)
#485: FILE: drivers/renesas/rcar/iic_dvfs/iic_dvfs.c:466:
patch: /tmp/patches/0011-rcar_gen3-drivers-io-emmc-mem.patch
patch: /tmp/patches/0012-rcar_gen3-drivers-console.patch
patch: /tmp/patches/0013-rcar_gen3-drivers-power-controller.patch
patch: /tmp/patches/0014-rcar_gen3-drivers-rom-api.patch
patch: /tmp/patches/0015-rcar_gen3-drivers-spi-multio-bus-controller.patch
patch: /tmp/patches/0016-rcar_gen3-drivers-serial-controller-interface.patch
patch: /tmp/patches/0017-rcar_gen3-drivers-watchdog.patch

ldts:
I believe the iic_dvfs errors are false positives due to the macro defining the functions

ldts:
@soby-mathew I hope this PR can now be accepted and further changes be done under maintainer-ship. Please let me know (I still would love to meet the ELCE deadline next week)

drivers/renesas/rcar/console/rcar_console.S
antonio-nino-diaz-arm: https://github.com/ARM-software/arm-trusted-firmware/wiki/TF-A-Release-information#removal-of-deprecated-interfaces

ldts: ah awesome. great effort on the documentation btw

soby-mathew:
@ldts, thanks for doing that. In fact this was going to be my suggestion, viz to create a staging folder within drivers directory. This will allow maintainers to stage their code and clean it up overtime.

soby-mathew:
jenkins: test this please

soby-mathew:
CI tests look good.

PR #1590, opened 2018-09-26
Update link to Linux coding style

sandrine-bailleux-arm:


Yann-lms:
I know that it won't be the official repo, but why not using the github clone (https://github.com/torvalds/linux/blob/master/Documentation/process/coding-style.rst) to benefit from github's formatting of .rst files?

PR #1597, opened 2018-10-01
plat/arm: Remove option ARM_BOARD_OPTIMISE_MEM

antonio-nino-diaz-arm:
This option makes it hard to optimize the memory definitions of all Arm platforms because any change in the common defines must work in all of them. The best thing to do is to remove it and move the definition to each platform's header.

FVP, SGI and SGM were using the definitions in board_arm_def.h. The definitions have been copied to each platform's platform_def.h. Juno was already using the ones in platform_def.h, so there have been no changes.

### DIFF
--- a/docs/user-guide.rst
+++ b/docs/user-guide.rst
@@ -707,14 +707,6 @@ Arm development platform specific build options
    sets the TSP location to DRAM and ignores the ``ARM_TSP_RAM_LOCATION`` build
    flag.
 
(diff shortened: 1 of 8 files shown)


antonio-nino-diaz-arm:
@npoushin, what do you think? (Also, your github user in the maintainers.rst file is wrong, as far as I can tell).

soby-mathew:
jenkins: test this please

PR #1612, opened 2018-10-04
tools: Make invocation of host compiler correct

antonio-nino-diaz-arm:
HOSTCC should be used in any of the tools inside the tools/ directory instead of CC. That way it is possible to override both values from the command line when building the Trusted Firmware and the tools at the same time. Also, use HOSTCCFLAGS instead of CFLAGS.

Also, instead of printing the strings CC and LD in the console during the compilation of the tools, HOSTCC and HOSTLD have to be used for clarity. This is how it is done in other projects like U-Boot or Linux.

### DIFF
--- a/tools/cert_create/Makefile
+++ b/tools/cert_create/Makefile
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2015-2017, ARM Limited and Contributors. All rights reserved.
+# Copyright (c) 2015-2018, ARM Limited and Contributors. All rights reserved.
 #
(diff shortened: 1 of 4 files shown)


Yann-lms:
OK for STM32MP1.
By the way this could conflict with the patch proposal made by @jan-kiszka for #631, but I've seen no PR for this, @antonio-nino-diaz-arm  maybe you could squash both patches?

Yann-lms:
Exact, I missed the patch from @kostapr, then it should be OK for me after the rebase

soby-mathew:
jenkins: test this please

PR #1613, opened 2018-10-04
psci: implement platform hook to migrate_info

ldts:
Even if the TrustedOS does not implement migrate restrictions, the
platform might.

This is the case when for instance a platform contains a migratable
TrustedOS but a certain CPU (tipically the boot cpu) can't be
hotpluged. This commit allows the implementation of such use case.

### DIFF
--- a/include/lib/psci/psci.h
+++ b/include/lib/psci/psci.h
@@ -322,12 +322,14 @@ typedef struct plat_psci_ops {
 	int (*write_mem_protect)(int val);
 	int (*system_reset2)(int is_vendor,
 				int reset_type, u_register_t cookie);
+	int (*migrate_info)(u_register_t *mpidr);
 } plat_psci_ops_t;
 
 /*******************************************************************************
  * Function & Data prototypes
  ******************************************************************************/
 unsigned int psci_version(void);
+int psci_migrate_info(u_register_t *mpidr);
 int psci_cpu_on(u_register_t target_cpu,
 		uintptr_t entrypoint,
 		u_register_t context_id);
--- a/lib/psci/psci_main.c
+++ b/lib/psci/psci_main.c
@@ -44,6 +44,23 @@ int psci_cpu_on(u_register_t target_cpu,
 	return psci_cpu_on_start(target_cpu, &ep);
 }
 
+int psci_platform_migrate_info(u_register_t *mpidr)
+{
+	int rc;
+
+	if (psci_plat_pm_ops->migrate_info == NULL) {
+		/* no migrate restrictions */
+		return PSCI_TOS_NOT_PRESENT_MP;
+	}
+
+	rc = psci_plat_pm_ops->migrate_info(mpidr);
+
+	assert((rc == PSCI_TOS_UP_MIG_CAP) || (rc == PSCI_TOS_NOT_UP_MIG_CAP) ||
+	       (rc == PSCI_TOS_NOT_PRESENT_MP) || (rc == PSCI_E_NOT_SUPPORTED));
+
+	return rc;
+}
+
 unsigned int psci_version(void)
 {
 	return PSCI_MAJOR_VER | PSCI_MINOR_VER;
@@ -276,8 +293,14 @@ int psci_migrate(u_register_t target_cpu)
 int psci_migrate_info_type(void)
 {
 	u_register_t resident_cpu_mpidr;
+	int rc;
 
-	return psci_spd_migrate_info(&resident_cpu_mpidr);
+	rc = psci_spd_migrate_info(&resident_cpu_mpidr);
+	if ((rc != PSCI_TOS_NOT_UP_MIG_CAP) && (rc != PSCI_TOS_UP_MIG_CAP))
+		return psci_platform_migrate_info(NULL);
+
+	return rc;
+
 }
 
 u_register_t psci_migrate_info_up_cpu(void)
@@ -290,8 +313,14 @@ u_register_t psci_migrate_info_up_cpu(void)
 	 * psci_spd_migrate_info() returns.
 	 */
 	rc = psci_spd_migrate_info(&resident_cpu_mpidr);
-	if ((rc != PSCI_TOS_NOT_UP_MIG_CAP) && (rc != PSCI_TOS_UP_MIG_CAP))
-		return (u_register_t)(register_t) PSCI_E_INVALID_PARAMS;
+	if ((rc != PSCI_TOS_NOT_UP_MIG_CAP) && (rc != PSCI_TOS_UP_MIG_CAP)) {
+		rc = psci_platform_migrate_info(&resident_cpu_mpidr);
+		if ((rc != PSCI_TOS_NOT_UP_MIG_CAP) &&
+		    (rc != PSCI_TOS_UP_MIG_CAP))
+			/* neither the SDP nor the platform have any migration
+			   restrictions: any core can be hotpluged */
+			return (u_register_t)(register_t) PSCI_E_INVALID_PARAMS;
+	}
 
 	return resident_cpu_mpidr;
 }


antonio-nino-diaz-arm:
IP Review +1

soby-mathew:
Hi @ldts,
The patch as it stands is a bit messy as it overrides the advertised TOS migrate behaviour to prevent CPU_OFF call on the boot CPU.  This doesn't prevent the actual CPU_OFF call (or the PSCI_MIGRATE call) from taking place if it were made by the normal world. Overriding the migrate info may also have side-effects with regards to how normal world interacts with the TOS.

We need to understand where this limitation arises from.  You mentioned previously in PR #1574 , that you suspect that the Trusted OS is introducing this limitiation. If that can be confirmed, then the right solution here would be for the SPD `svc_off` hook to reject the CPU_OFF request without changing the migrate info.

ldts:
hi @soby-mathew right but it only overwrites the behavior so when there are no migrate restrictions announced by the SPD.
Not sure what you mean? When a restriction is announced It does prevent the CPU_OFF call from being made by the Linux kernel.
Yes I was wrong and the TrustedOS does not impose any restrictions; in this case, it is the rcar-gen3 platform that has the restriction; during suspend all power is removed and the system can only boot from the boot cpu. if the boot cpu is hotplugged, the system wont be able to resume (the TF-A contains the bootcpu id hardcoded)

soby-mathew:
> When a restriction is announced It does prevent the CPU_OFF call from being made by the Linux kernel

Yes, but the firmware has to be prepared to with deal rogue requests from normal world. So, in addtion to the migrate behaviour, the CPU_OFF and MIGRATE calls also would need to be rejected. Anyway, this problem will not be present if we can implement my suggestion below.

> it is the rcar-gen3 platform that has the restriction; during suspend all power is removed and the system can only boot from the boot cpu. if the boot cpu is hotplugged, the system wont be able to resume (the TF-A contains the bootcpu id hardcoded)

I see, that changes the situation. In that case a new PSCI ops hook has to be introduced to allow to reject CPU OFF call (`int (*allow_cpu_off)(void)`). This should be an optional hook which can only return SUCCESS or DENIED and this hook needs to be called before the SPD `svc_off` is called.

The CPU_OFF caller (linux kernel) will get a DENIED error for a CPU that cannot be offlined. It would have been ideal if the DT could encode this info so that kernel doesn't need to make the CPU_OFF call to figure this out. That's an optimization and will need further discussion I guess.

ldts:
thanks @soby-mathew . shall I re-implement to that spec or do you or @antonio-nino-diaz-arm  plan to do this?

antonio-nino-diaz-arm:
@ldts, I think it's better if you do it. In my case my problem is easier because I don't have a trusted OS so I can just implement a hacky fix by adding a dummy ``spd_pm_ops_t`` that denies all CPU_OFF calls from CPU0.

ldts:
ok

ldts:
now that I remember, last time I tried this implementation (which I just did again), the kernel locks up - it cant handle a returning CPU_OFF (ie, DENIED) ; I know that the PSCI spec says that is should be handled but it isnt.. that is why I had to implement the more complex hook under the SPD...how do we move forward?

ldts:
btw I also tested this TF-A with the rcar-gen3 m3ulcb platform running automotive grade linux (AGL) and it is all good

ldts:
so I can use that new interface to prevent cpuX from being put in retention (which incidentally in rcar-gen3 is just offline so I'll need that interface for cpuidle!) but not to prevent the cpu from being hotplugged by the kernel.

soby-mathew:
>  the kernel locks up - it cant handle a returning CPU_OFF (ie, DENIED) ; I know that the PSCI spec says that is should be handled but it isnt.. that is why I had to implement the more complex hook under the SPD...how do we move forward? will you raise it with the kernel team (not sure if you are in the same building/floor)?

Hmm, that presents a problem. Looks like kernel is not prepared to handle failure of CPU_OFF : https://elixir.bootlin.com/linux/latest/source/arch/arm64/kernel/smp.c#L354

This will need wider discussion internally. We will get back to you on this.

> so I can use that new interface to prevent cpuX from being put in retention (which incidentally in rcar-gen3 is just offline so I'll need that interface for cpuidle!) but not to prevent the cpu from being hotplugged by the kernel.

Cpu Idle request (via PSCI_CPU_SUSPEND) can always be downgraded to a regular wfi() if there are any hardware restrictions. So I didn't get why you need to reject the retention request.

ldts:
yep, thanks. will wait for the cpu_off comments.

the reason why cpu idle needs to be disabled for the boot cpu (boot cpu detected at firmware level) is that it will prevent system suspend from happening (the platform can only enter suspend from the boot cpu);

antonio-nino-diaz-arm:
There is a problem when denying ``PSCI_CPU_OFF``, yes. When I tried ``echo 0 > /sys/devices/system/cpu/cpu0/online`` from the terminal I got a message that it had been denied but the terminal decided to hang anyway. I wanted to take a look at the kernel source but it looks like you have already done it for me, @ldts. And yes, as @soby-mathew says, this is a problem.

I actually managed to lock CPU0 in a ``WFE`` loop and release it with ``PSCI_CPU_ON`` from a different core, but this is just a workaround...

ldts:
hi, just to be able to pace my work, when should I expect an answer to the PR (and its client, #1582 )?

soby-mathew:
Hi @ldts,
Okay, we had a discussion regarding this and the conclusion is that Linux cannot handle an error from CPU_OFF because it is too late to undo all the teardown done. So returning error code is not an option. If there are CPU_OFF restrictions, the kernel should be able to discover it to avoid calling CPU_OFF on that CPU. This would mean some kind of PSCI spec update to provide guidance on this.

Currently platforms with CPU_OFF restrictions manage to handle it within firmware. i.e they can implement CPU_OFF as WFI (for instance by relying on the default `pwr_domain_pwr_down_wfi()`
implementation) and take care of the consequences in the CPU_ON handler. This hides the CPU_OFF issue and from Linux' POV the CPU is OFF, and only CPU_ON brings it back.

Now in case of R-car platform, the SYSTEM_SUSPEND call can only resume from Boot CPU which means you need to reject the call on all other CPUs. This can be acheived by modifying the check for output from `psci_plat_pm_ops->get_sys_suspend_power_state`.

```
diff --git a/lib/psci/psci_main.c b/lib/psci/psci_main.c
index fd822bc..b4a25fb 100644
--- a/lib/psci/psci_main.c
+++ b/lib/psci/psci_main.c
@@ -167,8 +167,14 @@ int psci_system_suspend(uintptr_t entrypoint, u_register_t context_id)
        /* Query the psci_power_state for system suspend */
        psci_query_sys_suspend_pwrstate(&state_info);
 
+       /*
+        * Check if platform allows suspend to Highest power level
+        * (System level)
+        */
+       if (psci_find_target_suspend_lvl(&state_info) < PLAT_MAX_PWR_LVL)
+               return PSCI_E_DENIED;
+
        /* Ensure that the psci_power_state makes sense */
-       assert(psci_find_target_suspend_lvl(&state_info) == PLAT_MAX_PWR_LVL);
        assert(psci_validate_suspend_req(&state_info, PSTATE_TYPE_POWERDOWN)
                                                == PSCI_E_SUCCESS);
```

The platform can populate the right `state_info` if invoked from boot CPU while `zeroing` it out for other CPUs.  This way, SYSTEM_SUSPEND can be rejected on all other CPUs except the boot CPU.

Could you let us know whether that would work for you ?

> the reason why cpu idle needs to be disabled for the boot cpu (boot cpu detected at firmware level) is that it will prevent system suspend from happening (the platform can only enter suspend from the boot cpu);

I didn't understand how CPU_IDLE can prevent SYSTEM_SUSPEND from happening. All CPUs except the boot CPU will be offlined prior to SYSTEM_SUSPEND, which leaves only the boot CPU available for scheduling. Even if it were to IDLE, it will be resumed by kernel for issuing SYSTEM_SUSPEND command.

ldts:
hi @soby-mathew the rcar platform (AFAIK) does not have any CPU_OFF restriction beyond system suspend. This means that your patch above + a minor modification to the platform's get_sys_suspend_power_state implementation is sufficient . So I think that if you push your patch I can then abandon this PR and resubmit the #1582; I just tested it all and it works for me. what do you think?

ldts:
@soby-mathew just updated the commit adding your proposed change to psci (trying to save you spending more time on this)

ldts:
https://github.com/ARM-software/arm-trusted-firmware/pull/1625

antonio-nino-diaz-arm:
Just to say that in my case https://github.com/ARM-software/arm-trusted-firmware/pull/1616 I've managed to emulate ``PSCI_CPU_OFF`` with a ``WFE`` loop and everything seems to be working fine. A few days ago I made a mistake during my first try and Linux started showing warnings all the time about not being able to change the frequency after "turning CPU 0 off" (because it's also controlled by SCP).

soby-mathew:
Thanks @ldts for the PR. Good to know it will solve your issue.

PR #1615, opened 2018-10-08
Makefile: Support totally quiet output with -s

Andre-ARM:
"-s" is a command line option to the make tool, to suppress normal output,
something to the effect of prepending every line with '@' in the Makefile.
However with our V={0|1} support, we now print the shortened command line
output in any case (even with V=1, in addition to the long line!).
Normally -s helps to not miss non-fatal warnings, which tend to scroll out
of the window easily.

Introduce a new Makefile variable ECHO, to control the shortened output.
We only set it in the (current default) V=0 case, and replace every
occurence of "@echo" with that variable.
When the user specifies "-s", we set ECHO to some magic string which
changes the output line into a comment, so the output is suppressed.

Beside suppressing every output for "-s", we also avoid the redundant
short output when compiling with V=1.

This changes the output to:
==========
$ make -s PLAT=.... bl31

Built build/.../release/bl31.bin

==========
$ make PLAT=.... bl31
...
  CC      lib/libc/strncmp.c
  CC      lib/libc/strnlen.c
...
==========
$ make V=1 PLAT=.... bl31
...
gcc -DDEBUG=0 .... -o build/.../release/libc/strncmp.o
gcc -DDEBUG=0 .... -o build/.../release/libc/strnlen.o
...
==========

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,11 +74,19 @@ CHECK_PATHS		:=	${ROOT_DIRS_TO_CHECK}			\
 # Verbose flag
 ifeq (${V},0)
         Q:=@
+        ECHO:=@echo
         CHECKCODE_ARGS	+=	--no-summary --terse
 else
         Q:=
+        ECHO:=@\#
 endif
-export Q
+
+ifneq ($(findstring s,$(filter-out --%,$(MAKEFLAGS))),)
+        Q:=@
+        ECHO:=@\#
+endif
+
+export Q ECHO
 
 # Process Debug flag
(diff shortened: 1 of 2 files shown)


antonio-nino-diaz-arm:
IP Review +1

In general I agree with the idea, but you are missing the messages in the tools' makefiles, under ``tools/``.

soby-mathew:
The `tools/` makefile can be fixed by a later patch. Merging this one.

PR #1616, opened 2018-10-08
Initial port of Amlogic Meson S905 (GXBB)

antonio-nino-diaz-arm:
The Amlogic Meson S905 is a SoC with a quad core Arm Cortex-A53 running at 1.5Ghz. It also contains a Cortex-M3 used as SCP.

This port is a minimal implementation of BL31 capable of booting mainline U-Boot and Linux:
- SCPI support.
- Basic PSCI support (CPU_ON, CPU_OFF, SYSTEM_RESET, SYSTEM_OFF). Note that CPU0
  can't be turned off and it will reject PSCI_CPU_OFF requests.
- GICv2 driver set up.
- Basic SIP services (read efuse data, enable/disable JTAG).

In order to build it:

```
CROSS_COMPILE=aarch64-linux-gnu- make DEBUG=1 PLAT=gxbb bl31
```

This port has been tested in a ODROID-C2. After building it, follow the instructions in the [U-Boot repository](https://github.com/u-boot/u-boot/blob/master/board/amlogic/odroid-c2/README), replacing the mentioned **bl31.bin** by the one built from this port.

Also, note that the final workaround for the issue with CPU0 will depend on the outcome of https://github.com/ARM-software/arm-trusted-firmware/pull/1613


plat/meson/gxbb/gxbb_pm.c
```diff
@@ -113,17 +130,54 @@ static void gxbb_pwr_domain_off(const psci_power_state_t *target_state)
...
+		while (gxbb_cpu0_go == 0)
+			wfe();
+
+		VERBOSE("BL31: CPU0 resumed.\n");
+
+		gxbb_cpu0_go = 0;
+		dsb();
+		isb();
+
+		disable_mmu_el3();
```
soby-mathew: Issuing an rmr_el3 to reset the core might work better. Also the reset of the variable 
`gxbb_cpu0_go` to 0 can be done in the `on_finish` hook since the CPU_ON call can still get through and set the variable to 1. Since the `on_finish` handler executes after the PSCI affinity info state of the CPU has been updated to RUN, the CPU_ON call can no longer reach the platform layer. This will prevent spurious wake-ups.

plat/meson/gxbb/gxbb_pm.c
```diff
@@ -113,17 +130,54 @@ static void gxbb_pwr_domain_off(const psci_power_state_t *target_state)
...
+		VERBOSE("BL31: CPU0 entering wait loop...\n");
+
+		while (gxbb_cpu0_go == 0)
+			wfe();
+
+		VERBOSE("BL31: CPU0 resumed.\n");
+
+		gxbb_cpu0_go = 0;
+		dsb();
+		isb();
```
soby-mathew: nit: isb() is not required.

drivers/meson/console/aarch64/meson_console.S
```diff
@@ -0,0 +1,262 @@
...
+	 *     w1 - Uart clock in Hz
+	 *     w2 - Baud rate
+	 * Out: return 1 on success else 0 on error
+	 * Clobber list : x0-x3
+	 * -----------------------------------------------
+	 */
+func console_meson_init
+	cmp	w0, #0
+	beq	init_fail
+	cmp	w1, #0 /* TODO: Test with clocks other than 24 MHz */
```
Andre-ARM: You set the MESON_REG5_USE_XTAL_CLK_BIT below, so you are always using the (fixed frequency) oscillator as the clock base. And for ATF's purposes this is totally sufficient, so I wouldn't even try to support the clk81 source. You could even think about dropping the second parameter, and use the oscillator frequency directly in here. Or at least amend the comment to change the TODO: to a note about the clock being actually fixed.

plat/meson/gxbb/gxbb_pm.c
antonio-nino-diaz-arm: You're right, I didn't even remember that register.

plat/meson/gxbb/gxbb_pm.c
```diff
@@ -113,17 +130,54 @@ static void gxbb_pwr_domain_off(const psci_power_state_t *target_state)
...
+	/* CPU0 can't be turned OFF/ON, emulate it with a WFE loop */
+	if (core == GXBB_PRIMARY_CPU) {
+		VERBOSE("BL31: CPU0 entering wait loop...\n");
+
+		while (gxbb_cpu0_go == 0)
+			wfe();
+
+		VERBOSE("BL31: CPU0 resumed.\n");
+
+		gxbb_cpu0_go = 0;
```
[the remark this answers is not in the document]
antonio-nino-diaz-arm: Done.

drivers/meson/console/aarch64/meson_console.S
antonio-nino-diaz-arm: Done.

Regarding the NULL checks, I basically copied code from the PL011 driver and adapted it, so I decided to leave them there. I don't think they hurt.

plat/meson/gxbb/aarch64/gxbb_helpers.S
```diff
@@ -0,0 +1,97 @@
+/*
+ * Copyright (c) 2018, Antonio Niño Díaz. All rights reserved.
```
soby-mathew: Its better to use the standard copyright header as in your other files.

soby-mathew: Its better to use the standard copyright header as in your other files.

soby-mathew: Its better to use the standard copyright header as in your other files.

soby-mathew: Its better to use the standard copyright header as in your other files.

soby-mathew: Its better to use the standard copyright header as in your other files.

Amit-Radur:
Amlogic seems to use legacy SCPI protocol, So would it be possible to reflect this fact either through some
Documentation or through Source code ?
For instance , Linux[1] makes a clear distinction between the two.

[1]: https://github.com/torvalds/linux/blob/master/drivers/firmware/arm_scpi.c#L130

antonio-nino-diaz-arm:
@Amit-Radur

I've added a comment to ``gxbb_scpi.c``. I don't think it makes sense to document this anywhere else because this file isn't shared with other platforms so anyone that wants to know how it works (command IDs and similar) will end up taking a look at that file.

Amit-Radur:
Ok, Thanks . 

I would test this patch-set on my Nano pi  k2.

antonio-nino-diaz-arm:
I've added your tag, @Amit-Radur.

soby-mathew:
jenkins: test this please

PR #1618, opened 2018-10-09
Fix misra warnings in SMC and power mgmt code

satheesbalya-arm:
Change-Id: Ia00eba2b18804e6498d935d33ec104953e0e5e03

### DIFF
--- a/include/common/runtime_svc.h
+++ b/include/common/runtime_svc.h
@@ -4,8 +4,8 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-#ifndef __RUNTIME_SVC_H__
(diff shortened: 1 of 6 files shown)


soby-mathew:
Just one MISRA error highlighted :
```
plat/arm/board/fvp/drivers/pwrc/fvp_pwrc.c:52:[MISRA C-2012 Rule 12.2 (required)]<278105> In expression "1U << 31", shifting more than 7 bits, the number of bits in the essential type of the left expression, "1U", is not allowed.  The shift amount is 31.
```
Looks like the BIT_32() macro needs to append `UL`  instead of `U`. This can be taken up separately.

PR #1624, opened 2018-10-10
PSCI cache flush and comment fixup

glneo:
Here are the two first patches of the pull request [here](https://github.com/ARM-software/arm-trusted-firmware/pull/1553). As these two patches are not the subject of that pull request and are a bit less controversial I'd like to get them in independently here.

Thanks,
Andrew

### DIFF
--- a/lib/psci/psci_common.c
+++ b/lib/psci/psci_common.c
@@ -267,7 +267,7 @@ static plat_local_state_t *psci_get_req_local_pwr_states(unsigned int pwrlvl,
 static plat_local_state_t get_non_cpu_pd_node_local_state(
 		unsigned int parent_idx)
 {
-#if !(USE_COHERENT_MEM || HW_ASSISTED_COHERENCY)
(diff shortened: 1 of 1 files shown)


lib/psci/psci_common.c
```diff
@@ -267,7 +267,7 @@ static plat_local_state_t *psci_get_req_local_pwr_states(unsigned int pwrlvl,
 static plat_local_state_t get_non_cpu_pd_node_local_state(
 		unsigned int parent_idx)
 {
-#if !(USE_COHERENT_MEM || HW_ASSISTED_COHERENCY)
+#if !(USE_COHERENT_MEM || HW_ASSISTED_COHERENCY || WARMBOOT_ENABLE_DCACHE_EARLY)
```
soby-mathew: I realized that when HW_ASSISTED_COHERENCY is enabled, there is no need for COHERENT_MEM because,

* There are no bakery locks
* This data structure is always accessed with caches ON.

So, there is a case to enforce that rule in Makefile. I can do that as a separate PR.

soby-mathew:
IP-Review +1

soby-mathew:
jenkins: test this please

PR #1625, opened 2018-10-11
psci: platform control of SYSTEM_SUSPEND entry

ldts:
Some platforms can only resume from system suspend from the boot
CPU, hence they should only enter that state from that same core.

The following commit presents an interface that allows the platform to
reject system suspend entry near its very last stage (last CPU).

### DIFF
--- a/lib/psci/psci_main.c
+++ b/lib/psci/psci_main.c
@@ -167,8 +167,14 @@ int psci_system_suspend(uintptr_t entrypoint, u_register_t context_id)
 	/* Query the psci_power_state for system suspend */
 	psci_query_sys_suspend_pwrstate(&state_info);
 
+	/*
+	 * Check if platform allows suspend to Highest power level
+	 * (System level)
+	 */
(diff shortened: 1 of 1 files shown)


lib/psci/psci_main.c
```diff
@@ -167,8 +167,10 @@ int psci_system_suspend(uintptr_t entrypoint, u_register_t context_id)
 	/* Query the psci_power_state for system suspend */
 	psci_query_sys_suspend_pwrstate(&state_info);
 
+	if (psci_find_target_suspend_lvl(&state_info) < PLAT_MAX_PWR_LVL)
```
soby-mathew: Would prefer a comment here, something like :
```
       /*
        * Check if platform allows suspend to Highest power level
        * (System level)
        */
```

antonio-nino-diaz-arm:
IP Review +1

lib/psci/psci_main.c
ldts: ok

soby-mathew:
jenkins: test this please

PR #1633, opened 2018-10-16
Allwinner fixes and PMIC support 

Andre-ARM:
I understand this is quite a series, but most of the patches depend on each other, so not sure that splitting it up is too useful. The patches are arranged to be bisectable (both for compilation and functionality), so earlier patches can always be merged as soon as they are ready.

This updates the Allwinner port to do some cleanups and minor improvements, also adding PMIC and proper SYSTEM_OFF and CPU_OFF support to bring it eventually on par (and beyond) the existing "legacy" ATF port, by:
- Unifying the platform specific Makefiles.
- Reducing the memory footprint to make room for feature enhancements.
- Adding generic code to reduce power consumption on SYSTEM_OFF without a PMIC.
- Flipping GPIOs on H5 reference design boards to turn off regulators on SYSTEM_OFF.
- Adding support for the RSB bus and the AXP803 PMIC for boards using the A64 SoC (for SYSTEM_OFF).
- Find U-Boot's DTB copy in memory and use that to enable and fix up essential power rails on the AXP803 PMIC.
- Introduce some means to execute code on the management processor and use that to properly turn off cores on CPU_OFF requests.

I tested this (booting Linux, hotplugging random cores 1000 times, powering off) on: Pine64, BananaPiM64, Emlid Neutis N5, OrangePi PC2, PineH64. Please have a test with your setup and report any issues!


antonio-nino-diaz-arm:
IP Review +1

plat/allwinner/common/sunxi_common.c
```diff
@@ -156,3 +158,52 @@ int platform_init_r_twi(uint16_t socid, bool use_i2c)
...
+		if (!(mmio_read_32(SUNXI_R_CPUCFG_BASE) & BIT(0)))
+			break;
+
+		bakery_lock_release(&arisc_lock);
+	} while (1);
+
+	/* Patch up the code to feed in an input parameter. */
+	if (patch_offset >= 0 && patch_offset <= (size - 4))
+		code[patch_offset] = (code[patch_offset] & ~0xffff) | param;
+	flush_dcache_range((uintptr_t)code, size);
```
antonio-nino-diaz-arm: Strictly speaking you don't need to flush, clean is enough. Same below.

plat/allwinner/sun50i_a64/sunxi_power.c
```diff
@@ -13,6 +13,7 @@
 #include <platform_def.h>
 #include <sunxi_def.h>
 #include <sunxi_mmap.h>
+#include <sunxi_helpers.h>
```
smaeul: Hmm, this file should already be including `sunxi_private.h` for `sunxi_power_down`.

plat/allwinner/common/sunxi_common.c
```diff
@@ -10,8 +10,6 @@
 #include <sunxi_def.h>
 #include <xlat_tables_v2.h>
 
-#include "sunxi_private.h"
```
smaeul: The file defining the function should also include the header (to guarantee consistency in the prototype)

plat/allwinner/common/sunxi_cpu_ops.c
```diff
@@ -11,8 +11,6 @@
 #include <sunxi_cpucfg.h>
 #include <utils_def.h>
 
-#include "sunxi_private.h"
```
smaeul: This file is including `sunxi_private.h` because it defines functions declared there.

plat/allwinner/sun50i_a64/sunxi_power.c
```diff
@@ -7,18 +7,110 @@
...
+	/* AHB1: use OSC24M/1, APB1 = AHB1 / 2 */
+	mmio_write_32(SUNXI_CCU_BASE + 0x54, 0x1000);
+	/* APB2: use OSC24M */
+	mmio_write_32(SUNXI_CCU_BASE + 0x58, 0x1000000);
+	/* AHB2: use AHB1 clock */
+	mmio_write_32(SUNXI_CCU_BASE + 0x5c, 0);
+	/* CPU: use OSC24M */
+	mmio_write_32(SUNXI_CCU_BASE + 0x50, 0x10000);
+
+	/** turn off PLLs **/
```
smaeul: > Turn off PLLs.

(capitalization/punctuation)

plat/allwinner/sun50i_a64/sunxi_power.c
```diff
@@ -7,18 +7,110 @@
...
+	for (i = 0; i < 6; i++)
+		mmio_clrbits_32(SUNXI_CCU_BASE + i * 8, BIT(31));
+	switch (socid) {
+		case SUNXI_SOC_H5:
```
smaeul: Indentation of `case` isn't consistent here.

plat/allwinner/common/sunxi_common.c
```diff
@@ -70,3 +70,32 @@ uint16_t sunxi_read_soc_id(void)
...
+void sunxi_set_gpio_out(char port, int pin, bool level_high)
+{
+	uintptr_t pio_base = SUNXI_PIO_BASE;
+
+	if (port < 'A' || port > 'L')
+		return;
+	if (port == 'L')
+		pio_base = SUNXI_R_PIO_BASE;
+	else
+		pio_base += (port - 'A') * 0x24;
```
smaeul: The use of `=` vs `+=` is confusing here. How about `pio_base = SUNXI_PIO_BASE + (port - 'A') * 0x24;`, and drop the initialization earlier? And maybe `port_base`, since it's only sometimes the PIO base.

plat/allwinner/sun50i_a64/sunxi_power.c
```diff
@@ -105,6 +107,25 @@ void __dead2 sunxi_power_down(void)
...
+		 * - PL8 (PWR-STB = CPU power supply) -> low
+		 * - PL9 (PWR-DRAM) ->low
+		 * - PL10 (power LED) -> low
+		 * Note: Clearing PL8 will reset the board, so keep it up.
+		 */
+		sunxi_set_gpio_out('L', 5, 1);
+		sunxi_set_gpio_out('L', 9, 0);
+		sunxi_set_gpio_out('L', 10, 0);
+
+		/* turn off pin controller now. */
```
smaeul: turn -> Turn

plat/allwinner/common/include/sunxi_helpers.h
```diff
@@ -8,6 +8,7 @@
 #define __SUNXI_HELPERS_H__
 
 uint16_t sunxi_read_soc_id(void);
+int platform_init_r_twi(uint16_t socid, bool use_i2c);
```
smaeul: Why does this function have a different prefix than all of the other sunxi shared functions?

plat/allwinner/common/sunxi_common.c
```diff
@@ -99,3 +101,58 @@ void sunxi_set_gpio_out(char port, int pin, bool level_high)
 			   0x7 << ((pin % 8) * 4),
 			   0x1 << ((pin % 8) * 4));
 }
+
+int platform_init_r_twi(uint16_t socid, bool use_i2c)
```
smaeul: Maybe `use_i2c` -> `use_rsb` would make it more clear what's going on. When I read this I was confused how you could have TWI without I²C :)

plat/allwinner/common/allwinner-common.mk
```diff
@@ -13,6 +13,8 @@ PLAT_INCLUDES		:=	-Iinclude/plat/arm/common		\
 				-I${AW_PLAT}/common/include		\
 				-I${AW_PLAT}/${PLAT}/include
 
+include lib/libfdt/libfdt.mk
```
smaeul: This should probably go at the top of the file with the other include

plat/allwinner/common/sunxi_bl31_setup.c
```diff
@@ -29,6 +30,47 @@ static const gicv2_driver_data_t sunxi_gic_data = {
...
+
+		if (u_boot_base[i] != PLAT_SUNXI_NS_IMAGE_OFFSET)
+			continue;
+
+		/* Does the suspected U-Boot size look anyhow reasonable? */
+		if (u_boot_base[i + 1] >= 256 * 1024 * 1024)
+			continue;
+
+		/* end of the image: base address + size */
+		dtb_base = (void *)u_boot_base + u_boot_base[i + 1];
```
smaeul: I'm not sure what ATF's position on void pointer math (a GCC extension), maybe `char *` instead.

plat/allwinner/common/sunxi_bl31_setup.c
```diff
@@ -86,6 +129,17 @@ void bl31_platform_setup(void)
 
 	generic_delay_timer_init();
 
+	fdt = sunxi_find_dtb();
+	if (fdt) {
```
smaeul: I think `!= NULL` is expected here

plat/allwinner/common/sunxi_bl31_setup.c
```diff
@@ -86,6 +129,17 @@ void bl31_platform_setup(void)
...
+		const char *model;
+		int length;
+
+		model = fdt_getprop(fdt, 0, "model", &length);
+		NOTICE("BL31: Found U-Boot DTB at %p, model: %s\n", fdt,
+		     model ?: "unknown");
+	} else
```
smaeul: Please use braces on the `else`, since the `if` needs them

plat/allwinner/sun50i_a64/sunxi_power.c
```diff
@@ -119,6 +120,51 @@ static int axp_setbits_8(uint8_t reg, uint8_t set_mask)
 	return rsb_write(AXP803_RT_ADDR, reg, regval);
 }
 
+static bool fdt_handle_regulator(const void *fdt, int node)
```
smaeul: This isn't very descriptive. I suggest something to the effect of `should_enable_regulator`, and a comment in the code that the existence of a phandle means it's used elsewhere in the DT

plat/allwinner/sun50i_a64/sunxi_power.c
```diff
@@ -76,15 +81,66 @@ void sunxi_turn_off_soc(uint16_t socid)
...
+
+	ret = rsb_set_bus_speed(SUNXI_OSC24M_CLK_IN_HZ, 3000000);
+	if (ret)
+		return ret;
+
+	return rsb_assign_runtime_address(AXP803_HW_ADDR,
+					  AXP803_RT_ADDR);
+}
+
+static int axp_setbits_8(uint8_t reg, uint8_t set_mask)
```
smaeul: You don't use `_8` for `axp_write` later, and the registers are always 8 bits anyway (unlike MMIO where the size could vary), so I think it makes sense to drop the `_8` here as well.

plat/allwinner/sun50i_a64/sunxi_power.c
```diff
@@ -150,10 +228,18 @@ static void setup_axp803_rails(const void *fdt)
 	for (node = fdt_first_subnode(fdt, node);
 	     node != -FDT_ERR_NOTFOUND;
 	     node = fdt_next_subnode(fdt, node)) {
+		struct axp_regulator *reg;
 		const char *name;
 		int length;
 
 		name = fdt_get_name(fdt, node, &length);
+		for (reg = regulators; reg->dt_name; reg++) {
```
smaeul: If you pull the `fdt_handle_regulator` call outside the loop, in addition to avoiding the loop in that case, you could swap the two `if` statements for `dc1sw` below, and only call it once.

plat/allwinner/common/arisc_off.S
```diff
@@ -0,0 +1,115 @@
...
+	l.ff1	r6, r3				# get core number from high mask
+	l.addi	r6, r6, -17			# convert to 0-3
+	l.slli	r6, r6, 2			# r5: core number*4 (0-12)
+	l.add	r6, r6, r13			# add to base address
+	l.ori	r5, r0, 0xff			# 0xff means all switches off
+	l.sw	0x1540(r6), r5			# core power switch registers
+
+	l.sw	0x1c00(r13),r0			# pull down our own reset line
+
+	l.j	.				# just in case ....
```
smaeul: To be extra sure the reset bit gets cleared, you could jump to the `l.sw` above (and same for H6).

plat/allwinner/common/arisc_off.S
```diff
@@ -0,0 +1,115 @@
...
+	l.nop	0x0				# (delay slot)
+
+# same as above, but with the MMIO addresses matching the H6 SoC
+_start_h6:
+	l.movhi	r3, 0				# FIXUP! with core mask
+	l.movhi r0, 0				# clear r0
+	l.movhi	r13, 0x901			# r13: CPU_CFG_BASE=0x01700000
```
smaeul: Comment is wrong

plat/allwinner/common/sunxi_cpu_ops.c
```diff
@@ -36,18 +41,41 @@ static void sunxi_cpu_enable_power(unsigned int cluster, unsigned int core)
 	mmio_write_32(SUNXI_CPU_POWER_CLAMP_REG(cluster, core), 0x00);
 }
 
+#include <core_off_arisc.h>
```
smaeul: Why not at the top?

plat/allwinner/common/sunxi_common.c
```diff
@@ -17,8 +17,12 @@ static mmap_region_t sunxi_mmap[PLATFORM_MMAP_REGIONS + 1] = {
...
 	MAP_REGION_FLAT(SUNXI_DEV_BASE, SUNXI_DEV_SIZE,
 			MT_DEVICE | MT_RW | MT_SECURE),
-	MAP_REGION_FLAT(SUNXI_DRAM_BASE, SUNXI_DRAM_SIZE,
-			MT_MEMORY | MT_RW | MT_NS),
+	MAP_REGION(SUNXI_DRAM_BASE, SUNXI_DRAM_VIRT_BASE, SUNXI_DRAM_SEC_SIZE,
+			MT_MEMORY | MT_RW | MT_SECURE),
+	MAP_REGION(PLAT_SUNXI_NS_IMAGE_OFFSET,
+		   SUNXI_DRAM_VIRT_BASE + SUNXI_DRAM_SEC_SIZE,
+		   SUNXI_DRAM_MAP_SIZE,
+		   MT_MEMORY | MT_RW | MT_NS),
```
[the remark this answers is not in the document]
antonio-nino-diaz-arm: Write permissions are required to modify the DTB that comes with U-Boot.

plat/allwinner/common/sunxi_bl31_setup.c
antonio-nino-diaz-arm: That's only to make Coverity happy. If in Allwinner platforms you don't care about it, don't worry about that kind of details.

plat/allwinner/common/sunxi_common.c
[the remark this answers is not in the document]
Andre-ARM: At the moment we don't modify the DTB, so MT_RO would be sufficient (and I will change this). But there are plans to amend the DT, so I guess we will amend this later on.

plat/allwinner/common/sunxi_cpu_ops.c
Andre-ARM: Yes, I always scratch my head about this header file. `sunxi_private.h` hints at that it contains functions private to the Allwinner platform, but its placement in the common/ directory makes it actually private to the _common_ code only. Thus the introduction of `sunxi_helpers.h` in a proper include directory to fix this. Should we move all prototypes into this one, renaming it to `sunxi_private.h`, then just include only this file from everywhere? The distinction between the two seems rather arbitrary to me.

plat/allwinner/common/sunxi_common.c
Andre-ARM: True, good catch. Which makes me think whether we need to do something in ATF about Cortex-A53 erratum 855873 (promoting cache cleans to invalidate in code for cores before r0p3).

antonio-nino-diaz-arm: It looks like we are safe as part of the workaround for 82631. However, we should make it clearer, or make it explicit in the workaround for 855873.

plat/allwinner/sun50i_a64/sunxi_power.c
```diff
@@ -7,18 +7,110 @@
...
+void sunxi_turn_off_soc(uint16_t socid)
 {
-	/* STUB */
-	NOTICE("BL31: STUB PMIC setup code called\n");
+	int i;
+
+	/** Turn off most peripherals, most importantly DRAM users. **/
+	/* Keep DRAM controller running for now. */
+	mmio_clrbits_32(SUNXI_CCU_BASE + 0x2c0, ~BIT_32(14));
+	mmio_clrbits_32(SUNXI_CCU_BASE + 0x60, ~BIT_32(14));
```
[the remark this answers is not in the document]
smaeul: Yes, though aren't they _immediately_ quiescent if we stop their clocks and assert their resets? And if not, and they try to write to DRAM and fail/hang/crash, do we care? Maybe I'm missing something, and it matters--in that case, it'd be a good idea to add a comment in the source explaining why.

plat/allwinner/common/sunxi_bl31_setup.c
```diff
@@ -29,6 +30,47 @@ static const gicv2_driver_data_t sunxi_gic_data = {
 	.gicc_base = SUNXI_GICC_BASE,
 };
 
+/*
+ * Try to find a DTB loaded in memory by previous stages.
+ *
+ * At the moment we implement a heuritistic to find the DTB attached to U-Boot:
```
smaeul: heuritistic -> heuristic here as well

plat/allwinner/sun50i_a64/sunxi_power.c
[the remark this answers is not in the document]
Andre-ARM: I consider it good practice to turn producers off first, especially DMA masters. I am not in expert in these areas, but I failed DRAM transfer could trigger some *bus* error condition, possibly signaled to the CPU? I am pretty sure ATF wouldn't really know how to handle an exception or even an SError at this point.
So if you can live with this, I'd keep it in. That wouldn't be compiled in for a crust version, so I don't see any need to save a few bytes.

Andre-ARM:
Many thanks to all the reviewers, this was very helpful. To keep the history here, I created a new PR #1646 with hopeful all the comments addressed. Changelog there.
Closing this PR.

PR #1635, opened 2018-10-16
Initialize platform for MediaTek mt8183

kennyliang-mtk:
- Add basic platform setup
- Add a generic UART driver
- Add generic CPU helper functions
- Add delay timer platform implementation

Signed-off-by: kenny liang <kenny.liang@mediatek.com>


plat/mediatek/mt8183/plat_mt_gic.c
```diff
@@ -0,0 +1,931 @@
...
+void gic_sgi_save_all(void)
+{
+	int cpuid = 0;
+	unsigned int base;
+	unsigned int cpu_offset = 0;
+
+	base = MT_GIC_RDIST_BASE;
+	base += SZ_64K;
+	for (cpuid = 0; cpuid < PLATFORM_CORE_COUNT; cpuid++) {
+		cpu_offset = cpuid*SZ_64K*2;
```
antonio-nino-diaz-arm: Leave spaces around ``*``.

plat/mediatek/mt8183/plat_mt_gic.c
```diff
@@ -0,0 +1,931 @@
...
+
+void gic_dist_restore(void)
+{
+	unsigned int gic_irqs;
+	unsigned int dist_base;
+	unsigned int ctlr;
+	unsigned int rdist_sgi_base;
+	unsigned int i = 0;
+
+	dist_base = MT_GIC_BASE;//get_plat_config()->gicd_base;
```
antonio-nino-diaz-arm: Remove comment.

plat/mediatek/mt8183/plat_mt_gic.c
```diff
@@ -0,0 +1,931 @@
...
+	uint8_t cpu = 0;
+	uint16_t tlist = 0;
+	uint64_t cluster_id = 0;
+	uint64_t mpidr = 0;
+
+	/*
+         * Ensure that stores to Normal memory are visible to the
+         * other CPUs before issuing the IPI.
+         */
+	__asm__ volatile("dmb sy" : : : "memory");
```
antonio-nino-diaz-arm:
```
#include <arch_helpers.h>
dmbsy();
```

plat/mediatek/mt8183/plat_mt_gic.c
```diff
@@ -0,0 +1,931 @@
...
+
+	for (cpu = 0; cpu < PLATFORM_CORE_COUNT ; ++cpu) {
+		/* only check those cores in map */
+		if ((((1 << cpu) & map) == 0) ||
+				(cpu_logical_map[cpu] == CPU_LOGIC_MAP_RESET_VAL)) {
+			continue;
+		}
+
+		/* gicv3 can only send sgi per cluster,
```
antonio-nino-diaz-arm: Fix comment format (``*/`` has to be on its own line).

plat/mediatek/mt8183/plat_mt_gic.c
```diff
@@ -0,0 +1,931 @@
...
+	write_icc_eoir0_el1(irq);
+}
+
+void mask_wdt_fiq()
+{
+	gicd_set_icenabler(MT_GIC_BASE, WDT_IRQ_BIT_ID);
+}
+
+
+//===========================================================================
```
antonio-nino-diaz-arm: Use ``/* */`` style comments.

plat/mediatek/mt8183/plat_mt_gic.c
```diff
@@ -0,0 +1,931 @@
...
+	/* Wake up this CPU redistributor */
+	val = mmio_read_32(rdist_base+GICR_V3_WAKER);
+	val &= ~GICR_V3_WAKER_ProcessorSleep;
+	mmio_write_32(rdist_base+GICR_V3_WAKER, val);
+
+	/* wait until our children wakeup */
+	while (mmio_read_32(rdist_base + GICR_V3_WAKER) &
+			GICR_V3_WAKER_ChildrenAsleep);
+
+	/* set all SGI/PPI as non-secure GROUP1 by default.
```
antonio-nino-diaz-arm: Fix comment style.

plat/mediatek/mt8183/plat_mt_gic.c
```diff
@@ -0,0 +1,931 @@
...
+}
+
+void setup_int_schedule_mode(enum irq_schedule_mode mode, unsigned int active_cpu)
+{
+	assert(mode <= HW_MODE);
+	assert(active_cpu <= 0xFF);
+
+	if(mode == HW_MODE)
+	{
+		mmio_write_32(GIC_INT_MASK, (mmio_read_32(GIC_INT_MASK) & ~(GIC500_ACTIVE_SEL_MASK)) | (0x1 << GIC500_ACTIVE_SEL_SHIFT));
```
antonio-nino-diaz-arm: Lines are too long, split them.

plat/mediatek/mt8183/plat_mt_gic.c
```diff
@@ -0,0 +1,931 @@
...
+	else
+		gicd_clr_igrpmodr(gicd_base, fiq_num);
+
+	/* give FIQ the highest priority */
+	gicd_set_ipriorityr(gicd_base, fiq_num, GIC_HIGHEST_SEC_PRIORITY);
+	/* use 1-N model, means as long as one of N core can handle, this will be handled */
+	gicd_write_irouter(gicd_base, fiq_num, GICD_V3_IROUTER_SPI_MODE_ANY);
+
+	/* set trigger type, falling edage */
+	/* mt_irq_set_sens(gicd_base, fiq_num, GET_INT_EDGE_TYPE(trigger_type)); */
```
antonio-nino-diaz-arm: Remove comments?

plat/mediatek/mt8183/plat_mt_gic.c
```diff
@@ -0,0 +1,931 @@
...
+
+	if (irq < 32) {
+		return 0;
+	}
+
+	if (irq > 1020) {
+		return 0;
+	}
+
+	dist_base = MT_GIC_BASE;//get_plat_config()->gicd_base;
```
antonio-nino-diaz-arm: Remove comment.

plat/mediatek/mt8183/plat_mt_gic.c
```diff
@@ -0,0 +1,931 @@
...
+{
+	unsigned int cpuid = 0;
+	unsigned int base = 0;
+	unsigned int scr_el3;
+
+	scr_el3 = read_scr_el3();
+	INFO("SCR_EL3:0x%x\n", scr_el3);
+
+	INFO("GICD_CTLR:0x%x\n", mmio_read_32(MT_GIC_BASE + GICD_CTLR));
+	/* INFO("ICC_CTLR_EL1:0x%lx\n", read_icc_ctlr_el1()); */
```
antonio-nino-diaz-arm: Remove comments?

plat/mediatek/mt8183/platform.mk
```diff
@@ -0,0 +1,62 @@
...
+# SPDX-License-Identifier: BSD-3-Clause
+#
+
+MTK_PLAT      := plat/mediatek
+MTK_PLAT_SOC  := ${MTK_PLAT}/${PLAT}
+
+PLAT_INCLUDES := -I${MTK_PLAT}/common/                            \
+                 -I${MTK_PLAT}/common/drivers/uart/               \
+                 -I${MTK_PLAT_SOC}/include/                       \
+                 -Iinclude/plat/arm/common                        \
```
antonio-nino-diaz-arm: Don't include ``include/plat/arm`` or  ``include/plat/arm/common``. ``drivers/arm`` is fine.

plat/mediatek/mt8183/platform.mk
```diff
@@ -0,0 +1,62 @@
...
+                 -Iinclude/plat/arm/common/aarch64                \
+                 -Idrivers/arm/gic/v3/
+
+PLAT_BL_COMMON_SOURCES := lib/xlat_tables/aarch64/xlat_tables.c	  \
+                          lib/xlat_tables/xlat_tables_common.c    \
+                          plat/arm/common/arm_gicv2.c             \
```
antonio-nino-diaz-arm: Same, don't use files under ``plat/arm``.

plat/mediatek/mt8183/platform.mk
```diff
@@ -0,0 +1,62 @@
...
+                   ${MTK_PLAT}/common/mtk_plat_common.c           \
+                   ${MTK_PLAT}/common/drivers/uart/8250_console.S \
+                   ${MTK_PLAT_SOC}/aarch64/plat_helpers.S         \
+                   ${MTK_PLAT_SOC}/aarch64/platform_common.c      \
+                   ${MTK_PLAT_SOC}/plat_pm.c                      \
+                   ${MTK_PLAT_SOC}/plat_topology.c                \
+                   ${MTK_PLAT_SOC}/plat_mt_gic.c                  \
+                   ${MTK_PLAT_SOC}/bl31_plat_setup.c
+
+# STATIC_LIBS += ${MTK_PLAT_SOC}/drivers/sec/lib/sha2.a
```
antonio-nino-diaz-arm: Remove comment?

antonio-nino-diaz-arm:
IP Review +1

plat/mediatek/mt8183/platform.mk
```diff
@@ -0,0 +1,62 @@
...
+
+# Enable workarounds for selected Cortex-A53 erratas.
+ERRATA_A53_826319 := 0
+ERRATA_A53_836870 := 1
+ERRATA_A53_855873 := 1
+
+# indicate the reset vector address can be programmed
+PROGRAMMABLE_RESET_ADDRESS := 1
+
+# Disable the PSCI platform compatibility layer
```
antonio-nino-diaz-arm: Remove the ENABLE_PLAT_COMPAT lines, that is not used anymore.

plat/mediatek/mt8183/platform.mk
```diff
@@ -0,0 +1,62 @@
...
+                   ${MTK_PLAT_SOC}/plat_pm.c                      \
+                   ${MTK_PLAT_SOC}/plat_topology.c                \
+                   ${MTK_PLAT_SOC}/plat_mt_gic.c                  \
+                   ${MTK_PLAT_SOC}/bl31_plat_setup.c
+
+# STATIC_LIBS += ${MTK_PLAT_SOC}/drivers/sec/lib/sha2.a
+
+# Flag used by the MTK_platform port to determine the version of ARM GIC
+# architecture to use for interrupt management in EL3.
+ARM_GIC_ARCH := 2
```
antonio-nino-diaz-arm: I'm not sure if you actually need this, I think this was used in the deprecated GIC driver.

drivers/arm/cci/cci.c
```diff
@@ -21,6 +21,7 @@
 #define CCI400_PART_NUM		0x420
 #define CCI500_PART_NUM		0x422
 #define CCI550_PART_NUM		0x423
+#define MT8183_CCI_PART_NUM	0xB00
```
mtk09422: make 0xB00 aligned with 0x423

antonio-nino-diaz-arm: This is aligned, it looks like it isn't because of the GitHub UI. However the changes to this driver should be in a different patch.

hungte:
@jwerner-chromium

plat/mediatek/mt8183/include/plat_macros.S
```diff
@@ -0,0 +1,78 @@
...
+	/* ---------------------------------------------
+	 * The below macro prints out relevant GIC and
+	 * CCI registers whenever an unhandled exception
+	 * is taken in BL3-1.
+	 * Clobbers: x0 - x10, x16, x17, sp
+	 * ---------------------------------------------
+	 */
+	.macro plat_crash_print_regs
+	mov_imm x16, BASE_GICD_BASE
+	mov_imm x17, BASE_GICC_BASE
```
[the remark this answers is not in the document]
antonio-nino-diaz-arm: This is copy-pasted from Arm platforms code. If you take a look at the macros there it explicitly says that x16 and x17 are used to pass the base addresses. The registers aren't used as a callee-saved registers, so it is fine.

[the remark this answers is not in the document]
antonio-nino-diaz-arm: This is copy-pasted from Arm platforms code. If you take a look at the macros there it explicitly says that x16 and x17 are used to pass the base addresses. The registers aren't used as a callee-saved registers, so it is fine.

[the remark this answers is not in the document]
antonio-nino-diaz-arm: This is copy-pasted from Arm platforms code. If you take a look at the macros there it explicitly says that x16 and x17 are used to pass the base addresses. The registers aren't used as a callee-saved registers, so it is fine. However, I agree that it isn't a bad idea to change the registers, using x26 and x27 makes the code more intuitive.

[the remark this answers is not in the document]
antonio-nino-diaz-arm: This is copy-pasted from Arm platforms code. If you take a look at the macros there it explicitly says that x16 and x17 are used to pass the base addresses. The registers aren't used as a callee-saved registers, so it is fine. However, I agree that it isn't a bad idea to change the registers, using x26 and x27 makes the code more intuitive.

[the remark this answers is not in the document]
antonio-nino-diaz-arm: This is copy-pasted from Arm platforms code. If you take a look at the macros there it explicitly says that x16 and x17 are used to pass the base addresses. The registers aren't used as a callee-saved registers, so it is fine. However, I agree that it isn't a bad idea to change the registers, using x26 and x27 makes the code more intuitive.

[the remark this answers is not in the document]
antonio-nino-diaz-arm: This is copy-pasted from Arm platforms code. If you take a look at the macros there it explicitly says that x16 and x17 are used to pass the base addresses. The registers aren't used as a callee-saved registers, so it is fine. However, I agree that it isn't a bad idea to change the registers, using x26 and x27 makes the code more intuitive.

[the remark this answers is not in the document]
antonio-nino-diaz-arm: This is copy-pasted from Arm platforms code. If you take a look at the macros there it explicitly says that x16 and x17 are used to pass the base addresses. The registers aren't used as a callee-saved registers, so it is fine. However, I agree that it isn't a bad idea to change the registers, using x26 and x27 makes the code more intuitive.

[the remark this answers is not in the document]
antonio-nino-diaz-arm: This is copy-pasted from Arm platforms code. If you take a look at the macros there it explicitly says that x16 and x17 are used to pass the base addresses. The registers aren't used as a callee-saved registers, so it is fine. However, I agree that it isn't a bad idea to change the registers, using x26 and x27 makes the code more intuitive.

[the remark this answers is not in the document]
antonio-nino-diaz-arm: This is copy-pasted from Arm platforms code. If you take a look at the macros there it explicitly says that x16 and x17 are used to pass the base addresses. The registers aren't used as a callee-saved registers, so it is fine. However, I agree that it isn't a bad idea to change the registers, using x26 and x27 makes the code more intuitive.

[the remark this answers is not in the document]
antonio-nino-diaz-arm: This is copy-pasted from Arm platforms code. If you take a look at the macros there it explicitly says that x16 and x17 are used to pass the base addresses. The registers aren't used as a callee-saved registers, so it is fine. However, I agree that it isn't a bad idea to change the registers, using x26 and x27 makes the code more intuitive.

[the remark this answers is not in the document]
antonio-nino-diaz-arm: This is copy-pasted from Arm platforms code. If you take a look at the macros there it explicitly says that x16 and x17 are used to pass the base addresses. The registers aren't used as a callee-saved registers, so it is fine. However, I agree that it isn't a bad idea to change the registers, using x26 and x27 makes the code more intuitive.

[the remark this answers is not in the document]
antonio-nino-diaz-arm: This is copy-pasted from Arm platforms code. If you take a look at the macros there it explicitly says that x16 and x17 are used to pass the base addresses. The registers aren't used as a callee-saved registers, so it is fine. However, I agree that it isn't a bad idea to change the registers, using x26 and x27 makes the code more intuitive.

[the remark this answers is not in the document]
antonio-nino-diaz-arm: This is copy-pasted from Arm platforms code. If you take a look at the macros there it explicitly says that x16 and x17 are used to pass the base addresses. The registers aren't used as a callee-saved registers, so it is fine. However, I agree that it isn't a bad idea to change the registers, using x26 and x27 makes the code more intuitive.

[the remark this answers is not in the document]
antonio-nino-diaz-arm: This is copy-pasted from Arm platforms code. If you take a look at the macros there it explicitly says that x16 and x17 are used to pass the base addresses. The registers aren't used as a callee-saved registers, so it is fine. However, I agree that it isn't a bad idea to change the registers, using x26 and x27 makes the code more intuitive.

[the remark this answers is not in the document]
antonio-nino-diaz-arm: This is copy-pasted from Arm platforms code. If you take a look at the macros there it explicitly says that x16 and x17 are used to pass the base addresses. The registers aren't used as a callee-saved registers, so it is fine. However, I agree that it isn't a bad idea to change the registers, using x26 and x27 makes the code more intuitive.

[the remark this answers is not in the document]
antonio-nino-diaz-arm: Oh, ok, I see what you mean. Sure, if you are using the multi console for crash output then you need to use callee-saved registers, I didn't realize about that problem because only rockchip platforms do that, as far as I know. However, will it be enough to move x16 and x17? I haven't checked all the code in this macro, but I'd be surprised if x16 and x17 were the only problems here.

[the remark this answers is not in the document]
antonio-nino-diaz-arm: In general we probably have to do a cleanup of all assembly code and make it AAPCS compliant. I think that in the crash handling code we didn't do it because we never assumed that we would need to use more than really simple functions after a crash.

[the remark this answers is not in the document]
jwerner-chromium: Yeah, you're right, the code has more problems than that. I think the rest of it barely manages to scoot around conflicts. I tried to pick the extra registers for multi-console carefully so that most code that was written for the existing crash handlers should still work (i.e. taking the extra registers from the back of the caller-saved block, x15, x14, x13, which are less likely to have already been used).

But the other problem is (if I understand the ABI right), x16 and x17 are reserved for the linker and could get clobbered at any BL instruction, regardless of the code on both sides. The linker is always allowed to automatically insert a veneer in the call which clobbers these registers. Maybe the situations where veneers are needed just don't happen inside TF in practice, but the ABI doesn't specify them so it still seems like a very bad idea to trust these to hold their value across branch instructions ever.

(It's also not great that this code has been copied and mangled across so many platforms by now... it would really be better to maintain a single version in include/drivers/arm/gic_helpers.S or something like that.)

drivers/arm/cci/cci.c
```diff
@@ -92,6 +93,7 @@ static int get_slave_ports(unsigned int part_num)
 	case CCI400_PART_NUM:
 		num_slave_ports = CCI400_SLAVE_PORTS;
 		break;
+	case MT8183_CCI_PART_NUM:
```
soby-mathew: Need a comment here `/* Fallthrough */`

plat/mediatek/mt8183/include/plat_macros.S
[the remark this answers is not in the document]
soby-mathew: > In general we probably have to do a cleanup of all assembly code and make it AAPCS compliant.

Just to clarify, AAPCS is only possible where you have a backing store (stack) available when functions calls are nested more than a level. Since there are many places (like crash reporting, early init code) where stack is not available/reliable, we have rely on custom register clobber list. 

> Some parts of the multi-console core explicitly make use of these as scratch registers, so that would definitely blow up.

I see that the console_cbmc_putc() uses x16. The crash console framework had the assumption that the console_core_* functions would only use a couple of registers but MULTI-CONSOLE changed that.

Usually, when not following AAPCS, the convention is that the top level functions use the higher registers while the inner functions works on a smaller subset with the leaf function using the minimal and lowest set of registers.  WIth MULTI-CONSOLE, this is no longer true and, as you say, it is very difficult to track which register to use.

I think it is about time the crash framework is re-implemented for the MULTI-CONSOLE so that x0-x17 can be used by multi-console drivers. I will create a tf-issue for this.

> But the other problem is (if I understand the ABI right), x16 and x17 are reserved for the linker and could get clobbered at any BL instruction, regardless of the code on both sides

I agree it is not safe to use these registers across function calls and should be changed. Factoring out the useful bits in ARM platform to comon code is something we have started doing and this is another bit that can be in common code.

plat/mediatek/mt8183/include/platform_def.h
```diff
@@ -0,0 +1,322 @@
+/*
+ * Copyright (c) 2018, ARM Limited and Contributors. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef __PLATFORM_DEF_H__
+#define __PLATFORM_DEF_H__
```
soby-mathew: nit: The guard macros need not have the double underscores `__` .

plat/mediatek/mt8183/plat_pm.c
```diff
@@ -0,0 +1,45 @@
...
+#include <scu.h>
+#include <mtk_plat_common.h>
+#include <power_tracer.h>
+#include <plat_private.h>
+
+/*******************************************************************************
+ * Export the platform handlers to enable psci to invoke them
+ ******************************************************************************/
+static const plat_pm_ops_t plat_plat_pm_ops = {
+	.affinst_standby		= NULL,
```
soby-mathew: This struct is not present anymore. Please use `plat_psci_ops`.

plat/mediatek/mt8183/include/plat_macros.S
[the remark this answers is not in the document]
jwerner-chromium: > Just to clarify, AAPCS is only possible where you have a backing store (stack) available when functions calls are nested more than a level.

I wouldn't completely agree with that. What you're saying is true if you want every individual function to be AAPCS compliant, which we just can't really do if we want a console framework that doesn't require a stack. But you can still have an API that is AAPCS compliant from the outside even though internally it is implemented in assembler and relying on custom ABIs between its individual functions. MULTI_CONSOLE works that way (and technically, the old console framework already worked that way too), and it has to because all the API functions can be called from either assembly or C runtime contexts. For example, the normal C function putchar() calls the assembly function console_putc(), so the latter needs to be AAPCS compliant on the outside for this to work.

> I think it is about time the crash framework is re-implemented for the MULTI-CONSOLE so that x0-x17 can be used by multi-console drivers. I will create a tf-issue for this.

Not quite sure what you mean by that. Are you suggesting that MULTI_CONSOLE framework functions should be using higher registers (e.g. x29) instead? Since it has no stack to save them, this would break calls from normal C functions as described above, so that seems like a bad idea.

Actually, I just noticed that someone has already done this: https://github.com/ARM-software/arm-trusted-firmware/commit/d35cc347. Unfortunately I didn't notice that change when it went in. This presumably breaks all MULTI_CONSOLE crash reporting (do_crash_reporting uses SP as a scratch register) which was one of the main use cases I implemented the framework for, so I'd prefer if we could undo that.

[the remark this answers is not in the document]
antonio-nino-diaz-arm: The problem with undoing that commit is that now there is one console driver written in C (in a AArch32 port, not AArch64). The console framework needs to use AAPCS at least in AArch32, so it would be unexpected if it didn't in AArch64.

https://github.com/ARM-software/arm-trusted-firmware/blob/9c10e3485b10ece95f9af8a6af79558397ee1d78/drivers/imx/uart/imx_uart.c

plat/mediatek/mt8183/include/mcucfg.h
```diff
@@ -0,0 +1,259 @@
...
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef __MT8183_MCUCFG_H__
```
antonio-nino-diaz-arm: Header guards mustn't have ``__``.

plat/mediatek/mt8183/include/mt_gic_v3.h
```diff
@@ -0,0 +1,153 @@
+/*
+ * Copyright (c) 2018, ARM Limited and Contributors. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef __MT_GIC_V3_H
```
antonio-nino-diaz-arm: Header guards mustn't have ``__``.

plat/mediatek/mt8183/include/plat_private.h
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright (c) 2018, ARM Limited and Contributors. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef __PLAT_PRIVATE_H__
```
antonio-nino-diaz-arm: Header guards mustn't have ``__``.

plat/mediatek/mt8183/include/power_tracer.h
```diff
@@ -0,0 +1,43 @@
...
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef __POWER_TRACER_H__
```
antonio-nino-diaz-arm: Header guards mustn't have ``__``.

plat/mediatek/mt8183/include/scu.h
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright (c) 2015, ARM Limited and Contributors. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef __SCU_H__
```
antonio-nino-diaz-arm: Header guards mustn't have ``__``.

plat/mediatek/mt8183/include/plat_macros.S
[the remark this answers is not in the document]
soby-mathew: > I wouldn't completely agree with that. What you're saying is true if you want every individual function to be AAPCS compliant, which we just can't really do if we want a console framework that doesn't require a stack.  But you can still have an API that is AAPCS compliant from the outside even though internally it is implemented in assembler and relying on custom ABIs between its individual functions.

I agree. That's what I meant. Not all the assembly functions can be made AAPCS. As long as the APIs exported are AAPCS, it can be invoked from C.

> Not quite sure what you mean by that. Are you suggesting that MULTI_CONSOLE framework functions should be using higher registers (e.g. x29) instead? Since it has no stack to save them, this would break calls from normal C functions as described above, so that seems like a bad idea.

What I meant is rework the to crash reporting to use the higher register set, and the MULTI_CONSOLE stays as it is. The MULTI_CONSOLE API should remain AAPCS compliant.

> Actually, I just noticed that someone has already done this: d35cc34. Unfortunately I didn't notice that change when it went in.

Sorry, didn't realise that. Unfortunately interaction between MULTI_CONSOLE and crash reporting is something that I didn't understand too well till recent. Since we override the plat_crash_console_* functions, we couldn't detect the problem on ARM reference platforms.

It is a an useful capability to allow platforms to implement their console drivers in C, as long as they don't use it for crash reporting which is what the patch is allowing to do.

@antonio-nino-diaz-arm and I have had a discussion and we think the solution here is to have a new API `crash_console_putc` which is equivalent `console_putc` but doesn't use the stack. The default `plat_crash_console_putc` will call into `crash_console_putc`.   The implementation of `crash_console_putc` can be simpler than `console_putc` because it doesn't need to check for the 
`console_state` as it is always invoked in the crash state.  This means the `plat_crash_console_init` can be empty. Similarly `crash_console_flush` can be implemented.

Please let us know of your opinion.

plat/mediatek/mt8183/plat_mt_gic.c
```diff
@@ -0,0 +1,908 @@
...
+
+#define ICC_GRPEN1_EL1  S3_0_c12_c12_7
+DEFINE_RENAME_SYSREG_WRITE_FUNC(icc_igrpen1_el1, ICC_GRPEN1_EL1)
+
+#define MPIDR_LEVEL_BITS_SHIFT	3
+#define MPIDR_LEVEL_BITS		(1 << MPIDR_LEVEL_BITS_SHIFT)
+#define MPIDR_LEVEL_MASK		((1 << MPIDR_LEVEL_BITS) - 1)
+
+#define MPIDR_LEVEL_SHIFT(level) \
+	(((1 << level) >> 1) << MPIDR_LEVEL_BITS_SHIFT)
```
antonio-nino-diaz-arm: Add parentheses to ``level``.

plat/mediatek/mt8183/plat_mt_gic.c
```diff
@@ -0,0 +1,908 @@
...
+
+#define MPIDR_AFFINITY_LEVEL(mpidr, level) \
+	((mpidr >> MPIDR_LEVEL_SHIFT(level)) & MPIDR_LEVEL_MASK)
```
antonio-nino-diaz-arm: Add parentheses to ``mpidr``.

plat/mediatek/mt8183/aarch64/plat_helpers.S
```diff
@@ -0,0 +1,62 @@
...
+
+	/* ---------------------------------------------
+	 * int plat_crash_console_flush()
+	 * Function to force a write of all buffered
+	 * data that has not been output.
+	 * Out : return -1 on error else return 0.
+	 * Clobber list : r0
+	 * ---------------------------------------------
+	 */
+func plat_crash_console_flush
```
antonio-nino-diaz-arm: You should be using the multi console API rather than the old one.

When you migrate to it, just remove the functions ``plat_crash_console_*`` from this file and add ``plat/common/aarch64/crash_console_helpers.S`` to the makefile, as @jwerner-chromium suggested (he didn't know about some changes in the code so he pointed you at a different file).

plat/mediatek/mt8183/aarch64/platform_common.c
```diff
@@ -0,0 +1,104 @@
...
+	{ 0 }
+
+};
+
+
+/*******************************************************************************
+ * Macro generating the code for the function setting up the pagetables as per
+ * the platform memory map & initialize the mmu, for the given exception level
+ ******************************************************************************/
+#define DEFINE_CONFIGURE_MMU_EL(_el)					\
```
antonio-nino-diaz-arm: You only have one version of this, remove the define.

plat/mediatek/mt8183/bl31_plat_setup.c
```diff
@@ -0,0 +1,111 @@
...
+#include <platform_def.h>
+
+static entry_point_info_t bl32_ep_info;
+static entry_point_info_t bl33_ep_info;
+
+static void platform_setup_cpu(void)
+{
+	mmio_write_32((uintptr_t)&mt8183_mcucfg->mp0_rw_rsvd0, 0x00000001);
+
+#if PLATFORM_DEBUG
```
antonio-nino-diaz-arm: ``PLATFORM_DEBUG`` isn't defined anywhere, as far as I see.

plat/mediatek/mt8183/platform.mk
```diff
@@ -0,0 +1,52 @@
...
+# Enable workarounds for selected Cortex-A53 erratas.
+ERRATA_A53_826319 := 0
+ERRATA_A53_836870 := 1
+ERRATA_A53_855873 := 1
+
+# indicate the reset vector address can be programmed
+PROGRAMMABLE_RESET_ADDRESS := 1
+
+COLD_BOOT_SINGLE_CPU := 1
+
```
antonio-nino-diaz-arm: You need to use ``MULTI_CONSOLE_API := 1``.

plat/mediatek/mt8183/include/platform_def.h
```diff
@@ -0,0 +1,311 @@
...
+#define PLAT_MAX_OFF_STATE		U(2)
+
+
+#define PLATFORM_CACHE_LINE_SIZE        64
+#define PLATFORM_SYSTEM_COUNT           1
+#define PLATFORM_CLUSTER_COUNT          2
+#define PLATFORM_CLUSTER0_CORE_COUNT    4
+#define PLATFORM_CLUSTER1_CORE_COUNT    4
+#define PLATFORM_CORE_COUNT             (PLATFORM_CLUSTER1_CORE_COUNT + \
+                                         PLATFORM_CLUSTER0_CORE_COUNT)
```
antonio-nino-diaz-arm: nit: use tabs in lines 267, 270 and 271 to remove the checkpatch.pl errors. If you want, you can also remove more warnings from the rest of the patch.

plat/mediatek/mt8183/include/power_tracer.h
```diff
@@ -0,0 +1,19 @@
...
+#define CPU_UP		0
+#define CPU_DOWN	1
+#define CPU_SUSPEND	2
+#define CLUSTER_UP	3
+#define CLUSTER_DOWN	4
+#define CLUSTER_SUSPEND	5
+
+void trace_power_flow(unsigned long mpidr, unsigned char mode);
+
+#endif
```
antonio-nino-diaz-arm: #endif /* POWER_TRACER_H */

plat/mediatek/mt8183/include/mt_gic_v3.h
```diff
@@ -0,0 +1,154 @@
...
+}
+
+static inline void gicd_v3_set_irouter(unsigned int base, unsigned int id, uint64_t aff)
+{
+    unsigned int reg = base + GICD_V3_IROUTER + (id*8);
+
+    mmio_write_64(reg, aff);
+}
+
+#endif /* end of MT_GIC_V3_H */
```
antonio-nino-diaz-arm: `` #endif /* MT_GIC_V3_H */ ``

plat/mediatek/mt8183/include/platform_def.h
```diff
@@ -0,0 +1,311 @@
...
+#define MT_EDGE_SENSITIVE    1
+#define MT_LEVEL_SENSITIVE    0
+
+/*******************************************************************************
+ * UART related constants
+ ******************************************************************************/
+#define UART0_BASE    (IO_PHYS + 0x01002000)
+
+#define UART_BAUDRATE (115200)
+/* #define UART_CLOCK    (10000000) */
```
antonio-nino-diaz-arm: Why is this commented? If it isn't used, remove it.

plat/mediatek/mt8183/include/scu.h
```diff
@@ -0,0 +1,13 @@
...
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef SCU_H
+#define SCU_H
+
+void disable_scu(unsigned long mpidr);
+void enable_scu(unsigned long mpidr);
+
+#endif
```
antonio-nino-diaz-arm: ``#endif /* SCU_H */``

plat/mediatek/mt8183/plat_mt_gic.c
```diff
@@ -0,0 +1,908 @@
...
+
+	for (i = 1; i < DIV_ROUND_UP(gic_irqs, 32); i++)
+		mmio_write_32(dist_base + GICD_ISENABLER + i * 4, gic_data[0].saved_enable[i]);
+
+	for (i = 1; i < DIV_ROUND_UP(gic_irqs, 32); i++)
+		mmio_write_32(dist_base + GICD_IGROUPR + i * 4, gic_data[0].saved_group[i]);
+
+	for (i = 1; i < DIV_ROUND_UP(gic_irqs, 32); i++)
+		mmio_write_32(dist_base + GICE_V3_IGRPMOD0 + i * 4, gic_data[0].saved_grpmod[i]);
+#if CFG_MICROTRUST_TEE_SUPPORT
```
antonio-nino-diaz-arm: Where is this build option set?

plat/mediatek/mt8183/plat_mt_gic.c
```diff
@@ -0,0 +1,908 @@
...
+	/* Init cpu logic map to avoid unnecessary SGI interrupt
+		to cpu0 if WDT happens before cpu1~cpu7 on */
+	for (i = 0 ; i < sizeof(cpu_logical_map)/sizeof(uint64_t); i++)
+		cpu_logical_map[i] = CPU_LOGIC_MAP_RESET_VAL;
+	gicd_base = MT_GIC_BASE;
+	gic_distif_init(gicd_base);
+	gic_cpuif_init();
+
+	/* Register WDT handler */
+	/* request_fiq(WDT_IRQ_BIT_ID, aee_wdt_dump_all_core, INT_EDGE_FALLING, INTR_GROUP0, NULL); */
```
antonio-nino-diaz-arm: Why is this commented out?

plat/mediatek/mt8183/plat_mt_gic.c
```diff
@@ -0,0 +1,908 @@
...
+			rc |= 0xffff << 14;
+		} else {
+			rc |= (1<<cpu) << 14;
+		}
+	}
+
+	return rc;
+}
+
+void dump_gic(void)
```
antonio-nino-diaz-arm: This function isn't used anywhere, why is it here?

plat/mediatek/mt8183/plat_pm.c
```diff
@@ -0,0 +1,76 @@
...
+
+/*******************************************************************************
+ * MTK_platform handler called when an affinity instance is about to be turned
+ * on. The level and mpidr determine the affinity instance.
+ ******************************************************************************/
+static uintptr_t secure_entrypoint;
+
+static const plat_psci_ops_t plat_plat_pm_ops = {
+	.cpu_standby			= NULL,
+	.pwr_domain_on			= NULL,
```
antonio-nino-diaz-arm: How do the secondary cores come online if all PSCI hooks are NULL?

```
#define PLATFORM_SYSTEM_COUNT           1
#define PLATFORM_CLUSTER_COUNT          2
#define PLATFORM_CLUSTER0_CORE_COUNT    4
#define PLATFORM_CLUSTER1_CORE_COUNT    4
```

antonio-nino-diaz-arm:
Any updates regarding this PR?

plat/mediatek/mt8183/plat_pm.c
kennyliang-mtk: Multi-cores function is under development. This patch is only for basic setup with one primary CPU.

antonio-nino-diaz-arm:
ping

antonio-nino-diaz-arm:
The multi console issues should be fixed now.

kennyliang-mtk:
Sorry, because I updated/merged the fork branch to latest, the new change is in another commit.

plat/mediatek/mt8183/aarch64/plat_helpers.S
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright (c) 2018, ARM Limited and Contributors. All rights reserved.
```
antonio-nino-diaz-arm: You can use MediaTek here if you want.

plat/mediatek/mt8183/aarch64/platform_common.c
```diff
@@ -0,0 +1,89 @@
+/*
+ * Copyright (c) 2018, ARM Limited and Contributors. All rights reserved.
```
antonio-nino-diaz-arm: You can use MediaTek here if you want.

plat/mediatek/mt8183/bl31_plat_setup.c
```diff
@@ -0,0 +1,123 @@
+/*
+ * Copyright (c) 2018, ARM Limited and Contributors. All rights reserved.
```
antonio-nino-diaz-arm: You can use MediaTek here if you want.

plat/mediatek/mt8183/bl31_plat_setup.c
antonio-nino-diaz-arm: This is still here, is there a reason for it?

plat/mediatek/mt8183/drivers/mcsi/mcsi.c
```diff
@@ -0,0 +1,223 @@
+/*
+ * Copyright (c) 2018, ARM Limited and Contributors. All rights reserved.
```
antonio-nino-diaz-arm: You can use MediaTek here if you want.

plat/mediatek/mt8183/drivers/mcsi/mcsi.c
```diff
@@ -0,0 +1,223 @@
...
+	int i;
+
+	assert(cci_base);
+	assert(num_cci_masters < MAX_CLUSTERS);
+
+	cci_base_addr = cci_base;
+
+	for (i = 0; i < num_cci_masters; i++)
+		cci_cluster_ix_to_iface[i] = SLAVE_IFACE_OFFSET(i);
+	//mmio_write_32(cci_base_addr, config);
```
antonio-nino-diaz-arm: Remove commented out line.

plat/mediatek/mt8183/include/platform_def.h
```diff
@@ -0,0 +1,309 @@
...
+ ******************************************************************************/
+#define PLATFORM_LINKER_FORMAT      "elf64-littleaarch64"
+#define PLATFORM_LINKER_ARCH        aarch64
+
+/*******************************************************************************
+ * Generic platform constants
+ ******************************************************************************/
+
+/* Size of cacheable stacks */
+#if DEBUG_XLAT_TABLE
```
antonio-nino-diaz-arm: ``DEBUG_XLAT_TABLE`` doesn't exist anymore, please remove it.

plat/mediatek/mt8183/include/platform_def.h
```diff
@@ -0,0 +1,309 @@
...
+
+#define FIQ_SMP_CALL_SGI    13
+#define WDT_IRQ_BIT_ID      174
+#define ATF_LOG_IRQ_ID      277
+
+#define ATF_AMMS_IRQ_ID     338
+#define PCCIF1_IRQ0_BIT_ID  185
+#define PCCIF1_IRQ1_BIT_ID  186
+
+#define DEBUG_XLAT_TABLE    0
```
antonio-nino-diaz-arm: Remove ``DEBUG_XLAT_TABLE``.

antonio-nino-diaz-arm:
Oh, and squash your commits into one (removing the merge commit).

plat/mediatek/mt8183/drivers/spmc/mtspmc_private.h
```diff
@@ -0,0 +1,236 @@
+#ifndef _MTSPMC_PRIVATE_H_
```
antonio-nino-diaz-arm: This file needs a license header.

plat/mediatek/mt8183/drivers/spmc/mtspmc.h
```diff
@@ -0,0 +1,56 @@
+/*
+ * Copyright (c) 2015, ARM Limited and Contributors. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
```
antonio-nino-diaz-arm: Use SPDX.

plat/mediatek/mt8183/drivers/spmc/mtspmc.c
```diff
@@ -0,0 +1,381 @@
+/*
+ * Copyright (c) 2015,  ARM Limited and Contributors. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms,  with or without
```
antonio-nino-diaz-arm: Use SPDX.

plat/mediatek/mt8183/include/plat_dcm.h
```diff
@@ -0,0 +1,77 @@
+/*
+ * Copyright (c) 2015, ARM Limited and Contributors. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
```
antonio-nino-diaz-arm: Use SPDX.

plat/mediatek/mt8183/include/plat_pm.h
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright (c) 2014-2015, ARM Limited and Contributors. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
```
antonio-nino-diaz-arm: Use SPDX.

plat/mediatek/mt8183/include/spm.h
```diff
@@ -0,0 +1,1738 @@
+/*
+ * Copyright (c) 2015, ARM Limited and Contributors. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
```
antonio-nino-diaz-arm: Use SPDX.

plat/mediatek/mt8183/plat_dcm.c
```diff
@@ -0,0 +1,178 @@
+/*
+ * Copyright (c) 2013-2015, ARM Limited and Contributors. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
```
antonio-nino-diaz-arm: Use SPDX.

plat/mediatek/mt8183/plat_pm.c
```diff
@@ -0,0 +1,186 @@
...
+#define MTK_LOCAL_STATE_OFF     2
+
+static uintptr_t secure_entrypoint;
+
+static void mp1_L2_desel_config(void)
+{
+	mmio_write_64(MCUCFG_BASE + 0x2200, 0x2092c820);
+
+	isb();
+	dsb();
```
antonio-nino-diaz-arm: isb() after dsb()

kennyliang-mtk:
> Squash your commits in one.

Hi Antonio,

I cannot comment on github and always get “You can't comment at this time.” Message.

Do you mean that I should rebase my forked branch and combine my two commits into one?
Thank you.


Regards,
Kenny Liang

antonio-nino-diaz-arm:
> Do you mean that I should rebase my forked branch and combine my two commits into one?

Yes, sorry if I wasn't clear.

kennyliang-mtk:
Hi Antonio,
According to your comment, I rebased my branch and squashed my commits to one.
But it seems to cause chaos... Sorry, then how can I do?
Thank you.

dp-arm:
@kennyliang-mtk it seems something went wrong with the rebase on integration, after reworking can you rebase again?

kennyliang-mtk:
> @kennyliang-mtk it seems something went wrong with the rebase on integration, after reworking can you rebase again?

sorry to cause chaos.
I have rebased again and it seems fine.
Thanks.

kennyliang-mtk:
> > Do you mean that I should rebase my forked branch and combine my two commits into one?
> 
> Yes, sorry if I wasn't clear.

Finally done.
I have squashed my commits to one.
Thanks.

plat/mediatek/mt8183/drivers/gpio/mtgpio.c
```diff
@@ -0,0 +1,370 @@
...
+			 */
+			mmio_setbits_32(PUPD_addr[pin].addr, 1L << PUPD_offset[pin].offset);
+			mmio_clrbits_32(PUPD_addr[pin].addr, 1L << (PUPD_offset[pin].offset + 1));
+		} else {
+			/* For PULLEN + PULLSEL Type */
+			mmio_setbits_32(PULLEN_addr[pin].addr, 1L << PULLEN_offset[pin].offset);
+		}
+	} else if (en == GPIO_PULL_ENABLE_R0) {
+		assert(!(PUPD_offset[pin].offset == (int8_t)-1));
+		/* IOConfig does not support MWR operation, so use CLR + SET */
```
jwerner-chromium: I'm confused what this comment is trying to tell me. Can't you use mmio_clrsetbits_32?

plat/mediatek/mt8183/drivers/gpio/mtgpio.h
```diff
@@ -0,0 +1,171 @@
...
+
+#define MAX_GPIO_PIN    (MT_GPIO_BASE_MAX)
+/******************************************************************************
+* Enumeration for GPIO pin
+******************************************************************************/
+
+typedef enum GPIO_PIN {
+	GPIO_UNSUPPORTED = -1,
+
+	GPIO0, GPIO1, GPIO2, GPIO3, GPIO4, GPIO5, GPIO6, GPIO7,
```
jwerner-chromium: nit: This seems like a pretty pointless list of enums. Why not just use raw numbers?

plat/mediatek/mt8183/drivers/mcsi/mcsi.c
```diff
@@ -0,0 +1,223 @@
...
+
+	/* Wait for the dust to settle down */
+	while (mmio_read_32(cci_base_addr + SNP_PENDING_REG) >> SNP_PENDING)
+		;
+
+	return;
+
+}
+
+#ifdef ERRATA_MCSIB_SW
```
jwerner-chromium: There should be no need to #ifdef out prototypes usually, they don't hurt.

plat/mediatek/mt8183/drivers/pmic/pmic_wrap_init.c
```diff
@@ -0,0 +1,165 @@
+/*
```
jwerner-chromium: Isn't the pwrap/rtc interface 100% identical with MT8173? Please factor out code like this into plat/mediatek/common, don't duplicate it.

plat/mediatek/mt8183/include/plat_debug.h
```diff
@@ -0,0 +1,38 @@
...
+ * Copyright (c) 2018, ARM Limited and Contributors. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef PLATFORM_DEBUG_H
+#define PLATFORM_DEBUG_H
+
+#define readl(addr)		mmio_read_32((addr))
+#define writel(addr, val)	mmio_write_32((addr), (val))
```
jwerner-chromium: Please use mmio_read/write_32() directly in your code so we always have the same accessors everyone knows across the codebase, don't define your own names for everything.

plat/mediatek/mt8183/plat_dcm.c
```diff
@@ -0,0 +1,154 @@
...
+
+void plat_dcm_rgu_enable(void)
+{
+}
+
+void plat_dcm_big_core_sync(short on)
+{
+	/* Check if Big cluster power is existed */
+#ifdef DCM_DEBUG
+	console_init(gteearg.atf_log_port, UART_CLOCK, UART_BAUDRATE);
```
jwerner-chromium: What are you trying to do here?

If all you want is to output some stuff to the UART at runtime, you can just do

    console_switch_state(CONSOLE_FLAG_BOOT);
    ...print your stuff...
    console_switch_state(CONSOLE_FLAG_RUNTIME);

Since you have this pattern a dozen times, you'll probably want to define a helper function/macro for it (e.g. DCM_DEBUG_PRINT(...) or something).

plat/mediatek/mt8183/drivers/gpio/mtgpio.c
```diff
@@ -0,0 +1,370 @@
...
+
+	assert(pin < MAX_GPIO_PIN);
+
+	assert(dir < GPIO_DIR_MAX);
+
+	bit = pin % MAX_GPIO_REG_BITS;
+
+	/* To reduce memory usage, we don't use DIR_addr[] array*/
+	if (dir == GPIO_DIR_IN)
+		mmio_write_32(DIR_addr[pin].addr + 0x8, 1U << bit);
```
jwerner-chromium: I am very confused about these arrays (like DIR_addr)... where do they come from? I don't see them defined anywhere.

npoushin:
I can't see any files in my subsystem being modified here, so will not review unless tagged again.

antonio-nino-diaz-arm:
> I can't see any files in my subsystem being modified here, so will not review unless tagged again.

There was a version of this PR that wasn't rebased correctly, added a lot of patches to the PR, and the PR checker script decided to notify everyone in said list of patches. Don't worry about this PR.

plat/mediatek/mt8183/aarch64/platform_common.c
```diff
@@ -0,0 +1,89 @@
...
+
+};
+
+
+/*******************************************************************************
+ * Macro generating the code for the function setting up the pagetables as per
+ * the platform memory map & initialize the mmu, for the given exception level
+ ******************************************************************************/
+void plat_configure_mmu_el3(unsigned long total_base,
+							unsigned long total_size,
```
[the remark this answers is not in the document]
kennyliang-mtk: I think it's related to GitHub display. Here are 7 tabs(28 spaces) in my IDE.
The length equals to "void plat_configure_mmu_el3("

plat/mediatek/mt8183/aarch64/plat_helpers.S
```diff
@@ -0,0 +1,47 @@
...
+	 */
+func plat_my_core_pos
+	mrs     x0, mpidr_el1
+	and     x1, x0, #MPIDR_CPU_MASK
+	and     x0, x0, #MPIDR_CLUSTER_MASK
+	add     x0, x1, x0, LSR #6
+	ret
+endfunc plat_my_core_pos
+
+func read_cpuectlr
```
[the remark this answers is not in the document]
kennyliang-mtk: arm-trusted-firmware\plat\mediatek\mt8183\drivers\spmc\mtspmc.c

plat/mediatek/mt8183/aarch64/plat_helpers.S
```diff
@@ -0,0 +1,47 @@
...
+#include <asm_macros.S>
+#include <platform_def.h>
+
+	.globl plat_report_exception
+	.globl platform_is_primary_cpu
+	.globl plat_my_core_pos
+	.globl read_cpuectlr
+	.globl write_cpuectlr
+
+func platform_is_primary_cpu
```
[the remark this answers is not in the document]
kennyliang-mtk: replace it with plat_is_my_cpu_primary.

plat/mediatek/mt8183/drivers/gpio/mtgpio.c
kennyliang-mtk: arm-trusted-firmware\plat\mediatek\mt8183\drivers\gpio\mtgpio_cfg.h

plat/mediatek/common/drivers/uart/8250_console.S
```diff
@@ -89,10 +98,40 @@ func console_core_init
 core_init_fail:
 	mov	w0, wzr
 	ret
-endfunc console_core_init
+endfunc console_8250_core_init
+
+#if MULTI_CONSOLE_API
```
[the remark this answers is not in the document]
soby-mathew: As @jwerner-chromium mentions, we intend to remove the old console API soon.

soby-mathew:
jenkins: test this please

kennyliang-mtk:
> There are some comments which are not clarified yet.
Yes, they are still under processing by owners. Will update them once done internally.

plat/mediatek/common/drivers/uart/8250_console.S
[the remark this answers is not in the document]
kennyliang-mtk: Since this PR is for platform mt8183 initialization,
could removing old console API of old MTK platforms be done by another PR?

kennyliang-mtk:
> jenkins: test this please

Seems build failed? but I couldn't access ARM jenkins to see details.
I think the reason might be that the forked branch keeps in some older commit.
So we didn't encounter the error.

plat/mediatek/common/drivers/uart/8250_console.S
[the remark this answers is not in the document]
jwerner-chromium: Well, you're adding MULTI_CONSOLE_API support to the driver here, so it would be nicer if you could just switch all the way right now. Shouldn't it just need a two-line change each to make the two old SoCs conform to it?

On the other hand, if you follow my suggestion above to switch to the default 16550 driver rather than rolling your own, you may be able to reuse the legacy console support from there for now.

plat/mediatek/mt8183/aarch64/plat_helpers.S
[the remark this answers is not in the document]
jwerner-chromium: Note that you also don't really need to implement plat_is_my_cpu_primary if you don't intend to use BL1 and BL2 (which it looks like from this PR?).

plat/mediatek/mt8183/aarch64/plat_helpers.S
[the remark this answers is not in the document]
jwerner-chromium: Ah, I see.

Shouldn't this be added to include/arch/aarch64/arch_helpers.h, though? It's not an architectural register but it seems to have been present on all AArch64 Cortex-A cores for now, so it wouldn't make sense that every platform has to define its own version.

Also, I think you should probably name it read_cpuectlr_el1() to be consistent with the naming scheme for architectural registers.

plat/mediatek/mt8183/drivers/gpio/mtgpio.c
jwerner-chromium: Ah... don't know why I overlooked so much stuff here last time.

Those arrays are going to waste a huge amount of memory and binary size. Isn't there some more programmatic way you could implement this? It seems like most of the registers (mode, out, in) are completely 100% regular according to a single GPIO base, so you should be able to get the addresses and bit positions with a single division. For the pull it looks like you just have a handful of consecutive regions each mapped to their own base address, so can't you just make a gpio_get_iobase() function that has a big switch statement with ranges or something?

plat/mediatek/mt8183/aarch64/platform_common.c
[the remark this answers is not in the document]
Yann-lms: Tabs are 8 spaces in TF-A like in Linux kernel. If you want to align with the opening parenthesis, you should have 3 tabs and 4 spaces.

antonio-nino-diaz-arm:
jenkins: test this please

antonio-nino-diaz-arm:
Regarding the multi console, I think this shouldn't be a blocker of this PR. As long as it supports the multi console, it's fine. It would be a problem if it didn't support it, but right now the only thing we would have to do is to remove the old console code, which is a trivial change.

antonio-nino-diaz-arm:
Could you do a test build with ``ERROR_DEPRECATED=1`` and fix the issues that appear?

antonio-nino-diaz-arm:
I've tried to build this platform with ``CROSS_COMPILE=aarch64-linux-gnu- make DEBUG=1 PLAT=mt8183 -j`` and I get some compilation errors. Which version of GCC are you using?

antonio-nino-diaz-arm:
The static checks have passed. The other failures happen because this patch is not based in the latest master and it is missing some configs.

kennyliang-mtk:
> Could you do a test build with `ERROR_DEPRECATED=1` and fix the issues that appear?

Done. Remove "BL31_SOURCES    += drivers/console/aarch64/console.S" from platform.mk.

kennyliang-mtk:
> I've tried to build this platform with `CROSS_COMPILE=aarch64-linux-gnu- make DEBUG=1 PLAT=mt8183 -j` and I get some compilation errors. Which version of GCC are you using?

I build within coreboot environment. Sorry, I am not quite sure but I think it is
"chroot/opt/coreboot-sdk/bin/aarch64-elf-gcc-8.2.0"

plat/mediatek/common/drivers/uart/8250_console.S
```diff
@@ -25,7 +34,7 @@
 	 * Clobber list : x1, x2, x3
 	 * -----------------------------------------------
 	 */
-func console_core_init
+func console_8250_core_init
```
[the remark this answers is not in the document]
kennyliang-mtk: I don't think it's appropriate to use 16550_console.S. In addition to the difference in highspeed settings, there are other differences: like FCR registers, and the interface to log is not the same.

kennyliang-mtk:
> The static checks have passed. The other failures happen because this patch is not based in the latest master and it is missing some configs.

Where can I see the failure log?
It's odd that I checked out ARM latest integration branch and add mt8183 platform.
I didn't find any build error. If there is failure log, it could help a lot.

kennyliang-mtk:
> > The static checks have passed. The other failures happen because this patch is not based in the latest master and it is missing some configs.
> 
> Where can I see the failure log?
> It's odd that I checked out ARM latest integration branch and add mt8183 platform.
> I didn't find any build error. If there is failure log, it could help a lot.

Or should I rebase my forked branch to be based on latest master branch?

plat/mediatek/common/drivers/uart/8250_console.S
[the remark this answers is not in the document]
jwerner-chromium: Is there a significant difference in FCR? You're just clearing the FIFOs (which is something we could do on all boards, although it also shouldn't be necessary), and the 16550 driver is setting a DMA bit which I think doesn't make a difference either way. Not sure what you mean by "the interface to log"?

I think it's valuable to avoid code duplication wherever possible, so if possible implementing this driver as a thin wrapper (or a compile-time flag option) to the 16550 driver would be preferable.

antonio-nino-diaz-arm:
> Or should I rebase my forked branch to be based on latest master branch?

Yes, that will take care of the CI issues.

plat/mediatek/mt8183/platform.mk
```diff
@@ -0,0 +1,71 @@
...
+                 -I${MTK_PLAT_SOC}/drivers/rtc/                   \
+                 -I${MTK_PLAT_SOC}/include/                       \
+                 -Iinclude/common/                                \
+                 -Iinclude/drivers/                               \
+                 -Idrivers/arm/gic/v3/
+
+PLAT_BL_COMMON_SOURCES := lib/xlat_tables/aarch64/xlat_tables.c	      \
+                          lib/xlat_tables/xlat_tables_common.c        \
+                          plat/common/plat_gicv2.c                    \
+                          plat/common/aarch64/platform_helpers.S      \
```
antonio-nino-diaz-arm: Remove ``plat/common/aarch64/platform_helpers.S`` from here. It's already added from the main Makefile, and it makes make give the following warnings:
```
Makefile:776: warning: overriding recipe for target 'build/mt8183/debug/bl31/platform_helpers.o'
Makefile:776: warning: ignoring old recipe for target 'build/mt8183/debug/bl31/platform_helpers.o'
```

plat/mediatek/mt8183/platform.mk
```diff
@@ -0,0 +1,71 @@
...
+                 -I${MTK_PLAT}/common/drivers/uart/               \
+                 -I${MTK_PLAT_SOC}/drivers/                       \
+                 -I${MTK_PLAT_SOC}/drivers/spmc/                  \
+                 -I${MTK_PLAT_SOC}/drivers/gpio/                  \
+                 -I${MTK_PLAT_SOC}/drivers/pmic/                  \
+                 -I${MTK_PLAT_SOC}/drivers/rtc/                   \
+                 -I${MTK_PLAT_SOC}/include/                       \
+                 -Iinclude/common/                                \
+                 -Iinclude/drivers/                               \
+                 -Idrivers/arm/gic/v3/
```
antonio-nino-diaz-arm: Also, remove the following lines:
```
                 -Iinclude/common/                                \
                 -Iinclude/drivers/                               \
                 -Idrivers/arm/gic/v3/
```

antonio-nino-diaz-arm:
As part of the migration to trustedfirmware.org https://github.com/ARM-software/tf-issues/issues/681 we have switched the main development repository to Gerrit (https://review.trustedfirmware.org/). It is unfortunate that the switch has happened while your PR was being reviewed, but we don't intend to merge more patches through GitHub to avoid confusion.

Could you please follow the instructions in https://github.com/ARM-software/arm-trusted-firmware/blob/master/contributing.rst to continue with the review? We're sorry for the inconvenience.

sandrine-bailleux-arm:
In addition to the [contribution guidelines](https://github.com/ARM-software/arm-trusted-firmware/blob/master/contributing.rst), the following [Getting starting guide](https://developer.trustedfirmware.org/w/tf_a/gerrit-getting-started/) might help in this process.

plat/mediatek/common/drivers/uart/8250_console.S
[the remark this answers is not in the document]
kennyliang-mtk: Have switched to 16550 uart driver.

plat/mediatek/mt8183/drivers/gpio/mtgpio.h
kennyliang-mtk: We just want to standardize  the use of parameters and GPIOx seems more meaningful than 0,1,2... And ENUM type doesn't waste more binary size.

plat/mediatek/common/drivers/uart/8250_console.S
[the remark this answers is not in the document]
kennyliang-mtk: > As part of the migration to trustedfirmware.org [ARM-software/tf-issues#681](https://github.com/ARM-software/tf-issues/issues/681) we have switched the main development repository to Gerrit (https://review.trustedfirmware.org/). It is unfortunate that the switch has happened while your PR was being reviewed, but we don't intend to merge more patches through GitHub to avoid confusion.
> 
> Could you please follow the instructions in https://github.com/ARM-software/arm-trusted-firmware/blob/master/contributing.rst to continue with the review? We're sorry for the inconvenience.

Have just updated the commit according to your suggestion and rebase forked branch on latest master branch.
Then, I will push the commit and continue with review on https://review.trustedfirmware.org/.
Thanks.

kennyliang-mtk:
> As part of the migration to trustedfirmware.org [ARM-software/tf-issues#681](https://github.com/ARM-software/tf-issues/issues/681) we have switched the main development repository to Gerrit (https://review.trustedfirmware.org/). It is unfortunate that the switch has happened while your PR was being reviewed, but we don't intend to merge more patches through GitHub to avoid confusion.
> 
> Could you please follow the instructions in https://github.com/ARM-software/arm-trusted-firmware/blob/master/contributing.rst to continue with the review? We're sorry for the inconvenience.

Have just updated the commit according to your suggestion and rebase forked branch on latest master branch.
Then, I will push the commit and continue with review on https://review.trustedfirmware.org/.
Thanks.

antonio-nino-diaz-arm:
Review continued here: https://review.trustedfirmware.org/#/c/TF-A/trusted-firmware-a/+/860/

plat/mediatek/common/drivers/uart/8250_console.S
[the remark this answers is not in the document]
macpaul-lin-mtk: @jwerner-chromium and @kennyliang-mtk @mtk-rex-bc-chen 
We've found 16550 drivers seems cannot do baud rate at 921600, however, 8250 can do 921600. We've got many request hope MediaTek could support 921600 in TF-A. If there is no any concern, MediaTek will submit patch for switching driver back to 8250. At least, we hope to add a compile option in Makefile to enable 8250 and disable 16550.
Could you provide any comment? Thanks.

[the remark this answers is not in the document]
jwerner-chromium: Sorry, I'm not sure I really understand. We're talking about two driver implementations that would each talk to the same hardware peripheral, right? I see no fundamental reason why your implementation should be able to support a baud rate that the generic 16550 driver cannot.

Please identify what exactly your driver implementation is doing differently from the generic 16550 driver that allows it to support this baud rate. Once we know that, we can hopefully find a way to fix the generic driver without requiring diverging implementations.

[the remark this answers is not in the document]
hungte: @macpaul-lin-mtk The real 16550 is actually a successor of 8250 IC, so I'd assume your '8250' (driver) is 'generic 8250-like driver that supports MTK extensions'. Like Julius said, in that case you should extend the generic UART driver (although it was named as 16550, it is actually 8250/16550 shared driver) so it can support the highspeed baudrate.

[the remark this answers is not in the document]
macpaul-lin-mtk: @hungte @jwerner-chromium 
Thanks for the clarification, I'll inform our colleague to extend 'generic 8250-like driver that supports MTK extensions' on 16550. I've found there might be some register conflict on offset 0x24 ((HIGHSEEED/FIFO) but I'm not familiar with this. Let's see review the patch from MediaTek later for 16550. Thanks .

PR #1636, opened 2018-10-17
 Deprecate weak crash console functions 

antonio-nino-diaz-arm:
The default behaviour of the plat_crash_console_xxx functions isn't obvious to someone that hasn't read all the documentation. As they are not mandatory, it is unlikely that the code will be checked when doing a platform port, which may mean that some platforms may not have crash console support at all.

The idea of this patch is to force platform maintainers to decide how the crash console has to behave so that the final behaviour isn't unexpected.

This PR also fixes all upstream platforms that need it.

### DIFF
--- a/docs/porting-guide.rst
+++ b/docs/porting-guide.rst
@@ -2554,8 +2554,13 @@ NOTE: This section assumes that your platform is enabling the MULTI_CONSOLE_API
 flag in its platform.mk. Not using this flag is deprecated for new platforms.
 
 BL31 implements a crash reporting mechanism which prints the various registers
-of the CPU to enable quick crash analysis and debugging. By default, the
-definitions in ``plat/common/aarch64/platform\_helpers.S`` will cause the crash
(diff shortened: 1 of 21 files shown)


glneo:
Clearly we were not exercising this crash path very well, what is the recommenced method for testing? Do we just inject some early fault into the code and look for the console output? What are we expecting to see here?

antonio-nino-diaz-arm:
@glneo: I would just go to ``bl1_entrypoint`` or similar and try to execute an undefined instruction or something before ``bl1_early_platform_setup``. That will test that the crash console is working even before initializing the normal console. In order to test it after initializing the console and MMU, something like ``*(int *)0 = 0;`` should do the trick. In your case initializing the console shouldn't matter, though. You should see a dump of all registers.

antonio-nino-diaz-arm:
jenkins: test this please

antonio-nino-diaz-arm:
The only failure is a MISRA defect.

PR #1637, opened 2018-10-17
rpi3: Add mem reserve region to DTB if present

antonio-nino-diaz-arm:
When a device tree blob is pressent at a know address, instead of, for example, relying on the user modifying the Linux command line to warn about the memory reserved for the Trusted Firmware, pass it on the DTB.

In order to add this information to the DTB the code looks for an already existent reserve region (which is most likely the one reserved for the official EL3 bootstrap) and reuses it for the Trusted Firmware.

This system replaces the previous one consisting on adding ``memmap=16M$256M`` to the Linux command line. It's also meant to be used by U-Boot and any other bootloader that understands DTB files.

### DIFF
--- a/docs/plat/rpi3.rst
+++ b/docs/plat/rpi3.rst
@@ -133,9 +133,12 @@ secure platform!
     0x40000000 +-----------------+
 
 The area between **0x10000000** and **0x11000000** has to be manually protected
-so that the kernel doesn't use it. That is done by adding ``memmap=16M$256M`` to
-the command line passed to the kernel. See the `Setup SD card`_ instructions to
-see how to do it.
+so that the kernel doesn't use it. The current port tries to modify the live DTB
+to add a memreserve region that reserves the previously mentioned area.
+
+If this is not possible, the user may manually add ``memmap=16M$256M`` to the
+command line passed to the kernel in ``cmdline.txt``. See the `Setup SD card`_
+instructions to see how to do it. This system is strongly discouraged.
 
 The last 16 MiB of DRAM can only be accessed by the VideoCore, that has
 different mappings than the Arm cores in which the I/O addresses don't overlap
@@ -384,13 +387,8 @@ untouched). They have been tested with the image available in 2018-03-13.
    bootloader will look for a file called ``armstub8.bin`` and load it at
    address **0x0** instead of a predefined one.
 
-4. Open ``cmdline.txt`` and add ``memmap=16M$256M`` to prevent the kernel from
-   using the memory needed by TF-A. If you want to enable the serial port
-   "Mini UART", make sure that this file also contains
+4. To enable the serial port "Mini UART" in Linux, open ``cmdline.txt`` and add
    ``console=serial0,115200 console=tty1``.
-
-   Note that the 16 MiB reserved this way won't be available for Linux, the same
-   way as the memory reserved in DRAM for the GPU isn't available.
 
 5. Open ``config.txt`` and add the following lines at the end (``enable_uart=1``
    is only needed to enable debugging through the Mini UART):
--- a/plat/rpi3/platform.mk
+++ b/plat/rpi3/platform.mk
@@ -4,12 +4,16 @@
 # SPDX-License-Identifier: BSD-3-Clause
 #
 
+include lib/libfdt/libfdt.mk
+include lib/xlat_tables_v2/xlat_tables.mk
+
 PLAT_INCLUDES		:=	-Iinclude/common/tbbr			\
 				-Iplat/rpi3/include
 
 PLAT_BL_COMMON_SOURCES	:=	drivers/console/aarch64/console.S	\
 				drivers/ti/uart/aarch64/16550_console.S	\
-				plat/rpi3/rpi3_common.c
+				plat/rpi3/rpi3_common.c			\
+				${XLAT_TABLES_LIB_SRCS}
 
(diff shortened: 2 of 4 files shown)


antonio-nino-diaz-arm:
@Andre-ARM

soby-mathew:
jenkins: test this please

Andre-ARM:
Nice one, actually this seems useful beyond RPi.
So can we move this to some generic place?
Marking regions used by BL32 for instance sounds like a common thing.

antonio-nino-diaz-arm:
> Nice one, actually this seems useful beyond RPi.
> So can we move this to some generic place?
> Marking regions used by BL32 for instance sounds like a common thing.

What did you have in mind? In this function I look for a specific entry in the memreserve region and overwrite it (which is actually a bit hacky, libfdt doesn't provide any function to write this kind of region, only to read them).

I guess I could just forget about looking for the region reserved for the default bootstrap firmware. I know it's always going to be there. I could create instead a helper to overwrite the nth entry of the table that returns an error code if it fails, would this work for you?

Andre-ARM:
Glad you asked ;-)

    /*
     * fdt_add_mem_reserve(): Add a region to the DTB memreserve structure
     *
     * @fdt: pointer to the beginning of the DTB blob
     * @start: physical start address of the reserved region
     * @size: size of the reserved region
     * @entry: Use (overwrite) this specific entry, if no free entry could be found.
     *
     * This adds a region to the list of /memreserve entries at the beginning of the
     * DTB. It scans the existing list of regions, and merges into one if there
     * is an overlap.
     * If @entry is not -1, it will use that specific entry in case there is no
     * free slot available. If @entry is -1 and no free slot can be found, we
     * return an error.
     */
    int fdt_add_mem_reserve(void *fdt, uint64_t start, uint64_t size, int entry)
    {
        /*
         * - check for valid FDT
         * - get space for memreserve entries: off_dt_struct - off_mem_rsvmap
         *     return error if < 2
         * - scan list of existing entries for overlaps:
         *     merge if found, return
         * - check for free slot (max entries - last entry > 1?)
         *     no found? use entry if not -1, return error otherwise
         * - write entry into slot
         */
    }

antonio-nino-diaz-arm:
Actually, it looks like all we have to do is to update libfdt. I've found the following two functions in the current master, and they were already available in v1.4.7:
```
int fdt_add_mem_rsv(void *fdt, uint64_t address, uint64_t size);
int fdt_del_mem_rsv(void *fdt, int n);
```
I will change this PR so that we include the latest version and I will use the functions in the library instead.

Andre-ARM:
Even better, good find! These function seem to be a bit heavy though, as they potentially move the structure and string blocks if necessary, to make room. Just make sure you don't use old pointers after calling this function. Also we would need to be sure to not overwrite something behind the DT.

antonio-nino-diaz-arm:
I have some patches ready but our CI is broken at the moment. I hope to push them later today.

antonio-nino-diaz-arm:
I've created an extra PR for the libfdt changes: https://github.com/ARM-software/arm-trusted-firmware/pull/1643 This PR should be focused on the rpi3 part.

antonio-nino-diaz-arm:
@Andre-ARM something to add?

Andre-ARM:
I didn't have the chance to take a closer look, especially at what libfdt's fdt_add_mem_rsv() does exactly. I see that there is code to move the structure and string blocks of the DT around, so I was wondering if just calling this function is all you need (and it takes care of merging).
I have an implementation of the prototype sketched above, which would automatically merge regions and should work for the RPi case, without using the new libfdt functions. But it can't cope with making room for a new region, as fdt_add_mem_rsv() does. Will reply later today.

Andre-ARM:
Disregard my last comment, I was talking complete and utter nonsense:
- There is no moving around with those functions, so you always need a free slot. If the original DT doesn't have a memreserve entry, you can't easily add one (short of rebuilding the whole DTB). I got confused by this tempting fdt_set_off_dt_struct() function.
- There is no overlap check or the like in those libfdt functions. I think the idea of them is to be used when building a dtb from scratch, fdt_add_mem_rsv() only adds to the end.
- Testing your code in my test program, your algorithm seems to work, but only for your particular case. For instance if there two empty slots (one empty, one terminator), you don't do anything, although you could easily use the first free slot. Actually I was wondering if you could just unconditionally do a:
```
    fdt_del_mem_rsv(dtb, 0);
    fdt_add_mem_rsv(dtb, SEC_SRAM_BASE, SEC_SRAM_SIZE);
```
What about you use your existing code (or this shortened version) for now, and I replace it with something more fancy once I get around to solve this problem for Allwinner? Making sure that your use case would still be covered?

antonio-nino-diaz-arm:
jenkins: test this please

Andre-ARM:
Stupid question: why did you actually need to update libfdt? The fdt_add/del_mem_rsv() functions are around for ages already, and it seems like just your patch builds fine with master.

antonio-nino-diaz-arm:
@Andre-ARM You are right... I think I just pulled in my clone of libfdt and saw the function, and I suspect I didn't even grep in the copy of the TF, which already had it! :S

antonio-nino-diaz-arm:
Anyway, we were using v1.4.2 and now we use v1.4.7. v1.4.2 was tagged two years ago, so updating it hasn't been a waste of time.

PR #1639, opened 2018-10-18
plat/arm/scmi: introduce plat_css_get_scmi_info API

chandnich:
The default values of 'plat_css_scmi_plat_info' is not applicable for
all the platforms. There should be a provision to let platform code to
register a platform specific instance of scmi_channel_plat_info_t.

Add a new API 'plat_css_get_scmi_info' which lets the platform to
register a platform specific instance of scmi_channel_plat_info_t and
remove the default values.

In addition to this, the existing 'plat_css_scmi_plat_info' structure
is removed from the common code and instantiated for the platforms that
need it. This allows for a consistent provisioning of the SCMI channel
information across all the existing and upcoming platforms.

Change-Id: I4fb65d7f2f165b78697b4677f1e8d81edebeac06
Signed-off-by: Chandni Cherukuri <chandni.cherukuri@arm.com>

### DIFF
--- a/plat/arm/css/drivers/scp/css_pm_scmi.c
+++ b/plat/arm/css/drivers/scp/css_pm_scmi.c
@@ -13,7 +13,6 @@
 #include <platform.h>
 #include <string.h>
 #include "../scmi/scmi.h"
-#include "../mhu/css_mhu_doorbell.h"
 #include "css_scp.h"
 
 /*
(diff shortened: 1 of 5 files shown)


soby-mathew:
Please rebase on the integraton branch.

chandnich:
have checked the ci failures and they are not related to the patch

antonio-nino-diaz-arm:
jenkins: test this please

plat/arm/css/drivers/scp/css_pm_scmi.c
```diff
@@ -330,7 +321,7 @@ static int scmi_ap_core_init(scmi_channel_t *ch)
 
 void __init plat_arm_pwrc_setup(void)
 {
-	channel.info = &plat_css_scmi_plat_info;
+	channel.info = plat_css_get_scmi_info();
```
npoushin: Suggest reordering these so that the call to plat_css_get_scmi_info() is after channel.lock =, to preserve declarations before code (it goes from being an assignment to a function call and assignment)

plat/arm/css/sgi/sgi_bl31_setup.c
```diff
@@ -9,6 +9,21 @@
 #include <plat_arm.h>
 #include <sgi_plat_config.h>
 #include <sgi_ras.h>
+#include "../../css/drivers/scmi/scmi.h"
```
npoushin: Can we just add the path to the makefile to make the path not relative, as we might well reorganize this code at a later date (maybe PLAT_INCLUDES or similar)?

plat/arm/css/sgm/sgm_bl31_setup.c
```diff
@@ -8,6 +8,21 @@
 #include <debug.h>
 #include <plat_arm.h>
 #include <sgm_plat_config.h>
+#include "../../css/drivers/scmi/scmi.h"
```
npoushin: Same as previous comment on relative paths

PR #1640, opened 2018-10-18
Multi-console: Deprecate the `finish_console_register` macro

soby-mathew:
The `finish_console_register` macro is used by the multi console
framework to register the `console_t` driver callbacks. It relied
on weak references to the `ldr` instruction to populate 0 to the
callback in case the driver has not defined the appropriate
function. Use of `ldr` instruction to load absolute address to a
reference makes the binary position dependant. These instructions
should be replaced with adrp/adr instruction for position independant
executable(PIE). But adrp/adr instructions don't work well with weak
references as described in GNU ld bugzilla issue 22589.

This patch defines a new version of `finish_console_register` macro
which takes the driver callbacks as arguments and deprecates the old
one. If any of the argument is left blank, then the macro populates
0 for that callback. Hence the functionality of the previous
deprecated macro is preserved. The USE_FINISH_CONSOLE_REG_2 define
is used to select the new variant of the macro and will be removed
once the deprecated variant is removed.

All the upstream console drivers have been migrated to use the new
macro in this patch.

NOTE: Platforms be aware that the new variant of the
`finish_console_register` should be used and the old variant is
deprecated.

Change-Id: Ia6a67aaf2aa3ba93932992d683587bbd0ad25259

### DIFF
--- a/drivers/console/aarch64/skeleton_console.S
+++ b/drivers/console/aarch64/skeleton_console.S
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 #include <asm_macros.S>
+#define USE_FINISH_CONSOLE_REG_2
 #include <console_macros.S>
 
 	/*
@@ -60,8 +61,12 @@ func console_xxx_register
 	 * Keep console_t pointer in x0 for later.
 	 */
 
-	/* Macro to finish up registration and return (needs valid x0 + x30). */
-	finish_console_register xxx
+	/*
+	 * Macro to finish up registration and return (needs valid x0 + x30).
+	 * If any of the argument is unspecified, then the corresponding
+	 * entry in console_t is set to 0.
+	 */
+	finish_console_register xxx putc=1, getc=1, flush=1
 
 	/* Jump here if hardware init fails or parameters are invalid. */
 register_fail:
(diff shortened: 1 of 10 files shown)


soby-mathew:
@jwerner-chromium

drivers/console/aarch64/skeleton_console.S
```diff
@@ -60,8 +61,12 @@ func console_xxx_register
 	 * Keep console_t pointer in x0 for later.
 	 */
 
-	/* Macro to finish up registration and return (needs valid x0 + x30). */
-	finish_console_register xxx
+	/*
+	 * Macro to finish up registration and return (needs valid x0 + x30).
+	 * If any of the argument is blank, then the corresponding callback
```
jwerner-chromium: nit: Maybe clarify that you'll need to use the _driver_xxx=function form in that case, just leaving one out in the middle will not work well.

drivers/coreboot/cbmem_console/aarch64/cbmem_console.S
```diff
@@ -39,7 +40,7 @@ func console_cbmc_register
 	ldr	w2, [x0]
 	str	w2, [x1, #CONSOLE_T_CBMC_SIZE]
 	mov	x0, x1
-	finish_console_register cbmc
+	finish_console_register console_cbmc_putc, console_cbmc_getc, console_cbmc_flush
```
jwerner-chromium: console_cbmc_getc doesn't exist

soby-mathew:
Thanks @jwerner-chromium for your review, I will rework the patch.

antonio-nino-diaz-arm:
jenkins: test this please

PR #1641, opened 2018-10-18
AArch64: Enable lower ELs to use pointer authentication

jeenu-arm:
Pointer authentication is an Armv8.3 feature that introduces
instructions that can be used to authenticate and verify pointers.

Pointer authentication instructions are allowed to be accessed from all
ELs but only when EL3 explicitly allows for it; otherwise, their usage
will trap to EL3. Since EL3 doesn't have trap handling in place, this
patch unconditionally disables all related traps to EL3 to avoid
potential misconfiguration leading to an unhandled EL3 exception.

Fixes ARM-software/tf-issues#629

Change-Id: I9bd2efe0dc714196f503713b721ffbf05672c14d

### DIFF
--- a/docs/firmware-design.rst
+++ b/docs/firmware-design.rst
@@ -2532,6 +2532,12 @@ This Architecture Extension is targeted when ``ARM_ARCH_MAJOR`` == 8 and
    translation table entries for a given stage of translation for a particular
    translation regime.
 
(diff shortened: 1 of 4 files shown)


antonio-nino-diaz-arm:
jenkins: test this please

PR #1643, opened 2018-10-19
Update libfdt to version 1.4.7

antonio-nino-diaz-arm:
Also, import ``strrchr()`` as it is needed for this version of libfdt.

The previous version had some custom modifications and that's why it has been removed and then imported instead of simply updated. The new version hasn't been modified at all, so future updates should be done by simply updating the files instead of removing the old ones.


antonio-nino-diaz-arm:
jenkins: test this please

antonio-nino-diaz-arm:
Juno + clang runs out of memory.

antonio-nino-diaz-arm:
jenkins: test this please

PR #1645, opened 2018-10-19
Makefile: Fix verbose builds on Windows

antonio-nino-diaz-arm:
Commit ee1ba6d4ddf1 ("Makefile: Support totally quiet output with -s") broke verbose (V=1) builds on Windows. This patch fixes it by adding helpers to silence echo prints in a OS-dependent way.


antonio-nino-diaz-arm:
jenkins: test this please

PR #1650, opened 2018-10-26
Sgiclark ares support

chandnich:


npoushin:
All good, the regression seen is not caused by this patch

PR #1651, opened 2018-10-26
Fix some MISRA defects

antonio-nino-diaz-arm:


antonio-nino-diaz-arm:
The CI failures are some MISRA defects.

PR #1652, opened 2018-10-26
poplar, warp7, ls1043: Decouple from plat/arm files

antonio-nino-diaz-arm:
plat/arm files should only be used by Arm platforms. If other platforms use them, they create dependencies that can introduce problems when updating Arm platforms.


bryanodonoghue:
lgtm - warp7

shawnguo2:
For "poplar: Decouple from plat/arm files", looks good, and tested on Poplar.

PR #1671, opened 2018-11-06
Standardise header guards across codebase

antonio-nino-diaz-arm:
All identifiers, regardless of use, that start with two underscores are reserved. This means they can't be used in header guards.

The style that this project is now to use the full name of the file in capital letters followed by 'H'. For example, for a file called ``uart_example.h``, the header guard is ``UART_EXAMPLE_H``.

The exceptions are files that are imported from other projects:

- CryptoCell driver
- dt-bindings folders
- zlib headers


Andre-ARM:
OK for allwinner parts

antonio-nino-diaz-arm:
It looks like the GitHub API limits the number of files modified by a patch and only part of the maintainers have been notified with the script. Anyway, this is the rest, in case someone wants to take a look: @mtk09422, @qoriq-open-source, @Anson-Huang, @jenswi-linaro, @TonyXie06, @rockchip-linux, @b49020, @glneo, @masahir0y, @sivadur.

shawnguo2:
OK for Poplar.

jenswi-linaro:
OK for QEMU.

PR #1674, opened 2018-11-08
hikey, hikey960, poplar: use new console APIs

jforissier:
Enables building with ERROR_DEPRECATED=1.


antonio-nino-diaz-arm:
IP Review +1

antonio-nino-diaz-arm:
jenkins: test this please

plat/hisilicon/hikey/hikey_bl1_setup.c
```diff
@@ -42,8 +42,11 @@ meminfo_t *bl1_plat_sec_mem_layout(void)
  */
 void bl1_early_platform_setup(void)
 {
+	static console_pl011_t console;
```
antonio-nino-diaz-arm: I'd rather move this out of the function. It should be clear that this isn't in the stack. Same for the rest of the files.

jforissier: Sure. FYI, this pattern appears twice in the current code (with `console_16550_t` in `rockchip` and `ti/k3`).

antonio-nino-diaz-arm: Ok, that's not ideal...

In any case, the CI is already reporting build failures for your modified platforms. You need to modify the ``plat_crash_console_xxx`` functions.

jforissier: How do I see the build failures?
```
jenkins.oss.arm.com’s server IP address could not be found.
```

antonio-nino-diaz-arm: Yeah, the CI isn't public (but hopefully that will change over the next few months).  All you need to do to see the errors is:
```
CROSS_COMPILE=aarch64-linux-gnu- make PLAT=hikey
CROSS_COMPILE=aarch64-linux-gnu- make PLAT=hikey960
```
That should fail to compile.

jforissier:
Updated, thanks.

antonio-nino-diaz-arm:
jenkins: test this please

antonio-nino-diaz-arm:
The failures are Coverity false positives.

shawnguo2:
Tested on Poplar.

antonio-nino-diaz-arm:
@jforissier, I've just taken a look at the commits in which you have mentioned this patch. Don't enable ``ERROR_DEPRECATED`` by default. We try to keep all upstream platforms up to date with non-deprecated interfaces, but sometimes (like with the multi console) we can't because there can be problems with it (for example, on Juno, a trivial port broke ``PSCI_SYSTEM_SUSPEND``).

jforissier:
@antonio-nino-diaz-arm thanks for commenting, that is quite helpful.

Well, `ERROR_DEPRECATED` should not be a problem for us (OP-TEE), because we're not *tracking* the TF-A branch. We update TF-A in a controlled way, typically before we make a release. So if something breaks the build, we will know and hold off the update. At least that was the idea.

Now, thinking a bit more about it... As a user of TF-A, I suppose we don't really care about the use of deprecated APIs *inside* TF-A. What we care about is external interfaces. Such as 
`FVP_USE_GIC_DRIVER=FVP_GICV3_LEGACY` which has been deprecated and removed from TF-A v2.0 (and therefore we need to enable `CFG_ARM_GICV3` in OP-TEE). But that is not addressed by `ERROR_DEPRECATED` as far as I can tell. So yeah, I think I'll just drop my deprecation patches.

@jbech-linaro FYI.

antonio-nino-diaz-arm:
Ok, in your case I agree that it's not really a problem. I thought it would be a similar case as we've had with uniphier and allwinner, that enabled ``ERROR_DEPRECATED`` in their makefiles by default and made it quite hard for us to do some changes (we had to disable it in uniphier, IIRC).

In any case, ``ERROR_DEPRECATED`` used to change a lot the codebase, so I agree that it may have made sense in the past. Now, not so much, it only removes a few helpers/definitions that aren't used by any upstream platform because we keep them up to date.

hzhuang1:
Tested it on both hikey and hikey960.

PR #1675, opened 2018-11-08
SPM priority level changes

SNG-ARM:
SPM priority setting changes.


antonio-nino-diaz-arm:
Could you change the title of the pull request?

antonio-nino-diaz-arm:
jenkins: test this please

SNG-ARM:
Sorry, i did not understand the changes requested. Where do i put this statement about build failure. Should it be part of the commit message.

Also, the build fails only if the EL3_EXCEPTION_HANDLING flag is not set while building SPM.

antonio-nino-diaz-arm:
In "SPM: EHF: Build EHF module along with Secure Partition Manager" don't add a check. If SPM depends on ``EL3_EXCEPTION_HANDLING``, ``spm.mk`` should set ``EL3_EXCEPTION_HANDLING`` to 1.

SNG-ARM:
This is on same lines as what is being done in bl31.mk for sdei code. In fact i was asked to have this logic by tf-a team.

ifeq (${EL3_EXCEPTION_HANDLING},0)
  $(error EL3_EXCEPTION_HANDLING must be 1 for SDEI support)
endif

i believe we need to change the CI script to reflect this change.

antonio-nino-diaz-arm:
Ok, the reason is that setting flags automatically in the build system has some problems that are hard to detect because the build system isn't good enough to do this properly. Then yes, we need to modify the CI configs and update the documentation of ENABLE_SPM to reflect that it requires EL3_EXCEPTION_HANDLING.

SNG-ARM:
I have added the documentation to reflect the EL3_EXCEPTION_HANDLING flag needed for building SPM code.

antonio-nino-diaz-arm:
Thanks, could you create a patch for the CI to fix all configs that set ``ENABLE_SPM=1``?

SNG-ARM:
I do not know how to create a patch for the CI. Can you help me with it, or point me to some page which has instructions for the same. Thanks.

antonio-nino-diaz-arm:
jenkins: test this please

antonio-nino-diaz-arm:
I've taken a look at the CI configs and all I had to do is add one line, don't worry about it.

SNG-ARM:
Thanks Antonio!

antonio-nino-diaz-arm:
The failures are just a few MISRA defects that we can ignore.

PR #1678, opened 2018-11-12
psci, rpi3: add CONSOLE_FLUSH_ON_POWEROFF build option

pbatard:
Some OSes (e.g. Ubuntu 18.04 LTS on Raspberry Pi 3) may disable
the UART after boot, in conditions that create an infinite loop
in console_flush(), thus preventing power down or reboot.

By adding a new CONSOLE_FLUSH_ON_POWEROFF option (enabled by
default), ATF can be built that avoids system freeze on reboot.

Option is added to both psci & rpi3, as both paths can be used
for OS boot on Raspberry Pi.

Fixes ARM-software/tf-issues#647

Signed-off-by: Pete Batard <pete@akeo.ie>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -559,6 +559,7 @@ endif
 ################################################################################
 
 $(eval $(call assert_boolean,COLD_BOOT_SINGLE_CPU))
+$(eval $(call assert_boolean,CONSOLE_FLUSH_ON_POWEROFF))
 $(eval $(call assert_boolean,CREATE_KEYS))
 $(eval $(call assert_boolean,CTX_INCLUDE_AARCH32_REGS))
 $(eval $(call assert_boolean,CTX_INCLUDE_FPREGS))
@@ -613,6 +614,7 @@ $(eval $(call assert_numeric,SMCCC_MAJOR_VERSION))
 
 $(eval $(call add_define,ARM_ARCH_MAJOR))
 $(eval $(call add_define,ARM_ARCH_MINOR))
+$(eval $(call add_define,CONSOLE_FLUSH_ON_POWEROFF))
 $(eval $(call add_define,COLD_BOOT_SINGLE_CPU))
 $(eval $(call add_define,CTX_INCLUDE_AARCH32_REGS))
 $(eval $(call add_define,CTX_INCLUDE_FPREGS))
--- a/docs/user-guide.rst
+++ b/docs/user-guide.rst
@@ -305,6 +305,13 @@ Common build options
    ``plat_secondary_cold_boot_setup()`` platform porting interfaces do not need
    to be implemented in this case.
 
+-  ``CONSOLE_FLUSH_ON_POWEROFF``: This option, which is enabled by default,
+   tells the system to issue a flush of the console on power operations (reset,
+   power off). However, some combinations of OSes + hardware may have disabled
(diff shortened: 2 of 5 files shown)


vwadekar:
Would it be better to just disable console from 'bl31_plat_runtime_setup' instead? AFAIK, the common platform files already do that.

pbatard:
1. Is not an option. We have no control whatsoever over what the OS, the OS installer or the OS user will do in terms of enabling/disabling hardware. This issue __must__ be addressed with the ATF, one way or another.
2. I still think we should work on trying to let console output go through if possible (i.e. if OS didn't disable the UART), and only bail out on that if we find that we can't flush.
3. Is out of scope. This is adding an extra feature on top of a fix. I'd rather leave that to a separate patch.

What about this then?

- No global option, but a `TI_16650_TX_EMPTY_TIMEOUT` that would default to `0`, but that a __platform__ could set (a bit like `TI_16550_MDR_QUIRK`)
- In `drivers/ti/uart/aarch64/16550_console.S`, for every loop where we check `UARTLSR` for `UARTLSR_TEMT | UARTLSR_THRE`:
  - if `TI_16650_TX_EMPTY_TIMEOUT` is `0`, no change of behaviour (infinite loop until we get a clear)
  - if `TI_16650_TX_EMPTY_TIMEOUT` is not zero, use it as a decrementing counter and loop until we get a clear or the counter is `0`.

Then we could set `TI_16650_TX_EMPTY_TIMEOUT` in the rpi3 configuration, to a value that would give us the best of both worlds, or, set it to `0`, and have a build option override it.

No, of course, I do realize that `TIMEOUT` is a misnomer, since we don't really have a clock to rely on, and the actual time spent waiting will depend on the current CPU frequency. Yet, since the option would be designed to force a timeout, I still think it's appropriate. Of course, I'm open to other naming suggestions.

The only possible caveat I can see with this option is that this will increase the clobber list of the relevant 16550 function by one register (`x2` or `x3`), since we'll need to store that counter somewhere. Do we envision any issue from having an extra register being clobbered?

If the solution above is something you think you can go with, I can work on submitting a new pull request. Let me know.

antonio-nino-diaz-arm:
The problem is that you are trying to apply a fix to a driver that isn't broken. This is a problem of the port, as @soby-mathew, not of the 16650 driver. I mean, even the datasheet of the rpi doesn't claim to be 16650-compliant, just 16650-like. The fix has to be in the platform port.

pbatard:
So that only leaves disabling the UART altogether then.

pbatard:
Yeah, I think it's `UART1`, as it appears to be initialized as `/dev/ttyAMA1` in Ubuntu (rather than `/dev/ttyAMA0`).

I'll work on a pull request that adds the `RPI3_RUNTIME_UART` option, with `DISABLED` and `UART1` as the only 2 options for now, and that defaults to `DISABLED`.

pbatard:
Pull request #1680 has now been submitted for `RPI3_RUNTIME_UART`.

Note that I decided to use `-1` as the value for disabled, as this was simpler and it should also allow us to specify the UART index directly (e.g. `RPI3_RUNTIME_UART=0` for `UART0`). For now `-1` (default) disables the runtime UART altogether, and any other value enables the default UART (`UART1`) for runtime.

Since this pull request has been superseded, I will close it.

PR #1679, opened 2018-11-13
ccn: Introduce API to set and read value of node register

pangupta:
Signed-off-by: Pankaj Gupta <pankaj.gupta@nxp.com>

### DIFF
--- a/drivers/arm/ccn/ccn.c
+++ b/drivers/arm/ccn/ccn.c
@@ -11,6 +11,7 @@
 #include <debug.h>
 #include <errno.h>
 #include <mmio.h>
+#include <stdbool.h>
 #include "ccn_private.h"
 
 static const ccn_desc_t *ccn_plat_desc;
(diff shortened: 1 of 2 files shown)


antonio-nino-diaz-arm:
IP Review +1

antonio-nino-diaz-arm:
jenkins: test this please

pangupta:
I have tested this patch. It is working as expected.

I am not able to see the details for the failing check.

antonio-nino-diaz-arm:
There are only a few MISRA defects that we can ignore in this PR.

drivers/arm/ccn/ccn.c
```diff
@@ -490,3 +491,98 @@ int ccn_get_part0_id(uintptr_t periphbase)
...
+
+	return region_id;
+}
+
+/*******************************************************************************
+ * This function sets the value 'val' to the register at register_offset from
+ * the base address pointed to by the region_id.
+ * where, region id is mapped to a node_id of node_type.
+ ******************************************************************************/
+void set_node_reg(node_types_t node_type, unsigned int node_id,
```
soby-mathew: The API should use the naming convention. Prefix `ccn_*` to the names.

drivers/arm/ccn/ccn.c
```diff
@@ -490,3 +491,98 @@ int ccn_get_part0_id(uintptr_t periphbase)
...
+	VERBOSE("Value is successfully written at address 0x%lx.\n",
+		(ccn_plat_desc->periphbase + region_id_to_base(region_id)));
+}
+
+/*******************************************************************************
+ * This function read the value 'val' stored in the register at register_offset
+ * from the base address pointed to by the region_id.
+ * where, region id is mapped to a node_id of node_type.
+ ******************************************************************************/
+unsigned long read_node_reg(node_types_t node_type, unsigned int node_id,
```
soby-mathew: Same comment as above.

pangupta: Accepted and changes done.

drivers/arm/ccn/ccn.c
pangupta: Accepted and changes done.

pangupta:
Do I miss something?

drivers/arm/ccn/ccn.c
```diff
@@ -490,3 +491,101 @@ int ccn_get_part0_id(uintptr_t periphbase)
...
+			+ region_id_to_base(region_id))
+			+ reg_offset);
+}
+
+/*******************************************************************************
+ * This function read the value 'val' stored in the register at register_offset
+ * from the base address pointed to by the region_id.
+ * where, region id is mapped to a node_id of node_type.
+ ******************************************************************************/
+unsigned long ccn_read_node_reg(node_types_t node_type, unsigned int node_id,
```
soby-mathew: `unsigned long long` for 64 bit values. Please apply change the local variables and return types which are 64 bit. Also see comment about restricting this API to aux control register.

drivers/arm/ccn/ccn.c
```diff
@@ -490,3 +491,101 @@ int ccn_get_part0_id(uintptr_t periphbase)
...
+
+/*******************************************************************************
+ * This function sets the value 'val' to the register at register_offset from
+ * the base address pointed to by the region_id.
+ * where, region id is mapped to a node_id of node_type.
+ ******************************************************************************/
+void ccn_set_node_reg(node_types_t node_type, unsigned int node_id,
+			unsigned int reg_offset, unsigned long val)
+{
+	unsigned int region_id = get_region_id_for_node(node_type, node_id);
```
soby-mathew: Since this is an API which you are intending to access Auxuliary Control Register, it would be remove the `reg_offset` argument and hard-code to the needed offset and restrict the API as `ccn_set_node_aux_reg()`.

include/drivers/arm/ccn.h
```diff
@@ -76,6 +76,16 @@ typedef struct ccn_desc {
 	uintptr_t periphbase;
 } ccn_desc_t;
 
+/* Enum used to loop through all types of nodes in CCN*/
+typedef enum node_types {
+	NODE_TYPE_RNF = 0,
```
soby-mathew: It seems RNF is not a valid node type. Would suggest to remove this.

drivers/arm/ccn/ccn.c
```diff
@@ -490,3 +491,101 @@ int ccn_get_part0_id(uintptr_t periphbase)
...
+	 * occupy contiguous 16 byte apart offsets.
+	 */
+	mn_reg_off = MN_RNF_NODEID_OFFSET + (node_type << 4);
+	node_bitmap = ccn_reg_read(ccn_plat_desc->periphbase,
+					MN_REGION_ID, mn_reg_off);
+
+	assert((node_bitmap & (1ULL << (node_id))) != 0U);
+
+	switch (node_type) {
+	case NODE_TYPE_RNI:
```
soby-mathew: What about NODE_TYPE_RND ?

drivers/arm/ccn/ccn.c
```diff
@@ -490,3 +491,101 @@ int ccn_get_part0_id(uintptr_t periphbase)
...
+						unsigned int node_id)
+{
+	unsigned int mn_reg_off, region_id;
+	unsigned long long node_bitmap;
+	unsigned int loc_node_id, node_pos_in_map = 0;
+
+	assert(node_type < NUM_NODE_TYPES);
+	assert(node_id < MAX_RN_NODES);
+
+	/*
```
soby-mathew: The lines 508 - 516 are better placed below the switch case statement, since RNF node type is not supported. The comment can be updated for the same.

pangupta: Shifted the comments below switch case statement. Comments are updated as follows:
  /*
   * RN-I, HN-F, HN-I, SN node registers in the MN region
   * occupy contiguous 16 byte apart offsets.
   *
   * RN-F and RN-D node are not supported as 
   * none of them exposes any memory map to
   * configure any of their offset registers.
   */

drivers/arm/ccn/ccn.c
pangupta: Accepted.
Used unsigned long long for 64 bit values.

pangupta: Regarding second comment "restricting this API to aux control register":

In NXP SoCs, apart from configuring Auxilary Control register, it is required to configure QoS Registers as well.
In case, I restrict its usage, then I have to create another API for the same.

pangupta:
I pushed the code with all the comments disposed off except one "restricting this API to aux control register".

In NXP SoCs, apart from configuring Auxilary Control register, it is required to configure QoS Registers as well.
In case, I restrict its usage, then I have to create another API for the same.

Please share your view on this. Thanks.

drivers/arm/ccn/ccn.c
```diff
@@ -490,3 +491,106 @@ int ccn_get_part0_id(uintptr_t periphbase)
...
+						unsigned int node_id)
+{
+	unsigned int mn_reg_off, region_id;
+	unsigned long long node_bitmap;
+	unsigned int loc_node_id, node_pos_in_map = 0;
+
+	assert(node_type < NUM_NODE_TYPES);
+	assert(node_id < MAX_RN_NODES);
+
+	mn_reg_off = MN_RNF_NODEID_OFFSET + (node_type << 4);
```
soby-mathew: Suggest to move these lines 508 - 512 after the comments at line 540.

soby-mathew:
Thanks for the rework. It is better to have a separate API for QoS I think.

soby-mathew:
hmm, I see. In that case, suggest to keep the generic functions you have now, but restrict the offset to <= 0xFF00.  Also rename `ccn_set_node_reg()` to `ccn_write_node_reg()`.

pangupta:
Renamed ccn_set_node_reg() to ccn_write_node_reg().

Added assert if the register_offset is > 0xFF00.

soby-mathew:
@pangupta, there is one comment not addressed https://github.com/ARM-software/arm-trusted-firmware/pull/1679#discussion_r234932568

pangupta:
All the comments have been disposed off.

soby-mathew:
Thanks for the rework.

jenkins: test this please

pangupta:
> Thanks for the rework.
> 
> jenkins: test this please

I tested this and it is working as expected.
Thanks.

include/drivers/arm/ccn.h
```diff
@@ -76,6 +76,16 @@ typedef struct ccn_desc {
...
+	NODE_TYPE_RNI,
+	NODE_TYPE_RND,
+	NODE_TYPE_HNF,
+	NODE_TYPE_HNI,
+	NODE_TYPE_SN,
+	NUM_NODE_TYPES
+} node_types_t;
```
antonio-nino-diaz-arm: In order to comply with MISRA it looks like you will have to turn this enum into a set of #defines. You are using this inside ``static unsigned int get_region_id_for_node(node_types_t node_type, unsigned int node_id)``: ``(node_type << 4)``

I'm afraid that MISRA doesn't like enums being used in arithmetic/logical operations.

pangupta: In the same file ccn.c, similar usage is present:

static unsigned int ccn_get_rn_master_info(uintptr_t periphbase,
                                           rn_info_t *info)
{
        unsigned int num_masters = 0;
        rn_types_t rn_type;

        assert (info);

        for (rn_type = RN_TYPE_RNF; rn_type < NUM_RN_TYPES; rn_type++) {
                unsigned int mn_reg_off, node_id;
                unsigned long long rn_bitmap;

                /*
                 * RN-F, RN-I, RN-D node registers in the MN region occupy
                 * contiguous 16 byte apart offsets.
                 */
                mn_reg_off = MN_RNF_NODEID_OFFSET + (rn_type << 4);
-------------------------------------------------------------------------------------------------

I guess, I am using the same way.

antonio-nino-diaz-arm: Hmmm... Ok, I guess we need to do a cleanup of this driver anyway.

PR #1681, opened 2018-11-14
allwinner: clock / power fixes

Andre-ARM:
Some smaller fixes:
- Two bus clocks are not setup properly by U-Boot (was always done by the legacy ATF). Bring AHB1 back to 200 MHz and AHB2 to 300 MHz. For instance this improves Ethernet performance from ~700MBit/s to the normal ~920MBit/s (some iperf3 values).
- The DRIVEVBUS power pin wasn't properly setup. This isn't so much of an issue, since most boards don't need it (see commit msg), but the Remix Mini PC powers its regular USB sockets via this line.
- Add support for the DCDC6 power rail. Normally the reset default is good, but the Remix Mini PC needs a different voltage.


antonio-nino-diaz-arm:
IP Review +1

Andre-ARM:
> For the clock change, why not fix U-Boot or SPL (where there's already clock handling code) instead of working around it here?

Yeah, I thought about it, but it seems easier to fix it here. I couldn't find a good place in QEMU to setup those clocks in U-Boot *proper*, there seems to be only one SPL routine. I will try to come up with a solution in U-Boot (since the clocks should indeed be handled in one place), but was wondering if we could take it here meanwhile, since I can't predict how long the U-Boot patch would take - and it's a performance regression against the "legacy" ATF port.

antonio-nino-diaz-arm:
@smaeul any comment?

PR #1685, opened 2018-11-15
rpi3: add RPI3_USE_UEFI_MAP build option

pbatard:
The default Raspberry Pi 3 memory mapping for ATF is geared towards
the use of uboot + Linux. This creates issues when trying to use
ATF with an UEFI payload and Windows on ARM64.

We therefore introduce new build option RPI3_USE_UEFI_MAP, that
enables the build process to use an alternate memory mapping that
is compatible with UEFI + Windows (as well as UEFI + Linux).

Fixes ARM-software/tf-issues#649

Signed-off-by: Pete Batard <pete@akeo.ie>

### DIFF
--- a/plat/rpi3/include/platform_def.h
+++ b/plat/rpi3/include/platform_def.h
@@ -63,6 +63,25 @@
  * secure DRAM. Note that this is all actually DRAM with different names,
  * there is no Secure RAM in the Raspberry Pi 3.
  */
+#if RPI3_USE_UEFI_MAP
+#define SEC_ROM_BASE			ULL(0x00000000)
+#define SEC_ROM_SIZE			ULL(0x00010000)
+
+/* FIP placed after ROM to append it to BL1 with very little padding. */
+#define PLAT_RPI3_FIP_BASE		ULL(0x00020000)
+#define PLAT_RPI3_FIP_MAX_SIZE		ULL(0x00010000)
+
+/* Reserve 2M of secure SRAM and DRAM, starting at 2M */
+#define SEC_SRAM_BASE			ULL(0x00200000)
+#define SEC_SRAM_SIZE			ULL(0x00100000)
+
+#define SEC_DRAM0_BASE			ULL(0x00300000)
+#define SEC_DRAM0_SIZE			ULL(0x00100000)
+
+/* Windows on ARM requires some RAM at 4M */
+#define NS_DRAM0_BASE			ULL(0x00400000)
+#define NS_DRAM0_SIZE			ULL(0x00C00000)
+#else
 #define SEC_ROM_BASE			ULL(0x00000000)
 #define SEC_ROM_SIZE			ULL(0x00020000)
 
@@ -80,6 +99,7 @@
 
 #define NS_DRAM0_BASE			ULL(0x11000000)
 #define NS_DRAM0_SIZE			ULL(0x01000000)
+#endif /* RPI3_USE_UEFI_MAP */
 
 /*
  * BL33 entrypoint.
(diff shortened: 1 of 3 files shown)


plat/rpi3/include/platform_def.h
```diff
@@ -64,12 +64,32 @@
  * there is no Secure RAM in the Raspberry Pi 3.
  */
 #define SEC_ROM_BASE			ULL(0x00000000)
+#if RPI3_USE_UEFI_MAP
```
antonio-nino-diaz-arm: Rather than this mess of ifdefs, I think it would be better to have two sets of definitions:
```
#if RPI3_USE_UEFI_MAP
...
#else
...
#endif
```
If we have to repeat some definitions, that's ok as long as the result is more readable.

antonio-nino-diaz-arm:
Ok, I see the problem. I guess that leaving the secure memory regions at ``0x10000000`` isn't a possibility, right?

antonio-nino-diaz-arm:
IP Review +1

plat/rpi3/include/platform_def.h
pbatard: Okay. I thought about that too, but I didn't want to duplicate the comments that come before `SEC_ROM_BASE` and `PLAT_RPI3_FIP_BASE`.

I'll modify the proposal to use a single `#if` when I get a chance.

pbatard:
> I guess that leaving the secure memory regions at `0x10000000` isn't a possibility, right?

I'm afraid not. Due to Windows being closed source, there was a fair bit of trial and error to get to the proposed memory map. But, as far as I could see, having the secure memory regions at `0x10000000` was a definite showstopper...

antonio-nino-diaz-arm:
Ok, fair enough.

pbatard:
PR has been updated to use a single `#if` in `platform_def.h`.

antonio-nino-diaz-arm:
jenkins: test this please

plat/rpi3/include/platform_def.h
```diff
@@ -63,6 +63,25 @@
...
+#define SEC_SRAM_BASE			ULL(0x00200000)
+#define SEC_SRAM_SIZE			ULL(0x00100000)
+
+#define SEC_DRAM0_BASE			ULL(0x00300000)
+#define SEC_DRAM0_SIZE			ULL(0x00100000)
+
+/* Windows on ARM requires some RAM at 4M */
+#define NS_DRAM0_BASE			ULL(0x00400000)
+#define NS_DRAM0_SIZE			ULL(0x00C00000)
+#else
```
antonio-nino-diaz-arm: checkpatch.pl:
```
plat/rpi3/include/platform_def.h:84: ERROR: trailing whitespace
```

pbatard:
Should be fixed now. Sorry about that.

plat/rpi3/include/platform_def.h
pbatard: Whitespace has been removed.

antonio-nino-diaz-arm:
jenkins: test this please

pbatard:
Thanks Antonio. Just checking on the docs change: Do you have any idea why the new option appears in bold + italic when the `.rst` is viewed from github (https://github.com/ARM-software/arm-trusted-firmware/blob/master/docs/plat/rpi3.rst#build-instructions)?:

image1

I'm not seeing anything I added that could trigger the bolding, and, looking at the HTML source, it seems this is due to `<dl>` and `<dt>` tags having been added.

I suspect that these are pure github processing artefacts, due to the preformatted section that follows. But if you have an idea on how to fix this unwanted bolding/italicising of the new option, and want a patch submitted, let me know.

antonio-nino-diaz-arm:
I suspect that it's because of the extra spaces you have left as indentation for the paragraph.

pbatard:
Oh, I see now that I added an extra space! You're probably right that this is the cause

I'll confirm that this is the issue and submit a docfix pull request when I have a chance.

PR #1687, opened 2018-11-19
rcar-gen3: lock RPC hyper-flash access

ldts:
This feature was enabled to allow u-boot reprograming the ATF firmware
using a FIP image instead of having to toggle numerous DIP switches on
the board.

The code being disabled with this commit should only be re-enabled for
debugging (_never_ on a product release)

Signed-off-by: Jorge Ramirez-Ortiz <jorge.ramirez.ortiz@gmail.com>


antonio-nino-diaz-arm:
This will permanently disable reprogramming from the U-Boot, right? Maybe it's worth adding a build option for people that want to play around with the TF? Or, at least, document that the lines you have added are locking the flash.

IP Review +1

antonio-nino-diaz-arm:
jenkins: test this please

PR #1688, opened 2018-11-20
Initial Spectre V1 mitigations (CVE-2017-5753).

JoelHutton:
Initial Spectre Variant 1 mitigations (CVE-2017-5753).
A potential speculative data leak was found in PSCI code, this depends
on a non-robust implementation of the `plat_get_core_pos_by_mpidr()`
function. This is considered very low-risk. This patch adds a macro to
mitigate this. Note not all code paths could be analyzed with current
tools.

Add a macro which makes a variable 'speculation safe', using the
 __builtin_speculation_safe_value function of GCC and llvm. This will be
available in GCC 9, and is planned for llvm, but is not currently in
mainline GCC or llvm. In order to implement this mitigation the compiler
must support this builtin. Support is indicated by the
__HAVE_SPECULATION_SAFE_VALUE flag.

The -mtrack-speculation option maintains a 'tracker' register, which
determines if the processor is in false speculation at any point. This
adds instructions and increases code size, but avoids the performance
impact of a hard barrier.

Without the -mtrack-speculation option, __builtin_speculation_safe_value
expands to a

    ISB
    DSB SY

sequence after a conditional branch, before the
speculation safe variable is used. With -mtrack-speculation a

    CSEL tracker, tracker, XZR, [cond];
    AND safeval,tracker;
    CSDB

sequence is added instead, clearing the vulnerable variable by
AND'ing it with the tracker register, which is zero during speculative
execution. [cond] are the status flags which will only be true during
speculative execution. For more information on
__builtin_speculation_safe_value and the -mtrack-speculation option see
https://developer.arm.com/support/arm-security-updates/speculative-processor-vulnerability/compiler-support-for-mitigations

The -mtracking option was not added, as the performance impact of the
mitigation is low, and there is only one occurence.

Change-Id: Ic9e66d1f4a5155e42e3e4055594974c230bfba3c
Signed-off-by: Joel Hutton <Joel.Hutton@Arm.com>

### DIFF
--- a/include/lib/utils_def.h
+++ b/include/lib/utils_def.h
@@ -162,4 +162,11 @@
 
 #define COMPILER_BARRIER() __asm__ volatile ("" ::: "memory")
 
+/* Compiler builtin of GCC >= 9 and planned in llvm */
+#ifdef __HAVE_SPECULATION_SAFE_VALUE
+# define SPECULATION_SAFE_VALUE(var) __builtin_speculation_safe_value(var)
+#else
+# define SPECULATION_SAFE_VALUE(var) var
+#endif
+
 #endif /* UTILS_DEF_H */
--- a/lib/psci/psci_stat.c
+++ b/lib/psci/psci_stat.c
@@ -206,9 +206,9 @@ static int psci_get_stat(u_register_t target_cpu, unsigned int power_state,
 
 	if (pwrlvl > PSCI_CPU_PWR_LVL) {
 		/* Get the power domain index */
-		parent_idx = psci_cpu_pd_nodes[target_idx].parent_node;
+		parent_idx = SPECULATION_SAFE_VALUE(psci_cpu_pd_nodes[target_idx].parent_node);
 		for (lvl = PSCI_CPU_PWR_LVL + 1U; lvl < pwrlvl; lvl++)
-			parent_idx = psci_non_cpu_pd_nodes[parent_idx].parent_node;
+			parent_idx = SPECULATION_SAFE_VALUE(psci_non_cpu_pd_nodes[parent_idx].parent_node);
 
 		/* Get the non cpu power domain stats */
 		*psci_stat = psci_non_cpu_stat[parent_idx][stat_idx];


include/lib/utils_def.h
```diff
@@ -160,5 +160,11 @@
  */
 #define ASSERT_SYM_PTR_ALIGN(sym) assert(((size_t)(sym) % __alignof__(*(sym))) == 0)
 
+/* Compiler builtin of GCC >= 9 and planned in llvm */
+#ifdef __HAVE_SPECULATION_SAFE_VALUE
```
danh-arm: I'm told this flag is automatically defined by compiler versions that include the builtin, in which case the mitigations here will be on by default if such a compiler is used. This should be clearly stated in the commit message.

lib/pmf/pmf_main.c
```diff
@@ -74,7 +74,7 @@ int pmf_setup(void)
 		 * PMF service, if it is defined.
 		 */
 		if (pmf_svc_descs[ii].init != NULL) {
-			rc = pmf_svc_descs[ii].init();
+			rc = SPECULATION_SAFE_VALUE(pmf_svc_descs[ii].init)();
```
danh-arm: I'm struggling to see what the threat is here. I guess it's possible that the memory just past the end of pmf_svc_descs[] array might be speculatively loaded into the cache, but what can an attacker do with that information? It cannot see that memory directly. This is in the cold boot path, before any attacker controllable code is executed, so it cannot induce different cache behaviour to observe timing differences and infer anything about that memory.

lib/psci/psci_common.c
```diff
@@ -893,7 +893,7 @@ void psci_print_power_domain_map(void)
 
 	INFO("PSCI Power Domain Map:\n");
 	for (idx = 0; idx < (PSCI_NUM_PWR_DOMAINS - PLATFORM_CORE_COUNT);
-							idx++) {
+					idx = SPECULATION_SAFE_VALUE(idx + 1)) {
```
danh-arm: Again, I'm struggling to see the threat as this is also in the cold boot path and is limited to speculatively reading off the end of the psci_non_cpu_pd_nodes[] array.

lib/psci/psci_common.c
```diff
@@ -904,7 +904,8 @@ void psci_print_power_domain_map(void)
 				psci_non_cpu_pd_nodes[idx].local_state);
 	}
 
-	for (idx = 0; idx < PLATFORM_CORE_COUNT; idx++) {
+	for (idx = 0; idx < PLATFORM_CORE_COUNT; idx =
+			SPECULATION_SAFE_VALUE(idx + 1)) {
```
danh-arm: Again, I'm struggling to see the threat.

lib/psci/psci_setup.c
```diff
@@ -96,7 +96,7 @@ static void __init psci_update_pwrlvl_limits(void)
 				psci_non_cpu_pd_nodes[nodes_idx[j]].cpu_start_idx
 					= cpu_idx;
 			}
-			psci_non_cpu_pd_nodes[nodes_idx[j]].ncpus++;
+			psci_non_cpu_pd_nodes[SPECULATION_SAFE_VALUE(nodes_idx[j])].ncpus++;
```
danh-arm: Again, I'm struggling to see the threat as this is also in the cold boot path.

lib/psci/psci_stat.c
```diff
@@ -206,9 +206,9 @@ static int psci_get_stat(u_register_t target_cpu, unsigned int power_state,
 
 	if (pwrlvl > PSCI_CPU_PWR_LVL) {
 		/* Get the power domain index */
-		parent_idx = psci_cpu_pd_nodes[target_idx].parent_node;
+		parent_idx = SPECULATION_SAFE_VALUE(psci_cpu_pd_nodes[target_idx].parent_node);
```
danh-arm: Here, I agree with the mitigation, although the likelyhood of an exploit is low. This is in the runtime path and `target_idx` depends on the attacker-controllable `target_cpu` and the platform function `plat_core_pos_by_mpidr`. There is potentially a large range of memory that could be speculatively loaded into the cache (over a number of runs), which an attacker might be able to infer something from.

Given that this is (currently) the only runtime function with v1 mitigation and this is not in the performance hot path, perhaps the commit message should explicitly recommend not using -mtrack-speculation to avoid the code size cost, even though this increases the performance cost of the barrier. I guess this patch implicitly recommends not using this build flag anyway as the flags in the Makefile are not updated. Even so, an indication of the performance cost in the commit message would be useful.

Finally, due to the very theoretical nature of this issue, I think the commit message should play down its severity. I don't think systems without this patch are significantly more vulnerable.

danh-arm: Actually, if we're going to recommend using the hard barrier, then we could implement this within the TF codebase and avoid the dependency on compiler version.

include/lib/utils_def.h
JoelHutton: Done.

lib/pmf/pmf_main.c
JoelHutton: As discussed offline, we have revised decisions around cold boot path.

lib/psci/psci_common.c
JoelHutton: As discussed offline, we have revised decisions around cold boot path.

lib/psci/psci_common.c
JoelHutton: As discussed offline, we have revised decisions around cold boot path.

lib/psci/psci_setup.c
JoelHutton: As discussed offline, we have revised decisions around cold boot path.

lib/psci/psci_stat.c
JoelHutton: Commit message updated to reflect comments. As discussed offline, implementing code within TF codebase as opposed to compiler limits flexibility of future mitigations and option for using tracking in future.

PR #1689, opened 2018-11-20
xlat v2: Support mapping regions with allocated VA

antonio-nino-diaz-arm:
Provide new APIs to add new regions without specifying the base VA.

- `mmap_add_region_alloc_va` adds a static region to mmap choosing as base VA the first possible address after all the currently mapped regions. It is aligned to an appropriate boundary in relation to the size and base PA of the requested region. No attempt is made to fill any unused VA holes.

- `mmap_add_dynamic_region_alloc_va` it adds a region the same way as `mmap_add_region_alloc_va` does, but it's dynamic instead of static.

- `mmap_add_alloc_va` takes an array of non const `mmap_region_t`, maps them in the same way as `mmap_add_region_alloc_va` and fills their `base_va` field. A helper macro has been created to help create the array, called `MAP_REGION_ALLOC_VA`.

Note that this code is already present in TF-A-Tests: https://git.trustedfirmware.org/TF-A/tf-a-tests.git/tree/lib/xlat_tables_v2/xlat_tables_core.c?id=7ec31b8021120437a7951a9944546e58170c87c5#n814

### DIFF
--- a/include/lib/xlat_tables/xlat_tables_v2.h
+++ b/include/lib/xlat_tables/xlat_tables_v2.h
@@ -34,6 +34,12 @@
 	MAP_REGION(_adr, _adr, _sz, _attr)
 
 /*
+ * Helper macro to define entries for mmap_region_t. It allows to define 'pa'
+ * and sets 'va' to 0 for each region. To be used with mmap_add_alloc_va().
+ */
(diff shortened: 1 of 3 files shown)


antonio-nino-diaz-arm:
The CI failures are infrastructure issues.

PR #1694, opened 2018-11-22
Revert "aarch32: Apply workaround for errata 813419 of Cortex-A57"

antonio-nino-diaz-arm:
This reverts commit 6f512a3dfd61662dbdae4912fb6a320ae4d754d5.

According to the 'Cortex-A57 MPCore Software Developers Errata Notice':

    This bug will only affect secure AArch64 EL3. If the above
    conditions occur, the CPU will not invalidate the targeted EL3 TLB
    entries and incorrect translations might occur.

For this reason it is not needed in AArch32.


antonio-nino-diaz-arm:
The CI fails because of infrastructure issues.

PR #1696, opened 2018-11-23
romlib: Add juno support for romlib

satheesbalya-arm:
This patch builds the combined BL1 and ROMLIB
binary file with filler bytes for the right page
size alignment in Juno. The build option
USE_ROMLIB=1 generates the combined file:
bl1_romlib.bin which has to be used instead of
bl1.bin for Juno.

Change-Id: I407efbe48d3e522fa6ef855538a9587193cb1919


satheesbalya-arm:
fvp-tspd-gicv2-symmetric-model failed in the first build. It passed when re-triggered.

PR #1700, opened 2018-11-28
MULTI_CONSOLE_API fixes and cleanups

jwerner-chromium:
@antonio-nino-diaz-arm @soby-mathew 

Fixes ARM-software/tf-issues#646


maintainers.rst
```diff
@@ -60,6 +60,23 @@ eMMC/UFS drivers
 :F: include/drivers/ufs.h
 :F: include/drivers/synopsys/dw_mmc.h
 
+Console API framework
```
antonio-nino-diaz-arm: nit: Move this above the eMMC driver.

drivers/console/multi_console.c
```diff
@@ -0,0 +1,118 @@
...
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include <assert.h>
+#include <drivers/console.h>
+
+console_t *console_list;
+uint8_t console_state = CONSOLE_FLAG_BOOT;
+
+int console_register(console_t *console);	/* for use from asm only */
```
antonio-nino-diaz-arm: Why do we need a declaration here?

drivers/console/multi_console.c
```diff
@@ -0,0 +1,118 @@
...
+ * Copyright (c) 2018, ARM Limited and Contributors. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include <assert.h>
+#include <drivers/console.h>
+
+console_t *console_list;
+uint8_t console_state = CONSOLE_FLAG_BOOT;
```
antonio-nino-diaz-arm: ``console_list`` and ``console_state`` can be static.

drivers/console/multi_console.c
```diff
@@ -0,0 +1,118 @@
...
+
+int console_register(console_t *console);	/* for use from asm only */
+int console_register(console_t *console)
+{
+	extern console_t __STACKS_START__[];
+	extern console_t __STACKS_END__[];
```
antonio-nino-diaz-arm: Use ``IMPORT_SYM()`` instead.

drivers/console/multi_console.c
```diff
@@ -0,0 +1,118 @@
...
+
+	/* Assert that the struct is not on the stack (common mistake). */
+	assert(console < __STACKS_START__ || console >= __STACKS_END__);
```
antonio-nino-diaz-arm: Add parentheses to both subexpressions.

drivers/console/multi_console.c
```diff
@@ -0,0 +1,118 @@
...
+	for (ptr = &console_list; *ptr; ptr = &(*ptr)->next)
+		if (*ptr == to_be_deleted) {
+			*ptr = (*ptr)->next;
+			return to_be_deleted;
+		}
+
+	return NULL;
+}
+
+int console_is_registered(console_t *to_find)
```
antonio-nino-diaz-arm: Maybe this should return bool instead.

drivers/console/multi_console.c
```diff
@@ -0,0 +1,118 @@
...
+
+	/* Return 1 for convenient tail-calling from console_xxx_register(). */
+	return 1;
+}
+
+console_t *console_unregister(console_t *to_be_deleted)
+{
+	console_t **ptr;
+
+	assert(to_be_deleted);
```
antonio-nino-diaz-arm: ``assert(to_be_deleted != NULL);``

drivers/console/multi_console.c
```diff
@@ -0,0 +1,118 @@
...
+
+	for (ptr = &console_list; *ptr; ptr = &(*ptr)->next)
```
antonio-nino-diaz-arm: ``*ptr != NULL``

drivers/console/multi_console.c
```diff
@@ -0,0 +1,118 @@
...
+		}
+
+	return NULL;
+}
+
+int console_is_registered(console_t *to_find)
+{
+	console_t *console;
+
+	assert(to_find);
```
antonio-nino-diaz-arm: ``assert(to_find != NULL);``

drivers/console/multi_console.c
```diff
@@ -0,0 +1,118 @@
...
+
+	for (console = console_list; console; console = console->next)
```
antonio-nino-diaz-arm: ``for (console = console_list; console != NULL; console = console->next)``

drivers/console/multi_console.c
```diff
@@ -0,0 +1,118 @@
...
+}
+
+void console_switch_state(unsigned int new_state)
+{
+	console_state = new_state;
+}
+
+void console_set_scope(console_t *console, unsigned int scope)
+{
+	assert(console);
```
antonio-nino-diaz-arm: ``assert(console != NULL);``

drivers/console/multi_console.c
```diff
@@ -0,0 +1,118 @@
...
+
+	console->flags = (console->flags & ~CONSOLE_FLAG_SCOPE_MASK) | scope;
+}
+
+int console_putc(int c)
+{
+	int err = ERROR_NO_VALID_CONSOLE;
+	console_t *console;
+
+	for (console = console_list; console; console = console->next)
```
antonio-nino-diaz-arm: ``for (console = console_list; console != NULL; console = console->next)``

drivers/console/multi_console.c
```diff
@@ -0,0 +1,118 @@
...
+	return err;
+}
+
+int console_getc(void)
+{
+	int err = ERROR_NO_VALID_CONSOLE;
+	console_t *console;
+
+	do {	/* Keep polling while at least one console works correctly. */
+		for (console = console_list; console; console = console->next)
```
antonio-nino-diaz-arm: ``for (console = console_list; console != NULL; console = console->next)``

drivers/console/multi_console.c
```diff
@@ -0,0 +1,118 @@
...
+
+	return err;
+}
+
+int console_flush(void)
+{
+	int err = ERROR_NO_VALID_CONSOLE;
+	console_t *console;
+
+	for (console = console_list; console; console = console->next)
```
antonio-nino-diaz-arm: ``for (console = console_list; console != NULL; console = console->next)``

drivers/console/multi_console.c
```diff
@@ -0,0 +1,118 @@
...
+
+int console_putc(int c)
+{
+	int err = ERROR_NO_VALID_CONSOLE;
+	console_t *console;
+
+	for (console = console_list; console; console = console->next)
+		if (console->flags & console_state) {
+			int ret = console->putc(c, console);
+			if (err == ERROR_NO_VALID_CONSOLE || ret < err)
```
antonio-nino-diaz-arm: ``if ((err == ERROR_NO_VALID_CONSOLE) || (ret < err))``

drivers/console/multi_console.c
```diff
@@ -0,0 +1,118 @@
...
+
+int console_flush(void)
+{
+	int err = ERROR_NO_VALID_CONSOLE;
+	console_t *console;
+
+	for (console = console_list; console; console = console->next)
+		if (console->flags & console_state) {
+			int ret = console->flush(console);
+			if (err == ERROR_NO_VALID_CONSOLE || ret < err)
```
antonio-nino-diaz-arm: ``if ((err == ERROR_NO_VALID_CONSOLE) || (ret < err))``

antonio-nino-diaz-arm:
IP Review +1

antonio-nino-diaz-arm:
I think that moving most of the multi console driver to C is a good idea. Sorry about all the nitpicks, but that driver is common code and we'd like to keep the newly introduced MISRA defects to a minimum (we don't care about platform code, that's for the maintainer to decide).

Yann-lms:
Tested OK on STM32MP1

drivers/console/multi_console.c
jwerner-chromium: No, they need to be exported to be accessible by crash_console_helpers.S.

drivers/console/multi_console.c
jwerner-chromium: Oh, right, guess we don't. I'm used to writing code for projects that set -Wmissing-prototypes. Will fix.

drivers/console/multi_console.c
jwerner-chromium: I'm following the existing prototype from <console.h>. I wanted to just 1:1 port the assembly implementation into C for now, we can make further changes in later commits.

include/drivers/console.h
```diff
@@ -53,7 +53,7 @@ typedef struct console {
  * registered that way can be unregistered/reconfigured with below functions.
  */
 /* Remove a single console_t instance from the console list. */
-int console_unregister(console_t *console);
+console_t *console_unregister(console_t *console);
```
soby-mathew: Looks like this was wrong before. Would be good to call this out in commit message.

plat/common/aarch64/crash_console_helpers.S
```diff
@@ -18,55 +18,142 @@
...
+func plat_crash_console_init
+	mov	x4, x30		/* x4 is not clobbered by spin_lock() */
+
+	mrs	x1, sctlr_el3
+	tst	x1, #SCTLR_C_BIT
+	beq	skip_spinlock	/* can't synchronize when cache disabled */
+
+	adrp	x0, crash_console_spinlock
+	add	x0, x0, :lo12:crash_console_spinlock
+	bl	spin_lock	/* may spin forever if not the first CPU here */
```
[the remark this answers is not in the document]
jwerner-chromium: Fair enough. I thought rebooting was about the only useful thing you could do in plat_panic_handler. But if there are real use cases where every CPU needs to reach it, I agree that your solution is better.

include/drivers/console.h
jwerner-chromium: Will put in a separate commit for clarity.

plat/common/aarch64/crash_console_helpers.S
```diff
@@ -18,55 +18,142 @@
...
 	 */
-	mov	x0, #0
-	ret
-#else	/* IMAGE_BL1 */
-	mov	x3, x30
-	mov	x0, #CONSOLE_FLAG_CRASH
-	bl	console_switch_state
+.section .data.crash_console_reg_stash
+	.align 3
+	crash_console_reg_stash: .quad 0, 0, 0, 0, 0
```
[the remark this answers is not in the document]
jwerner-chromium: I mean... I think 5 is plenty? Why not wait until we have a console that really needs more to bump this?

plat/common/aarch64/crash_console_helpers.S
```diff
@@ -18,55 +18,142 @@
...
+	/* --------------------------------------------------------------------
+	 * int plat_crash_console_init(void)
+	 * Takes the crash console spinlock (if possible) to prevent other CPUs
+	 * crashing at the same time from interfering. If the lock is already
+	 * taken, CPU will spin here until released. Note that no framework
+	 * function will ever release this lock -- if platforms desire to be
+	 * able to log the output from more than one concurrent crash, they'll
+	 * have to call spin_unlock(&crash_console_spinlock) explicitly from
+	 * their plat_panic_handler() implementation.
+	 * --------------------------------------------------------------------
```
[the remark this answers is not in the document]
jwerner-chromium: Will add.

plat/common/aarch64/platform_helpers.S
```diff
@@ -35,6 +35,8 @@ func plat_report_exception
 endfunc plat_report_exception
 
 #if !ERROR_DEPRECATED
+	.weak	crash_console_spinlock
+	.weak	crash_console_reg_stash
```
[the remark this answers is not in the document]
jwerner-chromium: When a platform has !ERROR_DEPRECATED but also includes crash_console_helpers.S explicitly, all this stuff gets built twice, so everything in the platform_helpers.S version needs to be weak.

plat/common/aarch64/platform_helpers.S
```diff
@@ -40,78 +35,10 @@ func plat_report_exception
...
-endfunc plat_crash_console_putc
-
-func plat_crash_console_flush
-	ret
-endfunc plat_crash_console_flush
-#endif
+	.weak	plat_crash_console_init
+	.weak	plat_crash_console_putc
+	.weak	plat_crash_console_flush
+#include "crash_console_helpers.S"
```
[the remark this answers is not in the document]
jwerner-chromium: I really don't want a solution where every platform has to add more boilerplate. The whole point of moving it to crash_console_helpers.S was that you could include it with one Makefile line. If you don't want this, I think the next best solution would just remove the MULTI_CONSOLE_API part here (it was already broken anyway) and keep the empty stubs duplicated.

However, I don't think there's any problem with this implementation. .weak always overrides .globl (or I think you could also say .weak implies .globl, since a weak symbol that's not visible to the linker would be pointless).

Here's how this object file looks assembled for me in readelf:

        61: 0000000000000000    36 FUNC    WEAK   DEFAULT    7 plat_crash_console_init 

...and in objdump:

    0000000000000000  w    F .text.asm.plat_crash_console_init      0000000000000024 plat_crash_console_init

plat/common/aarch64/crash_console_helpers.S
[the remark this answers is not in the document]
soby-mathew: OK, we can change in future if required.

plat/common/aarch64/platform_helpers.S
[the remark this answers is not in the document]
soby-mathew: I see. If the behaviour is defined when both .weak and .globl are specified, I am OK with what is done currently.

plat/common/aarch64/platform_helpers.S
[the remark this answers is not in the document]
soby-mathew: I see, thanks for the clarification.

soby-mathew:
IP Review +1

soby-mathew:
jenkins: test this please

Makefile
```diff
@@ -488,6 +488,12 @@ endif
 
 include lib/stack_protector/stack_protector.mk
 
+ifeq (${MULTI_CONSOLE_API},1)
+BL_COMMON_SOURCES	+=	drivers/console/${ARCH}/multi_console.S
+else
+BL_COMMON_SOURCES	+=	drivers/console/${ARCH}/deprecated_console.S
```
soby-mathew: Including the `deprecated_console` by default is causing build failure on rcar platform:

./build/rcar/debug/bl2/scif.o:/work/workspace/workspace/tf-worker@9/trusted_firmware/drivers/renesas/rcar/scif/scif.S:128: first defined here
11:45:09 ./build/rcar/debug/bl2/deprecated_console.o: In function `console_uninit':
11:45:09 (.text.asm.console_uninit+0x0): multiple definition of `console_uninit'
11:45:09 ./build/rcar/debug/bl2/scif.o:/work/workspace/workspace/tf-worker@9/trusted_firmware/drivers/renesas/rcar/scif/scif.S:128: first defined here

Since the `deprecated_console` is set to be removed soon, better not to bother with the change.

jwerner-chromium:
I think(?) I resolved all the issues. If there's still anything left please copy the full build errors from the CI output so I have a better idea of what's going wrong.

I think the generic crash code may still be useful for BL1/BL2 on certain platforms, so I don't want to kill it completely (platforms can still choose on their own whether to compile it in or not for each stage, after all). I just made it return success immediately for init(), since there should be no need to synchronize in stages where only one CPU is running code anyway.

antonio-nino-diaz-arm:
jenkins: test this please

antonio-nino-diaz-arm:
There are still a few problems. This is the output of the Marvell builds (a3700 builds fine, the rest don't):
```
./build/a80x0/debug/ble/marvell_console.o: In function `marvell_console_boot_init':
/work/workspace/workspace/tf-worker@10/trusted_firmware/plat/marvell/common/marvell_console.c:47: undefined reference to `console_flush'
/work/workspace/workspace/tf-worker@10/trusted_firmware/plat/marvell/common/marvell_console.c:47:(.text.marvell_console_boot_init+0x44): relocation truncated to fit: R_AARCH64_CALL26 against undefined symbol `console_flush'
/work/workspace/workspace/tf-worker@10/trusted_firmware/plat/marvell/common/marvell_console.c:50: undefined reference to `console_set_scope'
/work/workspace/workspace/tf-worker@10/trusted_firmware/plat/marvell/common/marvell_console.c:50:(.text.marvell_console_boot_init+0x5c): relocation truncated to fit: R_AARCH64_JUMP26 against undefined symbol `console_set_scope'
./build/a80x0/debug/ble/backtrace.o: In function `backtrace':
/work/workspace/workspace/tf-worker@10/trusted_firmware/common/backtrace/backtrace.c:247: undefined reference to `console_flush'
/work/workspace/workspace/tf-worker@10/trusted_firmware/common/backtrace/backtrace.c:247:(.text.backtrace+0x20): relocation truncated to fit: R_AARCH64_CALL26 against undefined symbol `console_flush'
./build/a80x0/debug/ble/platform_helpers.o: In function `plat_crash_console_putc':
/work/workspace/workspace/tf-worker@10/trusted_firmware/plat/common/aarch64/crash_console_helpers.S:120: undefined reference to `console_list'
/work/workspace/workspace/tf-worker@10/trusted_firmware/plat/common/aarch64/crash_console_helpers.S:121: undefined reference to `console_list'
./build/a80x0/debug/ble/platform_helpers.o: In function `plat_crash_console_flush':
/work/workspace/workspace/tf-worker@10/trusted_firmware/plat/common/aarch64/crash_console_helpers.S:158: undefined reference to `console_list'
/work/workspace/workspace/tf-worker@10/trusted_firmware/plat/common/aarch64/crash_console_helpers.S:159: undefined reference to `console_list'
./build/a80x0/debug/ble/16550_console.o: In function `console_16550_register':
/work/workspace/workspace/tf-worker@10/trusted_firmware/drivers/ti/uart/aarch64/16550_console.S:116: undefined reference to `console_register'
/work/workspace/workspace/tf-worker@10/trusted_firmware/drivers/ti/uart/aarch64/16550_console.S:116:(.text.asm.console_16550_register+0x4c): relocation truncated to fit: R_AARCH64_JUMP26 against undefined symbol `console_register'
/work/workspace/workspace/tf-worker@10/trusted_firmware/./build/a80x0/debug/lib/libc.a(assert.o): In function `__assert':
/work/workspace/workspace/tf-worker@10/trusted_firmware/lib/libc/assert.c:32: undefined reference to `console_flush'
/work/workspace/workspace/tf-worker@10/trusted_firmware/lib/libc/assert.c:32:(.text.__assert+0x28): relocation truncated to fit: R_AARCH64_CALL26 against undefined symbol `console_flush'
/work/workspace/workspace/tf-worker@10/trusted_firmware/./build/a80x0/debug/lib/libc.a(putchar.o): In function `putchar':
/work/workspace/workspace/tf-worker@10/trusted_firmware/lib/libc/putchar.c:13: undefined reference to `console_putc'
/work/workspace/workspace/tf-worker@10/trusted_firmware/lib/libc/putchar.c:13:(.text.putchar+0x14): relocation truncated to fit: R_AARCH64_CALL26 against undefined symbol `console_putc'
plat/marvell/a8k/common/a8k_common.mk:114: recipe for target 'build/a80x0/debug/ble/ble.elf' failed
```

This is the output with clang for FVP:
```
./build/fvp/debug/bl1/arm_helpers.o: In function `plat_crash_console_flush':
/work/workspace/workspace/tf-worker@10/trusted_firmware/include/common/asm_macros_common.S:358: multiple definition of `plat_crash_console_flush'
./build/fvp/debug/bl1/platform_helpers.o:/work/workspace/workspace/tf-worker@10/trusted_firmware/include/common/asm_macros_common.S:412: first defined here
./build/fvp/debug/bl1/arm_helpers.o: In function `plat_crash_console_init':
/work/workspace/workspace/tf-worker@10/trusted_firmware/include/common/asm_macros_common.S:334: multiple definition of `plat_crash_console_init'
./build/fvp/debug/bl1/platform_helpers.o:/work/workspace/workspace/tf-worker@10/trusted_firmware/include/common/asm_macros_common.S:337: first defined here
./build/fvp/debug/bl1/arm_helpers.o: In function `plat_crash_console_putc':
/work/workspace/workspace/tf-worker@10/trusted_firmware/include/common/asm_macros_common.S:334: multiple definition of `plat_crash_console_putc'
./build/fvp/debug/bl1/platform_helpers.o:/work/workspace/workspace/tf-worker@10/trusted_firmware/include/common/asm_macros_common.S:337: first defined here
Makefile:681: recipe for target 'build/fvp/debug/bl1/bl1.elf' failed
make: *** [build/fvp/debug/bl1/bl1.elf] Error 1
```

jwerner-chromium:
Hmmm... the clang thing really confuses me. When I build FVP with CC=clang, it fails the way you posted, and the symbols in platform_helpers.o are marked as global, not weak. But when I build with CC=aarch64-cros-linux-gnu-clang (the cross-compiler I normally use), it works fine the way I described before (with the symbols ending up as weak). According to --version both of my compilers were built from the exact same source commit, just that clang has the default target x86_64-pc-linux-gnu and aarch64-cros-linux-gnu-clang has the default target aarch64-cros-linux-gnu. The binary stills ends up as aarch64 either way (because of the -target aarch64-elf) but it seems that there are still weird behavior differences in the compiler depending on which default target it was configured for.

Anyway, I guess this just means that Soby was right and the behavior when both .globl and .weak are present is not reliable enough between compilers. I'll change it to only implement the MULTI_CONSOLE_API stuff in crash_console_helpers.S and leave weak no-op stubs in platform_helpers.S without the #include trick, then.

The Marvell problem is due to them building a non-standard BL ("ble") by calling MAKE_BL directly in their platform Makefile. MAKE_BL evaluates BL_COMMON_SOURCES, but the problem is that this happens at a point where the main Makefile code to add multi_console.S to BL_COMMON_SOURCES hasn't run yet. I also can't move that above the include line for the platform Makefile, because it depends on whether MULTI_CONSOLE_API is set which usually happens in the platform Makefile.

I think possible options to fix this would be to: 

1) Add multi_console.S to the files explicitly added to that non-standard BL (with this changes would be only limited to the platform Makefile, but it's very hacky)

2) Have a list of non-standard BL names that platforms may append to, e.g.:

  platform.mk:
    EXTRA_BLS += e

  Makefile (near the end):
    ifneq(${EXTRA_BLS},)
      foreach(bl,${EXTRA_BLS},$(eval $(call MAKE_BL,${bl})))
    endif

3) Take this even further to allow platform Makefiles to defer arbitrary stuff to be evaluated at the end:

  platform.mk:
    EVAL_AT_END += $$(eval $$(call MAKE_BL,e))

  Makefile (near the end):
    $(eval ${EVAL_AT_END})

Any preference?

soby-mathew:
Thanks for getting to the bottom of this.

> I'll change it to only implement the MULTI_CONSOLE_API stuff in crash_console_helpers.S and leave weak no-op stubs in platform_helpers.S without the #include trick, then.

Ok, sounds good.

> I think possible options to fix this would be to:
> Any preference?

Its not preferable to add evaluations that depend on the placement in the Makefile. How about adding the multi_console.c always to BL_COMMON_SOURCES and wrapping the implementation within #if MULTI_CONSOLE_API ? The build failures seem to be resolved with this change.

```diff
diff --git a/Makefile b/Makefile
index 391df13..eed7c7a 100644
--- a/Makefile
+++ b/Makefile
@@ -214,6 +214,7 @@ include lib/libc/libc.mk
 BL_COMMON_SOURCES      +=      common/bl_common.c                      \
                                common/tf_log.c                         \
                                common/${ARCH}/debug.S                  \
+                               drivers/console/multi_console.c         \
                                lib/${ARCH}/cache_helpers.S             \
                                lib/${ARCH}/misc_helpers.S              \
                                plat/common/plat_bl_common.c            \
@@ -488,10 +489,6 @@ endif
 
 include lib/stack_protector/stack_protector.mk
 
-ifeq (${MULTI_CONSOLE_API},1)
-BL_COMMON_SOURCES      +=      drivers/console/multi_console.c
-endif
-
 ################################################################################
 # Auxiliary tools (fiptool, cert_create, etc)
 ################################################################################
diff --git a/drivers/console/multi_console.c b/drivers/console/multi_console.c
index 6a0ceb9..d0d84e8 100644
--- a/drivers/console/multi_console.c
+++ b/drivers/console/multi_console.c
@@ -7,6 +7,8 @@
 #include <assert.h>
 #include <drivers/console.h>
 
+#if MULTI_CONSOLE_API
+
 console_t *console_list;
 uint8_t console_state = CONSOLE_FLAG_BOOT;
 
@@ -116,3 +118,5 @@ int console_flush(void)
 
        return err;
 }
+
+#endif

```

jwerner-chromium:
> Its not preferable to add evaluations that depend on the placement in the Makefile. How about adding the multi_console.c always to BL_COMMON_SOURCES and wrapping the implementation within #if MULTI_CONSOLE_API ? The build failures seem to be resolved with this change.

Good idea, that works too. (Although it's its own sort of ugly in a way, but hopefully we can get rid of the old console API eventually and then it'll not be a problem anymore.)

soby-mathew:
jenkins: test this please

PR #1704, opened 2018-12-03
Arm/master/memsize passing v1

marex:
This patchset adds support for passing DRAM configuration and board compatible string from ATF to next stages.

### DIFF
--- a/plat/renesas/rcar/bl2_plat_mem_params_desc.c
+++ b/plat/renesas/rcar/bl2_plat_mem_params_desc.c
@@ -8,6 +8,7 @@
 #include <desc_image_load.h>
 #include <platform.h>
 #include <platform_def.h>
+#include <xlat_tables_defs.h>
 
 #if (RCAR_BL33_EXECUTION_EL != 0) && (RCAR_BL33_EXECUTION_EL != 1)
 #error
@@ -18,6 +19,8 @@
 #else
 #define BL33_MODE MODE_EL2
 #endif
+
+extern uint64_t fdt_blob[PAGE_SIZE_4KB / sizeof(uint64_t)];
 
 static bl_mem_params_node_t bl2_mem_params_descs[] = {
 	{
@@ -69,6 +72,7 @@ static bl_mem_params_node_t bl2_mem_params_descs[] = {
 #ifdef RCAR_BL33_ARG0
 		.ep_info.args.arg0 = RCAR_BL33_ARG0,
 #endif
+		.ep_info.args.arg3 = (uintptr_t)fdt_blob,
 		SET_STATIC_PARAM_HEAD(image_info, PARAM_EP, VERSION_2,
 			image_info_t, 0),
 		.image_info.image_max_size =
--- a/plat/renesas/rcar/bl2_plat_setup.c
+++ b/plat/renesas/rcar/bl2_plat_setup.c
@@ -300,6 +300,42 @@ meminfo_t *bl2_plat_sec_mem_layout(void)
 	return &bl2_tzram_layout;
 }
 
+static void bl2_advertise_dram_size(uint32_t product, uint32_t cut)
+{
+	/* Later than H3 Ver.3.0 */
+	if (product == RCAR_PRODUCT_H3 && cut >= RCAR_CUT_VER30) {
+#if (RCAR_DRAM_LPDDR4_MEMCONF == 0)
+		/* 4GB(1GBx4) */
+		NOTICE("BL2: CH0: 0x400000000 - 0x43fffffff, 1 GiB\n");
+		NOTICE("BL2: CH1: 0x500000000 - 0x53fffffff, 1 GiB\n");
+		NOTICE("BL2: CH2: 0x600000000 - 0x63fffffff, 1 GiB\n");
+		NOTICE("BL2: CH3: 0x700000000 - 0x73fffffff, 1 GiB\n");
+#elif (RCAR_DRAM_LPDDR4_MEMCONF == 1) && \
+      (RCAR_DRAM_CHANNEL        == 5) && \
+      (RCAR_DRAM_SPLIT          == 2)
+		/* 4GB(2GBx2 2ch split) */
+		NOTICE("BL2: CH0: 0x400000000 - 0x47fffffff, 2 GiB\n");
+		NOTICE("BL2: CH1: 0x500000000 - 0x57fffffff, 2 GiB\n");
+#elif (RCAR_DRAM_LPDDR4_MEMCONF == 1) && (RCAR_DRAM_CHANNEL == 15)
+		/* 8GB(2GBx4: default) */
+		NOTICE("BL2: CH0: 0x400000000 - 0x47fffffff, 2 GiB\n");
+		NOTICE("BL2: CH1: 0x500000000 - 0x57fffffff, 2 GiB\n");
+		NOTICE("BL2: CH2: 0x600000000 - 0x67fffffff, 2 GiB\n");
+		NOTICE("BL2: CH3: 0x700000000 - 0x77fffffff, 2 GiB\n");
+#endif /* RCAR_DRAM_LPDDR4_MEMCONF == 0 */
+	}
+
+	if (product == RCAR_PRODUCT_E3) {
+#if (RCAR_DRAM_DDR3L_MEMCONF == 0)
+		/* 1GB(512MBx2) */
+		NOTICE("BL2: 0x400000000 - 0x43fffffff, 1 GiB\n");
+#elif (RCAR_DRAM_DDR3L_MEMCONF == 1)
+		/* 2GB(512MBx4) */
+		NOTICE("BL2: 0x400000000 - 0x47fffffff, 2 GiB\n");
+#endif /* RCAR_DRAM_DDR3L_MEMCONF == 0 */
+	}
+}
+
 void bl2_el3_early_platform_setup(u_register_t arg1, u_register_t arg2,
 				  u_register_t arg3, u_register_t arg4)
 {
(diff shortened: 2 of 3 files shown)


ldts:
thanks @marex. will try to do this before EOD (GMT) but all seems good to me.

plat/renesas/rcar/bl2_plat_setup.c
```diff
@@ -85,6 +86,32 @@ CASSERT((PARAMS_BASE + sizeof(bl2_to_bl31_params_mem_t) + 0x100)
 
 static meminfo_t bl2_tzram_layout __aligned(CACHE_WRITEBACK_GRANULE);
 
+/* FDT with DRAM configuration */
+static uint64_t fdt_blob[4096 / sizeof(uint64_t)];
```
antonio-nino-diaz-arm: 4096? Is this meant to be the page size, or is it just an arbitrary number?

plat/renesas/rcar/bl2_plat_setup.c
```diff
@@ -314,6 +343,54 @@ static void bl2_advertise_dram_entries(uint64_t dram_config[8])
...
+		size = dram_config[2 * chan + 1];
+		if (!size)
+			continue;
+
+		/*
+		 * Channel 0 is mapped in 32bit space and the first
+		 * 128 MiB are reserved
+		 */
+		if (chan == 0) {
+			start = 0x48000000;
```
antonio-nino-diaz-arm: It would be better to not use magic numbers.

plat/renesas/rcar/bl2_plat_setup.c
marex: Arbitrary number into which the DT fits. What better size than a page-aligned one.

plat/renesas/rcar/bl2_plat_setup.c
marex: The comment above explains those.

plat/renesas/rcar/bl2_plat_setup.c
ldts: why not then use PAGE_SIZE_4KB?

marex:
On 12/05/2018 10:53 PM, ldts wrote:
> *@ldts* commented on this pull request.
> 
> ------------------------------------------------------------------------
> 
> In plat/renesas/rcar/bl2_plat_setup.c
> <https://github.com/ARM-software/arm-trusted-firmware/pull/1704#discussion_r239251899>:
> 
>> @@ -85,6 +86,32 @@ CASSERT((PARAMS_BASE + sizeof(bl2_to_bl31_params_mem_t) + 0x100)
>  
>  static meminfo_t bl2_tzram_layout __aligned(CACHE_WRITEBACK_GRANULE);
>  
> +/* FDT with DRAM configuration */
> +static uint64_t fdt_blob[4096 / sizeof(uint64_t)];
> 
> why not then use PAGE_SIZE_4KB?

Fixed

-- 
Best regards,
Marek Vasut

marex:
On 12/06/2018 09:11 AM, ldts wrote:
> *@ldts* commented on this pull request.
> 
> it is not clear to me how the device tree makes it to bl33 in the
> ep_info.args. AFAICS you'd have to build with RCAR_BL33_ARG0 holding the
> value of the address? could you clarify please? I think we discussed
> this some weeks ago but I am not sure what you decided in the end.

The DT is being passed via arg3, except that code disappeared during the
rebase on the ATF master. Now my question is, where/how do I set the
arg3 in ATF/master ? The old function to do that is gone.

-- 
Best regards,
Marek Vasut

ldts:
what function were you using before rebasing? I think you should be able to just pass the dt address with a change in bl2_plat_mem_params_desc.c (using arg3 instead of arg0)

marex:
On 12/06/2018 09:58 AM, ldts wrote:
> what function were you using before rebasing? I think you should be able
> to just pass the dt address with a change in bl2_plat_mem_params_desc.c
> (using arg3 instead of arg0)

bl2_plat_set_bl33_ep_info() . I guess I can put it into the static
table, since the FDT buffer is also static chunk of memory.

-- 
Best regards,
Marek Vasut

marex:
On 12/06/2018 10:26 AM, ldts wrote:
> maybe you can call get_bl_mem_params_node(BL33_IMAGE_ID) from
> bl2_plat_handle_pre_image_load and update the node there with the fdt
> address in arg3?

Maybe I can put it into the static bl2_mem_params_desc[] array , since
it's a static chunk of memory anyway ?

-- 
Best regards,
Marek Vasut

ldts:
sounds good to me

ldts:
how this going?

ldts:
thanks Marex, looks good to me. @antonio-nino-diaz-arm ?

antonio-nino-diaz-arm:
Looks good to me.

antonio-nino-diaz-arm:
jenkins: test this please

antonio-nino-diaz-arm:
The CI failures are due to merging some patches in TF-A-Tests, this PR is correct.

plat/renesas/rcar/bl2_plat_mem_params_desc.c
```diff
@@ -19,6 +20,8 @@
 #define BL33_MODE MODE_EL2
 #endif
 
+extern uint64_t fdt_blob[PAGE_SIZE_4KB / sizeof(uint64_t)];
```
soby-mathew: Is the kernel DTB being created in secure memory ? How would BL33 access the secure memory ?

Also, would prefer to use arg1 for the kernel DTB to be inline with Dynamic config framework in TF-A. https://github.com/ARM-software/arm-trusted-firmware/blob/master/docs/firmware-design.rst#11dynamic-configuration-during-cold-boot

marex:
On 12/10/2018 04:13 PM, Soby Mathew wrote:
> *@soby-mathew* commented on this pull request.
> 
> ------------------------------------------------------------------------
> 
> In plat/renesas/rcar/bl2_plat_mem_params_desc.c
> <https://github.com/ARM-software/arm-trusted-firmware/pull/1704#discussion_r240248769>:
> 
>> @@ -19,6 +20,8 @@
>  #define BL33_MODE MODE_EL2
>  #endif
>  
> +extern uint64_t fdt_blob[PAGE_SIZE_4KB / sizeof(uint64_t)];
> 
> Is the kernel DTB being created in secure memory ? How would BL33 access
> the secure memory ?

I can just read the memory address passed in x3 from U-Boot .

> Also, would prefer to use arg1 for the kernel DTB to be inline with
> Dynamic config framework in TF-A.
> https://github.com/ARM-software/arm-trusted-firmware/blob/master/docs/firmware-design.rst#11dynamic-configuration-during-cold-boot
Is this something we want even if the DT is just a small fragment and
not a complete DT ?

-- 
Best regards,
Marek Vasut

soby-mathew:
> I can just read the memory address passed in x3 from U-Boot .

OK, I am not sure whether the fdt_blob[] is being copied to non-secure memory prior to handover to uboot. If x3 is pointing a memory within the secure address space (i.e within B2), and Uboot shouldn't be able to read it.

> Is this something we want even if the DT is just a small fragment and
not a complete DT ?

In this case, it can be either treated as config data to uboot (i.e NT_FW_CONFIG) or as kernel DT (HW_CONFIG). But I see you are already using x0 for other purpose, hence using x1 (HW_CONFIG) might be easier.  Although this is optional, it might be easier to do the changes now rather than later as this would become ABI between TF-A and UBoot for rcar.

marex:
On 12/12/2018 10:49 AM, Soby Mathew wrote:
>     I can just read the memory address passed in x3 from U-Boot .
> 
> OK, I am not sure whether the fdt_blob[] is being copied to non-secure
> memory prior to handover to uboot. If x3 is pointing a memory within the
> secure address space (i.e within B2), and Uboot shouldn't be able to
> read it.
> 
>     Is this something we want even if the DT is just a small fragment and
>     not a complete DT ?
> 
> In this case, it can be either treated as config data to uboot (i.e
> NT_FW_CONFIG) or as kernel DT (HW_CONFIG). But I see you are already
> using x0 for other purpose, hence using x1 (HW_CONFIG) might be easier.

So like this ?
https://github.com/marex/arm-trusted-firmware-1/commit/06ce6912b314fc32b7971a424417429c633d3df6

> Although this is optional, it might be easier to do the changes now
> rather than later as this would become ABI between TF-A and UBoot for rcar.

It's fine by me either way, better be consistent.

-- 
Best regards,
Marek Vasut

soby-mathew:
> So like this ?
> marex@06ce691

Yes, that should be OK. 

Also, could you clarify about the memory being pointed to ?  If it is memory within BL2, and it is accessible by BL33, it is a security violation and needs to be fixed (via suitable TrustZone settings).

marex:
On 12/14/2018 01:19 PM, Soby Mathew wrote:
>     So like this ?
>     marex/arm-trusted-firmware-1@06ce691
>     <https://github.com/marex/arm-trusted-firmware-1/commit/06ce691>
> 
> Yes, that should be OK.

Fine, let me send a PR.

(I presume git request-pull is not supported by ATF or github, right?)

> Also, could you clarify about the memory being pointed to ? If it is
> memory within BL2, and it is accessible by BL33, it is a security
> violation and needs to be fixed (via suitable TrustZone settings).

What am I looking for exactly ? I'm still getting into this ATF stuff.

-- 
Best regards,
Marek Vasut

soby-mathew:
A PR should be fine.

> What am I looking for exactly ? I'm still getting into this ATF stuff.

Basically all of the firmware is supposed to run in memory which is `Secure` access only. Software without `Secure` world privilege like U-Boot/Linux cannot access this memory.

The memory can be `secure` either via hardware configuration or runtime configurable by software. If configurable, this would be done with the aid of a TrustZone controller (TZC-400) or something similar. There are some security setting in `plat/renesas/rcar/bl2_secure_setting.c` but not sure whether this configures the security for system RAM as well.

soby-mathew:
@AlexeiFedorov gave me some inputs, but we didn't figure out whether System RAM is Secure-access only or not.

marex:
On 12/17/18 6:32 PM, Soby Mathew wrote:
> @AlexeiFedorov <https://github.com/AlexeiFedorov> gave me some inputs,
> but we didn't figure out whether System RAM is Secure-access only or not.

If I dump the xlat tables (v2) for the SRAM, it is marked as
MEM-RO-EXEC-S (and a chunk which is MEM-RW-XN-S), so I take it is secure
access only.

-- 
Best regards,
Marek Vasut

PR #1708, opened 2018-12-04
Add possibility to add compilation warnings

Yann-lms:
This is based on the same warnings lists as defined in kernel.
To activate this, the option W=x has to be provided in make command, with x=1, 2 or 3.
W=1 adds warning of the first list, W=2 adds warnings of the first and second list, and W=3 adds the warnings of the 3 lists.
When activating this, the -Werror is disabled, or else we would only have the first warning(s), and the compilation would end.

This series also includes 2 patches that corrects some warnings.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -226,8 +226,12 @@ else ifeq (${W},3)
 WARNINGS := $(WARNING1) $(WARNING2) $(WARNING3)
 endif
 
+ifneq (${E},0)
+ERRORS := -Werror
+endif
+
 CPPFLAGS		=	${DEFINES} ${INCLUDES} ${MBEDTLS_INC} -nostdinc		\
-				-Wmissing-include-dirs -Werror $(WARNINGS)
+				-Wmissing-include-dirs $(ERRORS) $(WARNINGS)
 ASFLAGS			+=	$(CPPFLAGS) $(ASFLAGS_$(ARCH))			\
 				-D__ASSEMBLY__ -ffreestanding 			\
 				-Wa,--fatal-warnings
--- a/Makefile
+++ b/Makefile
@@ -188,8 +188,46 @@ TF_CFLAGS_aarch64	+=	-mgeneral-regs-only -mstrict-align
 ASFLAGS_aarch32		=	$(march32-directive)
 ASFLAGS_aarch64		=	-march=armv8-a
 
+WARNING1 := -Wextra
+WARNING1 += -Wunused -Wno-unused-parameter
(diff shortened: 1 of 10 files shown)


hzhuang1:
Tested on hikey.

shawnguo2:
OK for Poplar.

Makefile
```diff
@@ -188,8 +188,48 @@ TF_CFLAGS_aarch64	+=	-mgeneral-regs-only -mstrict-align
...
+WARN_OR_ERROR := $(WARNING1) $(WARNING2)
+else ifeq (${W},3)
+WARN_OR_ERROR := $(WARNING1) $(WARNING2) $(WARNING3)
+else
+WARN_OR_ERROR := -Werror
+endif
+
 CPPFLAGS		=	${DEFINES} ${INCLUDES} ${MBEDTLS_INC} -nostdinc		\
-				-Wmissing-include-dirs -Werror
+				-Wmissing-include-dirs $(WARN_OR_ERROR)
```
antonio-nino-diaz-arm: It's a bit weird to leave ``-Wall`` in ``TF_CFLAGS``, right?

antonio-nino-diaz-arm:
IP Review +1

bryanodonoghue:
lgtm

Makefile
Yann-lms: With this patch I didn't want to change what is currently done, that's why I kept the -Werror if no W=x option is provided in the command line.
-Wall may be misleading, as it does not includes the warnings that are added in this patch, as stated here: https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html

Yann-lms:
I've pushed an updated version as we got compilation error in AARCH32 if BL2_IN_XIP_MEM was enabled

Makefile
soby-mathew: I think it would be useful to set W=3 and also -Werror, once all the warning are fixed. In this case, it would be better to disable -Werror with another variable and have it enabled by default.

Yann-lms: OK, I can then keep the -Werror as it is now, just add the warnings if W=x is provided. And then make another patch to put -Werror depending on a command line variable, why not E=0 to deactivate errors?

soby-mathew:
jenkins: test this please

Makefile
soby-mathew: OK, sounds good.

Yann-lms:
I've updated the patch set to manage -Werror

soby-mathew:
jenkins: test this please

PR #1715, opened 2018-12-07
ccn: for RN-I, used node id instead of node postion

pangupta:
For RN-I, node id is used instead of node postion in the bitmap
to calculate the region id.

Signed-off-by: Pankaj Gupta <pankaj.gupta@nxp.com>

### DIFF
--- a/drivers/arm/ccn/ccn.c
+++ b/drivers/arm/ccn/ccn.c
@@ -553,7 +553,14 @@ static unsigned int get_region_id_for_node(node_types_t node_type,
 		return REGION_ID_LIMIT;
 	}
 
-	region_id += node_pos_in_map;
+	/*
+	 * According to section 3.1.1 in CCN specification, region offset for
+	 * the RN-I components is calculated as (128 + NodeID of RN-I).
+	 */
+	if (node_type == NODE_TYPE_RNI)
+		region_id += node_id;
+	else
+		region_id += node_pos_in_map;
 
 	return region_id;
 }


antonio-nino-diaz-arm:
IP Review +1

drivers/arm/ccn/ccn.c
```diff
@@ -553,7 +553,13 @@ static unsigned int get_region_id_for_node(node_types_t node_type,
 		return REGION_ID_LIMIT;
 	}
 
-	region_id += node_pos_in_map;
+	/* For RN-I like nodes, region id depends on the node id
```
soby-mathew: Need more justification why this is the case. Could you clarify from the specification ? 

The assertion at line 539 seems to suggest that this node must be present in the bitmap. Is it the case that RN-I node_bitmap is (oly_rni_nodeid_list | oly_rnidvm_nodeid_list) ?

Also the comment should follow multi line comment format used at line line 526

pangupta: image

The above snapshot is taken from ARM CCN-508 TRM.
You can see that, for RN-I, the region offset uses node id instead of node position.

soby-mathew: would be good to have this reference in the code comments above.

pangupta: Comments updated.

soby-mathew:
jenkins: test this please

PR #1719, opened 2018-12-13
stm32mp1: remove useless compilation flags

Yann-lms:
On AARCH32, thumb is used by default, no need to redefine it.
As all our binaries are compiled with thumb, interwork is not needed.
The binaries compiled with or without those flags are the same,
except of course for the date.


antonio-nino-diaz-arm:
jenkins: test this please

antonio-nino-diaz-arm:
jenkins: test this please

PR #1720, opened 2018-12-14
GIC: Remove lowest priority constants

jeenu-arm:
The GIC lowest priority values for each world depends on the number of
priority values implemented in hardware. These constants currently
defined in gic_common.h only meant to enumerate lowest possible
architectural values. Since these values are not used in generic code or
upstream platforms, and that general use of these constants can be
wrong, remove these. Platforms should either define and use these as
appropriate, or determine correct values at run time.

Change-Id: I3805cea8ceb8a592b9eff681ea1b63b7496cec5f


soby-mathew:
Fixes ARM-Software/tf-issues#645

PR #1721, opened 2018-12-15
plat: rcar: Move FDT from x3 to x1

marex:
As suggested, pass the FDT to BL 33 via x1 instead of x3 , to be
consistent with the other platforms.


antonio-nino-diaz-arm:
IP Review +1

antonio-nino-diaz-arm:
jenkins: test this please

PR #1724, opened 2018-12-18
mmc: increase delay when initializing mmc

jbech-linaro:
Running TF-A 2.0 and later seems to cause a regression on the old CircuitCo HiKey 620 4GB (issue is not seen when running on LeMaker HiKey 620 8GB).

    NOTICE:  BL2: v2.0(release):v2.0
    NOTICE:  BL2: Built : 17:41:23, Dec 17 2018
    NOTICE:  acpu_dvfs_set_freq: set acpu freq success!ERROR:   CMD1 failed after 100 retries
    ERROR:   BL2: Failed to load image (-5)

The reason seems to be that during emmc enumeration when BL2 sends the command
    OCR_SECTOR_MODE | OCR_VDD_MIN_2V7 | OCR_VDD_MIN_1V7

it for some reason takes some more time to get a reply. So a delay with
mdelay(1), seems to not be enough any longer and therefore we increase it to
mdelay(10) instead which makes the device boot as expected again.

Signed-off-by: Joakim Bech <joakim.bech@linaro.org>

Note that I've tried with mdelay(5) which also seems to work, but since this is running in a loop running 100 times. I guess it doesn't actually matter that much if it's 1, 5 or 10. Since when the power state has changed it will take at most 10ms (instead of 1) to continue. This fixes the issue we've discussed in https://github.com/OP-TEE/optee_os/pull/2703#issuecomment-447924390.

### DIFF
--- a/drivers/mmc/mmc.c
+++ b/drivers/mmc/mmc.c
@@ -404,7 +404,7 @@ static int mmc_send_op_cond(void)
 			return 0;
 		}
 
-		mdelay(1);
+		mdelay(10);
 	}
 
 	ERROR("CMD1 failed after %d retries\n", SEND_OP_COND_MAX_RETRIES);


antonio-nino-diaz-arm:
IP Review +1

jbech-linaro:
@vchong , maybe that patch plays a role, but I still cannot make it work without the delay, this is what I have tried:
```bash
Commit          Flags                                  PATCH applied                                                        
===========================================================================                                          
dbc8d94         CFG_DRAM_SIZE_GB=1 CFG_FLASH_SIZE=4    NO       ---> ERROR                                           
dbc8d94         <NO FLAGS>                             NO       ---> ERROR                                           
dbc8d94         CFG_DRAM_SIZE_GB=1 CFG_FLASH_SIZE=4    YES      ---> ERROR                                           
dbc8d94         <NO_FLAGS>                             YES      ---> ERROR                                           
                                                                                                                     
e010654         CFG_DRAM_SIZE_GB=1 CFG_FLASH_SIZE=4    NO       ---> ERROR                                           
e010654         <NO_FLAGS>                             NO       ---> ERROR                                           
e010654         CFG_DRAM_SIZE_GB=1 CFG_FLASH_SIZE=4    YES      ---> OK!                                             
e010654         <NO_FLAGS>                             YES      ---> OK!  
```
There `dbc8d94` is the 2.0 which we are referring to in our manifest and `e010654` is the latest as of today.

vchong:
@jbech-linaro Sorry for not being clear. I mean, rather than
```
-		mdelay(1);
+		mdelay(10);
```
if it's possible to try something like
```
-		mdelay(1);
+		do {
+			ret = mmc_device_state();
+			if (ret < 0) {
+				return ret;
+			}
+		} while (ret != MMC_STATE_TRAN);
```
instead.

jbech-linaro:
@vchong , maybe and I can give it a try. Any preference from the TF-A maintainers?

jbech-linaro:
Just tried Victors patch:
4GB HiKey:
```
NOTICE:  BL2: v2.0(release):v2.0-rc0-442-g8310a2f5f
NOTICE:  BL2: Built : 08:56:39, Dec 28 2018
NOTICE:  acpu_dvfs_set_freq: set acpu freq success!ERROR:   BL2: Failed to load image (-5)
```

8GB HiKey:
```
NOTICE:  BL2: v2.0(release):v2.0-rc0-406-g8e00d04d
NOTICE:  BL2: Built : 08:42:30, Dec 28 2018
NOTICE:  acpu_dvfs_set_freq: set acpu freq success!ERROR:   BL2: Failed to load image (-5)
```

So it seems like that proposal didn't work. Would it make sense to merge the `mdelay(10)` patch that actually solves a problem and then come up with a new and optimized patch later?

hzhuang1:
Acked.
@Yann-lms Please help to verify it on your platform. Thanks.

dtischler:
@jbech-linaro 

Sorry that I didn't clearly state my error, only that I had the same behavior.

To be more precise, I get the same error message as you:

> NOTICE:  BL2: v2.0(release):v2.0
> NOTICE:  BL2: Built : 17:41:23, Dec 17 2018
> NOTICE:  acpu_dvfs_set_freq: set acpu freq success!ERROR:   CMD1 failed after 100 retries
> ERROR:   BL2: Failed to load image (-5)

And, this occurs immediately after applying power to the board, of course.

I use the 96Boards Serial Adapter to observe this behavior.

Yann-lms:
Sorry for late reply. The patch is OK for me and tested on STM32MP1.

antonio-nino-diaz-arm:
Any update regarding this patch?

jbech-linaro:
> Any update regarding this patch?

I'm not sure if you're asking me, but I haven't done any additional changes. If there is anything left to be done for me, please let me know.

antonio-nino-diaz-arm:
I was asking because of the last comment of @dtischler.

jbech-linaro:
I see, well I guess he just chimed in saying that he also have seen the same problem as I was facing (which was the reason for this patch).

antonio-nino-diaz-arm:
Ok. If the maintainers are happy with the change I guess there is no point in holding it back anymore.

jbech-linaro:
Thanks @antonio-nino-diaz-arm !

dtischler:
Yes, correct, I was simply chiming in to indicate that I was seeing the same behavior.  Once new images are built and available for download, I will retest.  I believe they end up here:  http://snapshots.linaro.org/96boards/reference-platform/components/uefi-staging/latest/hikey/debug/

PR #1726, opened 2018-12-18
Sanitise includes across codebase

antonio-nino-diaz-arm:
Enforce full include path for includes. Deprecate old paths.

The following folders inside include/lib have been left unchanged:

- include/lib/cpus/${ARCH}
- include/lib/el3_runtime/${ARCH}

The reason for this change is that having a global namespace for includes isn't a good idea. It defeats one of the advantages of having folders and it introduces problems that are sometimes subtle (because you may not know the header you are actually including if there are two of them).

For example, this patch had to be created because two headers were called the same way: e0ea092 ("Fix gpio includes of mt8173 platform to avoid collision."). More recently, this patch has had similar
problems: 46f9b2c ("drivers: add tzc380 support").

Platforms that want to preserve the way they include headers may add the removed paths to PLAT_INCLUDES, but this is discouraged.

(Thanks to @JoelHutton for reviewing the changes)


antonio-nino-diaz-arm:
The list of files is too big and the PR checker script hasn't got the full list (it looks like it's a GitHub API limitation), so this is the list of maintainers left to notify: @npoushin, @shawnguo2, @mtk09422, @qoriq-open-source, @Anson-Huang, @jenswi-linaro, @TonyXie06, @rockchip-linux, @b49020, @glneo, @masahir0y and @sivadur.

drivers/st/ddr/stm32mp1_ddr.c
```diff
@@ -4,22 +4,25 @@
...
-#include <mmio.h>
-#include <platform.h>
-#include <stddef.h>
-#include <stm32mp1_clk.h>
-#include <stm32mp1_ddr.h>
-#include <stm32mp1_ddr_regs.h>
+#include <lib/mmio.h>
+#include <plat/common/platform.h>
+
+#include <stm32mp1_def.h>
```
Yann-lms: I'd prefer you include platform_def.h here, as well as in the 2 or 3 other files that are modified the same way. And it avoids adding 1 line in plat/st/stm32mp1/platform.mk: 
`+PLAT_INCLUDES		+=	-Iplat/st/stm32mp1/`

But I can push a patch for this afterwards.

Otherwise, boot is OK on STM32MP1.

antonio-nino-diaz-arm: I think that you could push a patch for this and to move some headers to ``include/plat`` and then use the full path to include them, rather than adding the paths to ``PLAT_INCLUDES``. In general, I think that platforms should aim to have as few paths as possible in ``PLAT_INCLUDES``. If you can do all of the cleanup in one go, that would be great. I'd like this patch to be a cleanup of the common includes (drivers, lib, blx, etc). I mean, this patch doesn't even affect Arm platforms' headers. However, I didn't want to do that for all platforms in this PR because that's too much (I've only done the rpi3 patch because it was trivial).

I think that the only global platform headers should be the files ``platform_def.h`` and ``plat_macros.S`` (there is no way to get rid of them now anyway, it would break all platforms).

For example, I've noticed that some drivers include platform headers. That feels quite wrong to me. At most they should be including ``platform_def.h``, and that should have the needed definitions (or include a file that has them).

What do you think?

antonio-nino-diaz-arm: I'll explain what I mean a bit more clearly. Given that ``platform_def.h`` is the header file that the TF considers the master file for definitions of the platform, we should keep doing it. However, this file should include every platform file with definitions that are needed by drivers or other platform files. We are doing this wrong in Arm platforms.

``platform_def.h`` should include ``arm_def.h``, ``plat_arm.h`` and any other Arm platform header that we have. Then, no other file should include them: all other files should only include ``platform_def.h``. This would break the include loops that we already have, and it would let each board of each company to have a different set of includes in ``platform_def.h``. Maybe they want to include ``arm_watchdog_mmap.h`` in two platforms and not in the third one, for example. Including ``platform_def.h`` only would abstract all of this.

antonio-nino-diaz-arm: Also, this is useful for any driver that doesn't accept the base address as an argument. All they have to do is include ``platform_def.h``, which means that any platform would be able to reuse it. Right now, for example, ``drivers/st/ddr/stm32mp1_ddr.c`` includes ``stm32mp1_def.h``. This looks like a bad idea to me. In the case there is another platform that wants to use this driver, it is forced to have the same header. (Note that this is just an example because you have mentioned that file before!).

jenswi-linaro:
OK for plat/qemu

b49020:
Ok for plat/socionext/synquacer

drivers/st/ddr/stm32mp1_ddr.c
Yann-lms: I totally agree with all your remarks, that's why I didn't want to have stm32mp1_def.h in the drivers files. But I know some of our drivers still wrongly include platform header files. I have a patch internally to make some cleanup in this. But it will conflict with your patch. So the better is to first merge your patch, then I'll rework mine and push it afterwards.

antonio-nino-diaz-arm: I'd rather wait until this is merged, yes. I will have to rebase it right before the merge, and then platforms can start fixing their own includes in parallel.

In any case, most of us are going to be away for holidays for several days, and we imagine that this will be the case for other maintainers, so this PR will most likely be merged around 4th January to give more people the chance to see it.

Yann-lms: FYI, I've prepared a patch here: https://github.com/Yann-lms/arm-trusted-firmware/commit/eb4e0947e5eecb7007c869a1b3860bba254d15e2
It applies on top of your series of patches.
This is a first step to remove platform includes in ST drivers

antonio-nino-diaz-arm: The patch looks good to me.

antonio-nino-diaz-arm:
The only failures are some MISRA defects.

drivers/st/ddr/stm32mp1_ddr.c
antonio-nino-diaz-arm: You can create a PR now, @Yann-lms

PR #1731, opened 2018-12-25
docs: marvell: Fix typo in file build.txt

miyatsu:
The build option "SECURE" was never documented in the repo, and option "MARVELL_SECURE_BOOT" is never used as example. So I believe it must be a typo here.


antonio-nino-diaz-arm:
IP Review +1

PR #1740, opened 2019-01-07
plat/arm: Restrict PIE support to FVP

soby-mathew:
The patch SHA 55cf015c enabled PIE support when RESET_TO_BL31=1 for
all ARM platforms. But it seems n1sdp platform doesn't work with PIE
support yet. Hence restrict the ENABLE_PIE=1 to fvp platform.

Change-Id: If44e42528e4b0b57c69084503f346576fe0748bd


soby-mathew:
jenkins: test this please

PR #1757, opened 2019-01-15
Move BL1 and BL2 private defines to bl_common.h

antonio-nino-diaz-arm:
The definitions in bl1/bl1_private.h and bl2/bl2_private.h are useful for platforms that may need to access them.


jenswi-linaro:
OK for plat/qemu

PR #1760, opened 2019-01-16
rpi3: fix RPI3_PRELOADED_DTB_BASE usage

igoropaniuk:
In case if `RPI3_PRELOADED_DTB_BASE` isn't defined explicitly with
proper pre-loaded DTB address, `add_define` macro defined in
`make_helpers/build_macros.mk` still supplies this definition to the
compiler like `-DRPI3_PRELOADED_DTB_BASE`, and it's obviously is set to
default value 1.

This simply leads to the wrong `MAP_NS_DTB` region definition (base_va
is set `0x1` instead of `0x00010000`) in `plat/rpi3/rpi3_common.c`:

Which causes aligment check to fail in `mmap_add_region_check()`:
VERBOSE: base_pa: 0x00000001, base_va: 0x00000001, size: 0x00010000
...
ERROR:   mmap_add_region_check() failed. error -22

Signed-off-by: Igor Opaniuk <igor.opaniuk@linaro.org>


igoropaniuk:
Additional details about this issue:
https://github.com/OP-TEE/optee_os/pull/2735#issuecomment-454913371

jbech-linaro:
And our fix for it for the moment is to explicitly set it when doing our OP-TEE builds, please see: https://github.com/OP-TEE/build/pull/333

antonio-nino-diaz-arm:
Oh, right. Thanks for the fix.

antonio-nino-diaz-arm:
IP Review +1

jbech-linaro:
I see, too bad that we're (the OP-TEE team) no longer tracking upstream in our **non-stable** builds, since if we would have, then we also would have seen this much sooner. Anyhow, problem identified and there is a solution, thanks!

antonio-nino-diaz-arm:
jenkins: test this please

PR #1761, opened 2019-01-17
plat/arm: Save BL2 descriptors to reserved memory.

satheesbalya-arm:
On ARM platforms, the BL2 memory can be overlaid by BL31/BL32. The memory
descriptors describing the list of executable images are created in BL2
R/W memory, which could be possibly corrupted later on by BL31/BL32 due
to overlay. This patch creates a reserved location in SRAM for these
descriptors and are copied over by BL2 before handing over to next BL
image.

Fixes ARM-Software/tf-issues#626

Change-Id: I755735706fa702024b4032f51ed4895b3687377f


satheesbalya-arm:
Coverity has flagged two false positives about left shift of more than 7 bits.

In expression "1U << 12U", shifting more than 7 bits, the number of bits in the essential type of the left expression, "1U", is not allowed. The shift amount is 12.

antonio-nino-diaz-arm:
CC: @grqpnegre

PR #1703, opened 2018-11-29
hikey960: initialize EDMAC and channels

oscardagrach:
This is needed to utilize the DMA controller on the hikey960

Signed-off-by: Ryan Grachek <ryan@edited.us>

PR #1765, opened 2019-01-18
hikey960: add asp dma non-secure mode setting

xu-yiping:
init asp dma to non-secure mode.

Signed-off-by: Youlin Wang <wangyoulin1@hisilicon.com>
Signed-off-by: Tanglei Han <hantanglei@huawei.com>
Signed-off-by: Xu YiPing <xuyiping@hisilicon.com>

### DIFF
--- a/plat/hisilicon/hikey960/hikey960_bl31_setup.c
+++ b/plat/hisilicon/hikey960/hikey960_bl31_setup.c
@@ -159,6 +159,19 @@ static void hikey960_edma_init(void)
 	}
 }
 
+static void hikey960_asp_dma_init(void)
+{
+	int i;
+	uint32_t non_secure;
+
+	non_secure = ASP_DMAC_SEC_CTRL_INTR_SEC | ASP_DMAC_SEC_CTRL_GLOBAL_SEC;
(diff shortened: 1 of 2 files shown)


xu-yiping:
@hzhuang1 , please review, thanks

johnstultz-work:
@oscardagrach Can you review this to see how this is different from the patch you submitted earlier:
https://github.com/ARM-software/arm-trusted-firmware/pull/1703/files

Want to make sure nothing was missed.

oscardagrach:
@johnstultz-work This initializes the ASP DMA (DMA-a) controller as opposed to the peripheral DMA controller (DMA-p) my patch enabled. A similar patch will be required for the IOMCU DMA controller in the future. It is my anecdotal understanding the DMA-a controller does not require this initialization as it is accessible in non-secure mode by default.

johnstultz-work:
@oscardagrach Ah! Thanks for the clarification!  I too haven't observed the need for this DMA-A initialization yet, but doing it explicitly is probably better then assuming. It may have effects on other platforms that share the IP.

I'll follow up with @xu-yiping to see why this got closed.

PR #1773, opened 2019-01-21
Rpi3 gpio driver

grandpaul:
Hello. I'm adding this GPIO driver for reviewing. I plan to implement a SDHost driver in ATF so that we can load FIP from the SDCard directly. The SDHost driver needs to set the GPIO alt function and tweak the pull resistors. Thus this driver is needed. Many Thanks.

### DIFF
--- a/drivers/rpi3/gpio/rpi3_gpio.c
+++ b/drivers/rpi3/gpio/rpi3_gpio.c
@@ -1,0 +1,169 @@
+/*
+ * Copyright (c) 2019, Linaro Limited
+ * Copyright (c) 2019, Ying-Chun Liu (PaulLiu) <paul.liu@linaro.org>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include <string.h>
+#include <assert.h>
+#include <lib/mmio.h>
+#include <drivers/delay_timer.h>
+#include <drivers/rpi3/gpio/rpi3_gpio.h>
+
+static struct rpi3_gpio_params rpi3_gpio_params;
+
+static int rpi3_gpio_get_direction(int gpio);
+static void rpi3_gpio_set_direction(int gpio, int direction);
+static int rpi3_gpio_get_value(int gpio);
+static void rpi3_gpio_set_value(int gpio, int value);
+static void rpi3_gpio_set_pull(int gpio, int pull);
+
+static const gpio_ops_t rpi3_gpio_ops = {
+	.get_direction  = rpi3_gpio_get_direction,
+	.set_direction  = rpi3_gpio_set_direction,
+	.get_value      = rpi3_gpio_get_value,
+	.set_value      = rpi3_gpio_set_value,
+	.set_pull       = rpi3_gpio_set_pull,
+};
+
+/**
+ * Get selection of GPIO pinmux settings.
+ *
+ * @param gpio The pin number of GPIO. From 0 to 53.
+ * @return The selection of pinmux. RPI3_GPIO_FUNC_INPUT: input,
+ *                                  RPI3_GPIO_FUNC_OUTPUT: output,
+ *                                  RPI3_GPIO_FUNC_ALT0: alt-0,
+ *                                  RPI3_GPIO_FUNC_ALT1: alt-1,
+ *                                  RPI3_GPIO_FUNC_ALT2: alt-2,
+ *                                  RPI3_GPIO_FUNC_ALT3: alt-3,
+ *                                  RPI3_GPIO_FUNC_ALT4: alt-4,
+ *                                  RPI3_GPIO_FUNC_ALT5: alt-5
+ */
+int rpi3_gpio_get_select(int gpio)
+{
+	int ret;
+	uintptr_t reg_base = rpi3_gpio_params.reg_base;
+	int regN = gpio / 10;
+	int shift = 3 * (gpio % 10);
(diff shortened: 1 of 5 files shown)


antonio-nino-diaz-arm:
So what's your plan regarding the FIP? Unless you use https://github.com/christinaa/rpi-open-firmware and replace their Arm binary you still need to load BL1 from the filesystem.

grandpaul:
Hi. Yes. My plan is to load FIP from raw sectors right now. I currently separate two FIPS. One FIP with BL2 combines with BL1 as armstub8.bin which is smaller and loaded by rpi-open-firmware. And another FIP contains BL3* writes to raw sectors. Because there's no FAT drivers in ATF. If there's FAT drivers we can put the FIP into FAT partition too. Not sure why but the current armstub8.bin we have is just too large to be padded to rpi-open-firmware. VC4 doesn't load.

antonio-nino-diaz-arm:
The problem is that the open firmware is still very limited (as far as I remember, only the UART and the SD card are working). In practice this means that the installation is going to become more complicated for everyone: you need ``dd`` rather than to simply copy a file. Only the open firmware users are going to get something out of it. And I don't think there is anyone using the open firmware as main firmware, only the developers use it.

I'm not sure about this. Wouldn't it be better to fix the open firmware to load bigger files?

antonio-nino-diaz-arm:
Ok, that's fair enough.

drivers/rpi3/gpio/rpi3_gpio.h
```diff
@@ -0,0 +1,22 @@
...
+#include <stdint.h>
+
+typedef struct rpi3_gpio_params {
+	uintptr_t       reg_base;
+} rpi3_gpio_params_t;
+
+void rpi3_gpio_init(rpi3_gpio_params_t *params);
+int rpi3_gpio_get_select(int gpio);
+void rpi3_gpio_set_select(int gpio, int fsel);
+
```
antonio-nino-diaz-arm: You need to add defines here for the GPIO functions:
```
#define RPI3_GPIO_FUNC_INPUT	U(0)
#define RPI3_GPIO_FUNC_OUTPUT	U(1)
#define RPI3_GPIO_FUNC_ALT0	U(4)
#define RPI3_GPIO_FUNC_ALT1	U(5)
#define RPI3_GPIO_FUNC_ALT2	U(6)
#define RPI3_GPIO_FUNC_ALT3	U(7)
#define RPI3_GPIO_FUNC_ALT4	U(3)
#define RPI3_GPIO_FUNC_ALT5	U(2)
```

Then you can change the comments in your functions:
```
 * @param fsel The selection of pinmux. 0: input, 1: output, 2: alt-5,
 *                                      3: alt-4, 4: alt-0,  5: alt-1,
 *                                      6: alt-2, 7: alt-3
```
And you should replace all magic numbers by the defines.

drivers/rpi3/gpio/rpi3_gpio.c
```diff
@@ -0,0 +1,157 @@
...
+	mmio_write_32(reg_sel, sel);
+}
+
+static int rpi3_gpio_get_direction(int gpio)
+{
+	int result = rpi3_gpio_get_select(gpio);
+
+	if (result == 0)
+		return GPIO_DIR_IN;
+	else if (result == 1)
```
antonio-nino-diaz-arm: Replace 0 and 1 by the defines I've proposed, same in ``rpi3_gpio_set_direction()``.

drivers/rpi3/gpio/rpi3_gpio.h
```diff
@@ -0,0 +1,22 @@
+#ifndef __RPI3_GPIO_H__
```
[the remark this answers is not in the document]
grandpaul: Hi Antonio,
Sorry, I don't understand what I should use. It seems to me that I can only use ARM Limited and Contributors? Any other things I can use? 
Thanks.

grandpaul:
Hi. I've fixed almost all the issues based on the review. Thanks a lot. The only problem left is about the copyright header. What should I use for Copyright (c) <?> 
Thanks.

include/drivers/rpi3/gpio/rpi3_gpio.h
```diff
@@ -0,0 +1,38 @@
...
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef RPI3_GPIO_H
+#define RPI3_GPIO_H
+
+#include <stdint.h>
+#include <drivers/gpio.h>
+
+typedef struct rpi3_gpio_params {
```
antonio-nino-diaz-arm: Don't typedef new structs. Just use ``struct rpi3_gpio_params`` wherever needed:

https://github.com/ARM-software/arm-trusted-firmware/wiki/ARM-Trusted-Firmware-Coding-Guidelines#avoid-anonymous-typedefs-of-structsenums-in-header-files

antonio-nino-diaz-arm:
That looks fine to me. Can you squash the commits?

grandpaul:
Hi. I've squashed into 2 commits. Do you want me to squash as only one commits for this pull request? 
Thanks a lot.

antonio-nino-diaz-arm:
No, this is fine. I just wanted you to squash the commits to fix the struct thing.

antonio-nino-diaz-arm:
IP Review +1

antonio-nino-diaz-arm:
jenkins: test this please

PR #1777, opened 2019-01-22
ti: k3: common: Add support for runtime detection of GICR base address

glneo:
Valid addresses for GICR base are always a set calculable distance from
the GICD and is based on the number of cores a given instance of GICv3 IP
can support. The formula for the number of address bits is given by the
ARM GIC-500 TRM section 3.2 as 2^(18+log2(cores)) with the MSB set to
one for GICR instances. Holes in the GIC address space are also
guaranteed to safely return 0 on reads. This allows us to support runtime
detection of the GICR base address by starting from GIC base address plus
BIT(18) and walking until the GICR ID register (IIDR) is detected. We
stop searching after BIT(20) to prevent searching out into space if
something goes wrong. This can be extended out if we ever have a device
with 16 or more cores.

Signed-off-by: Andrew F. Davis <afd@ti.com>


antonio-nino-diaz-arm:
IP Review +1

antonio-nino-diaz-arm:
jenkins: test this please

glneo:
This has only been tested with GIC-500 instances. I have not studied other GIC models to know if similar techniques can be used, but if they do then this could be made more generic at some point.

PR #1778, opened 2019-01-23
multi_console: Check functions are not NULL

JoelHutton:
Change-Id: I2d67bb1bebd15e6a7d69ea5e7b6fda9c972f9d86
Signed-off-by: Joel Hutton <Joel.Hutton@Arm.com>


drivers/console/multi_console.c
```diff
@@ -77,7 +77,11 @@ int console_putc(int c)
 
 	for (console = console_list; console != NULL; console = console->next)
 		if (console->flags & console_state) {
```
jwerner-chromium: nit: wouldn't it look cleaner if you just changed this to

    if ((console->flags & console_state) && console->putc)

?

PR #1781, opened 2019-01-24
spd: trusty: trusty_setup should bail on unknown image

dtwlin:
When an unknown Trusty image is found, there's no point of still trying
to register the BL32 init handler. Instead, we just should bail out of
the trusty_setup() and allow the system to continue to boot.

Signed-off-by: David Lin <dtwlin@google.com>


antonio-nino-diaz-arm:
IP Review +1

antonio-nino-diaz-arm:
jenkins: test this please

PR #1784, opened 2019-01-25
plat/arm: Cleanup of includes and drivers

antonio-nino-diaz-arm:
This PR moves all the drivers that were scattered around plat/arm/ to drivers/. This avoids the problem of having several "drivers" folders in the tree and not knowing where to find a specific driver. It also encourages other platforms to re-use the drivers if possible, and to clean them up to allow re-utilization if not.

This PR also finishes what https://github.com/ARM-software/arm-trusted-firmware/pull/1726 and https://github.com/ARM-software/arm-trusted-firmware/pull/1756 started. Arm platforms now use full paths for includes and the number of relative/user includes has been reduced as much as possible (only 14 right now, and they include private headers).


antonio-nino-diaz-arm:
The failures are just 2 MISRA defects.

PR #1751, opened 2019-01-11
Tegra scatter file support

vwadekar:
This is actually a PR to get some thoughts on the approach. This has been verified for Tegra platforms only.

PR #1785, opened 2019-01-25
Tf2.0 tegra downstream rebase 1.25.19

vwadekar:


vwadekar:
False positives flagged by the checkpatch script

`plat/nvidia/tegra/soc/t210/plat_psci_handlers.c:493: ERROR: code indent should use tabs where possible
plat/nvidia/tegra/soc/t210/plat_psci_handlers.c:494: ERROR: code indent should use tabs where possible
`

plat/nvidia/tegra/soc/t210/plat_setup.c
```diff
@@ -1,15 +1,22 @@
...
-#include <bpmp.h>
 #include <cortex_a57.h>
 #include <common/bl_common.h>
+#include <common/debug.h>
+#include <common/interrupt_props.h>
 #include <drivers/console.h>
 #include <lib/xlat_tables/xlat_tables_v2.h>
+#include <gic_common.h>
+#include <gicv2.h>
+#include <interrupt_mgmt.h>
```
antonio-nino-diaz-arm:
```
#include <drivers/arm/gic_common.h>
#include <drivers/arm/gicv2.h>
#include <bl31/interrupt_mgmt.h>
```

services/spd/trusty/trusty.c
```diff
@@ -354,7 +354,7 @@ static void trusty_cpu_resume(uint32_t on)
 
 static int32_t trusty_cpu_off_handler(u_register_t unused)
```
antonio-nino-diaz-arm: Maybe you should rename this argument, now that it isn't unused. Same for the other functions in this patch.

plat/nvidia/tegra/platform.mk
```diff
@@ -1,9 +1,22 @@
 #
-# Copyright (c) 2015-2017, ARM Limited and Contributors. All rights reserved.
+# Copyright (c) 2015-2018, ARM Limited and Contributors. All rights reserved.
 #
 # SPDX-License-Identifier: BSD-3-Clause
 #
 
+# override with stdlib files used by the Tegra platform
+STDLIB_SRCS	:=	$(addprefix lib/stdlib/,	\
```
antonio-nino-diaz-arm: This can't be working, the folder is now called libc. ``tf_snprintf.c`` doesn't exist anymore.

plat/nvidia/tegra/common/tegra_pm.c
```diff
@@ -146,18 +146,36 @@ void tegra_get_sys_suspend_power_state(psci_power_state_t *req_state)
...
+	 * irrespective of the value of the PSTATE.I bit value.
+	 */
+	write_scr_el3(saved_scr_el3 | SCR_IRQ_BIT);
+
 	/*
 	 * Enter standby state
-	 * dsb is good practice before using wfi to enter low power states
+	 * isb & dsb is good practice before using wfi to enter low power states
 	 */
+	isb();
```
antonio-nino-diaz-arm: Normally you would add ``isb()`` after ``dsb()`` so that the effect of the ``dsb()`` is seen.

antonio-nino-diaz-arm:
IP Review +1

vwadekar:
@antonio-nino-diaz-arm 

Thanks for the review comments. Pushed updated PR

plat/nvidia/tegra/platform.mk
```diff
@@ -1,9 +1,19 @@
 #
-# Copyright (c) 2015-2017, ARM Limited and Contributors. All rights reserved.
+# Copyright (c) 2015-2018, ARM Limited and Contributors. All rights reserved.
 #
 # SPDX-License-Identifier: BSD-3-Clause
 #
```
antonio-nino-diaz-arm: I still don't know why this patch is needed. All libc files are compiled and packaged in a library archive. As far as I understand, linkers only pull in the object files that are actually used. This isn't even about removing individual functions in one object file, this is about removing the object file itself.

Also, this patch relies on the current order of inclusions in the Makefile. If the standard ``LIBC_SRCS`` is expanded before this file is included, it will have no effect.

antonio-nino-diaz-arm: I feel that this may result in a conversation similar to the one in https://github.com/ARM-software/arm-trusted-firmware/pull/1751. If you still think it is needed, maybe it's worth moving it to that PR. I wouldn't want to block this PR just because of a discussion in one patch that doesn't affect the others.

plat/nvidia/tegra/soc/t210/plat_setup.c
```diff
@@ -137,10 +146,76 @@ void plat_early_platform_setup(void)
 	}
 }
 
+/* Secure IRQs for Tegra186 */
+static const interrupt_prop_t tegra210_interrupt_props[] = {
+	INTR_PROP_DESC(TEGRA210_WDT_CPU_LEGACY_FIQ, GIC_HIGHEST_SEC_PRIORITY,
+			GICV2_INTR_GROUP0, GIC_INTR_CFG_EDGE),
+};
+
+void plat_late_platform_setup(void)
```
soby-mathew: I couldn't find where this function is invoked. Suggest the API name to be made specific to Tegra `plat_tegra_late_platform_setup()`.

plat/nvidia/tegra/platform.mk
vwadekar: @antonio-nino-diaz-arm 

This is required to get to zero MISRA defects for Tegra. Since we do not use these files, it reduces the amount of MISRA defects we have to deal with. Plus, this patch also reduces the code size.

> Also, this patch relies on the current order of inclusions in the Makefile. If the standard LIBC_SRCS is expanded before this file is included, it will have no effect.

That's true. We can either document that dependency or change this makefile to remove the files from BL31_SOURCES

plat/nvidia/tegra/soc/t210/plat_setup.c
vwadekar: @soby-mathew 

It will be used in subsequent patches.

plat/nvidia/tegra/common/tegra_bl31_setup.c
```diff
@@ -404,6 +406,14 @@ void bl31_plat_arch_setup(void)
 	 */
 	boot_profiler_add_record("[TF] arch setup entry");
 
+	/* add MMIO space */
+	plat_mmio_map = plat_get_mmio_map();
```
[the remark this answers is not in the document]
vwadekar: @soby-mathew 

This patch only moves the code to be called a bit earlier. I would like to update the function names as part of another PR

plat/nvidia/tegra/platform.mk
soby-mathew: Hi Varun,
Not sure if I understand the code size argument. The libc is built as a `library archive` and only referenced symbols will be included in the binary. Also the `--gc-sections` will take care of removing unused sections.

The MISRA defect argument is justified but you seem to have included most of files except some string utilities and I don't think there would be many MISRA defects in the other files excluded. Fixing the few remaining MISRA defects in the other files shouldn't be too difficult.

Allowing platforms to override makefile variables like these and adding dependency on the order of inclusion for building library archives will not allow us to make improvements/changes to the build system without affecting platform ports. Hence currently I don't see a convincing reason why this should be allowed.

vwadekar:
I believe platforms should be the ones deciding which files to include in
the build.

MISRA is one aspect, but you also have to come up with unit tests,
integration tests, design docs etc when you add even a single line of code.

That is another reason why we are trimming the files we use. We can come up
with another way of doing that, but the fact remains that we won't be
including unnecessary files in the build.

antonio-nino-diaz-arm:
>I believe platforms should be the ones deciding which files to include in the build.

There is a common core of functionality that all platforms need. libc is one of them. As it is a library and each function is compiled as a different .o file, the linker will only use the .o files that are actually needed. You don't even need special flags or anything. This means that platforms that don't need some functions of libc won't have the functions in the final binary.

> MISRA is one aspect, but you also have to come up with unit tests, integration tests, design docs etc when you add even a single line of code.

MISRA is not a valid reason for me, you can ignore MISRA defects if you can justify why. In this case "we don't use this function" would be a valid reason. Also, I don't see why you need tests for functionality that is not part of the binary. If you are not using it, you don't need tests.

The way I see it, this is creating more dependencies between makefiles, and dependencies that are not easy to spot. And, if we changed the behaviour during a patch, we wouldn't even notice because if our list of files overrides yours, it would still compile.

Also, right now platforms don't need to care about libc at all. If we started to allow this, it would mean that changes that we do in libc could affect platforms, and we have no way to fix non-upstream platforms.

> That is another reason why we are trimming the files we use. We can come up with another way of doing that, but the fact remains that we won't be including unnecessary files in the build.

The fact is that you are already compiling code that you don't need.

For example, xlat v2 has support for EL1 and EL2 tables, but your port is BL31-only, so you only need EL3 and you are adding this extra code that you don't need. Should we also start adding build flags to remove all the code you don't need? I don't think this is maintainable.

Letting platforms cherry-pick every file would also make it really hard for us to change things without potentially breaking them when we change something.

vwadekar:
> There is a common core of functionality that all platforms need. libc is one of them

I beg to differ. Tying platforms to internal libraries slows them down. e.g. we would like to completely remove usage of some libraries and use safety certified ones instead. How do we do that, if there is no way to turn off code?

> MISRA is not a valid reason for me, you can ignore MISRA defects if you can justify why.

NVIDIA has to go through the actual certification and have to jump through a lot of hoops to file a deviation. It actually takes a lot of man hours to get the deviation approved.

> The way I see it, this is creating more dependencies between makefiles, and dependencies that are not easy to spot. And, if we changed the behaviour during a patch, we wouldn't even notice because if our list of files overrides yours, it would still compile.

I agree. And we should use this opportunity to solve the problem. I can introduce new macros to allow disabling libc completely.

> The fact is that you are already compiling code that you don't need

I understand and we are cleaning it up as we speak.

> Should we also start adding build flags to remove all the code you don't need? I don't think this is maintainable

I dont want to fork out and maintain the entire code on github as we would like to be in sync with upstream, hence these discussions to find proper solutions that work for everyone.

soby-mathew:
Hi Varun,

> we would like to completely remove usage of some libraries and use safety certified ones instead. How do we do that, if there is no way to turn off code?

There are other ways we could provide that flexibility. For example, provide an `override lib` build option which prevents the top level Makefile from pulling in libc.mk by default and allow platform.mk to define the libraries to build.

Regarding certification, if you can prove that unused code is removed at link time, and is not part of the final binary, wouldn't that be a reasonable justification ?

vwadekar:
Hi Soby,

Can you point me to the docs to use 'override lib' option?

> if you can prove that unused code is removed at link time, and is not
part of the final binary, wouldn't that be a reasonable justification ?

Yes, it is reasonable AFAICT. We wont know for sure until we get through
certification

Are you suggesting that we compile the "native" libc code and then remove
it at link time? We dont see value in compiling such code and then removing
at link time. One example would be being MISRA clean, which wont be
possible if we are compiling unwanted code.

soby-mathew:
I was suggesting something simple. Define a build option to allow platform to override the standard libc, something like :

``` diff
index 3de7b3f..4f7d937 100644
--- a/Makefile
+++ b/Makefile
@@ -247,7 +247,9 @@ DTC_FLAGS           +=      -I dts -O dtb
 # Common sources and include directories
 include lib/compiler-rt/compiler-rt.mk
+ifeq ($(OVERRIDE_LIBC), 1)
 include lib/libc/libc.mk
+endif
 
 BL_COMMON_SOURCES      +=      common/bl_common.c                      \
```

Then in your platform.mk you can build the library you need. So you wouldn't compile the libc in the source tree.

This approach is only valid when you want to replace an entire component with a component of your choice. But if you want to avoid compiling parts of C Source file by adding `#if/#ifdefs` because it is not part of your platform build, that is something we do not agree.

antonio-nino-diaz-arm:
> I see. Then you will have to give us a commitment that you will clean up all MISRA issues in the code. And not just the MISRA defects you find fit, but ones that are also important to us. If we have to make this work, then everyone needs to be treated equally

The following wiki page has the information about the MISRA work we've been doing and a list of rules we intend to comply with and not: https://github.com/ARM-software/arm-trusted-firmware/wiki/ARM-Trusted-Firmware-Coverity-MISRA-Analysis

If you need a different set of rules, could you tell us what we need to do differently? What other rules would we need to enforce? All of them?

Removing all MISRA issues in the codebase is unrealistic. The following examples come to my mind:

- We use weak functions for platform interfaces. MISRA prohibits weak functions, and there are many platforms ports that would just break if we decided to remove the weak definitions and force them to implement them. This would also break all non-upstream platforms whenever we add a new interface, even if it is needed only by 1 platform that isn't even upstream.
- All the messages that we print go through macros that internally call printf. There are 2 MISRA issues for each message, then. Even ``assert()`` and ``CASSERT()`` raise MISRA defects, and we aren't going to remove checks because of MISRA, that would just be counter-productive. If we count all of them there are easily 2000 defects. We have simply disabled the rules that flag these defects because they just create noise.

vwadekar:
@soby-mathew 

Please let me know if this looks good. I have updated the PR as per our discussion so far to override libc

vwadekar:
@soby-mathew @antonio-nino-diaz-arm 

How would you like to proceed with this PR?

Makefile
```diff
@@ -392,6 +391,13 @@ endif
 
 endif
 
+################################################################################
+# Remove overridden libraries
```
soby-mathew: Suggest: "Include libc if not overridden"

docs/user-guide.rst
```diff
@@ -581,6 +581,10 @@ Common build options
    1 (do save and restore). 0 is the default. An SPD may set this to 1 if it
    wants the timer registers to be saved and restored.
 
+- ``OVERRIDE_LIBC``: This option allows platforms to remove lib/libc from the
+   final image. It can take either 0 (include) or 1 (remove). The default
```
soby-mathew: Suggest: "This option allows platforms to override the default libc for the BL image. It can take either 0 (include) or 1 (roverride)"

Makefile
```diff
@@ -392,6 +391,13 @@ endif
...
+################################################################################
+ifeq (${OVERRIDE_LIBC},0)
+include lib/libc/libc.mk
```
soby-mathew: I just realized that there are some more changes that need to be done for isolating building libraries from the assumptions in the main Makefile. We will do this separately later and apply necessary fixes to your platform as well.

vwadekar:
@soby-mathew 

Thanks. Added the suggested changes in the latest PR update

antonio-nino-diaz-arm:
Can you fix the commit message of "Introduce build option to override libc"? It has two sets of change-ids and sign-offs.

vwadekar:
> Can you fix the commit message of "Introduce build option to override libc"? It has two sets of change-ids and sign-offs.

Done

antonio-nino-diaz-arm:
jenkins: test this please

antonio-nino-diaz-arm:
The failure is the stratix10 platform build config, and that's because your PR isn't based on master, so it's ok.

PR #1787, opened 2019-01-27
Do not mix var declarations and code.

laroche:
Signed-off-by: Florian La Roche <Florian.LaRoche@gmail.com>


laroche:
Not sure if these things should be fixed or what your rules for the source code are. If you are with these
changes, there are a couple more files that need the same changes and I could submit them here
as well. Otherwise just close this request...

soby-mathew:
Hi @laroche,
Since `declaration-after-statement` is allowed in C99 (GNU99), this is not mandated in coding guidelines. It is currently left to individual developer preference.

Hence I would prefer to not make this change.

laroche:
Ok, sounds good to me. I am closing this request.

best regards,

Florian La Roche

PR #1790, opened 2019-01-28
Add a minimal reset framework

Yann-lms:
With this framework, the drivers could be more generic instead of directly calling platform reset drivers.

### DIFF
--- a/drivers/reset/reset.c
+++ b/drivers/reset/reset.c
@@ -1,0 +1,40 @@
+/*
+ * Copyright (c) 2019, STMicroelectronics - All Rights Reserved
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include <assert.h>
+
+#include <drivers/reset.h>
+
+static const struct reset_ops *reset_ops;
+
+void reset_assert(unsigned int id)
+{
+	assert((reset_ops != NULL) && (reset_ops->reset_assert != NULL));
+
+	reset_ops->reset_assert(id);
+}
+
+void reset_deassert(unsigned int id)
+{
+	assert((reset_ops != NULL) && (reset_ops->reset_deassert != NULL));
+
+	reset_ops->reset_deassert(id);
+}
+
+bool reset_asserted_status(unsigned int id)
+{
+	assert((reset_ops != NULL) &&
+	       (reset_ops->reset_asserted_status != NULL));
+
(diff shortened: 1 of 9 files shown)


drivers/reset/reset.c
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright (C) 2019, ARM Limited and Contributors. All rights reserved.
```
antonio-nino-diaz-arm: Don't use this copyright notice for new files. Use your name/company instead.

include/drivers/reset.h
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright (C) 2019, ARM Limited and Contributors. All rights reserved.
```
antonio-nino-diaz-arm: Don't use this copyright notice for new files. Use your name/company instead.

Yann-lms:
@antonio-nino-diaz-arm I've pushed a new version with the required changes

soby-mathew:
@Yann-lms , Could you explain more about what you mean by `platform reset` because I see the CPU still continues to proceed after the reset.   Is this a reset for the `mmc` device or is it for the whole platform ? 

If the reset is device specific, then I would expect the reset framework to allow registration for multiple devices.

Yann-lms:
@soby-mathew I agree this is not well described. It is clearly not a platform reset, but a reset for a device (MMC, UART...). On our platform, this is handled by one IP: RCC (Reset and Clock control). This framework should be seen as an abstraction of what was done previously in the driver stm32mp1_reset.c.
The reset line for a given device is defined with its ID. So I wanted to keep that simple, and fit with our previous code. But it then may not be easily re-usable for other platforms.
I will update the commit message, and check what could be done for multiple devices registration.

Yann-lms:
After discussion internally, it seems a too big change to allow registration for multiple devices. Our reset driver is simple and works with IDs, found in a device tree file. So registering a reset framework in each device driver will consume code size and execution time, not worth the change.
So I'll may just close this PR, and maybe make our reset driver a bit more generic for our platform.

Yann-lms:
After discussion internally, it seems a too big change to allow registration for multiple devices. Our reset driver is simple and works with IDs, found in a device tree file. So registering a reset framework in each device driver will consume code size and execution time, not worth the change.
So I'll may just close this PR, and maybe make our reset driver a bit more generic for our platform.

PR #1794, opened 2019-01-29
tools/fiptool: Fix UUID parsing in blob handling

apritzel:
Commit 033648652f ("Make TF UUID RFC 4122 compliant") changed the scanf
parsing string to handle endianness correctly.
However that changed the number of items sscanf handles, without
adjusting the sanity check just below.

Increase the expected return value from 11 to 16 to let fiptool handle
UUIDs given as blob parameters correctly again.


antonio-nino-diaz-arm:
jenkins: test this please

PR #1799, opened 2019-01-31
GICv3: Allow probe for fewer GICR interfaces than exposed by the frame

soby-mathew:
Previously the GICv3 redistributor probe function
(gicv3_rdistif_base_addrs_probe()) asserted that the number of
per-CPU redistributor interfaces expected to be probed by the
platform is equal to the number exported by the redistributor
frame. This is a problem in case the number of CPUs in the
platform is less than the number of redistributor interfaces
in the frame. Hence this patch removes the assertion check
and allows probe for fewer redistributor interfaces as required
by the platform.

Change-Id: I3449763a3ad70817224442cbe184d001030c9874


antonio-nino-diaz-arm:
The CI failures are due to the patch being based on an old branch with different SPM build option names than the current master (so the CI doesn't know about them).

PR #1805, opened 2019-02-05
drivers: generic_delay_timer: Assert presence of Generic Timer

antonio-nino-diaz-arm:
The Generic Timer is an optional extension to an ARMv7-A implementation. The generic delay timer can be used from any architecture supported by the Trusted Firmware. In ARMv7 it is needed to check that this feature is present. In ARMv8 it is always present.


Yann-lms:
Tested OK on STM32MP1 (ARMv7-A Cortex-A7)

PR #1815, opened 2019-02-12
gic: make sure ProcessorSleep bit clear successfully

Anson-Huang:
GICR_WAKER.ProcessorSleep can only be set to zero when:
— GICR_WAKER.Sleep bit[0] == 0.
— GICR_WAKER.Quiescent bit[31] == 0.

On some platforms, when system reboot with GIC in sleep
mode but with power ON, such as on NXP's i.MX8QM, Linux
kernel enters suspend but could be requested to reboot,
and GIC is in sleep mode and it is inside a power domain
which is ON in this scenario, when CPU reset, the GIC
driver trys to set CORE's redistributor interface to awake,
with GICR_WAKER.Sleep bit[0] and GICR_WAKER.Quiescent bit[31]
both set, the ProcessorSleep bit[1] will never be clear
and cause system hang.

This patch makes sure GICR_WAKER.Sleep bit[0] and
GICR_WAKER.Quiescent bit[31] are both zeor before clearing
ProcessorSleep bit[1].

Signed-off-by: Anson Huang <Anson.Huang@nxp.com>

### DIFF
--- a/plat/imx/common/plat_imx8_gic.c
+++ b/plat/imx/common/plat_imx8_gic.c
@@ -9,6 +9,8 @@
 #include <common/bl_common.h>
 #include <common/interrupt_props.h>
 #include <drivers/arm/gicv3.h>
+#include <drivers/arm/arm_gicv3_common.h>
+#include <lib/mmio.h>
 #include <lib/utils.h>
 #include <plat/common/platform.h>
 
@@ -52,8 +54,27 @@ void plat_gic_driver_init(void)
 #endif
 }
 
+static __inline void plat_gicr_exit_sleep(void)
+{
+	unsigned int val = mmio_read_32(PLAT_GICR_BASE + GICR_WAKER);
+
+	/*
+	 * ProcessorSleep bit can ONLY be set to zero when
+	 * Quiescent bit and Sleep bit are both zero, so
+	 * need to make sure Quiescent bit and Sleep bit
+	 * are zero before clearing ProcessorSleep bit.
+	 */
+	if (val & WAKER_QSC_BIT) {
+		mmio_write_32(PLAT_GICR_BASE + GICR_WAKER, val & ~WAKER_SL_BIT);
+		/* Wait till the WAKER_QSC_BIT changes to 0 */
+		while ((mmio_read_32(PLAT_GICR_BASE + GICR_WAKER) & WAKER_QSC_BIT) != 0U)
+			;
+	}
+}
+
 void plat_gic_init(void)
 {
+	plat_gicr_exit_sleep();
 	gicv3_distif_init();
 	gicv3_rdistif_init(plat_my_core_pos());
 	gicv3_cpuif_enable(plat_my_core_pos());


drivers/arm/gic/v3/gicv3_helpers.c
```diff
@@ -14,6 +14,7 @@
 
 #include "../common/gic_common_private.h"
 #include "gicv3_private.h"
+#include "arm_gicv3_common.h"
```
antonio-nino-diaz-arm: This header contains implementation-defined values. Including it here would prevent other platforms from using this driver.

Anson-Huang: Then how to handle this case? Add a function in our imx platform and if the function is avaiable, call it? Then other platforms will NOT be impacted?

antonio-nino-diaz-arm: I'm thinking that maybe a build option is more in-line with what we do in the codebase (check ``PL011_GENERIC_UART``). If the definitions of that header are present in most arm designs, maybe this build option makes sense. I'm not a fan of either option to be honest.

Anson-Huang:
Looks like NO better option, so I add a build option to enable sleep status check for GIC operation.

drivers/arm/gic/v3/gicv3_helpers.c
Anson-Huang: Looks like no better option, I added a build option for it.

soby-mathew:
@Anson-Huang, This patch is more complicated than what it seems.

First could you confirm on GICv3 implementation on your platform ?

Assuming that it is similar to GIC-500, there is " only 1 GICR_WAKER.Sleep and 
GICR_WAKER.Quiescent which can be read and written through the GICR_WAKER register of any redistributor".

This is to be only done when powering down the GIC and is meant to be done as part of Suspend-to-RAM or Suspend-to-Disk (corresponding to SYSTEM_SUSPEND PSCI API) by the last CPU. This shouldn't be done as part of regular idle suspend requests from kernel.

If you look at [arm_gicv3_distif_pre_save()](https://github.com/ARM-software/arm-trusted-firmware/blob/master/drivers/arm/gic/v3/arm_gicv3_common.c#L28), this sets the GICR_WAKER.Sleep and GICR_WAKER.Quiescent only for SYSTEM_SUSPEND cases which the platform determines. Similarly
[arm_gicv3_distif_post_restore()](https://github.com/ARM-software/arm-trusted-firmware/blob/master/drivers/arm/gic/v3/arm_gicv3_common.c#L72) is invoked when waking up from a SYSTEM_SUSPEND and it clears the respective bits if they are not already cleared.

Could you let me know how the  GICR_WAKER.Sleep and GICR_WAKER.Quiescent are set to 1 in your platform ? 

It seems that in your case, during a suspend, the kernel issued a reboot instruction. Is this idle or SYSTEM_SUSPEND ?

Also when the reboot is issued the SoC is expected to undergo a power cycle and all the peripherals including GIC are meant to be in reset state. A cold boot sequence is expected to be run after this on the primary CPU. But you mentioned that GIC is still in ON power domain, so I assume that in your platform, reboot doesn't cause GIC to be reset. In this case, I would expect the platform to do something similar to arm_gicv3_distif_post_restore() in its platform initialization sequence. 

But even with that there might be issues with LPI pending tables for your case, but perhaps it is not a problem if you don't use LPIs in your platform. 

BRs
Soby Mathew

Anson-Huang:
@soby-mathew 
Could you let me know how the GICR_WAKER.Sleep and GICR_WAKER.Quiescent are set to 1 in your platform ?
[Anson] The GICR_WAKER.Sleep is set to 1 when linux doing a suspend, at this point, our platform has a system controller which controls overall system's power/clock etc., and there are also other masters running, like other M4 CORE running RTOS, GIC is NOT powered down since it is side a power domain shared with other modules which are needed by other M4 masters. In this case, when system controller or M4 with RTOS issue a reboot command to reboot Linux, since GIC is NOT powered off/reset, then ATF boot up and try to initialize GIC and this issue happen.

To make it easy for understand, the case is, Linux go into suspend with GIC in sleep mode, other masters issue reboot to reboot Linux without GIC reset since it has NOT individual power gate or reset, so issue happen.

It seems that in your case, during a suspend, the kernel issued a reboot instruction. Is this idle or SYSTEM_SUSPEND ?
[Anson], reboot instruction is NOT issued by kernel since it is already suspended, it is issued by system
controller or other masters like M4 running a RTOS.

Also when the reboot is issued the SoC is expected to undergo a power cycle and all the peripherals including GIC are meant to be in reset state.
[Anson] Unfortunately, our platform has no individually power gate or reset for GIC and in this case GIC is in sleep and NOT reset when ATF got reboot.

 A cold boot sequence is expected to be run after this on the primary CPU. But you mentioned that GIC is still in ON power domain, so I assume that in your platform, reboot doesn't cause GIC to be reset. In this case, I would expect the platform to do something similar to arm_gicv3_distif_post_restore() in its platform initialization sequence.
[Anson] OK, I can do it in our platform's early setup stage, but that will need to do for all our platforms, I thought it is a common flow that common GIC driver should handle if the GIC impelmentation is same as our platforms.

But even with that there might be issues with LPI pending tables for your case, but perhaps it is not a problem if you don't use LPIs in your platform.
[Anson] I think we do NOT use LPIs currently, so if you thinkg putting it inside our platform's early setup phase is the best solution, OK, I will do it.

soby-mathew:
@Anson-Huang , Thanks for your clarification, the issue is clear to me now.

antonio-nino-diaz-arm:
jenkins: test this please

PR #1826, opened 2019-02-17
allwinner: A few minor improvements

smaeul:
In the process of [developing SCPI support](https://github.com/smaeul/arm-trusted-firmware/commits/crust) for the Allwinner platforms, I came up with a few small improvements that are independent of the PSCI changes.

* ~~Make indentation/alignment consistent (though the only really "wrong" indentation was the `NOTICE` in sunxi_bl31_setup.c).~~
* Constify data structures where possible, so they get put in `.rodata`
* Make the CPU ops functions simpler to call, saving about 50 bytes. This will also save some line wrapping when the SCPI case is added.

### DIFF
--- a/plat/allwinner/common/sunxi_bl31_setup.c
+++ b/plat/allwinner/common/sunxi_bl31_setup.c
@@ -11,6 +11,7 @@
 #include <platform_def.h>
 
 #include <arch.h>
+#include <arch_helpers.h>
 #include <common/debug.h>
 #include <drivers/arm/gicv2.h>
 #include <drivers/console.h>
@@ -101,7 +102,7 @@ void bl31_early_platform_setup2(u_register_t arg0, u_register_t arg1,
 	SET_SECURITY_STATE(bl33_image_ep_info.h.attr, NON_SECURE);
 
 	/* Turn off all secondary CPUs */
-	sunxi_disable_secondary_cpus(plat_my_core_pos());
+	sunxi_disable_secondary_cpus(read_mpidr());
 }
 
 void bl31_plat_arch_setup(void)
--- a/plat/allwinner/common/sunxi_cpu_ops.c
+++ b/plat/allwinner/common/sunxi_cpu_ops.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017-2018, ARM Limited and Contributors. All rights reserved.
+ * Copyright (c) 2017-2019, ARM Limited and Contributors. All rights reserved.
  *
  * SPDX-License-Identifier: BSD-3-Clause
  */
@@ -45,9 +45,10 @@ static void sunxi_cpu_enable_power(unsigned int cluster, unsigned int core)
 	mmio_write_32(SUNXI_CPU_POWER_CLAMP_REG(cluster, core), 0x00);
 }
 
-void sunxi_cpu_off(unsigned int cluster, unsigned int core)
+void sunxi_cpu_off(u_register_t mpidr)
 {
-	int corenr = cluster * PLATFORM_MAX_CPUS_PER_CLUSTER + core;
+	unsigned int cluster = MPIDR_AFFLVL1_VAL(mpidr);
+	unsigned int core    = MPIDR_AFFLVL0_VAL(mpidr);
 
 	VERBOSE("PSCI: Powering off cluster %d core %d\n", cluster, core);
 
@@ -55,9 +56,9 @@ void sunxi_cpu_off(unsigned int cluster, unsigned int core)
 	mmio_clrbits_32(SUNXI_CPUCFG_DBG_REG0, BIT(core));
 
 	/* We can't turn ourself off like this, but it works for other cores. */
-	if (plat_my_core_pos() != corenr) {
+	if (read_mpidr() != mpidr) {
 		/* Activate the core output clamps, but not for core 0. */
-		if (corenr != 0)
+		if (core != 0)
 			mmio_setbits_32(SUNXI_POWEROFF_GATING_REG(cluster),
 					BIT(core));
 		/* Assert CPU power-on reset */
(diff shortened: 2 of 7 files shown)


plat/allwinner/common/sunxi_bl31_setup.c
```diff
@@ -76,11 +77,11 @@ static void *sunxi_find_dtb(void)
 }
 
 void bl31_early_platform_setup2(u_register_t arg0, u_register_t arg1,
-				u_register_t arg2, u_register_t arg3)
+                                u_register_t arg2, u_register_t arg3)
```
apritzel: Sorry, but I don't like those white space changes. AFAIK using tabs and fill up the remaining "modulo" part with spaces is completely standard.  The issue with indentation in diffs is well known and people just learned to live with it. And I don't get the argument of using tabs != 8 spaces? Why would you do that?
We could have endless discussions about coding style, but the most important argument is for all to use the same one, as this simplifies review a lot.
And white space changes just for the sake of it are somewhat pointless anyway.

plat/allwinner/common/sunxi_cpu_ops.c
```diff
@@ -45,25 +45,27 @@ static void sunxi_cpu_enable_power(unsigned int cluster, unsigned int core)
...
 			mmio_setbits_32(SUNXI_POWEROFF_GATING_REG(cluster),
-					BIT(core));
+			                BIT(core));
 		/* Assert CPU power-on reset */
 		mmio_clrbits_32(SUNXI_POWERON_RST_REG(cluster), BIT(core));
-		/* Remove power from the CPU */
-		sunxi_cpu_disable_power(cluster, core);
+		/* Remove power from the CPU, but not for core 0. */
+		if (core != 0)
+			sunxi_cpu_disable_power(cluster, core);
```
apritzel: Interesting, how did you trigger this issue? A PSCI SYSTEM_OFF from a core other than core 0?
And shouldn't this hit in the OpenRISC code as well? I am pretty sure I offlined core 0 multiple times during my randomized offlining test.

plat/allwinner/common/sunxi_cpu_ops.c
```diff
@@ -77,11 +79,14 @@ void sunxi_cpu_off(unsigned int cluster, unsigned int core)
...
 	sunxi_execute_arisc_code(arisc_core_off, sizeof(arisc_core_off),
-				 0, BIT_32(core));
+	                         0, BIT_32(core));
 }
 
-void sunxi_cpu_on(unsigned int cluster, unsigned int core)
+void sunxi_cpu_on(u_register_t mpidr)
 {
+	unsigned int cluster = MPIDR_AFFLVL1_VAL(mpidr);
+	unsigned int core    = MPIDR_AFFLVL0_VAL(mpidr);;
```
apritzel: double semicolon

plat/allwinner/common/sunxi_cpu_ops.c
```diff
@@ -45,25 +45,27 @@ static void sunxi_cpu_enable_power(unsigned int cluster, unsigned int core)
 	mmio_write_32(SUNXI_CPU_POWER_CLAMP_REG(cluster, core), 0x00);
 }
 
-void sunxi_cpu_off(unsigned int cluster, unsigned int core)
+void sunxi_cpu_off(u_register_t mpidr)
 {
-	int corenr = cluster * PLATFORM_MAX_CPUS_PER_CLUSTER + core;
+	unsigned int cluster = MPIDR_AFFLVL1_VAL(mpidr);
+	unsigned int core    = MPIDR_AFFLVL0_VAL(mpidr);;
```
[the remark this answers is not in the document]
smaeul: Whoops, I'll fix that (and the other copy-paste)

plat/allwinner/common/sunxi_bl31_setup.c
smaeul: Fair enough. I'll drop the commit.

plat/allwinner/common/sunxi_cpu_ops.c
smaeul: > And is there a way to do a patch-by-patch review in Github? This whole PR-at-once process is quite painful ...

Yes, you can go to the "Commits" tab, click on one, and start reviewing inline from there. Though as far as I know, GitHub still likes to get the commits in the wrong order in PRs.

smaeul: Funny you should mention that :). No, it can't be (easily) triggered with the current code. `sunxi_cpu_off` is only called from 3 places:
* `sunxi_disable_secondary_cpus` from `bl31_early_platform_setup2`, which is always called from CPU0, so ignores CPU0
* `sunxi_pwr_down_wfi`, which is always called from the CPU being shutdown, so it hangs at the previous line.
* `sunxi_disable_secondary_cpus` from `sunxi_system_off`, which will only be called from CPU0 (Linux's "boot" CPU) if that CPU is online. And if that CPU is already offline... actually, let me check if it is hit in that case.

So you would only hit it when performing a poweroff, and you wouldn't notice that the CPU hung, because that's what it's supposed to do. But if you look at the A64 manual, the diagram in CPUCFG (section 3.4.2, page 148), there's no power switch for core 0. That's `DCDC_EN` for the whole cluster. That's corroborated by the `R_PRCM` documentation for the A83T, which has the core 0 power switch in bit 4, and the cluster power switch in bit 0. And I've experimentally tested this from the ARISC. Search for `R_PRCM_PWROFF_GATING_REG` in [this file](https://github.com/crust-firmware/crust/blob/9f3157394ec430d2198441fe062d00e734a554ef/platform/sun50i/css.c).

smaeul: Also see references to `set_power_switch` in that file. And finally that's what Allwinner does in their blob wrt the power switch (which I can't link directly because there's a few layers of indirection involved in that function).

smaeul: Huh, well the call to `sunxi_disable_secondary_cpus` from `sunxi_system_off` doesn't crash, even with CPU0 offline, so I must have something mixed up. I reverted that change as well for now, since it's not necessary.

antonio-nino-diaz-arm:
IP Review +1

antonio-nino-diaz-arm:
jenkins: test this please

antonio-nino-diaz-arm:
@apritzel: ping

PR #1827, opened 2019-02-18
allwinner: regulators: pick correct DT subnode

apritzel:
So far the DT node describing the AXP803 PMIC used in many Allwinner A64
boards had only one subnode, so our code just entering the first subnode
to find all regulators worked fine.

However recent DT updates in the Linux kernel add more subnodes *before*
that, so we need to make sure to explicitly enter the "regulators"
subnode to find the information we are after.

### DIFF
--- a/plat/allwinner/sun50i_a64/sunxi_power.c
+++ b/plat/allwinner/sun50i_a64/sunxi_power.c
@@ -229,8 +229,8 @@ static void setup_axp803_rails(const void *fdt)
 
 	/* locate the PMIC DT node, bail out if not found */
 	node = fdt_node_offset_by_compatible(fdt, -1, "x-powers,axp803");
-	if (node == -FDT_ERR_NOTFOUND) {
-		WARN("BL31: PMIC: No AXP803 DT node, skipping initial setup.\n");
+	if (node < 0) {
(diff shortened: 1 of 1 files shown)


smaeul:
The change itself looks good to me. However, I see that there's no error handling there. And after the `fdt_node_offset_by_compatible` above there, only one of the possible errors is handled. I think it should be okay to warn and return for either case if `node < 0` (the existing message is fine), and the `for` loop condition could also be made `node >= 0` to catch the other errors.

antonio-nino-diaz-arm:
IP Review +1

antonio-nino-diaz-arm:
@apritzel: ping

Andre-ARM:
@smaeul: Good point. I fixed that and improved the DT error handling in two other places. I can't test it though (traveling), but I will push it as soon as I got the chance to.

Andre-ARM:
Sorry @smaeul, don't mean to bother you, but I think there is no automatic Github notification for updates. So I fixed some error handling, can you please briefly check the changes I pushed? I managed to test it, and it now gracefully bails out in case the DT is botched. Thanks!

antonio-nino-diaz-arm:
jenkins: test this please

antonio-nino-diaz-arm:
The CI failures are unrelated to the changes of this PR (your branch doesn't include the latest platforms that have been merged).

PR #1835, opened 2019-02-20
Apply official names to new Arm Neoverse cores

jts-arm:
As the new Arm Neoverse platforms were announced today, this patch assigns the official names to the new cores.


lib/cpus/aarch64/neoverse_n1.S
```diff
@@ -1,24 +1,24 @@
...
- * This applies to revision r0p0 and r1p0 of Cortex-Ares.
+ * Errata Workaround for Neoverse N1 Errata
+ * This applies to revision r0p0 and r1p0 of Neoverse N1.
  * Inputs:
  * x0: variant[4:7] and revision[0:3] of current cpu.
  * Shall clobber: x0-x17
  * --------------------------------------------------
  */
-func errata_ares_1043202_wa
+func errata_n1_1043202_wa
```
thomas-arm: Just a minor question. Should this be errata_neoverse_n1_1043202_wa ?

jts-arm: Hello @thomas-arm ,

If you see in other similar files, e.g. cortex_a53.S, the errata functions and flags use the shortened name of the core, for example "errata_a53_...._wa" ("cortex" is missing). So we applied the same concept here to make the function name shorter and more readable. Makes sense?

thomas-arm: Ok, that is fine. Thank you. Just wanted to ensure that name qualifications are complete just in case there is a another n1 cpu of a different category.

thomas-arm:
Is it possible to get this patch merged as there are other changes that depend on this patch series.

lib/cpus/aarch64/neoverse_n1.S
jts-arm: > just in case there is a another n1 cpu of a different category
That's a valid point but, theoretically, that shouldn't happen since Cortex family uses A/R/M and Neoverse family uses N/E. Otherwise we'll indeed have to do more renamings in the future :)

jts-arm:
We will still give some time today to @npoushin just in case he has any objections. We will merge it ASAP.

antonio-nino-diaz-arm:
Nariman doesn't need to approve this patch, the script notifies him because of touching a few files in platforms he is the maintainer of, but the change is trivial.

npoushin:
Well, approved anyway but yes it's just trivial changes. Thanks

PR #1849, opened 2019-02-28
Cortex-A73: Implement workaround for errata 852427

loumay-arm:
In AArch32, execution of 2 instructions with opposite condition code
might lead to either a data corruption or a CPU deadlock. Set the bit
12 of the Diagnostic Register to prevent this.


antonio-nino-diaz-arm:
jenkins: test this please

PR #1853, opened 2019-03-01
Tegra: dummy support for the io_storage backend

vwadekar:
This patch provides dummy macros and platform files to compile
the io_storage driver backend. This patch is necessary to
remove the "--unresolved=el3_panic" linker flag from Tegra's
makefiles and allow us to revert this workaround, previously
suggested by the ARM toolchain team.

The "--unresolved=el3_panic" flag actually was a big hammer that
allowed Tegra platforms to work with armlink previously but it
masks legit errors with the code as well.

Change-Id: I0421d35657823215229f84231896b84167f90548


antonio-nino-diaz-arm:
IP Review +1

antonio-nino-diaz-arm:
jenkins: test this please

PR #1858, opened 2019-03-06
drivers: synopsys: Fix synopsys MMC driver

thloh85-intel:
There are some issues with synopsys MMC driver:
- CMD8 should not expect data
- CMD51 should expect data (MMC read)
- dw_prepare should not dictate size to be MMC_BLOCK_SIZE, block size is
now handled in the dw_prepare function
- after the CMD completes, when doing dw_read, we need to invalidate cache
- Need to set FIFO threshold, otherwise DMA will never get the interrupt to
read or write

### DIFF
--- a/drivers/synopsys/emmc/dw_mmc.c
+++ b/drivers/synopsys/emmc/dw_mmc.c
@@ -243,6 +243,11 @@ static int dw_send_cmd(struct mmc_cmd *cmd)
 		op = CMD_WAIT_PRVDATA_COMPLETE;
 		break;
 	case 8:
+		if (dw_params.mmc_dev_type == MMC_IS_EMMC)
+			op = CMD_DATA_TRANS_EXPECT | CMD_WAIT_PRVDATA_COMPLETE;
+		else
+			op = CMD_WAIT_PRVDATA_COMPLETE;
+		break;
 	case 17:
 	case 18:
 		op = CMD_DATA_TRANS_EXPECT | CMD_WAIT_PRVDATA_COMPLETE;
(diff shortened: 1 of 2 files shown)


hzhuang1:
I can't find CMD51 in eMMC v4.4 spec.

And CMD8 is used to fetch EXT CSD. If CMD8 can't read data, how could we fetch EXT CSD?

thloh85-intel:
> I can't find CMD51 in eMMC v4.4 spec.
My bad, I meant ACMD51(SEND_SCR) instead of CMD51. It is in the SD Specification.

> 
> And CMD8 is used to fetch EXT CSD. If CMD8 can't read data, how could we fetch EXT CSD?
Let me check on this. I might be mistaken on this because the drivers/mmc/mmc.c doesn't do read() after sending the CMD, thus I assumed data isn't needed. I'll submit a patch if this is a mistake I make.

antonio-nino-diaz-arm:
IP Review +1

thloh85-intel:
OK, so I've checked, CMD8 is fetch EXT CSD for eMMC, but for SD, it is SEND_IF_COMMAND, thus there's no data involved. I'll handle the cases respectively and push a new commit. Thanks for catching this @hzhuang1

thloh85-intel:
Updated with the fixes requested. Thanks!

soby-mathew:
jenkins: test this please

antonio-nino-diaz-arm:
The failures are due to this PR being based on an old master where some platforms are not present. The platforms introduced since then have (logically) failed to build. Otherwise, it's good.

PR #1861, opened 2019-03-06
Update checkpatch options

Yann-lms:
The first patch removes BRACES warning with conflicts with MISRA.
The second patch gives the possibility to use checkpatch with --strict option, and also removes some messages not compliant with TF-A or MISRA.

### DIFF
--- a/.checkpatch.conf
+++ b/.checkpatch.conf
@@ -73,3 +73,19 @@
 # BRACES reports this kind of messages:
 # braces {} are not necessary for any arm of this statement
 --ignore BRACES
+
+# PREFER_KERNEL_TYPES reports this kind of messages (when using --strict):
+# "Prefer kernel type 'u32' over 'uint32_t'"
+--ignore PREFER_KERNEL_TYPES
+
+# USLEEP_RANGE reports this kind of messages (when using --strict):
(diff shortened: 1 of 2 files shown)


antonio-nino-diaz-arm:
IP Review +1

.checkpatch.conf
```diff
@@ -73,3 +73,19 @@
 # BRACES reports this kind of messages:
 # braces {} are not necessary for any arm of this statement
 --ignore BRACES
+
```
antonio-nino-diaz-arm: I think you should have all the new ignored rules in the same patch. Or, at the very least, have one patch for ``CHECKPATCH_OPTS`` and one for the other options.

Yann-lms: For me the `--ignore BRACES` is to correct an issue that came when MISRA was adopted, that's why I put that in a separate patch.
I can split the second patch in two, one for the `CHECKPATCH_OPTS`, another one for the checks to ignore when using `--strict` option to checkpatch.

Yann-lms: I've pushed a new version in that way (3 patches)

Yann-lms:
Patch rebased

antonio-nino-diaz-arm:
jenkins: test this please

PR #1862, opened 2019-03-07
plat: intel: Improve ECC scrubbing performance

thloh85-intel:
We should be using zeromem to scrub memory instead of memset. This would
improve the performance by 200x


dp-arm:
IP Review +1

PR #1863, opened 2019-03-07
drivers: mmc: Fix some issues with MMC stack

thloh85-intel:
Some bugs in MMC stack needs to be fixed:
- scr cannot be local as this will cause cache issue when invalidating
after the read DMA transfer is completed
- ACMD41 needs to send voltage information in initialization, otherwise the
command is a query, thus will not initialize the controller
- when checking device state, retry until the retries counter goes to zero
before failing

### DIFF
--- a/drivers/mmc/mmc.c
+++ b/drivers/mmc/mmc.c
@@ -29,6 +29,7 @@ static unsigned char mmc_ext_csd[512] __aligned(16);
 static unsigned int mmc_flags;
 static struct mmc_device_info *mmc_dev_info;
 static unsigned int rca;
+static unsigned int scr[2]__aligned(16) = { 0 };
 
 static const unsigned char tran_speed_base[16] = {
 	0, 10, 12, 13, 15, 20, 26, 30, 35, 40, 45, 52, 55, 60, 70, 80
@@ -90,7 +91,8 @@ static int mmc_device_state(void)
 		ret = mmc_send_cmd(MMC_CMD(13), rca << RCA_SHIFT_OFFSET,
 				   MMC_RESPONSE_R1, &resp_data[0]);
 		if (ret != 0) {
-			return ret;
+			retries--;
+			continue;
 		}
 
 		if ((resp_data[0] & STATUS_SWITCH_ERROR) != 0U) {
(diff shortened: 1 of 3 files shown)


drivers/mmc/mmc.c
```diff
@@ -341,8 +342,9 @@ static int sd_send_op_cond(void)
 		}
 
 		/* ACMD41: SD_SEND_OP_COND */
-		ret = mmc_send_cmd(MMC_ACMD(41), OCR_HCS, MMC_RESPONSE_R3,
-				   &resp_data[0]);
+		ret = mmc_send_cmd(MMC_ACMD(41), OCR_HCS |
+			mmc_dev_info->ocr_voltage, MMC_RESPONSE_R3,
```
Yann-lms: That mmc_dev_info->ocr_voltage should be initialized somewhere? Could you give an example in your driver code?
With this patch, I fail to boot on SD-Card on STM32MP1 platform.

Yann-lms: There's an issue in STM32MP1 code (mmc_dev_info struct should be zeromem'ed before using it). I'll propose a patch for this.

thloh85-intel: Hi Yann,
On STM32MP1, without setting the voltage, does the controller gets initialized correctly? 

On the Stratix 10 platform, without sending the voltage, the check "if ((resp_data[0] & OCR_POWERUP) != 0U)" will fail.

Yann-lms: Yes, it is working. But I agree we should provide the supported voltage values in our driver, as it is done for example in STM32MP1 U-Boot driver.

Yann-lms: I've pushed a patch to correct the issue I had to boot (#1868).
And pushed a other branch with the required patches to manage ocr_voltage (https://github.com/Yann-lms/arm-trusted-firmware/tree/ocr_voltage).
I'll push the P-R for the last patch when the 2 other patches are merged.

Yann-lms:
Patch is OK for me, only if #1868 is merged prior to this one.

CC @hzhuang1 for additional review

hzhuang1:
It's fine to me.

hzhuang1:
Verified on hikey.

dp-arm:
IP Review +1

PR #1865, opened 2019-03-07
Fixup register handling in aarch32 reset_handler

mmind:
The BL handover interface stores the bootloader arguments in
registers r9-r12, so when the reset_handler stores the lr pointer
in r10 it clobers one of the arguments.

Adapt to use r8 and adapt the comment about registers allowed
to clober.

I've checked aarch32 reset_handlers and none seem to use higher
registers as far as I can tell.

Fixes: a6f340fe58b9 ("Introduce the new BL handover interface")
Cc: Soby Mathew <soby.mathew@arm.com>
Signed-off-by: Heiko Stuebner <heiko@sntech.de>

### DIFF
--- a/lib/cpus/aarch32/cpu_helpers.S
+++ b/lib/cpus/aarch32/cpu_helpers.S
@@ -21,9 +21,9 @@
 	 */
 	.globl	reset_handler
 func reset_handler
-	mov	r10, lr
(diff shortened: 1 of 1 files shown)


dp-arm:
jenkins: test this please

PR #1866, opened 2019-03-07
rockchip: add an fdt parsing stub for platform param

mmind:
The Rockchip ATF platform can be entered from both Coreboot and U-Boot.
While Coreboot does submit the list of linked parameter structs as
platform param, upstream u-boot actually always provides a pointer
to a devicetree as parameter.
This results in current ATF not running at all when started from U-Boot.

To fix this, add a stub that checks if the parameter is a fdt so we
can at least boot and not get stuck. Later on we can extend this with
actual parsing of information from the devicetree.

Signed-off-by: Heiko Stuebner <heiko@sntech.de>
Cc: Philipp Tomsich <philipp.tomsich@theobroma-systems.com>

### DIFF
--- a/plat/rockchip/common/params_setup.c
+++ b/plat/rockchip/common/params_setup.c
@@ -11,6 +11,7 @@
 #include <common/debug.h>
 #include <drivers/console.h>
 #include <drivers/gpio.h>
+#include <libfdt.h>
 #include <lib/coreboot.h>
 #include <lib/mmio.h>
 #include <plat/common/platform.h>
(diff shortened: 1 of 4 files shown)


antonio-nino-diaz-arm:
IP Review +1

soby-mathew:
jenkins: test this please

soby-mathew:
Please fix the checkpatch error.
```
plat/rockchip/common/params_setup.c:60: ERROR: "foo* bar" should be "foo *bar"
+static int dt_process_fdt(void* blob)
```

Other checks pass.

antonio-nino-diaz-arm:
Checkpatch errors:
```
plat/rockchip/common/params_setup.c:60: ERROR: "foo* bar" should be "foo *bar"
+static int dt_process_fdt(void* blob)
```

soby-mathew:
@mmind, if you can resolve the checkpatch error, this patch can be merged.

mmind:
I've moved the "*" as suggested now

PR #1868, opened 2019-03-07
stm32mp1: zeromem device_info struct

Yann-lms:
The change of the structure highlighted the fact that all fields are not
correctly initialized with zeroes.

Replace the other memset in the function with zeromem, as it is faster.

Change-Id: I27f45a64e34637f79fa519f486bf5936721ef396


dp-arm:
IP Review +1

PR #1875, opened 2019-03-08
fdts: stm32mp1: add bsec node

Yann-lms:
This node is added in a new file stm32mp157c-security.dtsi.
This node includes OTPs that should be shadowed and made readable
to non secure world.

The stgen node is also moved to this file.

Change-Id: I3c89a01588d2e411fecfc44997e1c5df2fc37cad

### DIFF
--- a/fdts/stm32mp157a-dk1.dts
+++ b/fdts/stm32mp157a-dk1.dts
@@ -173,6 +173,7 @@
 /* ATF Specific */
 #include <dt-bindings/clock/stm32mp1-clksrc.h>
 #include "stm32mp15-ddr3-1x4Gb-1066-binG.dtsi"
+#include "stm32mp157c-security.dtsi"
 
 / {
 	aliases {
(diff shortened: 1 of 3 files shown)


Yann-lms:
I've pushed a new version of the patch. First to rebase after the master/integration branch issue. And I've also updated the status and secure-status of the added nodes.

dp-arm:
Will wait until others have okay-ed this before merging.  We are nearing the code freeze period so hopefully we can merge this by the end of the week at the latest.

Yann-lms:
Hi @dp-arm, I just wonder who else should review this patch? It seems @Andre-ARM, @smaeul, @sivadur (sorry for the spam) were just notified after the force push of integration branch. It is the same I think for the other series for checkpatch.

soby-mathew:
The github script to notify other platforms maintainers seems to not work correctly for this case.  This change is specific to ST and we can merge this.

antonio-nino-diaz-arm:
I don't know why the others have been notified, but they are not maintainers affected by this PR.

Andre-ARM:
I think the original PR accidentally contained more patches (based on a wrong branch?), including the pending Allwinner patches. Maybe it was a integration vs. master hiccup?

antonio-nino-diaz-arm:
Yeah, that's likely what happened here. The script is pretty dumb and it only checks the list of files touched by a PR, which is enough in almost all cases.

PR #1876, opened 2019-03-11
Integration+bl2 el3 eret fix

bryanodonoghue:
On a subset of our Arm-v7 boards we have observed a failure at the final instruction "eret" to switch to OPTEE.

Drilling down into this it appears as if the usage of eret in bl2_el3_entrypoint.S relies on an undefined behavior.

The Arm documentation states the following with respect to the behaviour of the eret instruction:

>> http://infocenter.arm.com/help/topic/com.arm.doc.dui0588b/CJAEEEJG.html
>> ERET
>> When executed in **Hyp mode**, Exception Return loads the PC from ELR_hyp and loads the CPSR from SPSR_hyp.
>> When executed in a Secure or Non-secure PL1 mode, ERET behaves as:
>> •MOVS PC, LR in the ARM instruction set, see SUBS PC, LR and related instructions (ARM) on page B9-2010

And similarly states the following with respect to Hyp mode

>>https://static.docs.arm.com/ddi0406/c/DDI0406C_C_arm_architecture_reference_manual.pdf
>>**Hyp mode** is implemented only as part of the Virtualization Extensions, and **only in Non-secure state**. This means that:
>>
>>• implementations that do not include the Virtualization Extensions have only two privilege
levels, PL0 and PL1
>>
>>• execution in Secure state has only two privilege levels, PL0 and PL1.
>>In an implementation that includes the Security Extensions, the execution privilege levels are defined independently
in each security state, and there is no relationship between the Secure and Non-secure privilege levels.

Therefore relying on 'eret' to populate the CPSR on an 'eret' when in secure monitor mode is not architecturally valid since the defined behavior for populating CPSR in this fashion applies only to Hyp Mode, not to secure monitor mode.

This patch fixes the issue we have seen with this behavior by specifically setting the CPSR with the required value prior to executing eret, since in secure monitor mode, CPSR may not in fact be updated by the relevant micro-operations.

### DIFF
--- a/bl2/aarch32/bl2_el3_entrypoint.S
+++ b/bl2/aarch32/bl2_el3_entrypoint.S
@@ -74,11 +74,10 @@ func bl2_run_next_image
 
 	/*
 	 * Extract PC and SPSR based on struct `entry_point_info_t`
-	 * and load it in LR and SPSR registers respectively.
+	 * and load it in to the LR.
 	 */
 	ldr	lr, [r8, #ENTRY_POINT_INFO_PC_OFFSET]
-	ldr	r1, [r8, #(ENTRY_POINT_INFO_PC_OFFSET + 4)]
-	msr	spsr, r1
+	ldr	r4, [r8, #(ENTRY_POINT_INFO_PC_OFFSET + 4)]
 
 	/* Some BL32 stages expect lr_svc to provide the BL33 entry address */
 	cps	#MODE32_svc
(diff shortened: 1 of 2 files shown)


Yann-lms:
It seems you've pushed on master branch, that's why one of my patches came with your patch, and I was then notified.
Anyway, I also tested your patch as we use the same bl2_el3 file on STM32MP1. But now it fails to boot, in a sp_min configuration.

bryanodonoghue:
@Yann-lms hmm - thanks.
I'll have to rework this ..

bryanodonoghue:
I've read the arch documents wrong here

If I change
msr spsr, r1 to msr spsr_fsxc, r1 I can clear the error condition on my reference board and continue to use eret.

I'll repost this PR with an updated description
@Yann-lms thanks for the prompt test

PR #1877, opened 2019-03-11
bl2-el3: Fix exit to bl32 by ensuring full write to SPSR

bryanodonoghue:
We are observing a failure mode on a small minority of i.MX7 boards we have in our ATF setup.

Investigating this issue shows that the failure point is when 'eret' is executed to switch from BL2 to BL3.

On first blush, I thought it was down to misuse of eret intself for the particular mode the processor was in. With feedback from @Yann-lms though it was apparent my fix and diagnosis were wrong.

https://github.com/ARM-software/arm-trusted-firmware/pull/1876

Looking at OP-TEE and indeed other parts of the ATF codebase though we can see that the error is mis-setting of the SPSR, in the current case only setting bits 31:24 and 7:0 respectively.

Setting the full SPSR value though will negate the failure for me and ought to be correct for anybody else interested in bits 23:8 of the SPSR on eret.

### DIFF
--- a/bl1/aarch32/bl1_exceptions.S
+++ b/bl1/aarch32/bl1_exceptions.S
@@ -71,7 +71,7 @@ debug_loop:
 	 */
 	ldr	lr, [r8, #ENTRY_POINT_INFO_PC_OFFSET]
 	ldr	r1, [r8, #(ENTRY_POINT_INFO_PC_OFFSET + 4)]
-	msr	spsr, r1
+	msr	spsr_xc, r1
 
 	/* Some BL32 stages expect lr_svc to provide the BL33 entry address */
 	cps	#MODE32_svc
--- a/bl2/aarch32/bl2_el3_entrypoint.S
+++ b/bl2/aarch32/bl2_el3_entrypoint.S
@@ -78,7 +78,7 @@ func bl2_run_next_image
 	 */
 	ldr	lr, [r8, #ENTRY_POINT_INFO_PC_OFFSET]
 	ldr	r1, [r8, #(ENTRY_POINT_INFO_PC_OFFSET + 4)]
-	msr	spsr, r1
+	msr	spsr_xc, r1
 
 	/* Some BL32 stages expect lr_svc to provide the BL33 entry address */
 	cps	#MODE32_svc


Yann-lms:
With this patch, STM32MP1 is booting well.

soby-mathew:
jenkins: test this please

bryanodonoghue:
@antonio-nino-diaz-arm 

Yep I've found something in bl1 smc-handler that looks similar. I have a patch I'll post now

antonio-nino-diaz-arm:
Not just that. I've done a quick grep (``grep -rn spsr | grep -v spsr_ | grep "\.S"``) and that has shown that also the ``smccc_macros.S`` file is affected. Also, our ``DEFINE_SYSREG_RW_FUNCS(spsr)`` in ``arch_helpers.h`` is wrong, it will simply expand to doing a read from ``spsr``, which may not be what we want.

bryanodonoghue:
@antonio-nino-diaz-arm logically I have to agree.
I tend to agree with you there.

```
.smccc_save_gp_mode_regs
cps     #MODE32_irq
mrs     r2, spsr

```
but

```
.monitor_exit
cps     #MODE32_irq
ldm     r1!, {r2, sp, lr}
msr     spsr_fsxc, r2

```
On the face of it, it does appear as if we would drop the intervening bits...

I'll patch it out and we can discuss

include/arch/aarch32/smccc_macros.S
```diff
@@ -33,28 +33,28 @@
 	stm	r0!, {sp, lr}
 
 	cps	#MODE32_irq
-	mrs	r2, spsr
+	mrs	r2, spsr_fsxc
```
soby-mathew: hmm, OPTEE monitor seems to be doing the same thing:
https://github.com/OP-TEE/optee_os/blob/c3d0b15de7ece03f0b84b1d15379345b31674a61/core/arch/arm/sm/sm_a32.S#L26

@jenswi-linaro , @etienne-lms 
I would like some feedback from OPTEE folks regarding this.

bryanodonoghue: @soby-mathew 
Looking at Linux we see the same thing spsr is used for mrs ans spsr_fxsc is used for mrs.

```
deckard@event-horizon:~/Development/linux$ grep -r mrs * | grep spsr
arch/arm/boot/compressed/head.S:		mrs	r0, spsr
arch/arm/boot/compressed/head.S:		mrs	r0, spsr		@ Get saved CPU boot mode

deckard@event-horizon:~/Development/linux$ grep -r msr * | grep spsr
arch/arm/boot/compressed/head.S:		msr	spsr_cxsf, r9		@ Save the CPU boot mode in

```
Still seems to me as if you'd be dropping bit 8 for the mrs though...

bryanodonoghue: Ah OK I see, trying this on OPTEE we see mrs r1, spsr_fsxc is not a valid opcode.
```

core/arch/arm/kernel/thread_a32.S:290: Error: 'APSR', 'CPSR' or 'SPSR' expected -- `mrs r1,spsr_fsxc'
core/arch/arm/kernel/thread_a32.S:340: Error: 'APSR', 'CPSR' or 'SPSR' expected -- `mrs r1,spsr_fsxc'
core/arch/arm/kernel/thread_a32.S:984: Error: 'APSR', 'CPSR' or 'SPSR' expected -- `mrs r0,spsr_fsxc'
```

bryanodonoghue:
I've dropped the update to change "mrs r1, spsr" to "mrs r1, spsr_fsxc" the latter mnemonic is not a valid operation.

antonio-nino-diaz-arm:
It seems to me that we always need the 4 fields:
https://gnu.googlesource.com/binutils-gdb/+/82cb27ff6b4e1ae811e72a02dddc3db4c9ff97b8/gas/config/tc-arm.c#19196
``spsr_all`` hasn't meant all flags in a long time. Not specifying anything defaults to ``fc``:
https://gnu.googlesource.com/binutils-gdb/+/82cb27ff6b4e1ae811e72a02dddc3db4c9ff97b8/gas/config/tc-arm.c#6024

The problem is that all code generated in C using ``write_spsr()`` and ``read_spsr()`` is wrong as well. We aren't using this at all in C code in AArch32, so it's fine.

antonio-nino-diaz-arm:
Oh, ok, that one is invalid. That's strange, but I guess that's the right thing to do then.

antonio-nino-diaz-arm:
jenkins: test this please

etienne-lms:
Hi,
I agree with the 2 proposed changes. The value of SPSR written there (`bl2_run_next_image()` and `bl1_aarch32_smc_handler()`) is set by the platform (in `bl_mem_params_node_t` structures) and should set the whole expected processor state with entering target EL, including endianness etc... 
Note  however that flags and status bits are likely meaningless (and likely 0 if platform do not set it), hence `msr	spsr_cx, r1` should be enough.

Regarding other places where SPSR is manipulated, it depends on the purpose. Obviously when restoring a execution context, `msr	spsr_cxsf, <Rx>` shall be used as all SPSR fields reflect the resumed CPU state. This is the case in the OP-TEE OS code [1] referred above, function is `sm_restore_unbanked_regs()`.

[1] https://github.com/OP-TEE/optee_os/blob/c3d0b15de7ece03f0b84b1d15379345b31674a61/core/arch/arm/sm/sm_a32.S#L26

bryanodonoghue:
I've updated this to capture the XC bits of the SPSR alone, not the entire FSXC as suggested by @etienne-lms

PR #1886, opened 2019-03-14
Fix extra compilation warnings

ambroise-arm:
Fix some warnings when compiling the ARM reference platform with extra compiler warning flags enabled (W={1,2,3} in the makefile).

### DIFF
--- a/drivers/console/multi_console.c
+++ b/drivers/console/multi_console.c
@@ -13,11 +13,11 @@
 console_t *console_list;
 uint8_t console_state = CONSOLE_FLAG_BOOT;
 
+IMPORT_SYM(console_t *, __STACKS_START__, stacks_start)
+IMPORT_SYM(console_t *, __STACKS_END__, stacks_end)
+
 int console_register(console_t *console)
 {
-	IMPORT_SYM(console_t *, __STACKS_START__, stacks_start)
-	IMPORT_SYM(console_t *, __STACKS_END__, stacks_end)
-
 	/* Assert that the struct is not on the stack (common mistake). */
 	assert((console < stacks_start) || (console >= stacks_end));
 	/* Assert that we won't make a circle in the list. */
--- a/plat/arm/board/fvp/fvp_private.h
+++ b/plat/arm/board/fvp/fvp_private.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2014-2018, ARM Limited and Contributors. All rights reserved.
+ * Copyright (c) 2014-2019, ARM Limited and Contributors. All rights reserved.
  *
  * SPDX-License-Identifier: BSD-3-Clause
  */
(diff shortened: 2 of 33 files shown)


plat/arm/board/fvp/fvp_private.h
```diff
@@ -20,5 +20,7 @@ void fvp_interconnect_enable(void);
 void fvp_interconnect_disable(void);
 void tsp_early_platform_setup(void);
 
+void arm_gicv3_distif_pre_save(unsigned int rdist_proc_num);
+void arm_gicv3_distif_post_restore(unsigned int rdist_proc_num);
```
soby-mathew: Better to refer to the declaration in gicv3_private.h

drivers/console/console_private.h
```diff
@@ -0,0 +1,15 @@
...
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef CONSOLE_PRIVATE_H
+#define CONSOLE_PRIVATE_H
+
+#include <drivers/console.h>
+
+/* Private console_register() function. */
+int console_register(console_t *console);
```
jwerner-chromium: I think this should just go into console.h, there's no need to create a separate header. The "There is no publicly accessible console_register() function." comment in there is outdated -- since we support consoles written only in C now, those need to be able to call console_register().

drivers/console/multi_console.c
```diff
@@ -10,14 +10,16 @@
 
 #include <drivers/console.h>
 
+#include "./console_private.h"
+
 console_t *console_list;
 uint8_t console_state = CONSOLE_FLAG_BOOT;
 
+IMPORT_SYM(console_t *, __STACKS_START__, stacks_start)
+IMPORT_SYM(console_t *, __STACKS_END__, stacks_end)
```
jwerner-chromium: Why does this need to move?

ambroise-arm: The `IMPORT_SYM` macro declares an extern variable, which generates a nested-extern warning when used inside a function.
The way this macro is used in every other file in the project is outside of any function.

jwerner-chromium: Okay, fair enough.

drivers/console/console_private.h
ambroise-arm: Can you provide a description of `console_register` to put above its declaration in console.h?

jwerner-chromium: "Add a console_t instance to the console list. This should only be called by console drivers after they have initialized all fields in the console structure. Platforms seeking to register a new console need to call the respective console_<driver>_register() function instead."

antonio-nino-diaz-arm:
jenkins: test this please

antonio-nino-diaz-arm:
The failure is due to some MISRA issues that we can ignore in this PR.

PR #1895, opened 2019-03-18
Declare ENABLE_PAUTH build option as experimental

AlexeiFedorov:
Declare ENABLE_PAUTH and CTX_INCLUDE_PAUTH_REGS
build options as experimental.
Pointer Authentication is enabled for Non-secure world
irrespective of the value of these build flags if the
CPU supports it.
The patch also fixes the description of fiptool 'help' command.

Change-Id: I46de3228fbcce774a2624cd387798680d8504c38
Signed-off-by: Alexei Fedorov <Alexei.Fedorov@arm.com>

### DIFF
--- a/docs/user-guide.rst
+++ b/docs/user-guide.rst
@@ -358,11 +358,12 @@ Common build options
    registers to be included when saving and restoring the CPU context. Default
    is 0.
 
--  ``CTX_INCLUDE_PAUTH_REGS``: Boolean option that, when set to 1, will cause
-   the ARMv8.3-PAuth registers to be included when saving and restoring the CPU
-   context. Note that if the hardware supports this extension and this option is
-   set to 0 the value of the registers will be leaked between Secure and
-   Non-secure worlds if PAuth is used on both sides. The default is 0.
+-  ``CTX_INCLUDE_PAUTH_REGS``: Boolean option that, when set to 1, enables
+   Pointer Authentication for Secure world. This will cause the ARMv8.3-PAuth
+   registers to be included when saving and restoring the CPU context as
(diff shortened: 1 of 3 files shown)


docs/user-guide.rst
```diff
@@ -363,6 +363,7 @@ Common build options
    context. Note that if the hardware supports this extension and this option is
```
soby-mathew: add : "This enables pointer Authentication to be used in Secure World. "

docs/user-guide.rst
```diff
@@ -363,6 +363,7 @@ Common build options
...
    set to 0 the value of the registers will be leaked between Secure and
    Non-secure worlds if PAuth is used on both sides. The default is 0.
+   This is an experimental feature.
```
soby-mathew: Add: "Note that pointer Authentication is enabled in Non-Secure world irrespective of any build flag if the CPU supports it."

docs/user-guide.rst
```diff
@@ -417,6 +418,7 @@ Common build options
    extension, they are always saved if they are detected regardless of the value
    of this option. If enabled, it is needed to use a compiler that supports the
    option ``-msign-return-address``. It defaults to 0.
+   This is an experimental feature.
```
soby-mathew: Suggest: "Boolean option to enable ARMv8.3 Pointer Authentication
(``ARMv8.3-PAuth``) support for TF-A BL images itself. If enabled, it is needed to use a compiler that supports the
option ``-msign-return-address``. It defaults to 0 and this is an experimental feature.  Note that Pointer Authentication is enabled in Non-Secure world irrespective of any build flag if the CPU supports it."

PR #1903, opened 2019-03-22
driver: synosys: Fix SD MMC not initializing correctly

thloh85-intel:
dw_params.mmc_dev_type should be assigned before mmc_init, otherwise SDMMC
initialization will fail as the initialization treats the device as EMMC
instead of SD.


hzhuang1:
Acked

PR #1916, opened 2019-04-01
SPM: Some improvements to xlat handling code

antonio-nino-diaz-arm:
Move shim layer mapping to TTBR1_EL1 to allow partitions to reduce the address space size in TBBR0_EL0. This way they need less memory for translation tables as there can be less levels of translation.

Check the individual commits for more information.


antonio-nino-diaz-arm:
The CI failure is a false positive from Coverity.

PR #1929, opened 2019-05-16
plat/meson/gxl: BL31: remove BL2 dependency

khilman:
Remove an assert() that assumes a specific value being passed from
BL2.  This value is dependent on BL2 version, so makes this assert()
not portable.

Suggested-by: Remi Pommarel <repk@triplefau.lt>
Signed-off-by: Kevin Hilman <khilman@baylibre.com>


sandrine-bailleux-arm:
Pull request merged on trustedfirmware.org:
https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/1111

PR #1932, opened 2019-08-21
plat: imx8m: update OTP SIP

MrVan:
We could use one SIP function id for READ/WRITE,
no need use different function id.

Signed-off-by: Peng Fan <peng.fan@nxp.com>


joannafarley-arm:
Cleaning up.

PR #1939, opened 2021-01-28
mmc: Change mmc_dev_info from pointer to global variable

hudeqiWH:
1.Since in bl_platform_setup, the mmc_device_info  passed in are all temporary variables, avoid referencing the released space on the stack at runtime, so need to use global variables to store mmc_device_info in mmc.c
2.Delete redundant;

Signed-off-by: deqi.hu@siengine.com

### DIFF
--- a/drivers/mmc/mmc.c
+++ b/drivers/mmc/mmc.c
@@ -27,7 +27,7 @@ static unsigned int mmc_ocr_value;
 static struct mmc_csd_emmc mmc_csd;
 static unsigned char mmc_ext_csd[512] __aligned(16);
 static unsigned int mmc_flags;
-static struct mmc_device_info *mmc_dev_info;
+static struct mmc_device_info mmc_dev_info;
 static unsigned int rca;
 static unsigned int scr[2]__aligned(16) = { 0 };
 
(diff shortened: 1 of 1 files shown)


hudeqiWH:
as picture show mmc_read_blocks maybe called beyond function bl1_platform_setup


mmc_read_blocks maybe called beyond bl1_platform_setup

Yann-lms:
Hi,

This github is a read-only mirror for review.trustedfirmware.org.
Please check this link for submitting patches:
https://trustedfirmware-a.readthedocs.io/en/latest/process/contributing.html

That said, I'd prefer not to modify the framework.
The data should be kept either in dw_mmc.c (that's what is done in stm32_sdmmc2.c, but we do not expect to have 2 instances of MMC devices) or in hikey_bl1_setup.c, and other platform files using the dw_mmc driver.

Regards,
Yann

hudeqiWH:
> dw_mmc

Hi,

Thank you for your reply. Refer to other platform implementation(which all don`t have permanent instance), there may be a reference to the risk of free space, I will refer to your tips to submit patches in other repository.

Regards,
Deqi

hudeqiWH:
commit on https://review.trustedfirmware.org/

PR #1940, opened 2021-03-15
plat/allwinner: don't enable regulators disabled in DT

zavorka:
Since July 2020 (36766d39e8), u-boot began building all device trees
with symbols unless OF_LIBFDT_OVERLAY is explicitly disabled.

Consequentially, each node in a device tree is assigned a phandle,
regardless of it being actually referenced anywhere. This change made
the selective enablement of only those regulators for which a phandle
is present by the axp driver effectively useless.

In order to reinstate a way of achieving selectivity in what regulators
get enabled, check for a non-'okay' value of the 'status' property.

### DIFF
--- a/drivers/allwinner/axp/common.c
+++ b/drivers/allwinner/axp/common.c
@@ -96,8 +96,19 @@ static int setup_regulator(const void *fdt, int node,
 	return 0;
 }
 
+static int is_node_disabled(const void *fdt, int node)
+{
+	const char *cell;
+	cell = fdt_getprop(fdt, node, "status", NULL);
+	if (cell)
+		return strcmp(cell, "okay") != 0;
+	return 0;
+}
+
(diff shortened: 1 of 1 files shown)


zavorka:
Tried that, got an unexpected response:
```
 ! [remote rejected]     HEAD -> refs/for/integration/axp-reg-status (branch integration/axp-reg-status not found)
```

sandrine-bailleux-arm:
Hi @zavorka ,

What is the exact git command line you used that prompted this error message?

Yann-lms:
Hi,

It seems that the old way of using topic is no more supported, you should use: refs/for/integration%topic=axp-reg-status

Regards,
Yann

zavorka:
Okay, it is [there](https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/9207). Thank you, @Yann-lms

PR #1981, opened 2021-11-02
fix(auth): update with mbedtls development branch

BruceFangShengzhou:
the latest version of mbedtls implement sha256 based on sha224 and del rsa_internal.c


BruceFangShengzhou:
fine

PR #1986, opened 2022-07-27
Fix for PRI*PTR (inttypes.h)

katyo:
This fix solves problems with using PRI*PTR on aarch64 like the following:
```
error: format '%x' expects argument of type 'unsigned int', but argument 3 has type 'uintptr_t' {aka 'long unsigned int'}
```


odeprez:

According to the contributors' guide https://trustedfirmware-a.readthedocs.io/en/latest/process/contributing.html changes shall be submitted to https://review.trustedfirmware.org/

If this change is still required, can you please submit as advised above?

PR #1988, opened 2022-09-27
fix(rpi3_pm): fix RPi4 fails to boot secondary core at second time by PSCI

Luoject:
fix(rpi3_pm): fix RPi4 fails to boot secondary core at second time by PSCI

* the problem is that rpi3_pwr_down_wfi uses the same hook, which is not compatible with older function rpi3_pwr_domain_pwr_down_wfi.
* now we try to fix it.
* Issue: https://developer.trustedfirmware.org/T1003

Signed-off-by: luojects <luoyonglun@huawei.com>

### DIFF
--- a/plat/rpi/common/rpi3_pm.c
+++ b/plat/rpi/common/rpi3_pm.c
@@ -123,15 +123,6 @@ static void rpi3_pwr_domain_off(const psci_power_state_t *target_state)
 #endif
 }
 
-void __dead2 plat_secondary_cold_boot_setup(void);
-
-static void __dead2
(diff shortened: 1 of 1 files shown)


odeprez:

According to the contributors' guide https://trustedfirmware-a.readthedocs.io/en/latest/process/contributing.html changes shall be submitted to https://review.trustedfirmware.org/

If this change is still required, can you please submit as advised above?

odeprez:
Closing as no activity on this PR since 2022, please follow up by submitting a gerrit change to trustedfirmware.org if required, thanks.

PR #1995, opened 2023-03-25
Provide a mechanism to configure the initial parameters of the

rob-the-dude:
sdhost controller (initial clock rate, bus width), and then configure those parameters.

This change allows warm reboots of UEFI on Raspberry Pi 3B+ where existing code often fails with "unknown error".  See discussion at:

	https://github.com/pftf/RPi3/issues/24

The basic problem was that a number of SD commands are issued before sensible initialization configurations have been committed. I suspect that the primary culprit is setting the "slow card" bit, but the initial clock rate also seemed wrong, as did the missing voltage settings.


rob-the-dude:
This fix has been merged to TF-A via:

https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/20297

PR #2007, opened 2024-12-09
Add a SBOM template in CycloneDX format

hughsie:
Hi,

My name is Richard Hughes and I'm a developer at Red Hat. I'm the maintainer of fwupd and LVFS, and am trying to improve software supply chain security by encouraging OEMs, ODMs and IBVs to ship Software Bill of Materials with each firmware binary blob (SBOMs).

I'm working alongside lots of other companies proactively trying to do the right thing. The reason I've opened this pull request is because your project is either used in the *build process* of a firmware we care about (e.g. coreboot) or is built *into the firmware binary* itself. Although my personal focus is on firmware, the SBOM file is in CycloneDX format (one of the most popular industry standards) which makes it also useful when building containers or OS images too.

I would like to contribute this template SBOM file into your project that gets included into source control with substituted values that get populated automatically. I'm not super familiar with arm-trusted-firmware, and so I've done my best populating the project values -- but please point out any that are incorrect and I'll fix them up. I've also put the `sbom.cdx.json` file in what I feel is the right place, but please say if you want me to put it somewhere different or name it a different thing; the directory and `sbom` prefix are unimportant. I also wasn’t 100% sure whether to mark the component as a *library* or *application*, so advice is welcome.

The various firmware build tools will take these incomplete SBOM files and then build them into a complete composite SBOM to represent the firmware. Having an upstream reference to what the PURL and CPE values should be means we have something we can trust; I could quite easily spin up a web-service that we say "what CPE do we use for X" -> `cpe:2.3:a:Y:Z::::::::` but we don't actually know if that's still true, up to date, or what the maintainer actually wants them to be. Putting the template upstream means we can trust the values we find in the checked out code during the build process.

I've written a bit more about this proposal here https://blogs.gnome.org/hughsie/2024/11/14/firmware-sboms-for-open-source-projects/ and there's also lot more information about firmware SBOMs here: https://lvfs.readthedocs.io/en/latest/sbom.html – many thanks for your time.

### DIFF
--- a/docs/sbom.cdx.json
+++ b/docs/sbom.cdx.json
@@ -1,0 +1,43 @@
+{
+  "bomFormat": "CycloneDX",
+  "specVersion": "1.6",
+  "version": 1,
+  "metadata": {
+    "authors": [
+      {
+        "name": "@VCS_SBOM_AUTHORS@"
+      }
+    ]
+  },
+  "components": [
+    {
+      "type": "library",
+      "bom-ref": "pkg:github/ARM-software/arm-trusted-firmware@@VCS_TAG@",
+      "cpe": "cpe:2.3:a:arm:arm-trusted-firmware:@VCS_TAG@:*:*:*:*:*:*:*",
+      "name": "arm-trusted-firmware",
+      "version": "@VCS_VERSION@",
+      "description": "Reference implementation of secure world software for Arm A-Profile architectures",
+      "authors": [
+        {
+          "name": "@VCS_AUTHORS@"
+        }
+      ],
+      "supplier": {
+        "name": "Arm"
+      },
+      "licenses": [
+        {
+          "license": {
+            "id": "BSD-3-Clause"
+          }
+        }
+      ],
+      "externalReferences": [
+        {
+          "type": "vcs",
+          "url": "https://github.com/ARM-software/arm-trusted-firmware"
+        }
+      ]
+    }
+  ]
+}


odeprez:
Hi,

The official Trusted Firmware-A project mirror is now https://github.com/TrustedFirmware-A/trusted-firmware-a
I created a corresponding github issue https://github.com/TrustedFirmware-A/trusted-firmware-a/issues/32

Note we're not using the PR flow but accept submission through trustedfirmware.org https://review.trustedfirmware.org/
Would it be possible to submit your change through gerrit? Thanks.

hughsie:
> I created a corresponding github issue https://github.com/TrustedFirmware-A/trusted-firmware-a/issues/32

Thanks.

>  Would it be possible to submit your change through gerrit? Thanks.

IMHO (from someone that works with dozens of other projects), Gerrit is a the least preferred way to send contributions and it really makes it much harder for "an outsider" to contribute to open source projects.  The review.trustedfirmware.org `push your first change for code review` help text literally doesn't work and https://review.trustedfirmware.org/Documentation/user-upload.html doesn't actually use the correct domain. Guessing the values from other times I've used Gerrit, I tried

    git push ssh://hughsie@review.trustedfirmware.org:29418/TF-A/trusted-firmware-a HEAD:refs/for/master

which greeted me with:

	remote: You need 'Create Change' rights to upload code review requests.
	remote: Verify that you are pushing to the right branch.
	remote: User: hughsie
	remote: Contact an administrator to fix the permissions
	remote: Processing changes: refs: 1, done    
	To ssh://review.trustedfirmware.org:29418/TF-A/trusted-firmware-a
	 ! [remote rejected]     HEAD -> refs/for/master (prohibited by Gerrit: not permitted: create change on refs/heads/master)
	error: failed to push some refs to 'ssh://review.trustedfirmware.org:29418/TF-A/trusted-firmware-a'

odeprez:
https://github.com/TrustedFirmware-A/trusted-firmware-a/issues/32

PR #2011, opened 2025-01-28
feat(imx93): add BL31 PIE support

mgrzeschik:
Enable PIE support so the BL31 firmware can be loaded from anywhere within the OCRAM (SRAM). For the PIE support we only need to replace the BL31_BASE define by the BL31_START symbol which is a relocatable and we need to enable it by setting ENABLE_PIE := 1.


odeprez:

According to the contributors' guide https://trustedfirmware-a.readthedocs.io/en/latest/process/contributing.html changes shall be  submitted to https://review.trustedfirmware.org/ 

